PR #311, opened 2015-03-29
Pass the correct update marker to MXCFB_SEND_UPDATE ioctl.

erosennin:
I noticed that waiting for updates is broken on my Kobo Aura. According to strace, `MXCFB_SEND_UPDATE` is always called with `marker = 0`. This patch is supposed to fix that.

Tested only on Kobo Aura, but should work on other devices too, please test.


hwhw:
Oh dear, it's been broken since long ago. Thanks for catching this. I guess there are not many devices that actually need waiting.

Markismus:
That would explain why my Aura H2O doesn't update anymore. I'll try to test when I've got some time.

PR #312, opened 2015-03-30
Fix docToWindowRect fail for the last word of a line

chrox:


hwhw:
Travis' clang build has some issues, as it seems - missing include files? A bit funny looking, this. I'll merge anyway since it doesn't seem to be related to this patch.

PR #317, opened 2015-04-15
Build sdcv for Android

chrox:
by static linking of libz, libglib-2.0, libiconv and libintl

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,8 @@ build
 sdcv
 tar-1.28
 glib-2.6.6
+libiconv-1.14
+gettext-0.19
 openssl-1.0.1g
 openssl-1.0.1h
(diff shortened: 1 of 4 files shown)


hwhw:
Sorry, I merged a bit too fast: do you know why travis build is broken?

chrox:
Oh, I should test it earlier. The new glib has a dependency of pthread. I have sent a new PR on #318.

hwhw:
Though the busted tests do still fail. Hm. Did they ever work recently or are they only run for the main koreader repository?

chrox:
I noticed that the koptcontext sometimes failed the spec at `totable/fromtable` but it doesn't fail reliably. So probably it needs further investigation.

PR #322, opened 2015-04-26
Misc. bag o' changes!

NiLuJe:
* Nail in the coffin of the H2O puzzlement regarding waveform update modes.
* And update a bunch of third-party libs.


NiLuJe:
(Waiting for Travis, since I'm lazy and haven't actually double-checked the build)

PR #324, opened 2015-04-27
Massive bump of 3rd-party libs

NiLuJe:
... Including MUPDF, DVJU & LuaJIT, so, err, here be dragons? ^^.

(-> Travis will be my friend on this one ^^)

* Also a minor tweak to yesterday's update mode tweaks, including a fix for untracked update_marker (0 is invalid, which is potentially harmless, but not that great, and at the very least had rendered WAIT_FOR_SUBMISSION useless on Kindles).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,7 +36,7 @@ openssl-1.0.1g
 openssl-1.0.1h
 libpng-1.6.17
 thirdparty/gettext/gettext-0.*.*
-giflib-5.1.0
+giflib-5.1.1
 libjpeg-turbo-1.4.0
 jpeg
 lpeg*
--- a/Makefile
+++ b/Makefile
@@ -166,7 +166,7 @@ $(DJVULIBRE_LIB): $(JPEG_LIB)
 		&& CC="$(CC)" CXX="$(CXX)" CFLAGS="$(CFLAGS)" \
 		CXXFLAGS="$(CXXFLAGS)" LDFLAGS="$(LDFLAGS)" \
 		LIBS="$(STATIC_LIBSTDCPP)" \
-		../configure -q --disable-desktopfiles \
+		../autogen.sh -q --disable-desktopfiles \
 			--disable-static --enable-shared \
 			--disable-xmltools --disable-largefile \
 			--without-jpeg --without-tiff \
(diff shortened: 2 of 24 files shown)


NiLuJe:
Still WIP, I'm currently battling the MUPD API changes...

NiLuJe:
Okay, we're left with a single testsuite failure, that I'm not quite sure about, and can't really investigate further on my end, because luarocks is being terrible on my box.

If anyone more familiar with the mupdf testsuite has an idea, I'm all ears :).

NiLuJe:
Currently whipping up a Kindle build, so far, so good...

Sidenote: glib depends on zlib & libffi, which we absolutely never build. As such, the build fails on a barebones TC/sysroot.
I never caught that because I actually build those two for my own stuff, but this time I started the build from a blank sysroot...

NiLuJe:
Well, fuck me, it works. :D.

PDFs, ePubs & DjVu open without blowing up. Yeepee.

PDF annotation appear to be functional, so I'm really at a loss regarding the test failure.

Test build uploaded [here](http://files.ak-team.com/niluje/kopub/koreader-kindle-arm-kindle5-linux-gnueabi-v2013.03-2556-g31ea5a0.tar.gz) ;).

NiLuJe:
Epic VIM fail on ff9f52a commit message xD.

chrox:
The failed test case tried to do binary compare the annotated pdf file with the uploaded one which always fails after mupdf upgrade. Probably incremental update of the original file could be more stable.

chrox:
Crengine cannot load the new droidsansfallback. And all CJK characters won't be rendered properly for EPUB documents. 

screenshot from 2015-04-28 14 29 44

NiLuJe:
Ouch, my bad.

In fact, they're all messed up on my own box as well, so, not just a weird embedded thing. Never expected something like that, but should have checked nonetheless.

I'll pull older, saner copies ;).

chrox:
Never mind, it will be fixed in #326 and koreader/koreader#1510.

NiLuJe:
Agh. I pushed sane copies in #1511 if need be ;).

PR #326, opened 2015-04-28
Revert pulling droid fonts from mupdf resources

chrox:
and fix font filenames in urw.


NiLuJe:
Fantastic! Thank you :).

PR #344, opened 2015-10-04
if there is custom screen dpi also scale by dpi

chrox:


Frenzie:
Ah yes, I'd wanted to look into this myself sometime.

PR #345, opened 2015-10-06
patch mupdf to add support of password protected zip file

chrox:
now mupdf supports the traditional PLZIP encryption and a much stronger AES encryption for ZIP documents.


houqp:
cool, will the patch be contributed back to upstream?

chrox:
Yes, after we have adequate tests in koreader I will send this patch to upstream mupdf.

PR #348, opened 2015-10-11
Unbreak LuaJIT

NiLuJe:
Remind me not to play with git on 3 hours of sleep next time ;/.


houqp:
Maybe I should remind you get more sleep instead :P

PR #349, opened 2015-11-10
fix build in kobo

houqp:
add headers files in dependency

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ $(OUTPUT_DIR)/plugins:
 # (for directory and file name config, see Makefile.defs)
 
 # freetype, fetched via GIT as a submodule
-$(FREETYPE_LIB):
+build_freetype_lib:
 	mkdir -p $(FREETYPE_DIR)/build
 	cd $(FREETYPE_DIR) && sh autogen.sh
 	cd $(FREETYPE_DIR)/build && \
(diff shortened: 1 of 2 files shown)


houqp:
For unknown reason, my local checkout went into a weird state that kobo build keeps failing because of missing header files. And there was no way to trigger regen of the those headers in the build directory. I had to add these headers into dependencies to force the rebuild for 3rd party libraries.

chrox:
Sounds weird. Did you try `make fetchthirdparty`?

Makefile
```diff
@@ -122,18 +131,18 @@ $(AES_LIB):
...
 	$(MAKE) -j$(PROCESSORS) -C mupdf \
 		LDFLAGS="$(LDFLAGS) -L../$(OUTPUT_DIR)" \
-		XCFLAGS="$(CFLAGS) -DNOBUILTINFONT -I../$(JPEG_DIR)/include -I../$(FREETYPE_DIR)/include -I../$(ZLIB_DIR) -I../$(MINIZIP_DIR)" \
+		XCFLAGS="$(CFLAGS) -DNOBUILTINFONT -I../$(TURBOJPEG_DIR)/include -I../$(FREETYPE_DIR)/include -I../$(ZLIB_DIR) -I../$(MINIZIP_DIR)" \
 		CC="$(CC)" \
 		build="release" MUDRAW= MUTOOL= CURL_LIB= \
 		OS=$(if $(WIN32),,Other) verbose=1 \
 		FREETYPE_DIR=nonexisting \
-		JPEG_DIR=nonexisting \
+		TURBOJPEG_DIR=nonexisting \
```
chrox: `JPEG_DIR` is used in the build script of mupdf, `TURBOJPEG_DIR` is not.

Makefile
```diff
@@ -782,4 +795,5 @@ $(OUTPUT_DIR)/spec/base:
 test: $(OUTPUT_DIR)/spec $(OUTPUT_DIR)/.busted
 	cd $(OUTPUT_DIR) && busted -l ./luajit --exclude-tags=notest
 
-.PHONY: test
+.PHONY: test clean fetchthirdparty \
```
chrox: Since there is no file called `clean` `fetchthirdparty`, it's unnecessary to make them phony. Only `test` is necessary because we have a `test` directory.

houqp:
Sorry for the delay, I have been too busy moving recently :(

Yes I tried fetchthridparty couple times already. I think I found the cause. It's likely caused by thirdparty updates. So for example, libfoowrapper depends on libfoo and libfoo has an update. After the `fetchthridparty` update, libfoo will have a new version, but because the `fetch` command never remove the old `libfoo.so` file under OUTPUT dir, make will skip building the new libfoo library, which is required to generate the headers files used by libfoowrapper. And the build of libfoowrapper will fail because of that.

houqp:
@chrox this took longer than I expected. Got stuck in some bugs from CMake. PR #353 should fix this build issue together with others properly.

PR #353, opened 2015-12-21
fix: rebuild thridparty lib on version changes

houqp:
Changes in this PR:
1. switch to CMake for most of the thirdparty libraries
2. fix couple build issues for android and glib (pulled in libffi to fix the broken build)
3. fixed couple wrong dependencies (typo) in Makefile
4. bumped lua-serialize

Only tested for emulater, android, kindle-legacy, kindle5, kobo and ubuntu-touch. I don't have window machine, so hope someone can help test this change on windows too.

Some of the highlights from switching to cmake for thirdparty library builds:
1. Better change management. Whenever we bump a version or change compile flags in CMakeList.txt, the library will be rebuilt. Previously, this is not supported, which caused subtle build bugs, see https://github.com/koreader/koreader-base/pull/349.

2. All library built with cmake does out of source build now. I also made sure each arch has its own build dir so all platform builds are completely separated. Once we migrated everything over to cmake, it will vastly speed up the build process 

3. Cleaner source tree. Less scattered git submodules and tar balls. Everything is managed under `thirdparty` directory.

4. Faster clean up. Only need to run `rm -rf BUILD_DIR` to remove a build and start with a clean state. No more `make clean`, which is way slower than just removing the whole directory.


houqp:
Let me know what you guys think about this move. I will finish up the migration for the rest of the thirdparty libraries after this PR got merged.

chrox:
Really cool. I love it. :+1

NiLuJe:
Yup, pretty neat!

Also a new way for me to mess up commit hashes when bumping third-party libs :D.

Frenzie:
That's really cool indeed. :-)

PR #356, opened 2015-12-29
fix mupdf build and bump to latest release

houqp:


houqp:
@chrox , the annotation unit test broke because of the mupdf update. Looking at the sample output, I don't see any annotation on the page. So I don't know if the annotated output generated by the new mupdf is correct or not. Could you take a look? Or could you let me know what should I check on the new output?

chrox:
I'm afraid I won't have spare time to look into this until this weekend. But does it fail to save highlight annotation only on the unit test case or also in regular use of koreader? Anyway, I will merge this for now.

houqp:
I tested on 2col.pdf manually and it was able to save the highlight. I am just curious about the unit test sample output.

Issue #302, opened 2015-02-09
Why is -m32 used for building luajit?


lgeek:
On this line: https://github.com/koreader/koreader-base/blob/master/Makefile#L199

It's x86 specific and has to be removed to do native builds on ARM. Could it be removed and if not what would be the proper way to only enable it on x86_64 hosts?

Thanks

chrox:
According to Luajit [document](http://luajit.org/install.html#cross), it's necessary to add "-m32" flag in `HOST_CC` when cross compiling for 32 bit architecture.

Current `Makefile` is largely used for cross compiling except building the emulator that use SDL to handle input/output. I'm not sure if native compile is even supported without large portion of modification. But is there a particular reason to choose native compiling instead of cross compiling? I would never try to compile anything on a 800 MHz CPU plus 256 MB RAM of my Kindle device.

lgeek:
I can confirm this is the only change required to build koreader-base (and koreader) natively on ARM (for Kobo devices at least).

As to why, it's because I'm building koreader as part of a larger collection of software, some of which is easier to just build natively. I'm using a more powerful machine, not my reader, and the build time is entirely reasonable.

For my own use I can work around this with a local patch. However, since it seems to be the only issue blocking compilation on non-x86 hosts I think it might be worth fixing upstream.

chrox:
Then it seems easy to fix. Probably we can differentiate toolchains with `$(CC) -dumpmachine` (the MACHINE variable defined in Makefile.def) with `findstring` function.

houqp:
++ on what chrox's approach. @lgeek , feel free to send a PR

PR #360, opened 2016-01-01
build lpegn, evernote-sdk-lua and luajit with cmake

houqp:
also made the following changes:
* bumped luajit, tesseract, leptonica and libdjvu for bug fixes
* fixes https://github.com/koreader/koreader-base/issues/302
* split Makefile into third.Makefile and Makfile
* separate crengine build based on different archs


houqp:
If I haven't missed anything, the whole koreader-base build can be done out of source now. All out of source build is separated based on architecture, so it's going to take more storage space on your hard disk, but the the awesome part is we don't need to do `make clean` any more when building for a different architecture. For example, `make && make TARGET=kindle && make TARGET=kobo && make TARGET=android` should succeed without any error ;)

Frenzie:
Great! :-)

PR #367, opened 2016-01-10
build: cross compile kobo and kindle in travis

houqp:


houqp:
scratch this, container based build have all 32bit packages blacklisted, which are needed by luajit.

PR #375, opened 2016-02-02
minor: start SDL2 window at the center of screen

houqp:


Frenzie:
Can't check atm, but what's the previous behavior?

chrox:
Previously on GNOME 3 the SDL window will be placed at the last position where the window was closed. I don't know if centering the window on the screen each time koreader is launched is a good idea.

Frenzie:
Opening at last location sounds better to me.

houqp:
Ha, I see. My KDE4 has different behaviour, which it always start at the same location and covers the log in my terminal. It a little bit annoying that I have to manually move it away every time. I will close this PR for now.

PR #381, opened 2016-02-20
bump libpng for security bug fixes

houqp:


houqp:
Hmm... travis build is broken right now because of automake. Please don't merge it until I find a away to fix it.

thirdparty/libpng/CMakeLists.txt
```diff
@@ -16,14 +17,19 @@ set(CFG_ENV_VAR "CC=\"${CC}\" CXX=\"${CXX}\" CPPFLAGS=\"${CPPFLAGS}\" LDFLAGS=\"
...
 set(CFG_CMD_STR "${CFG_ENV_VAR} ${SOURCE_DIR}/configure ${CFG_OPTS}")
 
+ko_write_gitclone_script(
+    GIT_CLONE_SCRIPT_FILENAME
+    https://github.com/glennrp/libpng.git
+    v1.6.21
+    ${SOURCE_DIR}
+)
+
 set(LIBPNG_VER "1.6.19")
```
apletnev: remove it, or replace to 1.6.21

houqp:
OK... after countless retires, I finally got the travis build into a good shape... This should be ready to merge now.

Note that I switched back from docker build (sudo: false) to vm build (sudo: true). VM build will take roughly 20 more seconds to start up, but since we can now use docker command in the build, we can vastly speed up the build prepare process by using prebuilt docker images like https://github.com/koreader/virdevenv/tree/master/docker/ubuntu/baseimage. In the end, this should help us reduce the total build time. I will work on that later.

PR #382, opened 2016-02-20
Add dirname and basename function in util.lua

Hzj-jie:
This change is part of a new feature proposal, discussed @ https://github.com/koreader/koreader/issues/1838


houqp:
Thanks for the patch! Could you squash the commits into one to keep a cleaner history?

Hzj-jie:
Definitely.

Hzj-jie:
Will PR with one commit.

PR #384, opened 2016-02-26
build: fix android lpeg build

houqp:
luajit wasn't linked properly with lpeg, which causes crashes in andoird.

also clean up rpath for couple libraries.


thirdparty/lpeg/CMakeLists.txt
```diff
@@ -9,9 +9,10 @@ enable_language(C)
 assert_var_defined(CC)
 assert_var_defined(DYNLIB_CFLAGS)
 assert_var_defined(LUA_DIR)
-assert_var_defined(LUAJIT_LIB)
 
-set(BUILD_CMD_STR "${CC} ${DYNLIB_CFLAGS} -I${LUA_DIR}/src ${LUAJIT_LIB} -o lpeg.so lpcap.c lpcode.c lpprint.c lptree.c lpvm.c")
+# assuming lpeg.so is installed under $(OUTPUT_DIR)/rocks/lib/lua/5.1 and
+# libluajit.so is in $(OUTPUT_DIR)/libs
+set(BUILD_CMD_STR "${CC} ${DYNLIB_CFLAGS} -I${LUA_DIR}/src -Wl,-rpath,'../../../../libs' -lluajit -o lpeg.so lpcap.c lpcode.c lpprint.c lptree.c lpvm.c")
```
chrox: The linker option -rpath specifies run time path to look up shared objects and if it's a relative path it's relative to current working directory so "-Wl,-rpath,'libs'" should be OK.

houqp: Good point. Looks like the best practice is to use `$ORIGIN/libs`. I have no idea how this change fixed my android build then :-/

I did a quick audit of all libraries shipped in libs, many of the libraries don't even have `libs` in their rpath, for example, libmupdf.so only has `$ORIGIN` in its rpath. So how does it find our own libz, libjpeg and libfreetype builds?

chrox: The `$ORIGIN` path directs `ld` to look up linked shared objects just in the directory of the original binary file. In our case, `ld` will look for `libz` and `libjpeg` in the same directory as `libmupdf` and it will find them in the `libs`.

Aha, I see. We are talking about the Android platform. I can see that the `lpeg.so` is in `/data/data/org.koreader.launcher/files/rocks/lib/lua/5.1/lpeg.so` from which `../../../../libs` is  just `/data/data/org.koreader.launcher/files/libs` and the `libluajit.so` is right there.

But the recursive dependency loading function of shared libraries does differ a lot with `ld` on other Linux distributions. @hwhw implemented it here[https://github.com/koreader/android-luajit-launcher/blob/master/assets/dl.lua#L73]. From what I understand it doesn't respect the `rpath` at all. All it needs for a shared library are the dependent library names. It will look up the dependent libraries at [predefined paths](https://github.com/koreader/android-luajit-launcher/blob/master/assets/android.lua#L1350-L1353). And if an `so` file is found it will use `bionic`(Android's glibc)'s `dlopen` function to load it.

My explanation for your observation is that on your Android somehow hwhw's `dl` module cannot find `libluajit.so` while the `bionic` can find it with `rpath` in an extraordinary way.

NiLuJe:
On Linux at least, dlopen does honor RPATH/RUNPATH.

There are also various quirks with $ORIGIN that I can't quite remember (besides old crashers in some linkers).

thirdparty/lpeg/CMakeLists.txt
NiLuJe: On Linux at least, dlopen does honor RPATH/RUNPATH.

There are also various quirks with $ORIGIN that I can't quite remember (besides old crashers in some linkers).

chrox: @NiLuJe Yes, but it seems `bionic` never respects the `rpath` at least in my Android devices. That's why I feel confused with houqp's observation.

NiLuJe: Yeah, a quick Google check and a cursory glance at the [code](/android/platform_bionic/blob/master/linker/linker.cpp) would indeed seem to confirm that...

NiLuJe: I'm not quite following everything, but it would appear some some bionic versions [do seem to handle DT_RUNPATH entries](/android/platform_bionic/blob/master/linker/linker.cpp#L1542)... :?

NiLuJe: Woops, scratch that, that's not in *any* currently released Android version.

Yay. -_-"

NiLuJe: cf; https://github.com/android/platform_bionic/commit/6865082ca6d887766c6651ed2d079dca56a99d32

&

https://github.com/android/platform_bionic/commit/a1feb117e4d5614548574f28dede3443e073512b

which makes it work for dlopen().

But that's apparently targeting Android M, so, doesn't really help us...

houqp: Right... I forgot we have our own custom loader.. I can confirm that `-lluajit` is what actually fixes the crash. Without this, the custom loader will skip `libluajit.so`.

I was under the impression that RPATH is the path relative to the binary that first ran by user (in our case luajit). So looks like I am wrong and it should be the path for the dynamic libary that passed in to dlopen. That means for all libraries in `libs` dir, the correct RPATH should be `$ORIGIN` instead of `$ORIGIN/libs`?

I am confused because some of our RPATH are set to `libs` like filemq, libzyre etc. Shouldn't they be set to `$ORIGIN` too?

chrox: Both $ORIGIN and libs are OK because the "origin" path is actually the "libs" directory for those libraries.

houqp: I see. According to http://stackoverflow.com/questions/24573732/difference-between-relative-path-and-using-origin-as-rpath, current working path will be used as the base dir for relative path like `libs`, which in our case will be root of koreader directory.

So I think for us, the proper setup should be using `$ORIGIN` for all libraries in **koreader/libs** dir and `libs` for libraries that have deeper location like **koreader/common**. Thoughts?

NiLuJe: In a perfect world, I'd avoid both relative paths and $ORIGIN in rpath and stick to absolute paths, but, clearly, in our case, that's not practical ;).

So, yeah, $ORIGIN is potentially slightly less random. Just have to make sure it doesn't get mangled by any buildscripts/tools and/or expanded by a shell somewhere along the line... ;).

houqp: Yeah, took me awhile to get it work last nigh. The call chain we have in the current build system setup is painful to work with, so many layers of escape :-/

houqp:
OK, most of the RPATHs are based on `$ORIGIN` now:

```
koreader-emulator-x86_64-linux-gnu/koreader/common/libluacompat52.so
  RPATH                $ORIGIN
koreader-emulator-x86_64-linux-gnu/koreader/common/libtffi_wrap.so
  RPATH                $ORIGIN/../libs
koreader-emulator-x86_64-linux-gnu/koreader/common/serialize.so
  RPATH                $ORIGIN
koreader-emulator-x86_64-linux-gnu/koreader/common/ssl.so
  RPATH                $ORIGIN/../libs
koreader-emulator-x86_64-linux-gnu/koreader/libs/libcrengine.so
  RPATH                $ORIGIN
koreader-emulator-x86_64-linux-gnu/koreader/libs/libcrypto.so.1.0.0
  RPATH                $ORIGIN
koreader-emulator-x86_64-linux-gnu/koreader/libs/libczmq.so.1
  RPATH                $ORIGIN
koreader-emulator-x86_64-linux-gnu/koreader/libs/libdjvulibre.so.21
koreader-emulator-x86_64-linux-gnu/koreader/libs/libfmq.so.1
  RPATH                $ORIGIN
koreader-emulator-x86_64-linux-gnu/koreader/libs/libfreetype.so.6
koreader-emulator-x86_64-linux-gnu/koreader/libs/libgif.so.7
koreader-emulator-x86_64-linux-gnu/koreader/libs/libjpeg.so.8
koreader-emulator-x86_64-linux-gnu/koreader/libs/libk2pdfopt.so.2
  RPATH                libs
koreader-emulator-x86_64-linux-gnu/koreader/libs/libkoreader-cre.so
  RPATH                $ORIGIN
koreader-emulator-x86_64-linux-gnu/koreader/libs/libkoreader-djvu.so
  RPATH                $ORIGIN
koreader-emulator-x86_64-linux-gnu/koreader/libs/libkoreader-lfs.so
  RPATH                $ORIGIN
koreader-emulator-x86_64-linux-gnu/koreader/libs/liblept.so.3
  RPATH                /home/dave/Documents/code/kindle/koreader/base/thirdparty/libpng/build/x86_64-linux-gnu/libpng-prefix/src/libpng-build/lib:libs
koreader-emulator-x86_64-linux-gnu/koreader/libs/liblodepng.so
koreader-emulator-x86_64-linux-gnu/koreader/libs/libmupdf.so
  RPATH                $ORIGIN
koreader-emulator-x86_64-linux-gnu/koreader/libs/libpng16.so.16
  RPATH                $ORIGIN
koreader-emulator-x86_64-linux-gnu/koreader/libs/libssl.so.1.0.0
  RPATH                $ORIGIN
koreader-emulator-x86_64-linux-gnu/koreader/libs/libtesseract.so.3
  RPATH                /home/dave/Documents/code/kindle/koreader/base/thirdparty/libpng/build/x86_64-linux-gnu/libpng-prefix/src/libpng-build/lib:/home/dave/Documents/code/kindle/koreader/base/thirdparty/leptonica/build/x86_64-linux-gnu/leptonica-prefix/src/leptonica/src/.libs:/home/dave/Documents/code/kindle/koreader/base/thirdparty/leptonica/build/x86_64-linux-gnu/leptonica-prefix/src/leptonica/lib:$ORIGIN
koreader-emulator-x86_64-linux-gnu/koreader/libs/libturbojpeg.so
koreader-emulator-x86_64-linux-gnu/koreader/libs/libwrap-mupdf.so
  RPATH                $ORIGIN
koreader-emulator-x86_64-linux-gnu/koreader/libs/libzmq.so.4
koreader-emulator-x86_64-linux-gnu/koreader/libs/libz.so.1
koreader-emulator-x86_64-linux-gnu/koreader/libs/libzyre.so.1
  RPATH                $ORIGIN
```

I left out k2pdfopt since it also manages tesseract and lept, so it's easier to change the flags in its own repo.

PR #385, opened 2016-02-27
Remove legacy functions and variables in kobolight.lua

Hzj-jie:
After pending PR https://github.com/koreader/koreader/pull/1854 in koreader, we do not need these functions and variables in kobolight anymore.


chrox:
Great work. Thanks. :)

PR #387, opened 2016-02-29
Minor third-party tweaks

NiLuJe:
* Bump FT & LuaJIT
* Fix glib build on most envs (cf. [#1832](koreader/issues/1832))


NiLuJe:
On a sidenote, why the (out of date) FT git mirror?

houqp:
I think the mirror is out of date because gitlab is kinda broken... Hopefully it can fix it by itself.
We have mirror for couple of the projects that have less reliable git setup. We have been getting network failure in travis build for zsync/mupdf/freetype couple times. So I switched them to mirror. That way we also reduce the load on their git server ;)

NiLuJe:
Yeah, savannah was wonky a few weeks ago, I remember :).

PR #389, opened 2016-03-03
Minor buildsystem fixes

NiLuJe:
* Fix zmq to link to libstdc++ statically again
* Bump LuaJIT & OpenSSL (1.0.2 is still a no-go w/ Clang. There's a patch available in ChromeOS, but the current trunk is also OK, so, meh, let's wait for that to hit a release).


houqp:
Weird, travis did not kick off the build even after 30min >_<

PR #396, opened 2016-03-21
Add util.getsec shortcut to get current time in seconds

Hzj-jie:
This is part of to add an auto-suspend function for Kobo, which does not have power management function in kernel.
Discussed in #1356 #1565 & #1814.

### DIFF
--- a/ffi/util.lua
+++ b/ffi/util.lua
@@ -376,4 +376,10 @@ function util.unichar (value)
     end
 end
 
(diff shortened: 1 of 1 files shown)


Frenzie:
@houqp I thought you had objections to the specifics?

houqp:
Yeah, I merged this PR couple hours before I came up with a better way to get the second in that koreader PR though ;)

PR #398, opened 2016-03-23
Revert last change to add util.getsec()

Hzj-jie:
Sorry for the trouble, I should learn more about lua before making any changes.


houqp:
No worry, even I did not think of it at first glance ;)

PR #401, opened 2016-04-09
A bunch of build fixes

NiLuJe:
Plus a few third party bumps.

The luarocks thing might need some review. It works for me w/ luarocks 2.3.0, but then my Lua is installed in a sane and default path...


houqp:
Nice! For lua-spore, have you tried setting INCDIR to `"$(LUAJIT_DIR)/src"`? That's what's used in luasocket's build. Do we what it is looking for in LIBDIR? I wonder if it's just libluajit.so.

NiLuJe:
Ah, great, thanks ;p.

I was too tired last night to actually look for the right variable names that point around there...
I'll check it out.

And, yeah, it's probably -ljit, since it defaulted to /usr/lib64 over here... ;).

PR #407, opened 2016-04-18
fix mtime check for luasec and glib

houqp:


Makefile.third
```diff
@@ -223,7 +223,9 @@ $(LIBFFI_DIR)/include: $(THIRDPARTY_DIR)/libffi/CMakeLists.txt
 
 $(GLIB): $(LIBFFI_DIR)/include $(THIRDPARTY_DIR)/glib/CMakeLists.txt
 	install -d $(GLIB_BUILD_DIR)
-	-rm -f $(GLIB_DIR)/../glib-stamp/glib-install
+	# make install sometimes only update symlink, so purge lib dir here
+	# to update mtime for all the binaries
+	-rm -rf $(GLIB_DIR)/lib $(GLIB_DIR)/../glib-stamp/glib-install
```
Frenzie: Would it be faster to touch -m or does it not really matter?

houqp: It would be faster, but I am not sure how this would affect the install command. i.e. whether it will make the install step skip any file. Thus I use rm to make sure all files are fresh installs.

PR #408, opened 2016-04-19
bump crengine for segfault fix

houqp:
This PR relies on https://github.com/koreader/crengine/pull/41


houqp:
@Frenzie you missed the crengine PR: https://github.com/koreader/crengine/pull/41

PR #410, opened 2016-04-20
build(fix): use gitlab mirror for mupdf build

houqp:
The official git repo is breaking our ci tests...


houqp:
sigh... the mupdf's submodules are still using their own git server :(

houqp:
OK, I have asked upstream to fix the git server. But still won't hurt to clone the main repo from our own mirror ;)

PR #413, opened 2016-04-24
cmake: always force update remote incase of upstream repo url change

houqp:


thirdparty/cmake_modules/koreader_thirdparty_git.cmake
```diff
@@ -76,6 +76,14 @@ if(NOT \"\${curr_tag}\" STREQUAL \"${git_tag}\")
   )
   if(error_code)
     execute_process(
+      COMMAND \"${git_EXECUTABLE}\" remote rm origin
```
Frenzie: Is there a specific reason you're telling it to `rm` and `add` instead of just `fetch origin`?

PR #414, opened 2016-04-25
use native coordinates in wrectmap_native_inside

jedick:


houqp:
Thanks for fixing the bug! I am actually not very familiar with koptcontext code path. @chrox , can you take a look?

jedick:
This is also my first time with the code; sorry if I did something wrong! I tried a simple test, using debug statements to print the solved-for index (m) for the rectmap in each of the pair of functions (reflowToNativePosTransform and nativeToReflowPosTransform). Watching the values in the emulator when selecting, then highlighting a word in a reflowed PDF, I could see the value of m for the highlight (nativeToReflowPosTransform) was not matching the selection (reflowToNativePosTransform). On the PDFs I've tested (e.g. the one in the issue report) this change makes the values match, and the highlights placed better. For some reason the bug doesn't affect all PDFs, like the "Metropolis" article in the test directory.

chrox:
Looks good to me. Thanks for the patch.

PR #420, opened 2016-05-09
Bump k2pdfopt to v2.34, tesseract-ocr to 3.04 and leptonica to v1.72

chrox:


houqp:
Not entirely sure why the clang build is failing. Is this expected?

```patching file api/baseapi.cpp
Reversed (or previously applied) patch detected!  Skipping patch.
2 out of 2 hunks ignored -- saving rejects to file api/baseapi.cpp.rej
```

chrox:
Yeah, we silenced the `tesseract` build log to save limited output space on Travis CI. The problem is as follows: 

`libtool: compile:  /usr/bin/ccache clang++ -DHAVE_CONFIG_H -I. -I.. -O2 -DNDEBUG -I../cutil -I../ccutil -I../ccstruct -I../dict -I../viewer -I/home/chrox/dev/koreader/base/thirdparty/leptonica/build/x86_64-pc-linux-gnu/leptonica-prefix/src/leptonica/src/. -pthread -I/usr/include/pango-1.0 -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/cairo -I/usr/include/glib-2.0 -I/usr/lib/x86_64-linux-gnu/glib-2.0/include -I/usr/include/pixman-1 -I/usr/include/freetype2 -I/usr/include/libpng12 -O2 -ffast-math -pipe -fomit-frame-pointer -D__extern_always_inline=inline -Qunused-arguments -march=native -fPIC -O3 -MT shapeclassifier.lo -MD -MP -MF .deps/shapeclassifier.Tpo -c shapeclassifier.cpp  -fPIC -DPIC -o .libs/shapeclassifier.o
shapeclassifier.cpp:182:33: error: cannot pass object of non-POD type 'tesseract::ScoredFont' through variadic function; call will abort at runtime [-Wnon-pod-varargs]
        tprintf_internal(" %d", results[i].fonts[f]);
                                ^
1 error generated.`

PR #423, opened 2016-05-16
Really fix #421

### LINKED ISSUE #421: can not open pdf
get this error on my kindle and there is no problem on emulator
`# cannot open document /mnt/us/documents/1.pdf ./ffi/koptcontext.lua:13 /lib/libc.so.6 version 'GLIBC_2.22' not found (required by libs/liblept.so.3)`
compiled with GCC 5.3 shipped with ubuntu 16.04, and I dont know if this is the reason

NiLuJe:
Because my original fix broke builds on glibc < 2.22 (i.e., basically all we actually care about :D).


NiLuJe:
Note that this adds a dependency on coreutils >= 7 on Linux.
Thankfully, even Debian oldstable is using a decent version of that, so that shouldn't be an issue.
And, as the name might imply, it should be part of the base system of all desktop/server distro.

PR #424, opened 2016-05-18
fix highlight problem

frankyifei:
first page highlight.
except use "break", is there a better method?


houqp:
@frankyifei , any reason why you are closing this one?

frankyifei:
I should use `continue` instead of `break`. will create another pr soon

PR #425, opened 2016-05-24
fix highlight position

frankyifei:
ldomXpointer::getRectEX will return the width of the word, so it only runs one time for each part of the hyphenated word. It also considers paddings of the block when calculates position.
the build test will not pass until my commit for crengine gets merged.

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -1108,13 +1108,13 @@ static int getWordBoxesFromPositions(lua_State *L) {
 				lastx = wordRect.left;
 			} else {  // word is hyphenated
 				ldomWord word = words[i];
-				int y=-1;
+				int y = -1;
 				for (int j=word.getStart(); j < word.getEnd(); j++) {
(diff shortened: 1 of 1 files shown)


cre.cpp
```diff
@@ -1107,9 +1108,13 @@ static int getWordBoxesFromPositions(lua_State *L) {
 				lastx = wordRect.left;
 			} else {  // word is hyphenated
 				ldomWord word = words[i];
+				int y=-1;
 				for (int j=word.getStart(); j < word.getEnd(); j++) {
-					if (ldomXPointer(word.getNode(), j).getRect(charRect)) {
-						docToWindowRect(tv, charRect);
+					if (ldomXPointer(word.getNode(), j).getRectEX(charRect)) {
+						if(!docToWindowRect(tv,charRect)) continue;
```
houqp: Could you keep the coding style consistent with the rest of the code in this file? There are couple places that are missing spaces ;)

houqp:
Thanks for the fix. Looks good to me.

PR #426, opened 2016-05-25
fix highlight position

frankyifei:


chrox:
Could you rebase it on the master branch? The CI is fixed there.

PR #433, opened 2016-05-30
Fix build break of gettext library with android NDK 11c

Hzj-jie:
This issue has been discussed at koreader issue 2043 https://github.com/koreader/koreader/issues/2043

This change adds libc in gettext ld flags.

### DIFF
--- a/thirdparty/gettext/CMakeLists.txt
+++ b/thirdparty/gettext/CMakeLists.txt
@@ -13,6 +13,9 @@ ep_get_source_dir(SOURCE_DIR)
 ep_get_binary_dir(BINARY_DIR)
 
 set(CFG_ENV_VAR "CC=\"${CC}\"")
+if(${IS_ANDROID})
+    set(CFG_ENV_VAR "${CFG_ENV_VAR} && LDFLAGS=-lc")
+endif()
 set(CFG_OPTS "--with-threads=none --prefix=${BINARY_DIR} --with-libiconv-prefix=${LIBICONV_PREFIX} --enable-shared=false --enable-static=true ${CHOST_OPTS}")
 set(CFG_CMD sh -c "${CFG_ENV_VAR} ${SOURCE_DIR}/configure ${CFG_OPTS}")
 # remove HAVE_PWD_H for android build
--- a/thirdparty/tesseract/CMakeLists.txt
+++ b/thirdparty/tesseract/CMakeLists.txt
@@ -8,9 +8,9 @@ include("koreader_thirdparty_git")
 ep_get_source_dir(SOURCE_DIR)
 
 if ("$ENV{ANDROID}" STREQUAL "1")
-    set(PATCH_CMD sh -c "patch -N -p1 < ${CMAKE_CURRENT_SOURCE_DIR}/tesseract-android.patch")
+    set(PATCH_CMD COMMAND sh -c "patch -N -p1 < ${CMAKE_CURRENT_SOURCE_DIR}/tesseract-android.patch")
 else()
-    set(PATCH_CMD echo)
+    set(PATCH_CMD "")
 endif()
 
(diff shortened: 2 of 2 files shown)


thirdparty/tesseract/CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ ep_get_source_dir(SOURCE_DIR)
 if ("$ENV{ANDROID}" STREQUAL "1")
     set(PATCH_CMD sh -c "patch -N -p1 < ${CMAKE_CURRENT_SOURCE_DIR}/tesseract-android.patch")
```
houqp: You need to move `COMMAND` into PATCH_CMD. Otherwise, cmake will complain on argument not found for `COMMAND`.

Hzj-jie: Why is CMake changed significantly from 2.x to 3.x?

houqp: Haha, because it's a major version bump. Technically, they can introduce any backward incompatible change they want in a new major version :P

thirdparty/gettext/CMakeLists.txt
```diff
@@ -13,6 +13,9 @@ ep_get_source_dir(SOURCE_DIR)
 ep_get_binary_dir(BINARY_DIR)
 
 set(CFG_ENV_VAR "CC=\"${CC}\"")
+if(${IS_ANDROID})
+    set(CFG_ENV_VAR "${CFG_ENV_VAR} && LDFLAGS=-lc")
```
chrox: This breaks Android build as now on Android the `CC` variable won't be `${CC}` passed into CMake but raw default `gcc` instead.

```
set(CFG_ENV_VAR "${CFG_ENV_VAR} LDFLAGS=-lc")
```
should be OK.

chrox:
Without `CC=${CC}` gettext will be configured this:
```
=== configuring in gettext-runtime (/home/chrox/dev/koreader/base/thirdparty/gettext/build/arm-linux-androideabi/gettext-prefix/src/gettext-build/gettext-runtime)
configure: running /bin/bash /home/chrox/dev/koreader/base/thirdparty/gettext/build/arm-linux-androideabi/gettext-prefix/src/gettext/gettext-runtime/configure --disable-option-checking '--prefix=/home/chrox/dev/koreader/base/thirdparty/gettext/build/arm-linux-androideabi/gettext-prefix/src/gettext-build'  '--with-threads=none' '--with-libiconv-prefix=/home/chrox/dev/koreader/base/thirdparty/libiconv/build/arm-linux-androideabi/libiconv-prefix/src/libiconv-build' '--enable-shared=false' '--enable-static=true' '--host=arm-linux' 'host_alias=arm-linux' 'LDFLAGS=-lc' --cache-file=/dev/null --srcdir=/home/chrox/dev/koreader/base/thirdparty/gettext/build/arm-linux-androideabi/gettext-prefix/src/gettext/gettext-runtime
```

And `libintl.h` won't be installed:
```
koreader $ find . -name libintl.h
./base/toolchain/pocketbook-toolchain/arm-obreey-linux-gnueabi/sysroot/usr/include/libintl.h
./base/toolchain/pocketbook-toolchain/arm-obreey-linux-gnueabi/sysroot/usr/include_xtools/libintl.h
```
which causes errors when configuring `glib`:
```
checking libintl.h presence... no
checking for libintl.h... no
configure: error: 
*** You must have either have gettext support in your C library, or use the
*** GNU gettext library. (http://www.gnu.org/software/gettext/gettext.html

CMakeFiles/glib.dir/build.make:101: recipe for target 'glib-prefix/src/glib-stamp/glib-configure' failed
make[4]: *** [glib-prefix/src/glib-stamp/glib-configure] Error 1
```

PR #439, opened 2016-06-30
fix position independent executables (PIE) error on Android L

chrox:
Running a non-PIE binary on Android L will cause an error:
```
Error: only position independent executables (PIE) are supported
```
But if we link the binary with `-static` option it will have no dynamic section thus being PIE. Tested on MX4 Android 5.1.


houqp:
nice, how is the pass rate after this patch?

chrox:
I have not tested it yet, but it should have no difference as the PIE error only affects binary running like `sdcv`, `tar` and `zsync` in koreader.

PR #443, opened 2016-07-11
pull in lj-wpaclient library

houqp:
* get ready for the new network manager
* fixed a bug that can cause touch screen to lock up after a reader crash


NiLuJe:
Is CLOEXEC supported on legacy Kindle devices?

I seem to recall they're running Linux 2.6.22 or something, and the CLOEXEC stuff was added around that time (either 2.6.23 or 2.6.24).
AFAICT, we should be okay, but you never know... ;o).

Frenzie:
New network manager? Amazing. :)

houqp:
I was looking for a way to close all the inherited file descriptors in the child process, but was not able to find any. If you know any trick, please let me know!

PR #444, opened 2016-07-13
build(fix): use gitlab mirror for djvulibre

houqp:


houqp:
sigh, looks like sourceforge is partially down...

PR #457, opened 2016-09-20
Add missing dash

tobias-93:
In Makefile.defs, a dash is missing in the ARMV6_GENERIC_ARCH variable, making the build process stop working.


houqp:
Thank you for the fix :)

PR #461, opened 2016-10-12
Switch order in patching Zsync

tobias-93:
When building the Zsync package, first the configure script is generated and then the source is patched to work on Android.
Due to this, the patching doesn't have the wanted effect and the build fails.
If the patch is done first, and then the configure script is generated, it will work perfectly.


houqp:
Good catch, but looks like now reconfigure command got skipped. Could you change it to: `PATCH_COMMAND COMMAND ${PATCH_CMD1} COMMAND ${PATCH_CMD2} COMMAND ${RECONF_CMD}`?

tobias-93:
@houqp done

houqp:
Thanks!

PR #462, opened 2016-10-23
generate usb hotplug event for kobo

houqp:
By leveraging the netlink uevent message, we should now be able to use the same code for device hotplug detection for all Linux based platforms including kindle. I am only using this approach to handle usb hotplug in kobo for now since I don't want to break the kindle code by accident ;)


houqp:
Oh god, finally fixed the nasty build issue :)

houqp:
merge this for now since it's been open for a week :)

PR #463, opened 2016-11-05
fix: only build popen for kindle & bump lj-wpaclient for bug fix

houqp:
A bug in lj-wpaclient causing it to identify the wrong network as the currently connected one.


houqp:
merging this to fix the current koreader build

PR #468, opened 2016-12-04
Kobo input simulate

wesley6j:
fake tap input that works on kobo

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/468)

### DIFF
--- a/input/input.c
+++ b/input/input.c
@@ -39,6 +39,13 @@
 #define CODE_FAKE_CHARGING      10020
 #define CODE_FAKE_NOT_CHARGING  10021
 
+#define SEND(a,b,c){ \
+    gettimeofday(&ev.time, NULL);\
+    ev.type = (a);\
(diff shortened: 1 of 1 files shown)


houqp:
Thanks for the patch!

This is kobo specific code, could you refactor it like what I did with `generateFakeEvent`? i.e. move it into input-kobo.h (which will only be included for kobo builds) and rename it to `fakeTapInput` so all platforms can have the same interface.

You also need to move the current fakeTapInput function into another header file. You can name it `input-generic.h`, then include it in an `#else` macro branch here: https://github.com/koreader/koreader-base/blob/master/input/input.c#L51

houqp:
ping @wesley6j, let me know if there is anything I can help with updating the patch :)

Frenzie:
pinging @wesley6j ^ what @houqp said :-)

PR #470, opened 2017-01-14
Added hasXPointer() and getImageDataFromPosition()

poire-z:
Needed for 2 things i'll submit PR for to the koreader repository in a few days:

`CreDocument:hasXPointer(xp)` : will be used by `ReaderLink:onGotoLink(link)` to check that the link is actually an internal link. Currently, when clicking on an invalid or external link, we jump to page 1 of document.
(Mostly copied & pasted from `getPosFromXPointer()` , which I could have used as it returns 0 if the link is not internal, but it also returns 0 if the link is on the top line of the document.)

`CreDocument:getImageFromPosition(pos)` : will be used by `ReaderHighlight:onHold(arg,ges) `to check if we are holding on an image before calling `getWordFromPosition()`. (Currently, if holding on an image, you get a text with the last word before the image + the first word after the image.) That will allow getting the image data from crengine to create an ImageWidget with it, and display it fullscreen with a new ImageViewer widget.
(Mostly copied & pasted from `getLinkFromPosition()` and `getCoverPageImageData()` .)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -1281,6 +1281,9 @@ static int getCoverPageImageData(lua_State *L) {
 		unsigned size = stream->GetSize();
 		lvsize_t read_size = 0;
 		void *buffer = (void *)malloc(size);
+		/* This malloc'ed buffer NEEDs to be freed from lua after use with :
+		 *     ffi.C.free(data)
+		 * to not leak memory */
 		if (buffer != NULL) {
 			stream->Read(buffer, size, &read_size);
 			if (read_size == size) {
@@ -1392,6 +1395,49 @@ static int isXPointerInCurrentPage(lua_State *L) {
 	return 1;
 }
 
+static int isXPointerInDocument(lua_State *L) {
+    CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+    const char *xpointer_str = luaL_checkstring(L, 2);
+
+    bool found = true;
+    ldomXPointer xp = doc->dom_doc->createXPointer(lString16(xpointer_str));
+    lvPoint pt = xp.toPoint();
+    if (pt.y < 0) {
+        found = false;
+    }
+    lua_pushboolean(L, found);
+    return 1;
+}
+
+static int getImageDataFromPosition(lua_State *L) {
+    CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+    int x = luaL_checkint(L, 2);
+    int y = luaL_checkint(L, 3);
+    lvPoint pt(x, y);
+    ldomXPointer ptr = doc->text_view->getNodeByPoint(pt);
+    if (ptr.isNull())
+        return 0;
+    LVStreamRef stream = ptr.getNode()->getObjectImageStream();
+    if (!stream.isNull()) {
+        unsigned size = stream->GetSize();
+        lvsize_t read_size = 0;
+        void *buffer = (void *)malloc(size);
+        /* This malloc'ed buffer NEEDs to be freed from lua after use with :
+         *     ffi.C.free(data)
+         * to not leak memory */
+        if (buffer != NULL) {
+            stream->Read(buffer, size, &read_size);
+            if (read_size == size) {
+                lua_pushlightuserdata(L, buffer);
+                lua_pushinteger(L, size);
+                return 2;
+            }
+        }
+    }
+    return 0;
+}
+
+
 static const struct luaL_Reg cre_func[] = {
 	{"initCache", initCache},
 	{"initHyphDict", initHyphDict},
@@ -1453,10 +1499,12 @@ static const struct luaL_Reg credocument_meth[] = {
 	//{"drawCoverPage", drawCoverPage},
 	{"findText", findText},
 	{"isXPointerInCurrentPage", isXPointerInCurrentPage},
+	{"isXPointerInDocument", isXPointerInDocument},
 	{"getLinkFromPosition", getLinkFromPosition},
 	{"getWordFromPosition", getWordFromPosition},
 	{"getTextFromPositions", getTextFromPositions},
 	{"getWordBoxesFromPositions", getWordBoxesFromPositions},
+	{"getImageDataFromPosition", getImageDataFromPosition},
 	{"getPageLinks", getPageLinks},
 	{"getCoverPageImageData", getCoverPageImageData},
 	{"gotoLink", gotoLink},


cre.cpp
```diff
@@ -1392,6 +1392,48 @@ static int isXPointerInCurrentPage(lua_State *L) {
...
+    int y = luaL_checkint(L, 3);
+    lvPoint pt(x, y);
+    ldomXPointer ptr = doc->text_view->getNodeByPoint(pt);
+    if (ptr.isNull())
+        return 0;
+    LVStreamRef stream = ptr.getNode()->getObjectImageStream();
+    if (!stream.isNull()) {
+        unsigned size = stream->GetSize();
+        lvsize_t read_size = 0;
+        void *buffer = (void *)malloc(size);
```
houqp: It's generally not a good practice to mix LUA C binding with LUAJIT FFI. We had to use C binding for crengine because LUAJIT does not play well with C++.

The proper way to do memory management in LUA C binding is to use metatable with userdata. lightuserdata is mostly used when you need to do equality comparison, see: https://www.lua.org/pil/28.5.html.

Basically you will need to create a new metatable for the image stream in luaopen_cre. Then sets the `__gc` method in the metatable. Then set this metatable to your image stream userdata in getImageDataFromPosition. You can find an example in newDocView where it does the same thing for CreDocument. We set both `__gc` and `close` methods for CreDocument to the same function so it can be freed either manually or automatically by LuaJIT's GC.

houqp:
What's your thought about changing getPosFromXPointer() to make it return negative pos too? It's an internal API so should be safe to change. This way, we don't have to do the same set of crengine APIs twice.

houqp:
I like the ImageViewer widget idea!

poire-z:
> What's your thought on changing getPosFromXPointer() to make it return negative pos too? It's an internal API so should be safe to change. This way, we don't have to do the same set of crengine APIs twice.

We can do that, but it will just avoid having a new function in cre.cpp.
If you meant not doing twice the API calls, it would make things ugly if you want us to pass the y position all along. Currently:
```
ReaderLink:onGotoLink
 if self.ui.document:hasXPointer(link) then
   self.ui:handleEvent(Event:new("GotoXPointer", link))

function ReaderRolling:_gotoXPointer(xpointer)
    if self.view.view_mode == "page" then
        self:_gotoPage(self.ui.document:getPageFromXPointer(xpointer))
    else
        self:_gotoPos(self.ui.document:getPosFromXPointer(xpointer))
    end
```

cre.cpp
poire-z: Well, my first attempt code was simpler:
```
 /* This version would not need that and buffer would be freed by lua GC */
 void *buffer = lua_newuserdata(L, size);
 if (buffer != NULL) {
     stream->Read(buffer, size, &read_size);
     if (read_size == size) {
         lua_pushinteger(L, size);
         return 2;
     }
 }
```
Without any metatable stuff. I tested that (1000 calls from lua, no memory leak). Does that look fine to you ?
But as i'm not really at ease with C, I thought i'll do just like it is done in getCoverPageImageData() (with lightuserdata and ffi.C.free in lua code), to be conservative and avoid discussion on a topic i'm not confortable with :)
You tell me.

Btw, when reading the old issues about BlitBuffer leaks, I got the impression LUAJIT GC couldn't be trusted with them, and that is why we have to add explicite _bb:free() like i'm doing in ImageWidget.
(Or is there some C++ vs C differences here ?)

houqp: > Well, my first attempt code was simpler:

It's simpler to on the C side (callee), but more complicate/hacky on the Lua side (caller) ;)

> Without any metatable stuff. I tested that (1000 calls from lua, no memory leak). Does that look fine to you ?

I am not saying mixing C binding API with LUAJIT FFI won't work, it's just not a good practice.

> But as i'm not really at ease with C, I thought i'll do just like it is done in getCoverPageImageData() (with lightuserdata and ffi.C.free in lua code), to be conservative and avoid discussion on a topic i'm not confortable with :)
You tell me.

Ha, I see. If I was writing this patch, I will convert `getCoverPageImageData` to use C binding API too. But I also don't want to force you to do extra work here. If you want to stick to the current approach, please at least help add the same "caller is responsible for freeing memory" comment to getCoverPageImageData ;)

> Btw, when reading the old issues about BlitBuffer leaks, I got the impression LUAJIT GC couldn't be trusted with them, and that is why we have to add explicite _bb:free() like i'm doing in ImageWidget.

It couldn't be trusted in the sense that we don't know when it will free the memory. It will eventually free the object at some point. This is just the tradeoff you make when using automatic GC. Unfortunately, a lot of the embedded devices are memory constrained, so we don't always have the luxury to wait for a long time.

houqp:
You are right, it's better to validate the link in readerlink before inserting it to location_stack. We can also display visual hint as a feedback.

That said, I still consider it a bug in getPageFromXPointer and getPosFromXPointer. They should not return valid value for invalid/external links. But it's outside of the scope of this PR.

houqp:
`hasXPointer` is probably not an accurate name since it's argument is already the xpointer. Maybe rename it to something like `isInternalXPointer`?

cre.cpp
poire-z: > > Well, my first attempt code was simpler:

> It's simpler to on the C side (callee), but more complicate/hacky on the Lua side (caller) ;)

I meant the sample code in my comment above, using `lua_newuserdata()` but _without your suggesting metatable stuff_ (that I've seen in doc, but that I dont really understand), **does work**, and appears to not need anything from the lua side, and in my tests was freed immediately (or may be that was just cause i was repeating it 1000 times).
So, if you were to confirm this sample code look fine to you, I can replace it, there and in `getCoverPageImageData()` , and remove both `ffi.C.free(data)` from credocument.lua.

But if indeed, there's the chance that it is not gc'ed when it goes out of scope, we might as well, for these huge amount of image data, keep the explicite C.free().

houqp: Oh, sorry, I thought the code in your previous comment was just a repaste from the git diff. 

Yes, any memory allocated by lua_newuserdata will be tracked by LUAJIT VM and GCed when reference reached 0. That's why `closeDocument` only frees up text_view, but not the CreDocument struct.

In fact, you won't be able to manually free memory allocated with lua_newuserdata in LUA. Overloading `__gc` with metatable is the only safe way to make sure we have the freedom to manually free this memory while relying LUAJIT VM GC as a fallback when we choose not to.

I prefer explicit manual memory free in this case because we simply don't have luxury to wait for LUAJIT GC. So unless we go with the metatable route, let's keep it to `ffi.C.free` for now.

poire-z:
Dunno, `isInternalXPointer `could be mean to check if a xpointer starts with "#" or "/body", meaning it's not an external url. Whether it's present or not in the document. `hasXPointerTarget(xp) `? `isXPointerInDocument(xp)` ?

houqp:
I like isXPointerInDocument(xp) better :)

PR #472, opened 2017-01-21
Fix for links clickable out of bounds

poire-z:
See koreader issue #2412
crengine bumped, including fix for segfault when hold on list items.


houqp:
merge merge merge!

poire-z:
I take that as an invitation to bump base in koreader :)

PR #475, opened 2017-02-26
build(fix): ignore unused error for libzmq

houqp:


houqp:
Nop, this is for a new build error that shows up on my ubuntu 16.04. The android build failure from gitlab is likely caused by a bug from their build server because I was not able to reproduce it at all outside of gitlab. It's really strange that even building it with a docker image can still result in intermittent zmq error. I have no idea what's the source of the randomness :(

Frenzie:
Maybe you could investigate bumping it if you have the time?

houqp:
Yup, I am planning to get back to that. I watched the log for days when I was setting everything up initially, but didn't get anywhere. Travis CI uses the same docker image but never runs into this issue. I am actually tempting to set up gitlab CI locally to see if helps reproducing the issue.

PR #479, opened 2017-04-01
More accurate word detection and text selection

poire-z:
See https://github.com/koreader/crengine/pull/69 for details.
Bumped crengine.
Added fix for a correct display of old (before this crengine mod) saved highlights that included punctuation at edges.

Well, I didn't use highlights at all, so I'd be happy if one of you who do, and have such epubs near the emulator, could test on the emulator that everything feels right (old highlights display the same, new hightlights are fine, text selection is indeed better). Thanks !

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -1078,10 +1078,25 @@ static int getWordBoxesFromPositions(lua_State *L) {
 			return 0;
 		r.sort();
 
+	        // Old saved highlights may have included punctuation at
+	        // edges (they were not displayed in boxes because of
+	        // lvtinydom.cpp's ldomWordsCollector catching only letters)
+	        // and punctuation was considered part of the word
+	        // (a space was the sole word separator), so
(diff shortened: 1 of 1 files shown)


houqp:
Thanks! Works as expected during my quick tests :)

Frenzie:
Same here.

poire-z:
Thanks for the tests, and for the merge here and on the koreader side.

Frenzie:
No, thank you for hacking crengine! :-)

On Apr 2, 2017 21:21, "poire-z" <notifications@github.com> wrote:

Thanks for the tests, and for the merge here and on the koreader side.

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/koreader/koreader-base/pull/479#issuecomment-291008402>,
or mute the thread

houqp:
Indeed, thanks @poire-z for the fix!

PR #480, opened 2017-04-01
bump lj-wpaclient for bug fix

houqp:
@lgeek reported that many Linux systems report results message with warning level instead of info.


lgeek:
A bit off topic, but what's the policy on updating the koreader-base submodule in koreader? Should I open a pull request, just push it directly, etc?

Frenzie:
A PR with the submodule bump will do. :-)

PR #481, opened 2017-04-03
Bump czmq

Frenzie:
This will fail but I can't quite figure out the problem.

### DIFF
--- a/thirdparty/czmq/CMakeLists.txt
+++ b/thirdparty/czmq/CMakeLists.txt
@@ -16,9 +16,8 @@ assert_var_defined(HOST)
 ep_get_source_dir(SOURCE_DIR)
 ep_get_binary_dir(BINARY_DIR)
 
-#-Wno-error so we can build on glibc 2.24 in spite of readdir_r
-set(CFG_ENV_VAR "CC=\"${CC}\" LDFLAGS=\"${LDFLAGS}\" CFLAGS=\"-Wno-error ${CFLAGS}\" czmq_have_xmlto=no czmq_have_asciidoc=no")
-set(CFG_OPTS "-q --prefix=${BINARY_DIR} --with-gnu-ld --with-libzmq=${ZMQ_DIR} --disable-static --enable-shared --host=${HOST}")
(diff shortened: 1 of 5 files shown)


thirdparty/czmq/CMakeLists.txt
```diff
@@ -16,9 +16,8 @@ assert_var_defined(HOST)
 ep_get_source_dir(SOURCE_DIR)
 ep_get_binary_dir(BINARY_DIR)
 
-#-Wno-error so we can build on glibc 2.24 in spite of readdir_r
-set(CFG_ENV_VAR "CC=\"${CC}\" LDFLAGS=\"${LDFLAGS}\" CFLAGS=\"-Wno-error ${CFLAGS}\" czmq_have_xmlto=no czmq_have_asciidoc=no")
-set(CFG_OPTS "-q --prefix=${BINARY_DIR} --with-gnu-ld --with-libzmq=${ZMQ_DIR} --disable-static --enable-shared --host=${HOST}")
+set(CFG_ENV_VAR "CC=\"${CC}\" CFLAGS=\"${CFLAGS} -I${ZMQ_DIR}/include\" LDFLAGS=\"${LDFLAGS} -L${ZMQ_DIR}/lib\" czmq_have_xmlto=no czmq_have_asciidoc=no")
```
Frenzie: I got this change from [the czmq readme](https://github.com/zeromq/czmq/blob/eaac91a272779299d0b9ab8ac62c7dee1c34f567/README.md#building-and-installing):
```
export CFLAGS=-I$HOME/local/include
export LDFLAGS=-L$HOME/local/lib64
export PKG_CONFIG_PATH=$HOME/local/lib64/pkgconfig
./configure
```
Oddly enough it works although I'm really just restating what the --with-libzmq flag is supposed to do.

Anyway, the same fix doesn't work on filemq and zyre and I have no idea why it's not picking up on the libraries correctly.

Frenzie:
Manually sticking in headers (like `#include "../include/fmq_server.h"` in filemq_server.c) seems to make it compile, but somehow I think it's more likely that something's up with the "painful" build process (hehehe, thank https://github.com/koreader/koreader/issues/2798 for that phrase) over here than with upstream.

PR #482, opened 2017-04-11
Fix Android build

Frenzie:
Fixes https://github.com/koreader/koreader/issues/2647

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -417,18 +417,18 @@ K2PDFOPT_CFLAGS=-I$(K2PDFOPT_DIR)/willuslib \
 		-I$(K2PDFOPT_DIR)/k2pdfoptlib -I$(K2PDFOPT_DIR) \
 		-I$(K2PDFOPT_DIR)/include_mod
 
-FREETYPE_LIB_EXT=$(if $(WIN32),-6.dll,$(if $(DARWIN),.6.dylib,.so.6))
+FREETYPE_LIB_EXT=$(if $(WIN32),-6.dll,$(if $(DARWIN),.6.dylib,$(if $(ANDROID),.so,.so.6)))
 FREETYPE_LIB=$(OUTPUT_DIR)/libs/libfreetype$(FREETYPE_LIB_EXT)
 FREETYPE_BUILD_DIR=$(THIRDPARTY_DIR)/freetype2/build/$(MACHINE)
 FREETYPE_DIR=$(CURDIR)/$(FREETYPE_BUILD_DIR)/freetype2-prefix/src/freetype2-build
 
-JPEG_LIB_EXT=$(if $(WIN32),-8.dll,$(if $(DARWIN),.8.dylib,.so.8))
+JPEG_LIB_EXT=$(if $(WIN32),-8.dll,$(if $(DARWIN),.8.dylib,$(if $(ANDROID),.so,.so.8)))
 JPEG_LIB=$(OUTPUT_DIR)/libs/libjpeg$(JPEG_LIB_EXT)
 TURBOJPEG_LIB=$(OUTPUT_DIR)/libs/libturbojpeg$(if $(WIN32),.dll,$(if $(DARWIN),.dylib,.so))
 JPEG_BUILD_DIR=$(THIRDPARTY_DIR)/libjpeg-turbo/build/$(MACHINE)
 JPEG_DIR=$(CURDIR)/$(JPEG_BUILD_DIR)/libjpeg-turbo-prefix/src/libjpeg-turbo-build
 
-PNG_LIB_EXT=$(if $(WIN32),-16.dll,$(if $(DARWIN),.16.dylib,.so.16))
+PNG_LIB_EXT=$(if $(WIN32),-16.dll,$(if $(DARWIN),.16.dylib,$(if $(ANDROID),.so,.so.16)))
 PNG_LIB=$(OUTPUT_DIR)/libs/libpng16$(PNG_LIB_EXT)
 PNG_BUILD_DIR=$(THIRDPARTY_DIR)/libpng/build/$(MACHINE)
 PNG_DIR=$(CURDIR)/$(PNG_BUILD_DIR)/libpng-prefix/src/libpng-build
@@ -437,7 +437,7 @@ LODEPNG_LIB=$(OUTPUT_DIR)/libs/liblodepng$(LIB_EXT)
 LODEPNG_BUILD_DIR=$(THIRDPARTY_DIR)/lodepng/build/$(MACHINE)
 LODEPNG_DIR=$(CURDIR)/$(LODEPNG_BUILD_DIR)/lodepng-prefix/src/lodepng
 
-GIF_LIB_EXT=$(if $(WIN32),-7.dll,$(if $(DARWIN),.7.dylib,.so.7))
+GIF_LIB_EXT=$(if $(WIN32),-7.dll,$(if $(DARWIN),.7.dylib,$(if $(ANDROID),.so,.so.7)))
 GIF_LIB=$(OUTPUT_DIR)/libs/libgif$(GIF_LIB_EXT)
 GIF_BUILD_DIR=$(THIRDPARTY_DIR)/giflib/build/$(MACHINE)
 GIF_DIR=$(CURDIR)/$(GIF_BUILD_DIR)/giflib-prefix/src/giflib-build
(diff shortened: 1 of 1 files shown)


houqp:
@Frenzie I assume you have tested this on an android device? What NDK version did you test with?

houqp:
I think this patch fixed the random build failure on gitlab! Although I still don't know why it would randomly build libraries without version suffixes. It also explains why it was working for me previously. I really wonder where is the entropy coming from :(

Frenzie:
How does it explain that it was working for you?

houqp:
Because I was using the exact same toolchain to build android without issue and I was like why the hell is it broken for me now. If it will randomly produce builds without version suffix, then it sort of makes sense.

Frenzie:
Randomly produce versionless? Make sense? Lol!

houqp:
Makes sense in the sense that it breaks my local build ;)

chrox:
This definitely will not work. On Android we use a custom [`dl` function](https://github.com/koreader/android-luajit-launcher/blob/master/assets/dl.lua#L51) to recursively load dynamic libraries by parsing the ELF dynamic section and loading NEEDED shared library.

```
libs $ greadelf -d libmupdf.so

Dynamic section at offset 0x303a1c contains 35 entries:
  Tag        Type                         Name/Value
 0x00000003 (PLTGOT)                     0x304d98
 0x00000002 (PLTRELSZ)                   1208 (bytes)
 0x00000017 (JMPREL)                     0x24a48
 0x00000014 (PLTREL)                     REL
 0x00000011 (REL)                        0x13aa0
 0x00000012 (RELSZ)                      69544 (bytes)
 0x00000013 (RELENT)                     8 (bytes)
 0x6ffffffa (RELCOUNT)                   8690
 0x00000006 (SYMTAB)                     0x148
 0x0000000b (SYMENT)                     16 (bytes)
 0x00000005 (STRTAB)                     0x7968
 0x0000000a (STRSZ)                      33642 (bytes)
 0x00000004 (HASH)                       0xfcd4
 0x00000001 (NEEDED)                     Shared library: [libz.so.1]
 0x00000001 (NEEDED)                     Shared library: [libjpeg.so.8]
 0x00000001 (NEEDED)                     Shared library: [libfreetype.so.6]
 0x00000001 (NEEDED)                     Shared library: [libm.so]
 0x00000001 (NEEDED)                     Shared library: [liblog.so]
 0x00000001 (NEEDED)                     Shared library: [libc.so]
 0x00000001 (NEEDED)                     Shared library: [libdl.so]
 0x0000000e (SONAME)                     Library soname: [libmupdf.so]
 0x0000001a (FINI_ARRAY)                 0x30495c
 0x0000001c (FINI_ARRAYSZ)               8 (bytes)
 0x00000019 (INIT_ARRAY)                 0x304a18
 0x0000001b (INIT_ARRAYSZ)               4 (bytes)
 0x0000000f (RPATH)                      Library rpath: [$ORIGIN]
 0x00000010 (SYMBOLIC)                   0x0
 0x0000001e (FLAGS)                      SYMBOLIC BIND_NOW
 0x6ffffffb (FLAGS_1)                    Flags: NOW
 0x6ffffff0 (VERSYM)                     0x12b00
 0x6ffffffc (VERDEF)                     0x13a04
 0x6ffffffd (VERDEFNUM)                  1
 0x6ffffffe (VERNEED)                    0x13a20
 0x6fffffff (VERNEEDNUM)                 4
 0x00000000 (NULL)                       0x0
```

It won't find 
```
0x00000001 (NEEDED)                     Shared library: [libjpeg.so.8]
 0x00000001 (NEEDED)                     Shared library: [libfreetype.so.6]
```
these two libs any more as the version suffixes are omitted.

One solution might be making https://github.com/koreader/android-luajit-launcher/blob/master/assets/dl.lua#L51 handling `so` files without version suffixes.

houqp:
Weird, why is it linked to `libfreetype.so.6` when the actual dynamic library produced is `libfreetype.so`?

chrox:
Generally the file without version suffix is just a soft link to the one with suffix and the soname certainly contains version suffix.

houqp:
I know that's the case for standard gnu toolchain. I was under the impression that android NDK toolchain actually produces some of the dynamic libraries without the suffix. But I might completely wrong since the last time I touched android build was a long time ago ;)

chrox:
It can be easily verified with this:

```
libs $ ls
libcrengine.so      libfreetype.so      libkoreader-cre.so  libluacompat52.so   libpng16.so         libturbojpeg.so     libzyre.so
libcrypto.so.1.0.0  libgif.so           libkoreader-lfs.so  libluajit.so        libsqlite3.so       libwrap-mupdf.so
libczmq.so          libjpeg.so          liblept.so.4        liblualongnumber.so libssl.so.1.0.0     libz.so.1
libfmq.so           libk2pdfopt.so.2    liblodepng.so       libmupdf.so         libtesseract.so.3   libzmq.so
libs $ greadelf -d libfreetype.so

Dynamic section at offset 0x61df0 contains 28 entries:
  Tag        Type                         Name/Value
 0x00000003 (PLTGOT)                     0x62f64
 0x00000002 (PLTRELSZ)                   288 (bytes)
 0x00000017 (JMPREL)                     0x4aa8
 0x00000014 (PLTREL)                     REL
 0x00000011 (REL)                        0x2838
 0x00000012 (RELSZ)                      8816 (bytes)
 0x00000013 (RELENT)                     8 (bytes)
 0x6ffffffa (RELCOUNT)                   1099
 0x00000006 (SYMTAB)                     0x148
 0x0000000b (SYMENT)                     16 (bytes)
 0x00000005 (STRTAB)                     0xfc8
 0x0000000a (STRSZ)                      4007 (bytes)
 0x00000004 (HASH)                       0x1f70
 0x00000001 (NEEDED)                     Shared library: [libc.so]
 0x0000000e (SONAME)                     Library soname: [libfreetype.so.6]
 0x0000001a (FINI_ARRAY)                 0x6280c
 0x0000001c (FINI_ARRAYSZ)               8 (bytes)
 0x00000019 (INIT_ARRAY)                 0x62dec
 0x0000001b (INIT_ARRAYSZ)               4 (bytes)
 0x00000010 (SYMBOLIC)                   0x0
 0x0000001e (FLAGS)                      SYMBOLIC BIND_NOW
 0x6ffffffb (FLAGS_1)                    Flags: NOW
 0x6ffffff0 (VERSYM)                     0x262c
 0x6ffffffc (VERDEF)                     0x27fc
 0x6ffffffd (VERDEFNUM)                  1
 0x6ffffffe (VERNEED)                    0x2818
 0x6fffffff (VERNEEDNUM)                 1
 0x00000000 (NULL)                       0x0
```

`libfreetype.so` in the latest nightly build android release has a soname `libfreetype.so.6`.

Frenzie:
How old is libtool? See e.g. http://git.savannah.gnu.org/cgit/libtool.git/commit/?id=8eeeb00daef8c4f720c9b79a0cdb89225d9909b6 and https://github.com/moritz-wundke/Boost-for-Android/issues/44

houqp:
I think there are couple options here:

* Compile all the binaries without version suffix in soname for all platforms. For many libraries, there should be a configure flag to set it. For others, we can patch the binary. I think the version suffix is mostly useful when the libraries is installed and shared system wide. In our case, those are private libraries packaged with koreader, so the version doesn't provide that much value to us. That's one of the reason why android doesn't work well with soname versioning because libraries in each app is only supposed to be used by the app itself.

* Like @chrox said, change the loader to search for both

* Rename the library with version suffix on android to make our loader happy

Frenzie:
> Compile all the binaries without version suffix in soname for all platforms.

I'm not vehemently opposed or anything, but doing something like that on all platforms just for Android gets a definite "meh" from me.

> change the loader to search for both

Sounds good!

> Rename the library with version suffix on android to make our loader happy

That makes the linker unhappy (which is what this PR was about) so I'm not sure if that's really an option.

chrox:
It turns out to be the result of different versions of libtool that makes the discrepancy. Ubuntu 16.04 comes with libtool 2.4.6 which will generate a `soname_spec='$libname$release$shared_ext'` while  older libtool (on my machine, 2.4.2) will generate a `soname_spec='$libname$release$shared_ext$major'` with major version suffix. There seems to be no simple parameters to accommodate both versions of libtool. I would like to keep current fix and add a libtool version dependency in the README.

@houqp And we should upgrade the libtool in the building machine too. ref: https://www.gnu.org/software/libtool/

chrox:
> change the loader to search for both

Tried that, although it can load the freetype lib, it still failed on loading mupdf with system `dlopen`. It will fail when it cannot find the needed library.

PR #483, opened 2017-04-23
Handle 32-bit framebuffer depth

baskerville:


ffi/framebuffer_linux.lua
```diff
@@ -36,9 +36,12 @@ function framebuffer:init()
         self.fb_size = vinfo.xres_virtual * vinfo.yres_virtual * vinfo.bits_per_pixel / 8
     -- Newer eink framebuffer (Kindle Touch, Paperwhite, Kobo)
     elseif ffi.string(finfo.id, 11) == "mxc_epdc_fb" then
-        -- it seems that finfo.smem_len is unreliable on kobo
         -- Figure out the size of the screen in bytes
         self.fb_size = vinfo.xres_virtual * vinfo.yres_virtual * vinfo.bits_per_pixel / 8
+        -- Kobo hasn't updated smem_len when the depth was changed from 16 to 32
+        if self.fb_size > finfo.smem_len then
```
Frenzie: Since updates can go in multiple directions, wouldn't `~=` be better?

Frenzie:
Thanks!

ffi/framebuffer_linux.lua
baskerville:
```lua
self.fb_size = vinfo.xres_virtual * vinfo.yres_virtual * vinfo.bits_per_pixel / 8
if self.fb_size ~= finfo.smem_len then
	self.fb_size = finfo.smem_len
end
```
is equivalent to:
```lua
self.fb_size = finfo.smem_len
```

Frenzie: I suppose you mean the resulting behavior is equivalent in a 32-bit scenario, but based on the comment you removed (why?) it looks like my first thought was bad regardless. Unfortunately I have no idea what exactly was meant by `smem_len` being unreliable.

In what way does `vinfo.xres_virtual * vinfo.yres_virtual * vinfo.bits_per_pixel / 8` differ from `finfo.smem_len`? Is it half the size as you would expect when `vinfo.bits_per_pixel` is 16 or is it different in another way? Regardless of the result, does `vinfo.xres` (as opposed to `_virtual`) work because you mentioned `xres` is `1404` and `xres_virtual` is `1408`?

Basically just from looking at this I'm not sure if `self.fb_size > finfo.smem_len` is sufficiently generic. If it's not, I'm inclined to think the test should be something more specific like `vinfo.bits_per_pixel == 32`.

baskerville: `finfo.smem_len` is accurate in 16-bit mode but is half of what is should be in 32-bit mode.

If `self.fb_size` is greater that `finfo.smem_len` then writing to the `mmap`ed region will generate a segfault.

I agree with the more specific test you suggested and I've created #486 (because I forgot to create a branch in this PR).

Frenzie: > If self.fb_size is greater that finfo.smem_len then writing to the mmaped region will generate a segfault.

You'd think, but being "unreliable" could go either way. :-P

houqp: I feel like always use `smem_len` is probably the safest approach here. since it's literally the amount of framebuffer size that the kernel says you can use.

it's shitty that the kernel still reports double sized yres_virtual (around yres * 2) but only allocates a framebuffer with size around `xres * yres * 4`.

PR #489, opened 2017-04-30
Added thirdparty/nanosvg

poire-z:
For crengine support of SVG images. See https://github.com/koreader/koreader/issues/2668


houqp:
Nice!

PR #492, opened 2017-05-23
Fix memory leak with BB.fromstring()

poire-z:
Not used much by koreader, except when loading a cached TileCacheItem and in one unit test.
(But I witnessed huge memory leaks when playing with cached cover images and some thumbnails view - not sure i'll end up making something of that, but this bug should be fixed).
This setAllocated(1) is what will make the calls to free() actually do some freeing, it's what used in this file in all other places after a ffi.C.malloc() is made.


Frenzie:
I don't recall ever having seen that fromstring method. Thanks for the fix.

PR #493, opened 2017-06-06
[fix] patch libtool version < 2.4.6 to generate lib without version suffix for Android

chrox:
The `libtool` version >= 2.4.6 don't have this problem and won't be affected by the patch. See comments in #482.


thirdparty/czmq/CMakeLists.txt
```diff
@@ -21,6 +21,10 @@ set(CFG_ENV_VAR "CC=\"${CC}\" LDFLAGS=\"${LDFLAGS}\" CFLAGS=\"-Wno-error ${CFLAG
 set(CFG_OPTS "-q --prefix=${BINARY_DIR} --with-gnu-ld --with-libzmq=${ZMQ_DIR} --disable-static --enable-shared --host=${HOST}")
 set(CFG_CMD sh -c "${CFG_ENV_VAR} ${SOURCE_DIR}/configure ${CFG_OPTS}")
 
+if($ENV{ANDROID})
+    set(CFG_CMD "${CFG_CMD} && ${ISED} 's|\${shared_ext}\\\\\$major|\${shared_ext}|' libtool")
```
chrox: Funny thing is that we need exactly FOUR backslashs to escape one backslash in the original file.

PR #496, opened 2017-08-08
thirdparty: change sdcv source to upstream

Frenzie:
See discussion in https://github.com/koreader/koreader/issues/2951#issuecomment-313316115


Frenzie:
I'll bravely merge this. Everything seemed fine in my tests and upstream actually has unit testing to prevent regressions and such now.

PR #500, opened 2017-08-23
fix compile errors under Fedora due to -Werror

silberzwiebel:
fixes compilation errors reported in https://github.com/koreader/koreader/issues/3069
does not fix runtime errors under fedora (due to lua5.3 vs lua 5.1), though


Frenzie:
Thanks! I'll merge it once the checks finish.

PR #501, opened 2017-08-28
Bump crengine and add cre.setEmbeddedFonts()

poire-z:
crengine: fix parsing of long decimal values in css and new Polish and French hyph dictionaries
cre.setEmbeddedFonts() to make this crengine feature available to koreader (having a go at https://github.com/koreader/crengine/issues/76)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -881,6 +881,14 @@ static int setEmbeddedStyleSheet(lua_State *L) {
 	return 0;
 }
 
+static int setEmbeddedFonts(lua_State *L) {
+	CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+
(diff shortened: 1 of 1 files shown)


Frenzie:
Feel free to merge once the tests pass.

poire-z:
Any prefered option for the merge, among "Create a merge commit", "Squash and merge" and "Rebase and merge" (asking, as this would be my first, and I want to get it right :)

Frenzie:
All of them are basically fine, but squash & merge and rebase & merge are preferable because they make for a nicer Git log. There's no official policy but what I've been doing recently is squash & merge unless there are a bunch of nicely individually tailored commits inside the PR, even if there's only one commit. That way you also get a reference to the PR #.

PR #504, opened 2017-09-01
Bump crengine

Frenzie:
* Keep previous fallback font if new fallback font fails (https://github.com/koreader/crengine/pull/84)
*  Allow setting a bias to fonts for them to be prefered in CalcMatch algorithm (https://github.com/koreader/crengine/pull/83)


poire-z:
I need to add to proxy function `setAsPreferredFontWithBias()` to base/cre.cpp, i'll do that in a few minutes. No need for me to bump crengine then?

Frenzie:
Just merge this (provided it passes, yada yada) and open a PR for the other one. 👍

PR #508, opened 2017-09-14
[fix] ffi/framebuffer: memory use explosion in unit tests

Frenzie:
During the unit tests fb:getDPI() is hit multiple times and self.dpi has to be filled from scratch.
This results in multiple calls to `os.getenv("EMULATE_READER_DPI")`. No biggie you would think, but
for some reason this causes memory use to quickly balloon out of proportion and exceed the 2 GB limit.

### DIFF
--- a/ffi/framebuffer.lua
+++ b/ffi/framebuffer.lua
@@ -48,6 +48,8 @@ fb.ORIENTATION_LANDSCAPE = 1
 fb.ORIENTATION_PORTRAIT_ROTATED = 2
 fb.ORIENTATION_LANDSCAPE_ROTATED = 3
 
+local EMULATE_READER_DPI = tonumber(os.getenv("EMULATE_READER_DPI"))
+
 function fb:extend(o)
     o = o or {}
     setmetatable(o, self)
@@ -223,7 +225,7 @@ end
 
 function fb:getDPI()
     if self.dpi == nil then
-        self.dpi = tonumber(os.getenv("EMULATE_READER_DPI")) or G_reader_settings:readSetting("screen_dpi")
+        self.dpi = EMULATE_READER_DPI or G_reader_settings:readSetting("screen_dpi")
     end
     if self.dpi == nil and self.device then
         self.dpi = self.device.display_dpi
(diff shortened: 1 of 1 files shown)


poire-z:
Just curious as I was as perplexed as you, how did you figure that out ?

Frenzie:
There wasn't really anything *to* figure out as this was the only change, unless you mean the fix? The `os.getenv("EMULATE_READER_DPI")` in `fb:getDPI()` causes memory to quickly exceed 2 GB after being called multiple times, the one in `fb:scaleBySize(px)` still seems to cause significantly higher memory use than usual while running the unit tests (i.e., 300+ MB instead of 150-200).

If you mean how I figured out the fix, I already though calling `os.getenv` many times during GUI construction, while still super fast and probably completely harmless, wasn't the smartest thing to do. I figured I might as well try it… and by a lucky coincidence the first thing I tried work. I mean, I'd expect it to save a negligible number of CPU cycles and a negligible amount of memory… not to resolve some odd side effect in the cavernous and incomprehensible internals of LuaJIT. :-P

But yeah, I have no idea what's going on besides that some combination of busted, LuaJIT and os.getenv == memory spike.

poire-z:
Thanks for the story. (I meant: understood the cause in order to think of that fix. But I get the fix came first :)

Frenzie:
Wow, that was full of typos.

Frenzie:
Not entirely sure what you mean by the fix came first? a4d14e5abd22b6593cc2e1a8dd431b963d7946ed doesn't leave much room for doubt as to what change might be responsible for the issue. :-P

I suppose I left it a bit too implicit that first I tried them one at a time to see what was happening exactly, combined with a couple of print statement to see when these calls were being activated. I was slightly surprised to see how often `fb:getDPI()` was called, but still significantly fewer times than `fb:scaleBySize(px)`. But either way, you could just stick in a `for i = 1, 9999999 do bla = os.getenv("whatever") end` and you would barely even notice it, or at least not in memory use. (Sticking that in obviously causes the tests to slow down significantly, but it's not an insta-crash recipe.) It's not like a few dozen times more or less should make any real difference. And as if to underline it, `fb:getDPI()` was the troublemaker.

PR #510, opened 2017-09-16
Android NDK 15

Frenzie:
Updates some stuff.

`NDKABI?=14` is important for specifying something custom for testing.

Also see https://github.com/koreader/android-luajit-launcher/pull/59

### DIFF
--- a/toolchain/Makefile
+++ b/toolchain/Makefile
@@ -1,24 +1,20 @@
 SHELL:=/bin/bash
 
 
-SDK_DL_URL=https://dl.google.com/android/repository/sdk-tools-linux-3859397.zip
-SDK_TARBALL=sdk-tools-linux-3859397.zip
-SDK_SUM="444e22ce8ca0f67353bda4b85175ed3731cae3ffa695ca18119cbacef1c1bea0"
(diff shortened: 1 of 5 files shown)


toolchain/Makefile
```diff
@@ -1,28 +1,35 @@
 SHELL:=/bin/bash
 
 
-SDK_DL_URL=https://dl.google.com/android/android-sdk_r24.4.1-linux.tgz
-SDK_TARBALL=android-sdk_r24.4.1-linux.tgz
-SDK_SUM="978ee9da3dda10fb786709b7c2e924c0"
+SDK_DL_URL=https://dl.google.com/android/repository/sdk-tools-linux-3859397.zip
```
Frenzie: I tossed in this last-minute but it's better to stick with 24 (or at least no higher than `https://dl.google.com/android/repository/tools_r25.2.5-linux.zip`).

It does give you the nice sdkmanager --licenses thing, which on second thought might be somewhat questionable to stick in here, but it deprecates the android command and only prints a deprecated message.

So, for SDK 26 and up, `sdkmanager "build-tools;26.0.1" "platform-tools" "platforms;android-14" "platforms;android-16"`

For lower, `android update sdk -u --filter "platform-tools,build-tools-26.0.1,android-14,android-16"`.

PR #511, opened 2017-09-17
Added util.getNonBlockingReadSize(luafile)

poire-z:
Needed to implement cancelable dict lookup.
Checking if `static const int FIONREAD = 21531; ` does not fail here...

### DIFF
--- a/ffi/util.lua
+++ b/ffi/util.lua
@@ -214,6 +214,23 @@ function util.execute(...)
     end
 end
 
+--- Returns the length of data that can be read immediately without blocking
+function util.getNonBlockingReadSize(luafile)
+    -- returns 0 if not readable yet, otherwise len of available data
+    -- returns nil when unsupported: caller may read (with possible blocking)
+    if util.isAndroid() then
+        return
+    end
+    local fileno = ffi.C.fileno(luafile)
+    local available = ffi.new('int[1]')
+    local ok = ffi.C.ioctl(fileno, ffi.C.FIONREAD, available)
+    if ok ~= 0 then -- ioctl failed, not supported
+        return
+    end
+    available = tonumber(available[0])
+    return available
+end
+
 --- Gets UTF-8 charcode.
 function util.utf8charcode(charstring)
     local ptr = ffi.cast("uint8_t *", charstring)
--- a/ffi-cdecl/posix_decl.c
+++ b/ffi-cdecl/posix_decl.c
@@ -75,3 +75,5 @@ cdecl_func(printf)
 cdecl_func(sprintf)
 cdecl_func(fprintf)
 cdecl_func(fputc)
+
+cdecl_func(fileno)
--- a/ffi/posix_h.lua
+++ b/ffi/posix_h.lua
@@ -75,4 +75,6 @@ int printf(const char *, ...);
 int sprintf(char *, const char *, ...) __attribute__((nothrow));
 int fprintf(struct _IO_FILE *restrict, const char *restrict, ...);
 int fputc(int, struct _IO_FILE *);
+static const int FIONREAD = 21531;
+int fileno(struct FILE *stream);
 ]]


ffi/util.lua
```diff
@@ -214,6 +214,23 @@ function util.execute(...)
     end
 end
 
+--- Returns the length of data that can be read without blocking
+function util.availableDataForReadingSize(luafile)
+    -- returns 0 if not readable yet, otherwise len of available data
+    -- returns -1 when unsupported: caller can read (with possible blocking)
```
Frenzie: Is that better than plain old nil?

poire-z: Thought it would be more coherent to always return numbers.
`if util.availableDataForReadingSize(luafile) ~= 0 then go on`
It would mean getting nil would go on. But I don't know.

Frenzie: That's no different for what you just wrote. However, it would allow you to safely write `if util.availableDataForReadingSize(luafile) > 0 then go on`

It might make more sense to write it like this though (in Lua, at least)

```lua
local something = util.availableDataForReadingSize(luafile)
if something and something > 0 then
```

Frenzie: Hah, I just realized I this is exactly the same as https://github.com/koreader/koreader/pull/3231#discussion_r139309957

poire-z: Well, the most probable use case would be that if `util.availableDataForReadingSize()` is unsupported or not working and can't figure out if data is available, we would go on. So most probably `if something ~= 0`, which work with nil or -1. You tell me.
(So, I can assume you're fine with that long function name ? :)

Frenzie: I see. So either way I see no need to deviate from nil then. :-P

I hadn't really thought about the function name yet. What about `util.safeReadingSize`?

poire-z: safe alone don't say much, safe_for_non_blocking_reading would :)
I let you think :)
(sorry for bothering with function and file names, but names itch me till i find a good one)

poire-z: or just util.availableReadSize(luafile)

Frenzie: getSafeNonBlockReadingSize

ffi/util.lua
```diff
@@ -214,6 +214,23 @@ function util.execute(...)
     end
 end
 
+--- Returns the length of data that can be read immediately without blocking
+function util.availableReadSize(luafile)
```
Frenzie: I think "available" says less than "safe". Sure, if you just use "safe" you don't know safe for what without looking it up. But with just "available" the thought that going ahead anyway if it's 0 would have unintended consequences might not cross my mind.

poire-z: For me "safe" is too strong :) there no unintended consequences: the consequences are blocking or non-blocking, nothing unsafe, just blocking :)
util.availableNonBlockingReadSize() ?

Frenzie: I'd say getNonBlockingReadSize.

poire-z: OK, this one is fine :)

poire-z:
OK for merging (can't hurt till it is used) ?

Frenzie:
I'd prefer a unit test but I suppose that can wait for now.

poire-z:
Well, for this one, it may not be as frightening as frontlight on kobo without frontlight :)
Would this do if added to spec/unit/util_spec.lua ?
```lua
if not util.isAndroid() then
    describe("util.getNonBlockingReadSize", function()
        it("should read pipe when data is available", function()
            local std_out = io.popen("sleep 3; echo 'done'", "r")
            if std_out then
                assert.are.equal(util.getNonBlockingReadSize(std_out), 0)
                util.sleep(1)
                assert.are.equal(util.getNonBlockingReadSize(std_out), 0)
                util.sleep(1)
                assert.are.equal(util.getNonBlockingReadSize(std_out), 0)
                util.sleep(2)
                assert.are.equal(util.getNonBlockingReadSize(std_out), 5)
                local startsec = util.gettime()
                local data = std_out:read("*all")
                local donesec = util.gettime()
                assert.are.equal(data, "done\n")
                assert.is_true(donesec - startsec <= 1)
            end
        end)
    end)
end
```

PR #513, opened 2017-09-17
Added unit test for util.getNonBlockingReadSize()

poire-z:
Attempt at unit test for #511


poire-z:
Well, these are just tested on target EMULATE_READER ? Seems to pass ! ... although we don't see what happened.
Should I re-test a build with something erroneous to be sure ?

Frenzie:
Nice basic unit test. 👍 

I'd turn the Android test around though. Since that'll never be true in the testing environment, you could consider faking isAndroid to make sure it doesn't do anything. (Nah, you really don't have to do that. :-P)

And yes, it's always a good idea to check if it actually protects against what you think it does.

poire-z:
Failed as expected, so it must be ok.

spec/unit/util_spec.lua
```diff
@@ -160,4 +160,24 @@ describe("util module", function()
             assert.is_not_nil(err)
         end)
     end)
+
+    describe("util.getNonBlockingReadSize", function()
+        it("should read pipe when data is available", function()
+            if not util.isAndroid() then
+		local std_out = io.popen("sleep 3; echo 'done'", "r")
```
Frenzie: You accidentally used tabs here.

Frenzie:
Oh btw, since you're messing around in here anyway, would you mind addressing this by sticking in a `local`?

```
    spec/unit/util_spec.lua:144:13: setting non-standard global variable ok
    spec/unit/util_spec.lua:144:17: setting non-standard global variable err
    spec/unit/util_spec.lua:145:27: accessing undefined variable ok
    spec/unit/util_spec.lua:146:31: accessing undefined variable err
```

Also, in luacheckrc you remove this line:

```
    "spec/*",
```

and add this

```
files["spec/unit/*"].std = "+busted"
```

You can ignore it and I'll just do that tomorrow. But you know, since you're already messing around here anyway. :-P

poire-z:
Done. (Dunno if you can interrupt the old still running travis build, but if you can please do.)

Frenzie:
Awesome, thanks!

I intend to enable luacheck down here one of these days, but for the moment I'm just resolving the occasional minor issue.

PR #517, opened 2017-09-24
Add proper scaling to mupdf.renderImage()

poire-z:
mupdf.renderImage() may not scale the image to the width & height provided, see https://github.com/koreader/koreader/pull/3253#issuecomment-331702912.
I haven't tested that much for side effects, but it seems to work for menu images (they look more blury, less pixelized, dunno if it's actually better and good).
If you can test the rendering results on various sizes/dpi, upscale/downscale, I'm not that much able to appreciate that.
The code in mupdf for the scaling is in mupdf/source/fitz/draw-scale-simple.c fz_scale_pixmap_cached, and I don't even know what algorithm it used :|

### DIFF
--- a/ffi/mupdf.lua
+++ b/ffi/mupdf.lua
@@ -640,6 +640,15 @@ function mupdf.renderImage(data, size, width, height)
 
     local p_width = M.fz_pixmap_width(context(), pixmap)
     local p_height = M.fz_pixmap_height(context(), pixmap)
+    -- mupdf_new_pixmap_from_image() may not scale image to the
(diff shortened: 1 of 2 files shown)


Frenzie:
Looks better to me. :-)

robert00s:
Much better :)

poire-z:
I'll try to replace the 2 _bb:scale() in imagewidget by calls to this mupdf function , so our chevrons can also be shiny blury (and covers, and imageviewer a bit better).

robert00s:
It would be great. For now chevrons on my kindle is a little ugly :)

PR #518, opened 2017-09-24
Luacheck fixes

Frenzie:


ffi/pic.lua
```diff
@@ -151,11 +151,12 @@ function GifDocument:openPage(number)
     return page
 end
 function GifDocument:close()
-    if giflib.DGifCloseFile(self.giffile, err) ~= giflib.GIF_OK then
+    if giflib.DGifCloseFile(self.giffile, self.err) ~= giflib.GIF_OK then
         error(string.format("error closing/deallocating GIF: %s",
```
poire-z: Strange it works with a non ffi typed variable.
I would use there, like it is used in the next function:
`local err = ffi.new("int[1]")  -- luacheck: no unused`

ffi/pic.lua
```diff
@@ -170,7 +171,7 @@ function Pic.openGIFDocument(filename)
         error(string.format("Cannot parse GIF file: %s",
             ffi.string(giflib.GifErrorString(giffile.Error))))
     end
-    return GifDocument:new{giffile = giffile}
+    return GifDocument:new{giffile = giffile, err = err}
```
poire-z: Would just add `-- luacheck: no unused` , as GifDocument:new does not care.
I think it does not need to be the same err in both functions, it's just a receptacle for some error code.

PR #522, opened 2017-09-28
add --all flag to android update sdk command

silberzwiebel:
I got `Error: Ignoring unknown package filter 'build-tools-26.0.1` otherwise and in result a non-functioning build

### DIFF
--- a/toolchain/Makefile
+++ b/toolchain/Makefile
@@ -13,7 +13,7 @@ android-sdk:
 		 { echo "$(SDK_TARBALL) has the wrong checksum!"; exit 1; }
 	tar xvf $(SDK_TARBALL)
 	pushd android-sdk-linux/tools && \
-		./android update sdk -u --filter "platform-tools,build-tools-26.0.1,android-14,android-19" && \
+		./android update sdk --all -u --filter "platform-tools,build-tools-26.0.1,android-14,android-19" && \
 		popd
 	rm -f $(SDK_TARBALL)
 


Frenzie:
Could you give this a little more context?

```
November 20, 2015
Do you accept the license 'android-sdk-license-c81a61d9' [y/n]: y

Installing Archives:
  Preparing to install archives
  Downloading Android SDK Platform-tools, revision 26
  Installing Android SDK Platform-tools, revision 26
  Stopping ADB server failed (code -1).
    Installed Android SDK Platform-tools, revision 2691%)
  Downloading SDK Platform Android 4.4.2, API 19, revision 4
  Installing SDK Platform Android 4.4.2, API 19, revision 4
    Installed SDK Platform Android 4.4.2, API 19, revision 497%)
  Downloading SDK Platform Android 4.0, API 14, revision 4
  Installing SDK Platform Android 4.0, API 14, revision 4
    Installed SDK Platform Android 4.0, API 14, revision 496%)
    Stopping ADB server succeeded.
    Starting ADB server succeeded.
  Done. 3 packages installed.
~/src/kobo/koreader-base/toolchain
rm -f android-sdk_r24.4.1-linux.tgz
```

silberzwiebel:
The code in the makefile is requesting four packages:
`platform-tools,build-tools-26.0.1,android-14,android-19`

I got the rather silent (it doesn't stop) error `Error: Ignoring unknown package filter 'build-tools-26.0.1'`. Only later on, when creating the apk, the android sdk was complaining that the build-tools were not properly installed and it failed.
I could try to replicate so that you get the full error message?

Frenzie:
Oh, you're right. So --all will just ensure it actually downloads some build-tools and it won't do anything more than that? I tailored that command to download only that which is required and not a single byte more. (Unlike, say, Android Studio which downloads several Gigabytes right off the bat.)

I used build-tools-specific version because just `build-tools` also gives you that error. Do you know if there's something sensible that can be stuck in there that wouldn't necessitate --all? Basically I just want like how platform-14 gives you the most updated version of platform-14 (but not platform-15).

silberzwiebel:
I was successful in downloading the correct build-tools also only by trial and error.
These stackoverflow comments hinted me at including the counter-intuitive `--all` flag:


https://stackoverflow.com/questions/17963508/how-to-install-android-sdk-build-tools-on-the-command-line#comment44756210_23196528

https://stackoverflow.com/questions/17963508/how-to-install-android-sdk-build-tools-on-the-command-line#comment41317102_22862021

It seems that Google wants that people use SDKs as new as possible which would explain why it is so complicated to get an older version. The `--all` flag seems to do the job. Since there is still the filter active, only the required packages are installed.

Frenzie:
I suppose that'd be mostly alright if they acted like Microsoft with regard to compatibility. But it doesn't help that they took GNU/Linux and then destroyed it completely with all kinds of inferior nonsense because eww GNU. (Isn't there eww GNU perfectly serviceable BSD stuff?)

Anyway, sounds good to go then, thanks for the fix!

PR #523, opened 2017-09-29
Fix colors on the emulator

poire-z:
Red and Blue where previously inverted.
Our blitbuffer are RGBA, and SDL, using SDL_PIXELFORMAT_ARGB8888 (reverted, cause of endianess I guess), was getting BGRA. By using SDL.SDL_PIXELFORMAT_ABGR8888, it now gets them in the correct order.
See https://github.com/koreader/koreader/issues/3274
Some references:
https://wiki.libsdl.org/SDL_PixelFormatEnum
https://forums.libsdl.org/viewtopic.php?p=39284


Frenzie:
Awesome!

PR #524, opened 2017-09-30
mupdf, cre: allow for optional colored rendering

poire-z:
Support for colored rendering is already there for all engines/formats. It just needs a few fixes, and enabling it. This is the stuff needed in base.


cre.cpp
```diff
@@ -1286,15 +1286,28 @@ static int clearSelection(lua_State *L) {
 static int drawCurrentPage(lua_State *L) {
 	CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
 	BlitBuffer *bb = (BlitBuffer*) lua_topointer(L, 2);
+	bool color = false;
+	if (lua_isboolean(L,3)) {
```
Frenzie: nitpick: space

PR #525, opened 2017-10-01
Add KODEBUG environment variable to trigger debug build

Frenzie:
@poire-z Any thoughts on the name?

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -143,7 +143,9 @@ endif
 # Makefiles do some tricky stuff to differentiate HOST/TARGET
 BASE_CFLAGS:=-O2 -ffast-math -pipe -fomit-frame-pointer
 # Use this for debugging:
-#BASE_CFLAGS:=-O0 -g
+ifdef KODEBUG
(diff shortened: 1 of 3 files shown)


Makefile.defs
```diff
@@ -143,7 +143,9 @@ endif
 # Makefiles do some tricky stuff to differentiate HOST/TARGET
 BASE_CFLAGS:=-O2 -ffast-math -pipe -fomit-frame-pointer
 # Use this for debugging:
-#BASE_CFLAGS:=-O0 -g
+ifdef KODEBUG
+    BASE_CFLAGS:=-Og -g
```
Frenzie: Wait, is there a reason this line was different? Shouldn't it just be the same as regular (`-ffast-math` etc) but with `0g -g`?

poire-z: Dunno, may be updating it was just forgotten when updating the main one (check commit history maybe ?)
OK with `KODEBUG` , it sounds nice :)

Frenzie:
@poire-z Any thoughts on the additions?

poire-z:
Just the thought that I can't comment on the last commit, as I don't know much about luajit (and Makefile business). But OK with the other stuff.

PR #529, opened 2017-10-02
Makefile: add luajit-clean target

Frenzie:
Some automation should probably be added, but this is to aid switching between builds with and without KODEBUG. LuaJIT requires a manual `make clean` if you do.


Frenzie:
Superseded by #531.

PR #530, opened 2017-10-02
[fix] Ignore deprecated SSLv2 and SSLv3 encryptions

Frenzie:
* Bumping luasec takes care of proper TLS encryption.

* Bump OpenSSL to the latest 1.0.2l and disable insecure deprecated SSLv2 and SSLv3.


Frenzie:
I wanted to update it to `OpenSSL_1_0_2l`, but the Travis Clang build failed for mysterious reasons. `1.0.1u` seems to work, so updating OpenSSL to a more recent version can wait for another day.

PR #531, opened 2017-10-04
Makefile: add luajit-clean target and auto-call it

Frenzie:
This is to aid switching between builds with and without KODEBUG. LuaJIT requires a manual `make clean` if you do.


Frenzie:
I just realized the target should be included as well.

PR #537, opened 2017-10-06
Luacheck fixes

Frenzie:


ffi/util.lua
```diff
@@ -313,7 +318,9 @@ function util.idiv(a, b)
     return (q > 0) and math.floor(q) or math.ceil(q)
 end
 
-function util.orderedPairs(t)
+--- Equivalent of the pairs() function on tables. Allows to iterate in order.
+function util.orderedPairs(t) return end
+do
```
poire-z: What/why is that `do` doing there? And why an empty function ? (must be explained in comments if there's an explanation, cause I would wonder :)

Frenzie:
goto is relatively new (5.2).

ffi/util.lua
Frenzie: It's not really something that needs explanation. Sticking something in `do…end` limits the scope. ;-)

The luacheck complaint is valid in my eyes:
```
    ffi/util.lua:317:39: shadowing upvalue argument 't' on line 316
    ffi/util.lua:327:32: shadowing upvalue argument 't' on line 316
```
That means that while reading the code, you have to be really careful or you might get confused about the specifics of what `t` is. I don't like that, and I suspect neither does anybody else. So by using `do…end` I'm doing the same thing as embedding it in the function. Nothing outside that scope has access to it.

Normally I'd probably just write `local _whatever` (with `_` clearly indicating local use only elsewhere in the code) but in this case it seemed fairly elegant. Anyway, happy to change it. It's just one of those "all ways lead to Rome" kind of things. :-P

poire-z: Thanks for the explanation, didn't know that trick. May be just add:
`do -- just to limit the scope of the following, and not pollute namespace` or something :)

poire-z:
(Thought all this time we were with lua 5.1 ...)

Frenzie:
LuaJIT 2.0 already implements some conveniences from 5.2 and it's less incompatible than vanilla Lua 5.1: http://luajit.org/extensions.html#lua52

However, since we're on 2.1-beta3 we've actually got much more 5.2 stuff (e.g. more recently, see [here](https://github.com/LuaJIT/LuaJIT/commit/c1981676907cedde9ffe2bbdfb28d2f786ff69d9)). I even had to update a deprecated feature (one single issue) for the 2.1-beta3 update (https://github.com/koreader/koreader/commit/16f8f7226ef80035d1ac01a510e66fc3143ad6c5). It was already deprecated in 5.1, but it wasn't removed until 5.2. LuaJIT 2.1-beta3 also removed it.

So you see, we're in some mysterious lala-land. :-P

PR #538, opened 2017-10-06
[chore] blitbuffer luacheck fixes

Frenzie:
Also remove what appears to be an unused legacy function,
`BB_mt.__index:blitFromRotate(source, degree)`. We now
have the much better `BB:rotate()` and related functions.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,4 +1,5 @@
 unused_args = false
+unused_secondaries = false
 std = "luajit"
 -- ignore implicit self
 self = false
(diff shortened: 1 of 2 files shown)


Frenzie:
Right, that's why luacheck flagged it. I wouldn't have noticed otherwise. (That is, I'm not doing a giant cleanup; just an "enable luacheck as fail" cleanup.)

Btw, a second pair of eyes on #537 probably couldn't hurt either except there I haven't really changed anything.

PR #539, opened 2017-10-07
[CI] Enable LuaCheck

Frenzie:
* No more allowed failure for shellchecks
* Fix up final `koptcontext.lua`

### DIFF
--- a/ffi/koptcontext.lua
+++ b/ffi/koptcontext.lua
@@ -75,13 +75,10 @@ function KOPTContext_mt.__index:dstToBlitBuffer()
 end
 
 function KOPTContext_mt.__index:getWordBoxes(bmp, x, y, w, h, box_type)
-    local box = ffi.new("BOX[1]")
     local boxa = ffi.new("BOXA[1]")
     local nai = ffi.new("NUMA[1]")
-    local max_val = ffi.new("float[1]")
-    local last_index = ffi.new("int[1]")
     local counter_l = ffi.new("int[1]")
-    local nr_line, nr_word, current_line
+    local nr_word, current_line
     local counter_w, counter_cw
     local l_x0, l_y0, l_x1, l_y1
 
@@ -100,9 +97,6 @@ function KOPTContext_mt.__index:getWordBoxes(bmp, x, y, w, h, box_type)
     if boxa == nil or nai == nil then return end
 
     --get number of lines in this area
-    leptonica.numaGetMax(nai, max_val, last_index)
-    nr_line = max_val[0]
-    --get number of lines in this area
     nr_word = leptonica.boxaGetCount(boxa)
     assert(nr_word == leptonica.numaGetCount(nai))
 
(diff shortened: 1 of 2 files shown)


ffi/koptcontext.lua
```diff
@@ -117,7 +111,7 @@ function KOPTContext_mt.__index:getWordBoxes(bmp, x, y, w, h, box_type)
         counter_cw = 0
         l_x0, l_y0, l_x1, l_y1 = 9999, 9999, 0, 0
         while current_line == counter_l[0] and counter_w < nr_word do
-            box = leptonica.boxaGetBox(boxa, counter_w, ffi.C.L_CLONE)
+            local box = leptonica.boxaGetBox(boxa, counter_w, ffi.C.L_CLONE)
```
poire-z: If it's the same box that you remove the ffi declaration above, better add it back just there:
```lua
       local box = ffi.new("BOX[1]")
       box = leptonica.boxaGetBox(boxa, counter_w, ffi.C.L_CLONE)
```

ffi/koptcontext.lua
```diff
@@ -100,9 +97,6 @@ function KOPTContext_mt.__index:getWordBoxes(bmp, x, y, w, h, box_type)
     if boxa == nil or nai == nil then return end
 
     --get number of lines in this area
-    leptonica.numaGetMax(nai, max_val, last_index)
-    nr_line = max_val[0]
-    --get number of lines in this area
```
poire-z: I suppose nr_line was not used anywhere. Hoping leptonica.numaGetMax() don't do any stuff on the side

Frenzie: I could also add luacheck ignores to 'em but selecting words didn't seem to act any differently for any of this.

poire-z: Ok then (I rarely read pdf, so you'll probably notice sooner than me if it has rare side effects)

ffi/koptcontext.lua
Frenzie: It seems redundant. My own testing seemed fine, the unit tests didn't break,[1] and elsewhere in the code it doesn't do that either.[2] :-/

[1]
https://github.com/koreader/koreader-base/blob/b73bfe5f8b9e94c27d8af6272ae39a536347dc96/spec/unit/koptcontext_spec.lua#L86-L109

[2]
https://github.com/koreader/koreader-base/blob/f8caa6ab3d1eb8fa0e70d2232082f2957c418a0d/ffi/koptcontext.lua#L241
https://github.com/koreader/koreader-base/blob/f8caa6ab3d1eb8fa0e70d2232082f2957c418a0d/ffi/koptcontext.lua#L269

poire-z: OK then (some experiments of mine with python ctypes (the equivalent of lua ffi) lead me to think that it's always best - to avoid segfaults - to declare the ctype type of the returned value - may be lua ffi is better with that)

poire-z: OK, I guess it's the necessary include of things like require("ffi/leptonica_h") that helps ffi to make the right lua object as it knows what type of things a function returns.

Frenzie: Lua is definitely better than Python (in such matters).

PR #543, opened 2017-10-12
Bump crengine

poire-z:
for https://github.com/koreader/crengine/commit/febc7595e71dd17056eb1a65fa121823eefb0104 (epub) Return Series Number as string


poire-z:
Please wait before merging. The other & amp; problem can be solved on the crengine side too.

PR #545, opened 2017-10-14
fix build on Mac OSX and fix SDL input issues on OSX

chrox:


thirdparty/openssl/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ endif()
 set(CFG_OPTS "${CFG_OPTS} no-asm no-idea no-mdc2 no-rc5 no-ssl2 no-ssl3")
 set(CFG_CMD sh -c "${CFG_ENV_VAR} ${SOURCE_DIR}/${CONFIG_SCRIPT} ${CFG_OPTS}")
 
-set(BUILD_CMD "$(MAKE) depend && $(MAKE) CC=\"${CC}\" SHARED_LDFLAGS=\"${SHARED_LDFLAGS}\" LD=\"${LD}\" RANLIB=\"${RANLIB}\"")
+set(BUILD_CMD "$(MAKE) CC=\"${CC}\" SHARED_LDFLAGS=\"${SHARED_LDFLAGS}\" LD=\"${LD}\" RANLIB=\"${RANLIB}\"")
```
Frenzie: It's probably okay, although this message doesn't inspire much confidence, but what does it do on Mac OS other than save some time?

```
*** Because of configuration changes, you MUST do the following before
*** building:

        make depend
```

Frenzie:
You can add SHELLCHECKS to the front of the stack in .travis.yml. The reason I originally added it at the back a few months ago is because it was an allowed_fail for informational and motivational purposes. I only made it a requirement last week in #539.

Regarding the complaint, a ` -- luacheck: ignore 542` will do.

ffi/SDL2_0.lua
```diff
@@ -114,7 +114,11 @@ function S.waitForEvent(usecs)
 
 		-- if we got an event, examine it here and generate
 		-- events for koreader
-		if event.type == SDL.SDL_KEYDOWN then
+		if ffi.os == "OSX" and (event.type == SDL.SDL_FINGERMOTION or
+			event.type == SDL.SDL_FINGERDOWN or
```
Frenzie: Btw, what's the problem with the FINGERDOWN event on Mac OS?

Frenzie:
Btw, Travis (and CircleCI, should we want to switch over base as well)[1] can also do an OS X build to prevent regressions in this regard. If we simply want to duplicate all the tests we could do:

```yaml
os:
  - linux
  - osx
```

But that might be a bit much. It'd probably make more sense to do a sensible matrix setup.

https://docs.travis-ci.com/user/multi-os/#Example-Multi-OS-Build-Matrix

```yaml
matrix:
    include:
        - os: linux
          sudo: required
          env: EMULATE_READER=1 CC=gcc
        # repeat basic principle for our relevant envs
        # and some basic osx tests, only gcc and clang?
        # NB I have no idea what changes and additions might be required for that
        - os: osx
          env: EMULATE_READER=1 CC=gcc
        - os: osx
          env: EMULATE_READER=1 CC=clang
```

[1] I think our four free CircleCI containers are per organization, not per repo, so keeping the current setup might be advantageous. In any case it's a lot easier for now to keep things the way they are.

ffi/SDL2_0.lua
```diff
@@ -117,7 +117,8 @@ function S.waitForEvent(usecs)
 		if ffi.os == "OSX" and (event.type == SDL.SDL_FINGERMOTION or
 			event.type == SDL.SDL_FINGERDOWN or
 			event.type == SDL.SDL_FINGERUP) then
-			-- no op
+			-- noop for trackpad finger inputs which interfere with emulated mouse inputs
+			do end
```
Frenzie: Or that. :-)

ffi/SDL2_0.lua
chrox: Both FINGER inputs and emulated MOUSE inputs will be generated with Mac trackpad so we will just ignore the FINGER ones.

thirdparty/openssl/CMakeLists.txt
chrox: The following `depend` target will do that. Or does the first `$(MAKE) depend` do other tricks?

` set(BUILD_CMD sh -c "${BUILD_CMD} --silent depend build_crypto build_ssl >/dev/null 2>&1")`

ffi/SDL2_0.lua
Frenzie: Hah, it caught you! ;-D

```
    ffi/SDL2_0.lua:121:4: empty do..end block
```

Like I said, just stick a `-- luacheck: ignore 542` after the then. Or just writing return true or nil should do the trick as well.

chrox: Yeah, seems `luacheck` does not buy it.

thirdparty/openssl/CMakeLists.txt
Frenzie: My bad. It was just written the opposite way from how I would intuitively write it (targets first, then vars) so I stupidly overlooked it.

zhliu0124:
@chrox @Frenzie 
Thanks, guys! I successfully built KOReader on my mac, but I got these errors while I was trying to run it.

```
users-MacBook-Pro:koreader user$ ./kodev run

[*] Building KOReader…
************ Building for MACHINE: "x86_64-apple-darwin15.6.0-debug" **********
************ PATH: "/opt/local/bin:/Users/user/anaconda/bin:/usr/local/opt/qt/bin:/Users/user/anaconda/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin" **********
************ CHOST: "" **********
/Library/Developer/CommandLineTools/usr/bin/make -C base
************ Building for MACHINE: "x86_64-apple-darwin15.6.0-debug" **********
************ PATH: "/opt/local/bin:/Users/user/anaconda/bin:/usr/local/opt/qt/bin:/Users/user/anaconda/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin" **********
************ CHOST: "" **********
gcc -L/Users/user/koreader/base/build/x86_64-apple-darwin15.6.0-debug/libs -Og -g -ffast-math -pipe -fomit-frame-pointer  -march=native -fPIC -I/Users/user/koreader/base/thirdparty/luajit/build/x86_64-apple-darwin15.6.0-debug/luajit-prefix/src/luajit/src -shared -dynamiclib -undefined dynamic_lookup -o build/x86_64-apple-darwin15.6.0-debug/libs/libblitbuffer.so blitbuffer.c
error: invalid integral value 'g' in '-Og'
error: invalid integral value 'g' in '-Og'
make[1]: *** [build/x86_64-apple-darwin15.6.0-debug/libs/libblitbuffer.so] Error 1
make: *** [all] Error 2
```

PR #546, opened 2017-10-14
[Android] Build libdjvu

Frenzie:
References https://github.com/koreader/koreader/issues/1534

PR #547, opened 2017-10-15
djvu: allow for color rendering

poire-z:
See https://github.com/koreader/koreader-base/pull/546#issuecomment-336674759 and around.


Frenzie:
Wonderful. 👍

PR #548, opened 2017-10-15
Change JPG rendering to RGB in openJPGDocumentFromMem

robert00s:
This should fix incorrect rendering color in goodreads plugin picture (thumbnail)
https://github.com/koreader/koreader/blob/master/plugins/goodreads.koplugin/goodreadsbook.lua#L198


Frenzie:
@poire-z Do you know if this is safe as far as possible side effects go?

poire-z:
Should be fine, it's just invertion of BGR to RGB, like I had to do on SDL.
And check it fixes colors on both the emulator and android.

PR #550, opened 2017-10-15
[build] sdcv: change to upstream 0.5.2

Frenzie:
Deprecates our fork because we've got all improvements included upstream now.

<hr>

Tagged WIP because there are a couple of things remaining that need to be fixed, probably only for Android.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ env:
 addons:
   apt:
     packages:
+      - cmake3
       - g++-4.8
       - libsdl1.2-dev
       - luarocks
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -309,8 +309,8 @@ ifdef EMULATE_READER
 	CXXFLAGS+= $(HOST_ARCH)
 else
 	# CMake is difficult about CC and CXX, especially when it's not just a full path
-	CMAKE_C_COMPILER:=$(shell which $(CC))
-	CMAKE_CXX_COMPILER:=$(shell which $(CXX))
+	CMAKE_C_COMPILER:=$(shell which $(strip $(CC)))
+	CMAKE_CXX_COMPILER:=$(shell which $(strip $(CXX)))
 	CMAKE_C_COMPILER_ARG1:= -static-libstdc++
 	CMAKE_CXX_COMPILER_ARG1:= -static-libstdc++
 	ifdef CCACHE
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -309,8 +309,8 @@ ifdef EMULATE_READER
 	CXXFLAGS+= $(HOST_ARCH)
 else
 	# CMake is difficult about CC and CXX, especially when it's not just a full path
-	CMAKE_C_COMPILER:=$(shell which $(strip $(CC)))
-	CMAKE_CXX_COMPILER:=$(shell which $(strip $(CXX)))
+	CMAKE_C_COMPILER:=$(CC)
+	CMAKE_CXX_COMPILER:=$(CXX)
 	CMAKE_C_COMPILER_ARG1:= -static-libstdc++
 	CMAKE_CXX_COMPILER_ARG1:= -static-libstdc++
 	ifdef CCACHE
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -289,6 +289,14 @@ endif
 # in that case.
 
 ifdef EMULATE_READER
+	# CMake is difficult about CC and CXX, especially when it's not just a full path
+	CMAKE_C_COMPILER:=$(shell which $(strip $(HOSTCC)))
+	CMAKE_CXX_COMPILER:=$(shell which $(strip $(HOSTCXX)))
+	ifdef CCACHE
+		CMAKE_C_COMPILER_LAUNCHER:=$(CCACHE)
+		CMAKE_CXX_COMPILER_LAUNCHER:=$(CCACHE)
+	endif
+	# Regular
 	HOSTCC:=$(strip $(CCACHE) $(HOSTCC))
 	HOSTCXX:=$(strip $(CCACHE) $(HOSTCXX))
 	HOSTAR:=$(strip $(CCACHE) $(HOSTAR))
@@ -300,6 +308,15 @@ ifdef EMULATE_READER
 	CFLAGS+= $(HOST_ARCH)
 	CXXFLAGS+= $(HOST_ARCH)
 else
+	# CMake is difficult about CC and CXX, especially when it's not just a full path
+	CMAKE_C_COMPILER:=$(shell which $(CC))
+	CMAKE_CXX_COMPILER:=$(shell which $(CXX))
+	CMAKE_C_COMPILER_ARG1:= -static-libstdc++
+	CMAKE_CXX_COMPILER_ARG1:= -static-libstdc++
+	ifdef CCACHE
+		CMAKE_C_COMPILER_LAUNCHER:=$(CCACHE)
+		CMAKE_CXX_COMPILER_LAUNCHER:=$(CCACHE)
+	endif
 	# Don't let libtool piss on our parade...
 	# See what was mentioned around STATIC_LIBSTDCPP on ~L#93 for more details
 	# (cf. https://www.gnu.org/software/libtool/manual/html_node/Stripped-link-flags.html#Stripped-link-flags)
--- a/Makefile.third
+++ b/Makefile.third
@@ -290,19 +290,19 @@ $(OUTPUT_DIR)/sdcv: $(if $(ANDROID),$(GLIB_STATIC),$(GLIB)) $(ZLIB_STATIC) $(THI
 	cd $(SDCV_BUILD_DIR) && \
 		$(CMAKE) -DHOST="$(if $(EMULATE_READER),,$(CHOST))" \
 		$(PAGE_SIZE_CFG) -DPKG_CONFIG_PATH="$(GLIB_DIR)/lib/pkgconfig" \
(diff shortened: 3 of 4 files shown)


poire-z:
Could you please check this https://github.com/koreader/sdcv/commit/fdf7da2cbf52f40724c0146b9bd31de376901708 (Consider -u parameters to be already utf8) is no more needed (it was added here after 0.5.2 release there).
Can be tested with any "Dictionnaire de l'Acad**é**mie fran**ç**aise"": should be part of the results if at least one other dict is disabled.

Frenzie:
Either my system's sdcv or my system's dictionaries are getting in the way of that. What's the precise command I should run?

Frenzie:
Ah, you misled me. It's

> Dictionnaires de l’Académie française : 8ème édition

It returns an enormously long list of results if I run `sdcv -u "Dictionnaires de l’Académie française : 8ème édition" temps`. Is that sufficient?

poire-z:
As soon as you disable one dict, koreader should give all -u options to sdcv, ie:
`sh -c ./sdcv --utf8-input --utf8-output "-nj" "publications" --data-dir "./data/dict" -u 'Dictionnaire de l???Acad??mie Fran??aise, 8??me ??dition (1935).' -u '????????????-????????????' -u 'Petit Robert 2003'; echo`
(sorry, my terminal is latin1, but you get the idea)
KOreader gives these names as utf8 (like they should be in the .ifo), but before this patch, sdcv wanted to convert them from some "local" locale to utf8, and this failed, and so no dict with that name, and no result from those dict specified.
Better to test it from koreader to be sure (and exclude terminal encoding from this business).

(could be that i corrected the Dictionnaires in my .ifo once)

Frenzie:
I can't test it properly from my KOReader because it seems intent on using my `~/.stardict`. I can give you a Kobo build or perhaps you can just test this minor change yourself?

In other news, I accidentally deleted an `-DLDFLAGS="$(LDFLAGS)"` too many.

```
CMake Error at /home/travis/build/koreader/koreader-base/thirdparty/cmake_modules/koreader_thirdparty_common.cmake:36 (message):
  LDFLAGS variable not defined!
```

Frenzie:
This'll probably work. I'm not 100% sure all of my local other libs are the same versions and such.

[sdcv-x86_64-linux-gnu.zip](https://github.com/koreader/koreader-base/files/1385618/sdcv-x86_64-linux-gnu.zip)
[sdcv-arm-linux-gnueabihf.zip](https://github.com/koreader/koreader-base/files/1385619/sdcv-arm-linux-gnueabihf.zip)

poire-z:
Just tested on the emulator with sdcv-x86_64-linux-gnu.zip
With all my 4 dicts enabled : it works fine, showing results from the 4 dicts.
With Académie and another one disabled,: it works fine, showing results from 2/4 dicts.
With Académie enabled and another one disable: no result, and single-line error: `Internal error: map::at`

Frenzie:
Ah, so it's not my fault? I think I probably built it with #549 included.

Frenzie:
I won't see that problem though if I just run something like this:

```
$ sdcv -u "Dictionnaires de l’Académie française : 8ème édition" -u "nonsense" search
Found 1 items, similar to search.
-->Dictionnaires de l’Académie française : 8ème édition
-->anarchie

[ color="blue">ANARCHIE</font></b>.</p> n. f. Absence de gouvernement; état d'un peuple qui n'a ni chef, ni autorité, à laquelle on obéisse, ni lois auxquelles il soit soumis. <i>Tomber dans l'anarchie. Sortir de l'anarchie. Un État en proie à l'anarchie. Réprimer, dompter l'anarchie. L'anarchie des esprits,</i> La discorde, la confusion. <i>Fauteur d'anarchie.</i>]
```
(Also, anarchie similar to search? There's a stretch. :-P)

poire-z:
It's your fault that you bumbed sdcv :)
But wait, if the system locale is UTF8, it seems to work:
`LANG=en_US.utf8 koreader-emu` and my 3rd test work.

Frenzie:
See edit. I accidentally ran the wrong executable that last time. :-P

Anyway, while bumping sdcv has been a very laborious undertaking, both in getting all the patches submitted upstream (thanks @ecraven!) and working out the bulk this new cmake stuff (with some remaining Android hangups), this unforking is still a Good Thing™ because it means less work in the future.

In the end we've mostly got GitHub to thank for making contributing so much easier than sending over patch files to Sourceforge. Terrible stuff, that.

Frenzie:
Sometimes you really need a break to spot the silliest basic mistakes.

```c
set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -L$(LIBICONV_DIR)/lib -L$(GETTEXT_DIR)/lib")
```

$(VAR) is normal make. ${VAR} is cmake.

Frenzie:
Okay, I've got it all working on my computer now (i.e., I get tested and working x86_64 emultaor, Kobo ARM and Android ARM/x86 sdcv executables) but something's still off on Travis.

Frenzie:
Might have to "install" sdcv to get the RPATH if required.

https://www.semipol.de/2012/02/16/relative-rpath-settings-with-cmake.html

CMake is such a joker btw. I'm not sure what's going wrong in Travis precisely, but…

```
  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.
```

If only that worked. /scoffs

Frenzie:
@poire-z It looks like the Travis builds aren't picking up on the whole static thing, unlike my local build (including after starting from scratch). Any clue why that nightmarish combination is not playing along?

poire-z:
Dunno.
From a quick look at the log (you probably have seen it, but just in case):
In the configure-like stuff (must be cmake?), there is:
```
-- Performing Test CXX_SUPPORTS_CXX11 - Success
-- Found ZLIB: /home/travis/build/koreader/koreader-base/thirdparty/zlib/build/i686-linux-android/zlib-prefix/src/zlib/lib/libz.so (found version "1.2.8") 
```
which decided a `.so` was usable. But:
```
[0m[31m[1mLinking CXX executable sdcv
[0m/home/travis/build/koreader/koreader-base/toolchain/android-toolchain-x86/bin/../lib/gcc/i686-linux-android/4.9.x/../../../../i686-linux-android/bin/ld: error: cannot mix -static with dynamic object /home/travis/build/koreader/koreader-base/thirdparty/zlib/build/i686-linux-android/zlib-prefix/src/zlib/lib/libz.so
```
So, I guess if you want full static, you can't let it use this one, and may be the configure/cmake needs patching to know it's static and it should look for a `.a` ?
Or don't even allow it to look into this thirdparty dir?
(Just some quick ideas, haven't look at the build scripts, can't do much this evening)

Frenzie:
Yes, and on my own computer that says ".a" like I want it to. Hence all of that stupid duplication.

> Or don't even allow it to look into this thirdparty dir?

They're all in the same directory to be picked at will as it should be. This annoying CMake is "intelligent" so it doesn't find anything at all. (Or at least not the right stuff.) In spite of that I managed to pull the right switches so that it works… except for some reason not in Travis.

cmake 3.2.2 is quite a bit older than what I've got but 2.8.something should suffice.

Frenzie:
> From a quick look at the log (you probably have seen it, but just in case):

PS Otherwise I would've assumed everything was hunky dory. Because it is on my own system. But I know now to check *extremely* carefully whether CMake doesn't @#$ things up by automatically finding stuff where it shouldn't be looking in the first place.

Frenzie:
I may have found something: https://cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html

Now to see if it'll have an effect…

Frenzie:
Well, at least we got a semi-sensible error out of one of them this time around…

```
[0m/home/travis/build/koreader/koreader-base/toolchain/android-toolchain-x86/bin/../lib/gcc/i686-linux-android/4.9.x/../../../../i686-linux-android/bin/ld: error: cannot mix -static with dynamic object /home/travis/build/koreader/koreader-base/thirdparty/zlib/build/i686-linux-android/zlib-prefix/src/zlib/lib/libz.so
```
Yes, no kidding! Now quit trying to use dynamic objects. >_<

Frenzie:
Great, cmake 3.2.2 doesn't work and 3.5.1 does… I'm inclined to say forget about annoying buggy ancient CMake 3.2.x and earlier and just to make 3.5.1 the minimum since it seems to be at least almost usable. I'll contact @houqp to see if there are any alternative solutions.

poire-z:
Glad you finally worked it out :)
What exactly happened? You just upgraded cmake and it is finally using libz.a instead of libz.so ?
(We don't even see what it does when it prints `Linking CXX executable sdcv` :|)

Frenzie:
I suspected the only relevant difference was the cmake versions from the first moment of failure in Travis. It's 3.7.2 in Debian Jessie and Ubuntu 17.04, 3.9.1 in Ubuntu 17.10. But I figured it'd be a pain to use a more recent version, when it's actually as simple as picking the `cmake3` package in Ubuntu 14.04.

Of course we'll need to update the Docker images to use the `cmake3` package too, so I can't merge this just yet. It's a pity too, because updated sdcv simultaneously fixes both the Ubuntu 17.10 build[1] and the sdcv segfault on Kobo 4.6 firmware.

[1] Where you should take "build" very literally. It completes, but segfaults in crengine when it starts. I took a quick peek in gdb but nothing jumped out at me. MuPDF is fine though.

PR #557, opened 2017-10-21
sdcv patch: allow utf8 encoded or not-found -u dictnames

poire-z:
(temporary) patch to have Enabling/Disabling dictionaries working with sdcv 0.5.2.
See https://github.com/koreader/koreader/pull/3378


Frenzie:
May it be removed soon.

poire-z:
Thanks. Yes, sorry for forcing it, but as I'm going away in a few days for a week, I'd like to have a fully functional koreader with working sdcv (and a working enable/disable, for disabling some more dicts till 0.5.2 slowness gets better).
I'll check tomorrow how this patch does on kobo and android, which will allow for another iteration if it still fails.
Good news is that crengine has not yet crashed since this morning :) (or was it only on your own pc that if segfaulted and it shouldn't impact nightlys?)

Frenzie:
I mostly upgraded from Xubuntu 17.04 to 17.10 when I did to check up on the segfault report. Also I didn't bother to check if it happens when cross-compiling. (You don't hate my laptop's battery, do you? :-P) But yes, that's far away from anything on the production side of things. ;-)

I run desktop Debian Xfce stable until I get tired of it and switch to testing (which then turns into stable after a while…), laptop mostly Xubuntu du jour. Netbook Xubuntu LTS.

PR #558, opened 2017-10-21
[chore] framebuffer:setViewport ability to set multiple times

Frenzie:
References https://github.com/koreader/koreader/pull/3392

### DIFF
--- a/ffi/framebuffer.lua
+++ b/ffi/framebuffer.lua
@@ -143,19 +143,16 @@ function fb:setViewport(viewport)
     -- we might be switching the viewport on the fly, so don't overwrite self.full_bb
     if not self.full_bb then
         self.full_bb = self.bb
-    else
-        -- reset before applying new viewport
-        self.bb = self.bb:viewport(
-            -self.viewport.x, -self.viewport.y,
-            self.screen_size.w, self.screen_size.h)
     end
     self.debug("fb:setViewport() setting viewport to",
                viewport.x, viewport.y,
                viewport.w, viewport.h)
-    self.bb = self.bb:viewport(
+    self.bb = self.full_bb:viewport(
         viewport.x, viewport.y,
         viewport.w, viewport.h)
     self.viewport = viewport
+    self.full_bb:fill(Blitbuffer.COLOR_WHITE)
+    self:refreshFull()
 end
 
 function fb:calculateRealCoordinates(x, y)
--- a/ffi/framebuffer.lua
+++ b/ffi/framebuffer.lua
@@ -138,9 +138,20 @@ function fb:setViewport(viewport)
     or viewport.w < 0 or viewport.x+viewport.w > self.screen_size.w
     or viewport.h < 0 or viewport.y+viewport.h > self.screen_size.h
     then
-        error("bad viewport")
+        error("fb:setViewport() bad viewport")
     end
-    self.full_bb = self.bb
+    -- we might be switching the viewport on the fly, so don't overwrite self.full_bb
+    if not self.full_bb then
+        self.full_bb = self.bb
+    else
+        -- reset before applying new viewport
+        self.bb = self.bb:viewport(
+            -self.viewport.x, -self.viewport.y,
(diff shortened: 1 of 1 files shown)


ffi/framebuffer.lua
```diff
@@ -138,9 +138,20 @@ function fb:setViewport(viewport)
...
-        error("bad viewport")
+        error("fb:setViewport() bad viewport")
     end
-    self.full_bb = self.bb
+    -- we might be switching the viewport on the fly, so don't overwrite self.full_bb
+    if not self.full_bb then
+        self.full_bb = self.bb
+    else
+        -- reset before applying new viewport
+        self.bb = self.bb:viewport(
```
Frenzie: @poire-z I'm not sure if this is quite the right way to go, although a viewport doesn't duplicate any memory and such.

poire-z: Is there a real need to reset?
https://github.com/koreader/koreader-base/blob/0b4c48a8e3fb64024e8d7d91090744c6bb8eeb39/ffi/blitbuffer.lua#L1135-L1141
Can't you just call it (so, no `else`) on self.full_bb :  `self.bb = self.full_bb:viewport()` ?

Frenzie: I tried, but that just keeps pushing it down.

Frenzie: Pardon, that's a bit incomplete. But my test case is a top "status bar height" on the Android emulator. I'll give provide some screenshots.

Frenzie: Okay, so I was preparing my screenshots and I guess the problem I was trying to work around was caused by something incorrect in front? Or maybe it was just too late last night… but I'd swear what you said was the first thing I tried. Seems to be working now though.

There is this remaining beauty flaw. I know there's a draw white thing called somewhere…

ffi/framebuffer.lua
```diff
@@ -138,13 +138,21 @@ function fb:setViewport(viewport)
...
+        self.full_bb = self.bb
+    end
+    self.debug("fb:setViewport() setting viewport to",
+               viewport.x, viewport.y,
+               viewport.w, viewport.h)
+    self.bb = self.full_bb:viewport(
         viewport.x, viewport.y,
         viewport.w, viewport.h)
     self.viewport = viewport
+    self.full_bb:fill(Blitbuffer.COLOR_WHITE)
```
Frenzie: @poire-z Okay, it all looks good this way. Do you think these extra two lines are a self-evident addition to setViewport or some kind of overreach?

poire-z: From nowhere, I'd ask: why do you need them? You're setting a viewport on an existing full_bb, why would you need to erase it and repaint it? But if you did it, it obviously helps in some way, no?

Frenzie: You didn't see my screenshot? The existing full_bb already has content on it. It'll just remain there in perpetuity.

The only question is where these lines should be called from. Not if.

poire-z: Oh, ok, I probably don't get what viewport do (thought you had a full_bb, and viewport was the part of it that was displayed, so stuff outside would not matter).
So, you're more knowing than me to answer this question :)

Frenzie: Traditionally what you say is essentially correct.

1. We have a device, such as the H2O.
2. We set the viewport on init combined with a white fill. This is enough. The white will stay white.

But now we add:

3. We dynamically adjust the viewport on Android to adjust our display for the statusbar/toolbar. (I'd prefer to separate them out rather than just a generic "fullscreen".)

Whether this is 100% for sure the best way to do it I'm not entirely sure, but it's the easiest to implement for now and all roadblocks along the way are relevant with or without viewport (e.g., resizable SDL window).

Only the ReaderUI is currently capable of dynamic adjustment due to portrait/landscape. It'll need to be added to the FileManager. But I'll ignore that problem for now. Especially since there are some issues to be worked out on the Android side of things that I want to be as little involved in as possible.

PR #559, opened 2017-10-22
sdcv patch: allow utf8 encoded in -u dictnames (try 2)

poire-z:
Second attempt at solving this problem. See https://github.com/koreader/koreader/pull/3378#issuecomment-338478405

### DIFF
--- a/thirdparty/sdcv/sdcv.patch
+++ b/thirdparty/sdcv/sdcv.patch
@@ -1,16 +1,25 @@
 diff --git a/src/sdcv.cpp b/src/sdcv.cpp
-index 0c75eb1..f0dee2b 100644
+index 0c75eb1..da3ff4c 100644
 --- a/src/sdcv.cpp
 +++ b/src/sdcv.cpp
-@@ -62,7 +62,7 @@ using StrArr = ResourceWrapper<gchar *, gchar *, free_str_array>;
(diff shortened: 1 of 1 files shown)


Frenzie:
*cough*`./kodev release kobo`*cough*

poire-z:
I haven't set up the whole arm cross compiling stuff since I upgraded to debian stretch - avoiding that till I really need it.
And I won't ask you to launch a nightly build, so I will only be disapointed tomorrow if this doesn't work :)

Frenzie:
I can quickly compile it for you to test if you like.

Frenzie:
This ought to do it.

[sdcv-allow-utf8-encoded-in-u-kobo.zip](https://github.com/koreader/koreader-base/files/1404958/sdcv-allow-utf8-encoded-in-u-kobo.zip)

poire-z:
Thank you.
Seems to work!
Your binary is 1,1 MB vs 684 KB nightly - vs 198KB our fork in old nightlys, hope it's the same anyway, nor just stripped.
May be I somehow notice it is faster than yesterday's nightly, dunno if you build it some specific way.

Frenzie:
Nothing specific, but this is my regular desktop Debian Stretch. As you know the nightly build has a very different toolchain.

PR #560, opened 2017-10-23
sdcv patch: allow utf8 encoded in -u dictnames (try 3)

poire-z:
Force LC_ALL=en_US.UTF-8, which may be empty on Android
Added comments around previous modifications.
See https://github.com/koreader/koreader/pull/3378#issuecomment-338623549


poire-z:
Nothing against merging this ?

Frenzie:
No, go ahead.

PR #564, opened 2017-10-26
[CI] SHELLCHECKS first

Frenzie:
Also trying some other stuff.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@
 
 dist: trusty
 sudo: required
-# sudo: false
 
 service:
   - docker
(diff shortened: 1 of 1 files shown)


Frenzie:
Mac resources on Travis are too limited for free use to experiment with Mac OS, let alone actually put up as a regular test.

The very basics would add something like this, but I haven't been able to check if that's anywhere near enough. After 20 minutes the build hadn't even started yet. (And I don't mean because it was blocked by any regular Linux builds on our side.)

```yml
before_install:
  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update          ; fi
  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install nasm binutils libtool autoconf automake sdl2 lua51; fi
  - if [ "$EMULATE_READER" = 1 ]; then source .ci/before_install.sh; fi

#snip

matrix:
  include:
    - os: osx
      env:
        - EMULATE_READER=1 CC=gcc
    - os: osx
      env:
        - EMULATE_READER=1 CC=clang
```

So yeah, I'll just limit this PR to SHELLCHECKS first.

PR #571, opened 2017-11-05
C blitbuffer implementation

chrox:
The C implementation is only built for the Android platform and the Emulator. With this patch [mcode allocation failure](https://github.com/koreader/koreader/issues/1416) on Android don't degenerate the performance. And we can even disable `jit` on Android that also means we can port Koreader to iOS on which `jit` is disabled by the system.

This implementation does not change the `blitbuffer` API and is transparent to the `frontend`.

### DIFF
--- a/blitbuffer.c
+++ b/blitbuffer.c
@@ -52,25 +52,25 @@
 #define GET_BB_TYPE(bb) (((MASK_TYPE & bb->config) >> SHIFT_TYPE))
 #define SET_ALPHA_FROM_A(sbb_type, src) \
     if (sbb_type == TYPE_BB8) { \
-        Color8 *srcptr = (Color8*)(src->data) + (o_y * src->w) + o_x; \
+        Color8 *srcptr = (Color8*)(src->data + o_y * src->pitch) + o_x; \
         alpha = srcptr->a; \
     } else if (sbb_type == TYPE_BB8A) { \
-        Color8A *srcptr = (Color8A*)(src->data) + (o_y * src->w) + o_x; \
+        Color8A *srcptr = (Color8A*)(src->data + o_y * src->pitch) + o_x; \
         alpha = srcptr->a; \
     } else if (sbb_type == TYPE_BBRGB16) { \
-        ColorRGB16 *srcptr = (ColorRGB16*)(src->data) + (o_y * src->w) + o_x; \
+        ColorRGB16 *srcptr = (ColorRGB16*)(src->data + o_y * src->pitch) + o_x; \
         alpha = ColorRGB16_To_A(srcptr->v); \
     } else if (sbb_type == TYPE_BBRGB24) { \
-        ColorRGB24 *srcptr = (ColorRGB24*)(src->data) + (o_y * src->w) + o_x; \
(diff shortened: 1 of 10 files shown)


ffi/mupdf.lua
```diff
@@ -766,7 +766,7 @@ local function get_pthread()
     else
         -- Kobo devices strangely have no libpthread.so in LD_LIBRARY_PATH
         -- so we hardcode the libpthread.so.0 here just for Kobo.
-        for _, libname in ipairs({"libpthread.so", "libpthread.so.0"}) do
+        for _, libname in ipairs({"pthread", "libpthread.so.0"}) do
```
Frenzie: I saw this before and wondered about changing it, but I figured there might've been a reason for it (though I can't think of any). :-)

ffi/blitbuffer.lua
```diff
@@ -1091,9 +1173,14 @@ dim color values in rectangular area
 @param by dim by this factor (default: 0.5)
 --]]
 function BB_mt.__index:dimRect(x, y, w, h, by)
-    return self:paintRect(x, y, w, h,
-        Color8A(255, 255*(by or 0.5)),
-        self.setPixelBlend)
+    local color = Color8A(255, 255*(by or 0.5))
+    if use_cblitbuffer then
+        c = color:getColorRGB32()
```
Frenzie: I think this line should probably be removed?

blitbuffer.c
```diff
@@ -0,0 +1,940 @@
...
+                dstptr->b = srcptr->b ^ 0xFF;
+                o_y += 1;
+            }
+            o_x += 1;
+        }
+    }
+}
+
+void BB_color_blit_from(BlitBuffer *dst, BlitBuffer *src,
+        int dest_x, int dest_y, int offs_x, int offs_y, int w, int h, ColorRGB32 *color) {
```
Frenzie: nitpick: indentation

ffi/mupdf.lua
chrox: Yeah, benefit is that it can load `libpthread.dylib` on Mac now. Btw, this PR is developed solely on my Mac. Koreader is stable enough on Mac OSX now for development purpose. :)

Frenzie: Cool. :)

Btw, I wanted to look into that Travis Mac build business, but at least on Travis that's not an option due to insufficient resources. See https://github.com/koreader/koreader-base/pull/564#issuecomment-339693544 (I imagine everything would fail spectacularly as written there, but that aside. I just wanted to play around with it a bit.)

chrox: I'm not sure if the brew install stuff is sufficient. Probably there are hidden hacks on my Mac to make it work. Anyway I will verify that on a clean Mac when I have time.

poire-z:
Just curious: why did you decide to not enable it for other platform? Isn't pure C faster that JIT compiled lua?

Frenzie:
It's FFI data structures so they're faster (or should be) than Lua C API I think. Easier to maintain in any case. But the problem here is memory allocation on Android, not blitting performance.

chrox:
Actually I didn't make a real performance benchmark on the two blitbuffer implementations. The lua implementation is elegant and easy to extend. This one is intended to be used on platforms on which jit does not work.

PR #572, opened 2017-11-05
[feat] Add SDL clipboard interface

Frenzie:
References https://github.com/koreader/koreader/issues/3441

### DIFF
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -174,4 +174,16 @@ function S.waitForEvent(usecs)
 	end
 end
 
+function S.hasClipboardText()
+    return SDL.SDL_HasClipboardText()
+end
+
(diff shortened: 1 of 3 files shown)


Frenzie:
@poire-z Since you're back, any comments? :-P

poire-z:
Looks fine. But it looks to me these functions are not actively used anywhere? Or did I miss something?

Frenzie:
How could one actively use what isn't there yet? :-p

poire-z:
Not yet _PR-ly used_ then ? :)

Frenzie:
As for the actual use, my main thought right now is just to auto-copy selection, although there's also this empty button:

screenshot_2017-11-06_18-33-04

I'm not really sure if it could be sensibly added to the dictionary widget though.

PR #574, opened 2017-11-25
[build] Supports cross-compiling Android port on macOS

chrox:
We export a new env var `DARWINHOST` and the previous `DARWIN` env var specifically indicates darwin native build.


thirdparty/turbo/CMakeLists.txt
```diff
@@ -12,7 +12,13 @@ assert_var_defined(LDFLAGS)
 
 ep_get_source_dir(SOURCE_DIR)
 
-set(BUILD_CMD sh -c "$(MAKE) CC=\"${CC}\" LDFLAGS=\"${LDFLAGS}\" all")
+set(BUILD_CMD sh -c "$(MAKE) CC=\"${CC}\" CFLAGS=\"${CFLAGS}\" LDFLAGS=\"${LDFLAGS}\" all")
+
+if (APPLE AND $ENV{ANDROID})
```
Frenzie: Should that be  $ENV{DARWIN}?

chrox: The condition is evaluated to `(APPLE AND )` and throws an error. The right way should be `DEFINED ENV{ANDROID}` to check if the env var is defined.

PR #575, opened 2017-11-29
[build] Fixed zlib has no soname on arm-linux-gnueabi

chrox:
This should fix koreader/koreader#3523.


Makefile.third
```diff
@@ -276,11 +276,18 @@ $(GLIB_STATIC): $(LIBICONV) $(LIBGETTEXT) $(LIBFFI_DIR)/include $(THIRDPARTY_DIR
...
+ifdef ANDROID
 	cd $(ZLIB_BUILD_DIR) && \
 		$(CMAKE) -DCC="$(CC)" -DCHOST="$(CHOST)" \
-		-DLDFLAGS="$(LDFLAGS) $(if $(ANDROID),-shared $(ZLIB_LDFLAGS),) " \
+		-DLDFLAGS="$(LDFLAGS) -shared $(ZLIB_LDFLAGS)" \
 		$(CURDIR)/thirdparty/zlib && \
 		$(MAKE)
+else
+	cd $(ZLIB_BUILD_DIR) && \
+		$(CMAKE) -DCC="$(CC)" \
```
Frenzie: CHOST is only necessary for Android? :-)

chrox: If we specify CHOST we have to also specify the soname otherwise it will cause #3523. The CHOST is only necessary for Mac cross-compiling of Android port.

PR #578, opened 2017-12-20
Support tar in fz_archive_needs_password

TnS-hun:
fz_archive_needs_password didn't check the type of the archive, made an unsafe downcasting to fz_zip_archive. The result was a password prompt for tars (cbts).

This will be needed for https://github.com/koreader/koreader/issues/3533.


Frenzie:
Thanks!

PR #579, opened 2017-12-29
cblitbuffer: fix alpha in some bb conversions, and add a few more

poire-z:
This is about the cblitbuffer code used on the emulator (and android, although for some reason, I didn't notice the following problems on my android phone with the latest nightly...) and the new bbtypes we may get from the new mupdf 1.12.

Mupdf 1.8 used to always return BB8A or RGB32 from renderImage code.
Mupdf 1.12 now can return BB8 (from a greyscale jpeg) and RGB24 (from a color jpeg).
When viewing images from epub with ImageViewer, they would be fine when not scaled. But when scaled, they need to be blitted to a RGB32, and the cblitbuffer code would blit these BB8 or RGB24 with a full transparency, making the image full white. This sets a correct alpha of 0xFF so that no transparency is added (as BB8 and RGB24 do not carry any alpha info).
@chrox : do these additions look ok to you? (I'm not really at ease with all this alpha stuff - and I guess there may be other places that need this kind of alpha fixes - but these were the only ones I could somehow test).

Also added support for alpha_blitting from BB8 (greyscale jpeg) to RGB32, as this can happen when viewing greyscale jpeg cover images, and it would otherwise cause a crash.

Also added a note in mupdf.lua about a regression in Mupdf1.12 with some jpeg images:
https://upload.wikimedia.org/wikipedia/commons/0/0e/Death_of_Darius_by_Andre_Castaigne_cropped.jpg would display fine, but the resized (by wikipedia servers) version https://upload.wikimedia.org/wikipedia/commons/thumb/0/0e/Death_of_Darius_by_Andre_Castaigne_cropped.jpg/440px-Death_of_Darius_by_Andre_Castaigne_cropped.jpg is displayed as:
bad
Tested that this happens too with the mupdf 1.12 windows applet, while the 1.8 windows applet displays it fine (so, it's not some libjpeg mismatch on our side).

### DIFF
--- a/blitbuffer.c
+++ b/blitbuffer.c
@@ -509,6 +509,7 @@ void BB_blit_to_BB32(BlitBuffer *src, BlitBuffer *dst,
                 dstptr->r = srcptr->a;
                 dstptr->g = srcptr->a;
                 dstptr->b = srcptr->a;
+                dstptr->alpha = 0xFF;
                 o_y += 1;
             }
             o_x += 1;
@@ -524,6 +525,7 @@ void BB_blit_to_BB32(BlitBuffer *src, BlitBuffer *dst,
                 dstptr->r = srcptr->a;
                 dstptr->g = srcptr->a;
                 dstptr->b = srcptr->a;
+                dstptr->alpha = srcptr->alpha; // if bad result, try: 0xFF - srcptr->alpha
                 o_y += 1;
             }
             o_x += 1;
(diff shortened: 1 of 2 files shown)


poire-z:
It happened that Mupdf 1.12 can also return a BB8A (from a greyscale jpeg), so allow for alpha_blitting them too rather than crashing.

poire-z:
Also fix MuPDF.renderImage() as MuPDF 1.12 fz_scale_pixmap() behave strangely (on ARM only!) if it is given non-integer width and height, and returns a corrupted image with a different ncomp than original pixmap...)
No problem on the emulator, but reproducible on Kobo and my Android phone.
Looks like there is some ARM assembly code in [source/fitz/draw-scale-simple.c](https://github.com/ArtifexSoftware/mupdf/blob/7307a7f232b27b22176b8f68364e43e09ce393cc/source/fitz/draw-scale-simple.c#L557) that may not like being given floats instead of integers.

PR #581, opened 2017-12-31
MuPDF document loading from memory, Noto Sans support

TnS-hun:
MuPDF related changes:
* support loading HTML from a string
* layout HTML for a given resolution
* use Noto Sans font when requested

Needed for https://github.com/koreader/koreader/issues/1776

### DIFF
--- a/ffi-cdecl/mupdf_decl.c
+++ b/ffi-cdecl/mupdf_decl.c
@@ -86,10 +86,12 @@ cdecl_struct(fz_document_s)
 cdecl_struct(fz_page_s)
 
 cdecl_func(mupdf_open_document)
+cdecl_func(mupdf_open_document_with_stream)
 cdecl_func(fz_needs_password)
 cdecl_func(fz_authenticate_password)
 cdecl_func(fz_drop_document) // NOTE: libk2pdfopt uses old fz_free_document symbol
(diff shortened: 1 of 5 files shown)


poire-z:
Looks OK to me.

May be you could drop a few (all?) of the empty lines in mupdf.lua 's `mupdf.openDocumentFromText` , for a consistent style with the other functions ?
(If you do, could you also remove the line `M.fz_keep_image(context(), image)` in `mupdf.renderImage()` ? It's a line from 2014, and I think it may introduce some memory leaks, making the following `M.fz_drop_image` actually not releasing it, and everything seem to work fine without it.)

Frenzie:
At a quick skim it seems fine to me too. Also pinging @houqp @chrox @NiLuJe

TnS-hun:
> May be you could drop a few (all?) of the empty lines in mupdf.lua 's mupdf.openDocumentFromText , for a consistent style with the other functions ?

I removed two empty lines, if I remove more then the formatting will be out of sync with openDocument. :)

> (If you do, could you also remove the line M.fz_keep_image(context(), image) in mupdf.renderImage() ? 

No problem.

poire-z:
Merging, as it's mostly new code that's not yet used by current code.

PR #585, opened 2018-01-07
MuPDF font fixes

TnS-hun:
* fall back to FreeSans.ttf if needed
* font filenames were not freed everywhere
* font properties (bold, italics, ...) were not set

### DIFF
--- a/thirdparty/mupdf/external_fonts.patch
+++ b/thirdparty/mupdf/external_fonts.patch
@@ -11,20 +11,47 @@ index 114a6eb6..73880a64 100644
  	$(XPS_OBJ) \
  	$(SVG_OBJ) \
 diff --git a/include/mupdf/fitz/font.h b/include/mupdf/fitz/font.h
-index ef4cd74d..9dfc8ed8 100644
+index ef4cd74d..6e6ea124 100644
 --- a/include/mupdf/fitz/font.h
 +++ b/include/mupdf/fitz/font.h
-@@ -624,4 +624,8 @@ void fz_hb_lock(fz_context *ctx);
+@@ -624,4 +624,9 @@ void fz_hb_lock(fz_context *ctx);
  */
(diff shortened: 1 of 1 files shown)


thirdparty/mupdf/external_fonts.patch
```diff
@@ -379,23 +404,25 @@ index be69ce1c..a24b2122 100644
...
  			fz_throw(ctx, FZ_ERROR_GENERIC, "cannot find substitute font");
  
  		fontdesc->font = fz_new_font_from_memory(ctx, fontname, data, len, 0, 1);
 +#else
-+		char * filename = pdf_lookup_substitute_font_from_file(ctx, mono, serif, bold, italic);
++		char *filename = pdf_lookup_substitute_font_from_file(ctx, mono, serif, bold, italic);
 +		if (!filename)
 +			fz_throw(ctx, FZ_ERROR_GENERIC, "cannot find substitute font");
-+
++
```
Frenzie: I feel like a total nitpicker for pointing out this spare tab but GH kinda highlights it in the diff. >_>

TnS-hun: I should turn on trailing whitespace trimming. :)

poire-z:
This PR introduced using FreeSans.ttf as a fallback font (and there's been no related change since).
I had a try at using FreeSerif.ttf, and it brings proper arabic or hebrew in footnote popups and HTML dict lookups, while still not breaking chinese in footnotes popups (so, I guess CJK glyphs are looked up in the other CJK font the external_fonts.patch registers).
As this PR introduced FreeSans.ttf in our MuPDF to make HTML dicts work, I guess it has no effect on PDF files, so the risk of breaking/changeing PDF reading is low.
So, should we switch from FreeSans.ttf to FreeSerif.ttf ?

Frenzie:
Sure, why not?

PR #586, opened 2018-01-16
Added util.runInSubProcess() and related functions

poire-z:
These were already used in coverbrowser plugin, and are a bit extended here to support communication via pipe. All this work on my Android 6.0.1, so I expect them to work on all Android as it's all basic unix stuff.
Needed to implement Trapper:dismissableRunInSubprocess() and have wikipedia queries interruptible.

Also removed the restriction `not isAndroid()` in util.getNonBlockingReadSize(), as I believe it should work on all android.


poire-z:
Thanks. Merging now so I can bump base in an upcoming koreader PR.

PR #587, opened 2018-01-21
util.runInSubProcess(): give child its own process group

poire-z:
`util.runInSubProcess()`: give child its own process group and use `kill(-pid)` in `terminateSubProcess `to kill child and its children.
Should allow us to safely run `os.execute()` in sub-process.
(No current project in which I would need that, I just had a try with `os.execute("sleep 60")` and it has a few side effects that this solves.)


Frenzie:
Interesting idea.

PR #590, opened 2018-01-28
Fix some bb conversion so that white stays white

poire-z:
Blitting pure white from RGB32 and RGB16 to BB8 (which is the framebuffer format on Kindle) would result in white becoming some light gray (we got 0xFE instead of 0xFF). (Former values from #93.)
See https://github.com/koreader/koreader/issues/3631
(Pushing first a commit with some boring unit tests that should fail till next commit)


spec/unit/blitbuffer_spec.lua
```diff
@@ -31,6 +50,191 @@ describe("Blitbuffer unit tests", function()
             local c4u_32 = cRGB32:getColor4U()
             assert.are.equals(0xB0, c4u_32.a)
         end)
+
+        it("should have pure white stay pure white when converting", function()
+            print("Converting from RGB32")
```
Frenzie: I think it would be slightly nicer to write that as follows, don't you? :-)

```lua
describe("should have pure white stay pure white when converting", function()
    it("from RGB32"), function()
        -- etc
```

Frenzie: PS Very nice!

poire-z: I do (this unit test was obviously dedicated to you, as you would have requested one :) - and seeing if other conversions need a fix.)
(Can't really test the whole describe/it stuff, so hoping for no typo.)

poire-z:
I won't fix the few tests commented out among the unit tests: as explained above each: a simple method is used and loses the bit we changed - I'd rather not make things more complicated and risk side effects.

These failing tests mean that on our kobo with RGB16 framebuffer, light RGB24 gray does become white. But I had never guessed and have never been bothered.

spec/unit/blitbuffer_spec.lua
poire-z: Does the white dot (instead of green dot) on a unique line means anything?
Blitbuffer unit tests Color conversion should have pure white stay pure white when converting from RGB8 ... ●

poire-z: OK, no more white dot on last travis run.

Frenzie:
I'm not really sure what our levels of gray on E Ink correspond to but something as subtle as FFFFFE should in all likelihood turn into just white when you have only 16 levels of grayscale available. For example, I just did this test on Windows 98, one of the easiest ways I know to play around with 16 colors (which is something quite different from 16 shades of gray, but what can I do):

virtualbox_windows 98 se_28_01_2018_20_05_26

PS VirtualBox 5.2 is so. much. faster. (Not for Windows 98 obviously; I'm talking about King Slow Windows 8/10/Ubuntu 17.10.)

Frenzie:
PS You can clearly see Windows 98 doing some dithering, but that aside. ;-)

PR #595, opened 2018-02-01
cre.findText(): fix inconsistencies in results

poire-z:
See details in https://github.com/koreader/koreader/pull/3656 and https://github.com/koreader/crengine/pull/94.

Don't use `rc.height()` for the height to search in, as it may be small for some pages when there is a page-break (it would then be the height of the rendered text only) - and we would only search in this area of the other page, missing occurences in the visible but not searched other area.
Better to use the view height.

Don't use `+1` or `-1` with `rc.bottom` or `rc.top`, as there may be no node at this +/-1 location, and search would abort. So, use them directly.

Still, even when using them directly, we may miss some words at edges of page, and even some words not at edge of pages, but included in a node/tag that is itself at edge of page or spanning previous or
next page.
The surest way to not miss any such edge cases is to not do what it does currently:
 - forward: find next page's top `y`, search in `y ⇒ y + pageHeight`
 - reverse: find prev page's bottom, seach in `y - pageHeight ⇒ y`

but to include current page and search in 3*pageHeight:
 - forward: search from current page's top : search in `y ⇒ y + 3*pageHeight`
 - reverse: search from current page's bottom: search in `y ⇒ y - 3*pageHeight`

These `3 x pageHeight` areas are:
  1) the current page that may have part of the occurence, or a node spanning next/previous page,
  2) the next/previous page where we may find our interesting occurences
  3) the page after/before that one, where some node shown on 2) may start in

This will return more results, including those on current page, but we can filter them in `readersearch.lua` (which we would have to do anyway even with current algorithm, as edge cases could give results on current page and get us blocked on this page, unable to search backwards).

It may be a little bit more expensive, but that's negligeable:
- if occurence in next page, instead of parsing 1 page, we'll parse 3.
- if occurence in page 20 after current, instead of parsing 20 page, we'll parse 22.

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -1382,7 +1382,25 @@ static int findText(lua_State *L) {
     LVArray<ldomWord> words;
     lvRect rc;
     doc->text_view->GetPos( rc );
-    int pageHeight = rc.height();
+    int pageHeight = doc->text_view->GetHeight();
+    // To not miss any occurence, and have them all highlighted,
+    // we need to search in 3*pageHeight:
+    // 1) the current page that we may include in start-end below,
+    // 2) the next/previous page where we may find our interesting occurences
+    // 3) the page after/before that one, where some node shown on 2) may start or end in
+    // (We only need to use "2 * pageHeight" here, because current page (1) is
+    // not accounted for searchHeight, thanks to the searchHeightCheckStartY
+    // we set below when needed)
+    int searchHeight = 2 * pageHeight;
+    // And to not be stopped from searching further if there is a hit on current page,
+    // but none on 2) & 3): let crengine not start ensuring searchHeight before
+    // we pass beyond current page.
+    int searchHeightCheckStartY = -1;
+    // So, below, we include current page when needed, and adjust searchHeightCheckStartY
+    // accordingly.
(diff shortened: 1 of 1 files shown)


poire-z:
There was one unit test that failed in one build (gcc no debug) but not the others (gcc debug, clang):
```
KOPTContext module should convert koptcontext to/from table
./spec/base/unit/koptcontext_spec.lua:180: Expected objects to be the same.
Passed in:
(string) ''
Expected:
(string) 'eng'
```
Strange (this PR code should have no impact). (Unless I poke into other libraries' data - which I think I don't - but just curious: do these unit tests each run in a single luajit process, or are there multiple launched for each test?)

Frenzie:
Just a single one.

poire-z:
Any comment on this?
I'd like to trigger a rebuild (would push a comment change commit) just to have unit tests run again, but I guess it will fail like this morning nightlys as http://pkgs.fedoraproject.org/repo/pkgs/mingw-zlib/ is unavailable.

Frenzie:
Since you have write access you should be able to restart the job directly from Travis. I triggered a rebuild yesterday but apparently it had a Travis illegal instruction fluke instead.

PR #598, opened 2018-02-07
getLinkFromPosition(), getPageLinks(): better xpointer for link origin

poire-z:
Bumps crengine for the new `getHRef(a_xpointer)` methods, see https://github.com/koreader/crengine/pull/96
Follow up to #597.
Should still have no impact on current code that does not use it.
I just want it in a nightly so I can test on my kobo for some time a followup to https://github.com/koreader/koreader/pull/3202 (_I would also have liked to have that when back from link, to again see where we were before jumping to link, but that's more complicated_)

I shall use that like this:
```lua
-        local link = self.ui.document:getLinkFromPosition(ges.pos)
-        if link ~= "" then
-            return link
+        local link_xpointer, a_xpointer = self.ui.document:getLinkFromPosition(ges.pos)
+
+        if a_xpointer and not self:isXpointerCoherent(a_xpointer) then
+            -- On some documents (at least one Wikipedia saved as EPUB), crengine
+            -- may give wrong a_xpointer, that points to some other <A> element
+            -- in the same parapgraph. If followed then back, we could get to a
+            -- different page. So, we check here how valid it is, and if not,
+            -- we just discard it so that addCurrentLocationToStack() is used.
+            a_xpointer = nil
+        end
+
+        if link_xpointer ~= "" then
+            return {
+                -- this link's target xpointer is more precise than a classic
+                -- xpointer to top of a page: we can show a marker at its
+                -- y-position in target page
+                xpointer = link_xpointer,
+                marker_xpointer = link_xpointer,
+                from_xpointer = a_xpointer,
+            }
[...]
+                    table.insert(self.location_stack, {
+                            xpointer = link.from_xpointer,
+                            marker_xpointer = link.from_xpointer,
+                        }

```
with:
```lua
--- Check if xpointer really points to itself
function ReaderLink:isXpointerCoherent(xpointer)
    -- Get screen coordinates of xpointer
    local doc_margins = self.ui.document._document:getPageMargins()
    local doc_y, doc_x = self.ui.document:getPosFromXPointer(xpointer)
    local top_y = self.ui.document:getCurrentPos()
    -- (strange, but using doc_margins.top is accurate even in scroll mode)
    local screen_y = doc_y - top_y + doc_margins["top"]
    local screen_x = doc_x + doc_margins["left"]
    logger.dbg("getPosFromXPointer: doc_y  ", doc_y)
    logger.dbg("getCurrentPos:      top_y -", top_y)
    logger.dbg("doc_margins[top] margin_y +", doc_margins["top"])
    logger.dbg("                 screen_y =", screen_y)
    logger.dbg("getPosFromXPointer: doc_x  ", doc_x)
    logger.dbg("doc_margins[left]margin_x +", doc_margins["left"])
    logger.dbg("                 screen_x =", screen_x)
    -- Get again link and a_xpointer from this position
    local re_link_xpointer, re_a_xpointer = self.ui.document:getLinkFromPosition({x = screen_x, y = screen_y})
    logger.dbg("# or_a_xpointer:", xpointer)
    logger.dbg("# re_a_xpointer:", re_a_xpointer)
    logger.dbg("# re_link_xpointer:", re_link_xpointer)

    -- We should get the same a_xpointer. If not, crengine has messed up
    -- and we should not trust this xpointer to get back to this link.
    if re_a_xpointer ~= xpointer then
        logger.info("xpointer is NOT coherent")
        return false
    end
    logger.info("good xpointer")
    return true
end
```

(I'll merge it later tonight if no one stops me, so I can have it in next nightly :)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -523,7 +523,11 @@ static int getPosFromXPointer(lua_State *L) {
 	}
 	lua_pushinteger(L, pos);
 
-	return 1;
+	// Also returns the x value (as the 2nd returned value, as its
+	// less interesting to current code than the y value)
+	lua_pushinteger(L, pt.x);
+
+	return 2;
 }
 
 static int getCurrentPos(lua_State *L) {
@@ -947,10 +951,11 @@ static int getLinkFromPosition(lua_State *L) {
 
 	lvPoint pt(x, y);
 	ldomXPointer p = doc->text_view->getNodeByPoint(pt, true);
-	lString16 href = p.getHRef();
+	ldomXPointer a_p;
+	lString16 href = p.getHRef(a_p);
 	lua_pushstring(L, UnicodeToLocal(href).c_str());
-	if (!p.isNull()) {// return position's xpointer too
-		lua_pushstring(L, UnicodeToLocal(p.toString()).c_str());
+	if (!a_p.isNull()) { // return xpointer to <a> itself
+		lua_pushstring(L, UnicodeToLocal(a_p.toString()).c_str());
 		return 2;
 	}
 	return 1;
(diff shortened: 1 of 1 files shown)


poire-z:
Thanks !

PR #601, opened 2018-02-27
cre: Allow for increasing some crengine cache sizes

poire-z:
Bump crengine for cache re-use fix and warnings when cache sizes reached.
Full details in https://github.com/koreader/crengine/pull/100

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -610,10 +610,12 @@ static int walkTableOfContent(lua_State *L, LVTocItem *toc, int *count) {
 		lua_pushnumber(L, toc_tmp->getPage()+1);
 		lua_settable(L, -3);
 
+		// Note: toc_tmp->getXPointer().toString() and toc_tmp->getPath() return
+		// the same xpath string. But when just loaded from cache, the XPointer
+		// is not yet available, but getPath() is. So let's use it, which avoids
+		// having to build the XPointers until they are needed to update page numbers.
 		lua_pushstring(L, "xpointer");
 		// lua_pushstring(L, UnicodeToLocal( toc_tmp->getXPointer().toString()).c_str());
-		// As an optimisation, we use _path, which is available when loading from
-		// cache, so we can delay the expensive ->getXPointer() till next rendering
 		lua_pushstring(L, UnicodeToLocal(toc_tmp->getPath()).c_str());
 		lua_settable(L, -3);
 
--- a/cre.cpp
+++ b/cre.cpp
@@ -302,10 +302,15 @@ static int initCache(lua_State *L) {
     if (lua_isboolean(L, 3)) {
         compress_cached_data = lua_toboolean(L, 3);
     }
+    float storage_max_uncompressed_size_factor = (float)luaL_optnumber(L, 4, 1.0);
 
     // Setting this to false uses more disk space for cache,
     // but speed up rendering and page turns quite a bit
     compressCachedData(compress_cached_data);
+
+    // Increase the 4 hardcoded TEXT_CACHE_UNPACKED_SPACE, ELEM_CACHE_UNPACKED_SPACE,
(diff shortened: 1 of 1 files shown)


cre.cpp
```diff
@@ -606,9 +611,10 @@ static int walkTableOfContent(lua_State *L, LVTocItem *toc, int *count) {
 		lua_settable(L, -3);
 
 		lua_pushstring(L, "xpointer");
-		lua_pushstring(L, UnicodeToLocal(
-							toc_tmp->getXPointer().toString()).c_str()
-							);
+		// lua_pushstring(L, UnicodeToLocal( toc_tmp->getXPointer().toString()).c_str());
+		// As an optimisation, we use _path, which is available when loading from
+		// cache, so we can delay the expensive ->getXPointer() till next rendering
```
Frenzie: Does that imply cache will always be available for `walkTableOfContent`?

poire-z: No.
(Just had a quick check that TOC works with small books and no cache).

cre.cpp's `getTableOfContent()` calls crengine `doc->text_view->getToc()` before using `walkTableOfContent()` (which then takes some properties from each TocItem to make them available to lua).
```c
LVTocItem * LVDocView::getToc() {
        if (!m_doc)
                return NULL;
        if (!m_doc->isTocFromCacheValid())
            // Avoid expensive calls to getXPointer() when the toc items' _path
            // and _page are available and valid when just loaded from cache
            updatePageNumbers(m_doc->getToc());
        return m_doc->getToc();
}
```
(I guess this comment should be above the `if`)

So, `getToc()` will always call `updatePageNumbers()` (except when just loaded from cache).
`upadtePageNumbers()` calls:
```c
ldomXPointer LVTocItem::getXPointer()
{
    if ( _position.isNull() && !_path.empty() ) {
        _position = _doc->createXPointer( _path );
        if ( _position.isNull() ) {
            CRLog::trace("TOC node is not found for path %s", LCSTR(_path) );
        } else {
            CRLog::trace("TOC node is found for path %s", LCSTR(_path) );
            // CRLog::trace("           gives xpointer: %s", UnicodeToLocal(_position.toString()).c_str());
        }
    }
    return _position;
}
```
which is just there, if `_position` is null, to uses the `_path` (a xpath string, from the epub toc.xml file) and make a `XPointer` object out of it, stored as `_position`.
It turns out that `_position.toString()` just returns `_path` :) (at least, on the few tests i did with the commented `CRLog:trace` above)
Then, it just builds the other properties (_page, _percent).
```c
void LVDocView::updatePageNumbers(LVTocItem * item) {
        if (!item->getXPointer().isNull()) {
                lvPoint p = item->getXPointer().toPoint();
                int y = p.y;
                int h = GetFullHeight();
                int page = getBookmarkPage(item->_position);
                if (page >= 0 && page < getPageCount())
                        item->_page = page;
                else
                        item->_page = -1;
                if (y >= 0 && y < h && h > 0)
                        item->_percent = (int) ((lInt64) y * 10000 / h); // % * 100
                else
                        item->_percent = -1;
        } else {
                //CRLog::error("Page position is not found for path %s", LCSTR(item->getPath()) );
                // unknown position
                item->_page = -1;
                item->_percent = -1;
        }
        for (int i = 0; i < item->getChildCount(); i++) {
                updatePageNumbers(item->getChild(i));
        }
}
```
So, no change when no cache.

When saving to cache, the XPointer object can't be serialized (it's an object), but all the other properties can (they are strings or numbers).
When loading from cache, we get them back, but not the _position Xpointer. But _path is there, and the page/percent properties are still valid, so we can use them as is.
When re-rendering (font size changes, layout changes), _path and _position XPointer do not change, and they are then used to pick up the new _page and _percent. This is expensive. At this point, a XPointer object, if not already present, will be made from _path.
When no cache, the XPointer is there since the loading of the book, and then used anytime it is needed.

Frenzie:
Alright, merge when ready I'd say. Maybe move the comment slightly. :-)

PR #609, opened 2018-03-11
cre: added getGammaLevel()

poire-z:
So we can show in the Notification the real gamma value (0.8, 1, 2.5, 15) instead of the index (5, 15, 31, 56) of the possible values array.

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -394,6 +394,12 @@ static int setStringProperty(lua_State *L) {
     return 0;
 }
 
+static int getGammaLevel(lua_State *L) {
+	lua_pushnumber(L, fontMan->GetGamma());
+
+	return 1;
+}
+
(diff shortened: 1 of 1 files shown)


cre.cpp
```diff
@@ -394,6 +394,12 @@ static int setStringProperty(lua_State *L) {
     return 0;
 }
 
+static int getGammaLevel(lua_State *L) {
+	lua_pushnumber(L, fontMan->GetGamma());
```
Frenzie: Pssht, these are tabs. I realize the document is a mixture but I'd say to stick with our style guide for new stuff. ;-)

poire-z: They were tabs in the 2 functions related to Gamma just below - just cut and pasted from them.

poire-z:
Can you cancel the CI still running for earlier commit?

PR #614, opened 2018-03-18
[feat] SDL2: resize framebuffer implementation

Frenzie:
@poire-z There seems to be some kind of memory leak present (unless it's something in front), which is not the biggest issue for what's more of an experimental proof of concept, but I might be missing something obvious.

In front I'm not quite finished with the proof of concept, but basic resizing works as you'd expect (besides the memory leak).

```diff
diff --git a/frontend/device/input.lua b/frontend/device/input.lua
index f595d25c..76de4578 100755
--- a/frontend/device/input.lua
+++ b/frontend/device/input.lua
@@ -345,7 +345,33 @@ function Input:handleKeyBoardEv(ev)
 end
 
 function Input:handleMiscEv(ev)
-    -- should be handled by a misc event protocol plugin
+    -- bit of a hack for passing SDL window resize events
+    local SDL_WINDOWEVENT_RESIZED = 5
+    local x = 0
+    local y = 1
+print("FDSFSDFSDFSDFSD")
+logger.dbg(ev)
+    if ev.code == x then
+        self.new_x = ev.value
+    elseif ev.code == y then
+        self.new_y = ev.value
+    elseif ev.code == SDL_WINDOWEVENT_RESIZED then
+    print("attempting to resize")
+    self.device.screen.screen_size.w = self.new_x
+    self.device.screen.screen_size.h = self.new_y
+        self.device.screen.resize(self.device.screen, self.new_x, self.new_y)
+
+print(self.device.screen:getScreenWidth())
+
+return Event:new("SetDimensions", self.device.screen:getSize())
+    end
 end
```

### DIFF
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -66,8 +66,8 @@ function S.open()
 end
 
 function S.createTexture(w, h)
-    local w = w or S.w
-    local h = h or S.h
+    w = w or S.w
+    h = h or S.h
 
     return SDL.SDL_CreateTexture(
         S.renderer,
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -62,11 +62,22 @@ function S.open()
     )
 
     S.renderer = SDL.SDL_CreateRenderer(S.screen, -1, 0)
-    S.texture = SDL.SDL_CreateTexture(
-                    S.renderer,
-                    SDL.SDL_PIXELFORMAT_ABGR8888,
-                    SDL.SDL_TEXTUREACCESS_STREAMING,
-                    S.w, S.h)
+    S.texture = S.createTexture()
+end
+
+function S.createTexture(w, h)
+    local w = w or S.w
+    local h = h or S.h
+
+    return SDL.SDL_CreateTexture(
+        S.renderer,
+        SDL.SDL_PIXELFORMAT_ABGR8888,
+        SDL.SDL_TEXTUREACCESS_STREAMING,
+        w, h)
+end
+
+function S.destroyTexture(texture)
+    SDL.SDL_DestroyTexture(texture)
 end
 
 -- one SDL event can generate more than one event for koreader,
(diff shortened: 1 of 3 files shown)


Frenzie:
A couple of screenshots:

## Just opened


## Randomly resized


## Maximized

ffi/SDL2_0.lua
```diff
@@ -170,6 +181,19 @@ function S.waitForEvent(usecs)
...
+        elseif event.type == SDL.SDL_WINDOWEVENT
+            and (event.window.event == SDL.SDL_WINDOWEVENT_RESIZED
+                 or event.window.event == SDL.SDL_WINDOWEVENT_SIZE_CHANGED) then
+            local x = 0
+            local y = 1
+            local new_size_x = event.window.data1
+            local new_size_y = event.window.data2
+
+            if new_size_x and new_size_y then
+                genEmuEvent(ffi.C.EV_MSC, SDL.SDL_WINDOWEVENT_RESIZED, 0)
```
Frenzie: This should go at the bottom, not the top.

ffi/framebuffer_SDL2_0.lua
```diff
@@ -33,6 +22,44 @@ function framebuffer:init()
...
 end
 
+function framebuffer:resize(w, h)
+    local w = w or SDL.w
+    local h = h or SDL.h
+
+    if not self.dummy then
+        self:_newBB(w, h)
+    else
+        self.bb = BB.new(600, 800)
```
poire-z: In that else (if it ever happens), aren't the `self.bb` and `self.sdl_bb` (otherwise freed by _newBB()) not freed ?
Nothing else obvious for the cause of memory leaks...

Frenzie: The else is just for unit tests, but true. I was thinking it should just be a no-op and then I changed my mind 'cause maybe you'd want to do a unit test on some resizing type stuff and then I apparently ended up doing neither in my actual commit. >_>

poire-z:
No such memory leak when you switch portrait <> landscape ?

ffi/framebuffer_SDL2_0.lua
Frenzie: Actually I said that as if I were sure, but I don't know if this self.dummy is even used anywhere.

Frenzie:
> No such memory leak when you switch portrait <> landscape ?

No, (un?)fortunately not. But I found the culprit: I missed the invert_bb.

ffi/SDL2_0.lua
```diff
@@ -170,6 +181,19 @@ function S.waitForEvent(usecs)
             genEmuEvent(ffi.C.EV_SYN, ffi.C.SYN_REPORT, 0)
         elseif event.type == SDL.SDL_MULTIGESTURE then -- luacheck: ignore 542
             -- TODO: multi-touch support
+        elseif event.type == SDL.SDL_WINDOWEVENT
+            and (event.window.event == SDL.SDL_WINDOWEVENT_RESIZED
+                 or event.window.event == SDL.SDL_WINDOWEVENT_SIZE_CHANGED) then
+            local x = 0
```
Frenzie: @poire-z I should probably call this w and h, right?

poire-z: yep

Frenzie:
@poire-z With the memory issue resolved, any further comments besides the ones I'll address within the next few minutes.?

ffi/SDL2_0.lua
poire-z: btw, why w=0 and h=1 (and not both 0 or 1) ?

poire-z:
Nope. You can merge and bump both PRs when you feel like it (to avoid CI overload).

ffi/SDL2_0.lua
Frenzie: For front I pretend it's some kind of custom kernel event.

```lua
-- type (MSC), property (w), value (new_size_w)
genEmuEvent(ffi.C.EV_MSC, w, new_size_w)
```

Would you rather I try passing a non-standard value (like a serialized Lua string) so it can be done as a single event instead of three separate events?

poire-z: No no, no pb with that.
Just curious why `w=0, h=1` . And why not the more symetrical `w=1, h=1`   or   not `w=0, h=0`

poire-z: oh, ok, these are just codes, 0 means width, 1 means height - and not actual width and height !

PR #615, opened 2018-03-21
[feat] SDL2: framebuffer:setWindowTitle()

Frenzie:
References https://github.com/koreader/koreader/issues/3713

### DIFF
--- a/ffi/framebuffer.lua
+++ b/ffi/framebuffer.lua
@@ -306,6 +306,18 @@ function fb:setScreenMode(mode)
     end
 end
 
+function fb:getWindowTitle()
+    return self.window_title
+end
+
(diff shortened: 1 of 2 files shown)


Frenzie:
@poire-z Any comments?

poire-z:
None, looks proper and fine.
(you could have saved some electrons by including it in the crengine bump :)

Frenzie:
Well, I'm still hoping to figure out Mac OS without putting in any actual effort but new roadblocks keep presenting themselves: https://github.com/koreader/koreader/pull/3787

virtualbox_macos_21_03_2018_18_31_02

I'm stuck in czmq, which is unfortunately earlier in the build process than @cramoisi's glib issue that I actually wanted to investigate.

PR #616, opened 2018-03-21
[fix] MuPDF Mac Build

Frenzie:
References https://github.com/koreader/koreader/issues/3783


cramoisi:
https://github.com/koreader/koreader/issues/3783#issuecomment-374935328

PR #620, opened 2018-03-22
git error for libk2pdfopt

cramoisi:


Frenzie:
See https://github.com/koreader/koreader/pull/3094#issuecomment-323542865 for context.

PR #621, opened 2018-03-22
prevent error behind firewall

cramoisi:
ok it was the last !


Frenzie:
And with any luck this will fix the Docker Kobo/Kindle/PocketBook builds randomly failing with the same libffi FTP as well. :-)

PR #622, opened 2018-03-22
[feat] SDL2: drop file handling

Frenzie:


Frenzie:
@poire-z Btw, we should bump crengine here before I bump base in front.

poire-z:
You will include crengine bump in this PR?
Or you want me to do a quick independant one?

Frenzie:
I'll include it as a second commit in this one. But no squashing then. ;-)

PR #628, opened 2018-03-30
[feat] SDL2: preliminary gamepad support

Frenzie:
This is a "dumb" implementation that spits out fake keyboard events.

* Left stick & d-pad: arrow keys
* Bumpers and right stick: page up/down
* Menu button: menu
* A: enter
* B: back

This is sufficient to use most of the program.

Made possible by @onde2rock's recent efforts in https://github.com/koreader/koreader/pull/3796
https://github.com/koreader/koreader/pull/3785 https://github.com/koreader/koreader/pull/3774
https://github.com/koreader/koreader/pull/3765 and https://github.com/koreader/koreader/pull/3745

### DIFF
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -28,6 +28,30 @@ local S = {
     SDL = SDL,
 }
 
+local function openGameController()
+    local num_joysticks = SDL.SDL_NumJoysticks()
+
+    if num_joysticks < 1 then
+        S.controller = nil
+        io.write("SDL: no gamecontrollers connected", "\n")
+        return
+    end
+
+    for joystick_counter = 0, num_joysticks-1 do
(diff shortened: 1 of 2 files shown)


ffi/SDL2_0.lua
```diff
@@ -28,6 +28,30 @@ local S = {
...
+        return
+    end
+
+    for joystick_counter = 0, num_joysticks-1 do
+        if SDL.SDL_IsGameController(joystick_counter) ~= 0 then
+            S.controller = SDL.SDL_GameControllerOpen(joystick_counter);
+            if S.controller ~= nil then
+                io.write("SDL: opened gamecontroller ",joystick_counter, ": ",
+                         ffi.string(SDL.SDL_GameControllerNameForIndex(joystick_counter)), "\n");
+                break
```
Frenzie: Ideally there'd be some kind of controller stack but for now this'll do.

PR #634, opened 2018-04-04
[build] FreeType: bump to 2.9

Frenzie:


poire-z:
Any lucky obvious visual change in callgrind output with freetype GlyphInfo performance after this?

Frenzie:
Haha, I'll take a look. The change log speaks about stuff like stem hints and incorrect widths though, not speed. ;-)

Incidentally, I wonder if this could be of interest somehow.

> 2017-11-14  Nikolaus Waxweiler  <madigens@gmail.com>

	Use Adobe hinting engine for `light' hinting of both CFF and Type 1.

	Since Ewald Hew factored the Adobe hinting engine out of the CFF
	driver code, we can now use it on Type 1 (and CID) font formats, as
	both have the same hinting philosophy.

	This change activates the Adobe hinter when in LIGHT mode, and
	therefore always unless explicitly asking for the auto-hinter.  This
	makes LIGHT behavior consistent with CFF fonts.  As of this commit,
	the hinting engine table looks as follows.

	             LIGHT  NORMAL
	  -------------------------
	   TrueType  Auto   v40
	   CFF       Adobe  Adobe
	   Type 1    Adobe  Adobe

This might be of interest speed-wise, but we'd need to add harfbuzz first. (For RTL stuff.)

> 2017-10-14  Behdad Esfahbod  <behdad@behdad.org>

	[afshaper] Delay creating `hb_set' objects until needed.

	In runs on Noto Naskh Arabic, this results in 89 sets created
	instead of 340 before.  Makes auto-hinter setup with HarfBuzz
	enabled 20% to 30% faster.

poire-z:
Just for info: no problem noticed yet on my glohd with this bump and the other ones.

PR #638, opened 2018-04-06
djvu.c: return internal djvulibre error

Frenzie:
All of the `GURL` stuff in djvulibre is absurdly complicated and this does nothing to help.

References https://github.com/koreader/koreader/issues/542

```
04/06/18-10:20:48 WARN  cannot open document /mnt/onboard/.books/koreader_test/Михайлов ТФКП.djvu frontend/document/djvudocument.lua:42: ddjvu: ** Unrecognized DjVu Message:
	** Message name:  ByteStream.open_fail
	   Parameter: %D0%9C%D0%B8%D1%85%D0%B0%D0%B9%D0%BB%D0%BE%D0%B2%20%D0%A2%D0%A4%D0%9A%D0%9F.djvu
	   Parameter: No such file or directory
ddjvu: 'ByteStream.cpp:704'
```

@poire-z Not quite sure if a situation is possible where there's no internal error message. in that case it'd return `0` which is no good… 

Basically I want the `handle()` function with a `return luaL_error(L, "cannot open DjVu file <%s>", filename);` at the end instead of `return 0`, I suppose. Any bright suggestions? :-P

### DIFF
--- a/djvu.c
+++ b/djvu.c
@@ -95,8 +95,9 @@ static int openDocument(lua_State *L) {
 	ddjvu_cache_set_size(doc->context, (unsigned long)cache_size);
 
 	doc->doc_ref = ddjvu_document_create_by_filename_utf8(doc->context, filename, TRUE);
-	if (! doc->doc_ref)
-		return luaL_error(L, "cannot open DjVu file <%s>", filename);
+	if (! doc->doc_ref) {
+		return handle(L, doc->context, FALSE);
+	}
 	while (! ddjvu_document_decoding_done(doc->doc_ref))
 		handle(L, doc->context, True);
 
--- a/djvu.c
+++ b/djvu.c
@@ -96,7 +96,10 @@ static int openDocument(lua_State *L) {
 
 	doc->doc_ref = ddjvu_document_create_by_filename_utf8(doc->context, filename, TRUE);
 	if (! doc->doc_ref) {
-		return handle(L, doc->context, FALSE);
+		int res = handle(L, doc->context, FALSE);
+		if (res != 0) return res;
+		// in case we didn't get a more detailed error message
+		return luaL_error(L, "cannot open DjVu file <%s>", filename);
 	}
 	while (! ddjvu_document_decoding_done(doc->doc_ref))
 		handle(L, doc->context, True);


poire-z:
Mhhh, I never looked at this djvu.c.
But it seems that if you get `DDJVU_ERROR`, you'll return `luaL_error` (and I suppose `msg->m_error.message` is set to something or nothing)
And if you don't, there's no error :)

Or, but I don't get why you want that, if you want to call `handle()` and you know there's an error, and you expect it to return `luaL_error`: well, you could add another parameter `int return_error`, or just use the existing `int wait` (which seems to always be called with `true` in this file): if `false`, return `luaL_error` ?

Frenzie:
We know for sure there's an error because `! doc->doc_ref`. I just want to return a more detailed error without any risk of returning `0` because an error wasn't captured correctly internally. So yes, I want the following logic so as to be assured of returning the *specific* non-specific error. Is there something lacking in my reasoning?

```c
if (! doc->doc_ref) {
 	const ddjvu_message_t *msg;

 	while ((msg = ddjvu_message_peek(ctx)))
 	{
 	  switch(msg->m_any.tag)
 		{
 		case DDJVU_ERROR:
 			if (msg->m_error.filename) {
 				return luaL_error(L, "ddjvu: %s\nddjvu: '%s:%d'\n",
 					msg->m_error.message, msg->m_error.filename,
 					msg->m_error.lineno);
 			} else {
 				return luaL_error(L, "ddjvu: %s\n", msg->m_error.message);
 			}
 		default:
 		  break;
 		}
 	  ddjvu_message_pop(ctx);
 	}

 	// in principle this point should never be reached, I suppose
 	return luaL_error(L, "cannot open DjVu file <%s>", filename);
}
```

Perhaps instead of worrying about supposed elegance I should just do that?

poire-z:
OK, then you could just do:
```c
	if (! doc->doc_ref)
		int res = handle(L, doc->context, FALSE);
		if (res != 0) return res;
		// in case we didn't get any proper error message
		return luaL_error(L, "cannot open DjVu file <%s>", filename);
	}
```

Frenzie:
It's that simple? No type issues?

poire-z:
`static int handle(lua_State *L, ddjvu_context_t *ctx, int wait)` advertises itself as returning an int. So, without looking further, I assume `luaL_error()` returns an `int` too (which could be 0 or not, dunno, you could check - may be it's just pushing the error on the lua stack and returning 0 too - in which case, you would need to return yourself something else, so you know the error has already been set and avoid adding another).

Frenzie:
In any case it seems to work. :-)

PR #639, opened 2018-04-06
[build] djvulibre: add Kobo always UTF-8 filenames patch

Frenzie:
Fixes https://github.com/koreader/koreader/issues/542

This working patch exists largely thanks to browsing the history for the relevant files, see djvulibre commit [03f86480](https://gitlab.com/koreader/djvulibre/commit/03f86480458edc9199da47075de06df545021118).

Some context and potential caveats for the commit referenced above [here](https://sourceforge.net/p/djvu/bugs/197/#ef32).

> Explanation: it seems that launching a program by double clicking on your machine sets the "C" locale. Things are very confusing under the "C" locale because some Mac library functions assume UTF8, and some assume ASCII only (or maybe ISO8859-1 in fact.) This completely confuses the DjVu character code conversion routines (which are unfortunately easy to confuse). The fix essentially means that filenames on the mac are always UTF8 encoded, regardless of the locale. This is not true on other Unix derivatives.


thirdparty/djvulibre/kobo-always-utf-8.patch
```diff
@@ -0,0 +1,35 @@
...
+   }
+   return retval?retval:fopen((const char *)url.NativeFilename(),mode);
+-#elif defined(__APPLE__)
+-  return fopen((const char *)url.UTF8Filename(),mode);
+-#else
+-  return fopen((const char *)url.NativeFilename(),mode);
+ #endif
++//#elif defined(__APPLE__)
++  return fopen((const char *)url.UTF8Filename(),mode);
++//#else
```
Frenzie: @poire-z Patch better so that it's obvious what happened in the patched source or better as a clean removal?

poire-z: It was a bit hard reading `kobo-always-utf-8.patch` without any other context, I had to look at `djvulibre/libdjvu/ByteStream.cpp` to see what was that now-standalone #endif about. But I got it finally.
I would have gone with just adding something like (dunno the correct syntax, but you get the idea):
```
-#elif defined(__APPLE__)
+#elif defined(__APPLE__) or 1
```
or
```
-#elif defined(__APPLE__)
+#elif 1
```
or
```
-#elif defined(__APPLE__)
+#elif defined(__APPLE__) or defined(__KOREADER__) // or any existing var we already define
```
so there's less commenting, and it's obvious we want that APPLE behaviour of UTF8 here.

Frenzie: That's what I'm saying though. I think it's much clearer when you see the lines commented out in context, in `ByteStream.cpp` in this case. It just looks less pretty as a patch. :-) But fine, I'll change it to `|| 1` or `|| defined(KOBO)` or something like that. :-P

Of course I did just what you suggest in my Android patch for djvulibre, but I think there's a poetic quality about those lines.

```c
#ifdef __ANDROID__ 
# undef UNIX 
#endif 
```

https://github.com/koreader/koreader-base/blob/031eb1583a243b3dbe1cec48807af33e9138e31a/thirdparty/djvulibre/android.patch#L1-L31

PR #643, opened 2018-04-10
bump crengine

Frenzie:
* Fix possible segfault with LVIndexedRefCache.cache() https://github.com/koreader/crengine/pull/144
* Fix decoding of URL with percent-encoding https://github.com/koreader/crengine/pull/143


poire-z:
Thanks for the bump!

PR #647, opened 2018-04-14
[feat] SDL2: support SDL_SetWindowIcon

Frenzie:
References https://github.com/koreader/koreader/issues/3858

Edit: a couple of screenshots of this in action using one of the available icons (probably the worst one for this purpose since it doesn't scale down very recognizably).

screenshot_2018-04-14_13-50-24
screenshot_2018-04-14_13-50-47

### DIFF
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -386,9 +386,9 @@ function S.setWindowIcon(icon)
                                                            icon_bb:getWidth(), icon_bb:getHeight(),
                                                            icon_bit_depth, icon_bb.pitch,
                                                            SDL.SDL_PIXELFORMAT_ABGR8888)
-    icon_bb:free()
-
     SDL.SDL_SetWindowIcon(S.screen, surface)
+    SDL.SDL_FreeSurface(surface)
+    icon_bb:free()
 end
 
 return S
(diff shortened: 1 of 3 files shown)


ffi/SDL2_0.lua
```diff
@@ -370,4 +370,25 @@ function S.setClipboardText(text)
...
+    end
+
+    local BB = require("ffi/blitbuffer")
+    local icon_bb = BB.new(re.width, re.height, BB.TYPE_BBRGB32, re.data)
+    local icon_bit_depth = 32
+    local surface = SDL.SDL_CreateRGBSurfaceWithFormatFrom(icon_bb.data,
+                                                           icon_bb:getWidth(), icon_bb:getHeight(),
+                                                           icon_bit_depth, icon_bb.pitch,
+                                                           SDL.SDL_PIXELFORMAT_ABGR8888)
+    icon_bb:free()
```
Frenzie: I was just reading the docs and realized I'm doing this wrong.

https://wiki.libsdl.org/SDL_CreateRGBSurfaceWithFormatFrom
> No copy is made of the pixel data. Pixel data is not managed automatically; you must free the surface before you free the pixel data. 

So the correct order is something like:

```lua
SDL.SDL_SetWindowIcon(S.screen, surface)
SDL.SDL_FreeSurface(surface)
icon_bb:free()
```

PR #651, opened 2018-04-18
Some minor build tweaks

NiLuJe:
Encountered during my investigation for #3302

Mostly CMake being weird and failing to build stuff on my system.


Frenzie:
My rant on CMake here: https://github.com/koreader/koreader/issues/3496#issuecomment-372042304 :-P

thirdparty/sdcv/CMakeLists.txt
```diff
@@ -29,16 +32,10 @@ if(DEFINED ENV{ANDROID})
...
-if(NOT DEFINED ENV{ANDROID} AND NOT DEFINED ENV{DARWIN})
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -lpthread -lrt")
-endif()
-
 # because cmake needs all kinds of annoying special cmake variables
 set(CMAKE_EXE_LINKER_FLAGS "${LDFLAGS} -static-libstdc++")
 if(DEFINED ENV{DARWIN})
     set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -framework CoreFoundation -framework Security")
-else()
-    set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -static-libgcc")
```
Frenzie: Looks like it originates with this commit: https://github.com/koreader/koreader-base/commit/db68822f9f1fd2472701b8d0c0677fe531b56333

That might mean it should be added to that Android line with the `-static` flag below.

NiLuJe:
Yeah, I came to the same commit hash (that was a fun rabbit hole to follow, btw :D).

I wanted to see how the various CI would break ;).

And indeed, that was my idea too for Android, since it appeared clear from the doc that -static doesn't mean -static-libgcc, but I wanted to double-check anyway ;).

NiLuJe:
Wait, the CIs all passed? Man, that's no fun :D.

NiLuJe:
One day I'll get the itch to see how CMake handles the ```-l:libname.a``` syntax to pick stuff up for static linking... Because those absolute paths to a .a are hacky as hell ;p.

Frenzie:
If I'd known in advance just how messed up CMake was going to handle this stuff I would've just done something like https://github.com/koreader/koreader-base/blob/54e7eae63fa6c9396735133e8b82911bca672568/thirdparty/kpvcrlib/CMakeLists.txt

But by the time it turned out that CMake was refusing to do a proper static build I'd already spent so much time on it I figured it was better to just hack the Android build together into something that worked than to start over.

Also if I did start over I'd be more inclined to revamp the whole system with Meson because having to read the CMake source code to understand wtf is going on is a major waste of time. (Yeah, it has "[docs](https://cmake.org/documentation/)"… sure, if that's what you want to call it…)

NiLuJe:
Yeah, CMake is weird. It's either incredibly straightforward, or incredibly obtuse and cryptic, and there's approximately a billion different toggles and switches that do the exact same thing but not quite, which makes it a massive PITA when it's not heaving as you would expect it to (which, sadly, is often ^^).

Frenzie:
I think CMake seems to behave very well and appears quite elegant when you're compiling for your own system on your own system, and probably also for your general architecture on your own system. But I really don't understand why it's more popular than Autotools. I mean, I probably don't even half understand Autotools ffs and it has that annoying Perl dependency for Automake, so maybe that's what it's all about but at least it does what it's supposed to when you tell it to!

Incidentally, to me this is probably the definitive Autotools guide: https://autotools.io/index.html

(I'm glad I'm not alone here. I thought everybody was in love with CMake. :-P)

NiLuJe:
Preaching to the choir ;p.

I mean, autotools is weird as hell, too. But it's so ubiquitous that you're bound to find someone who already encountered the exact same thing, so it's usually easier to figure it out when things go wrong.
Also, docs, yeah. Docs are fine. I love docs. :D.

Frenzie:
Haha, ranter. I've been subscribed to his blog since I came across his patch for Elecom trackballs. I ordered the Huge from Japan a few months ago. In the end I've reverted to my Kensington Expert Mouse, unfortunately. I've been meaning to write up a review/rant about the thing.

I've been aware of the Autotools guide for much longer. Actually I didn't even make the connection at first.

PR #657, opened 2018-04-21
Adds Pic.openGIFDocumentFromData()

poire-z:
This will allow using giflib to render GIF images, which is better than MuPDF for rendering the first frame of an animated GIF, and will allow getting all its frames.
Also fix alpha value when combining animated GIF frames (although I don't really get how alpha works - I thought it would matter only on RGB32A, not on RGB32) but using 0xFF instead of 0x00 gives always the expected image (with 0x00, I sometime got a blank image, or just some shadows - might be related to the GIF transparency...)

### DIFF
--- a/ffi/pic.lua
+++ b/ffi/pic.lua
@@ -166,7 +166,7 @@ function GifDocument:openPage(number)
         local palette={}
         for c=0, cmap.ColorCount-1 do
             local color = cmap.Colors[c]
-            palette[c] = BB.ColorRGB32(color.Red, color.Green, color.Blue, 0)
+            palette[c] = BB.ColorRGB32(color.Red, color.Green, color.Blue, 0xFF)
         end
 
         -- Draw current frame on our bb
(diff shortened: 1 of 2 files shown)


ffi/pic.lua
```diff
@@ -221,6 +221,37 @@ function Pic.openGIFDocument(filename)
...
+    end
+    local read_cb = ffi.cast('GifInputFunc', data_reader(data, size))
+    local err = ffi.new("int[1]")
+    local giffile = giflib.DGifOpen(nil, read_cb, err)
+    if giffile == nil then
+        read_cb:free()
+        error(string.format("Cannot read GIF file: %s",
+            ffi.string(giflib.GifErrorString(err[0]))))
+    end
+    if giflib.DGifSlurp(giffile) ~= giflib.GIF_OK then
```
Frenzie: Now I'm trying to think of an excuse to write a slurp function somewhere.

ffi/pic.lua
```diff
@@ -221,6 +221,37 @@ function Pic.openGIFDocument(filename)
     return GifDocument:new{giffile = giffile}
 end
 
+function Pic.openGIFDocumentFromData(data, size)
+    -- Create GIF from data pointer (from https://github.com/luapower/giflib)
```
Frenzie: Oh, I recently came across LuaPower. It looks pretty neat. They also have a harfbuzz FFI binding which I hope I recently linked to somewhere relevant.

Other stuff of similar potential interest includes http://torch.ch/ and https://github.com/Planimeter/lgf

PR #665, opened 2018-05-05
[fix] SDL2: resolve rotation issues

Frenzie:
Embarrassingly, I recently introduced two (!) regressions in rotation handling.

1. The new blitbuffer created when resizing didn't take rotation into account. Cf. https://github.com/koreader/koreader-base/pull/614
2. The feature to emulate E Ink refresh didn't properly distinguish between BB rectangles and "physical" SDL rectangles. Cf. https://github.com/koreader/koreader-base/pull/626

Reported by @poire-z in https://github.com/koreader/koreader/pull/3812#issuecomment-386806443

### DIFF
--- a/ffi/framebuffer_SDL2_0.lua
+++ b/ffi/framebuffer_SDL2_0.lua
@@ -46,10 +46,12 @@ function framebuffer:_newBB(w, h)
     w = w or SDL.w
     h = h or SDL.h
 
+    local rotation
     local inverse
 
     if self.sdl_bb then self.sdl_bb:free() end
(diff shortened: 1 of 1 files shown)


Frenzie:
@poire-z Any comments before I merge?

poire-z:
Code-wise, I'm too far from being familiar or remembering that part to say a word.
But it solves the problem I reported, thanks :)

Frenzie:
It just occurred to me that the first issue technically isn't a regression at all. The window never resized before, after all. ;-)

PR #667, opened 2018-05-05
[build] bump glib to 2.56.1

Frenzie:
Includes commit https://github.com/GNOME/glib/commit/987bf5bbeb25e846617d7bc1ee9b78f049dc0d60 which should fix the problem that prevented 2.56.0 from working on certain platforms in https://github.com/koreader/koreader-base/pull/635


Frenzie:
/me glares at Android.

Oh well, like I said before, another time.

PR #672, opened 2018-05-10
Update reference to NotoSansCJK filename

NiLuJe:
Depends on [KO#3942](https://github.com/koreader/koreader/pull/3942)


poire-z:
This `drawCoverPage()`  looks like it's not used, so this update won't hurt :)
https://github.com/koreader/koreader-base/blob/f4a56a7a319496f207f9bd93d5f21bdf07aa3508/cre.cpp#L1682

PR #674, opened 2018-05-11
Fix build on Archlinux

onde2rock:
- On archlinux, `luarocks` is symlinked to `luarocks-5.3` 
Now the makefile use `luarocks-5.1` if found in the path, else it default to `luarocks` as before.

- Look like the path used by luarocks is now `luarocks/rocks-5.1/` instead of `luarocks/rocks`
This cause `make` to always rebuild lua-spore.
Does this also happen on other distrbution like ubuntu or my installation is messed up ?

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -30,6 +30,16 @@ endif
 # CMAKE=cmake --no-warn-unused-cli
 CMAKE=cmake
 
+ifneq (, $(DARWINHOST))
+LUAROCKS = luarocks-5.1
+else ifneq (, $(shell command -v luarocks-5.1 2> /dev/null))
+# On some distribution, multiple version of luarocks are installed at the same time, so choose
+# the right one.
+LUAROCKS = luarocks-5.1
+else
+LUAROCKS=luarocks
+endif
+
 # set this to your ARM cross compiler:
 # set CC CXX AR LD RANLIB
 ifeq ($(TARGET), arm-generic)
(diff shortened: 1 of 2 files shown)


Frenzie:
> Does this also happen on other distrbution like ubuntu or my installation is messed up ?

The bleeding user version of LuaRocks 3 may have moved to always versioned as the default. But Debian Buster and Ubuntu 18.04 use the stable LuaRocks 2.4.x (as they should).

For the moment I don't think that second commit should be part of this PR.

Makefile.defs
```diff
@@ -30,6 +30,16 @@ endif
 # CMAKE=cmake --no-warn-unused-cli
 CMAKE=cmake
 
+ifneq (, $(DARWINHOST))
+LUAROCKS = luarocks-5.1
```
Frenzie: Could we have four spaces for clarity? :-)

NiLuJe:
Luarocks is horribly outdated on Gentoo, so, no issues there ;).

```
/usr/bin/luarocks -> luarocks-5.1
/usr/bin/luarocks-5.1
/usr/bin/luarocks-admin -> luarocks-admin-5.1
/usr/bin/luarocks-admin-5.1
```

Not familiar with it, so no idea about the path... Is it this?

```
CONFIGURATION
        Lua version: 5.1
        Configuration files:
                System: /etc/luarocks/config-5.1.lua (ok)
                User  : /home/niluje/.luarocks/config-5.1.lua (not found)

        Rocks trees in use: 
                /home/niluje/.luarocks ("user")
                /usr/lib64/lua/luarocks ("system")
```

In which case, there's a `--tree=<tree>` option...?

onde2rock:
> The bleeding user version of LuaRocks 3 may have moved to always versioned as the default. But Debian Buster and Ubuntu 18.04 use the stable LuaRocks 2.4.x (as they should).

Ah yes, archlinux is building with `--versioned-rocks-dir` : https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/luarocks#n46

I will remove the last commit.

PR #675, opened 2018-05-15
[build] Pull in thirdparty repos with --depth 50

Frenzie:
Uses a fairly conservative value of 50 for now. Also implements a fallback to pull in the entire repo if that's not enough.

This will mainly help to speed up the nightly builds and the CI builds, but it will also reduce the space and time investment for potential hackers.

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -564,7 +564,7 @@ GLIB=$(GLIB_DIR)/lib/libglib-2.0$(if $(DARWIN),.dylib,.so.0)
 GLIB_STATIC=$(GLIB_DIR)/lib/libglib-2.0.a
 ZLIB=$(OUTPUT_DIR)/libs/$(if $(WIN32),zlib1.dll,$(if $(DARWIN),libz.1.dylib,libz.so.1))
 
-LEPT_LIB_EXT=$(if $(WIN32),-5.dll,$(if $(DARWIN),.5.dylib,.so.5))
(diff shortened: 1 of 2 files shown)


Frenzie:
I suspect the Docker images might be failing because iirc they only have Git 1.9.1, whereas Travis has Git 2.15.1, but I'm not sure how that would lead to failure.

On the plus side, in plain Travis, where it does work, you can see it shaves anywhere from 30 to 90 seconds off the individual Travis jobs, which will basically be equivalent to one job less across all jobs.

NiLuJe:
Could it be because the failed checkout tree still exists and is confusing the hell out of git?

And random unrelated thought: any reasons behind the --depth=999999999999? Couldn't we just drop the depth flag altogether?

Frenzie:
Unless that's a bug in the 1.9.x line, I don't think so. :-) Presumably it's simply that newer versions of Git are slightly more persistent/better about finding stuff. Later this week I'll do some other checks, like if `git fetch --tags` makes any difference.

If you don't override the depth then it won't fetch any more than the originally specified --depth argument. It'd be pretty annoying otherwise because then you'd have to keep remembering to stick --depth in every single command you type to prevent it from fetching everything! ;-)

In newer Gits you have an `--unshallow` flag. Since we shouldn't be dealing with any Git prior to 1.8.x we might be able to use it.

NiLuJe:
Right, that makes sense, thanks ;).

PR #676, opened 2018-05-16
Proper REAGL handling on Kobo Aura

NiLuJe:
Needs testing, I don't have the device ;).
Based on @cramoisi's logs from [KO#3130](https://github.com/koreader/koreader/issues/3130).

### DIFF
--- a/.ci/test_script.sh
+++ b/.ci/test_script.sh
@@ -12,7 +12,7 @@ export TESSDATA_PREFIX
 cd build/* && TESSDATA_PREFIX=$(pwd)/data && mkdir -p data/tessdata
 mv ../../tesseract-ocr/tessdata/* data/tessdata/ && cd ../../ || exit
 # fetch font for base test
-travis_retry wget https://github.com/koreader/koreader/raw/master/resources/fonts/droid/DroidSansMono.ttf
+travis_retry wget https://github.com/koreader/koreader-fonts/raw/master/droid/DroidSansMono.ttf
 export OUTPUT_DIR
 if [ -n "${KODEBUG+x}" ]; then
     KODEBUG_SUFFIX=-debug
--- a/ffi-cdecl/mupdf_decl.c
+++ b/ffi-cdecl/mupdf_decl.c
@@ -151,6 +151,9 @@ cdecl_func(fz_device_gray)
 cdecl_func(fz_device_rgb)
 cdecl_func(fz_device_bgr)
 
+cdecl_struct(fz_color_params_s)
+cdecl_func(fz_default_color_params)
+
 /* device, rendering */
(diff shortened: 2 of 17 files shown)


NiLuJe:
Sneaked in the last bit to make my life less of a nightmare in https://github.com/NiLuJe/koreader/tree/uimanager-wip

Should be harmless, and is arguably more correct anyway.

NiLuJe:
On a slightly related note, the whole marker handling for tracked marker feels a bit more complex than it should, we're basically doing wait_for_marker ourselves instead of relying on the kernel.
We could pretty much fire & forget a rotating 10 marker ids and be done with it ;).

NiLuJe:
Okay, rewrote REAGL handling, period. :D.

It behaves properly on my PW2, I need to check what happens on non-REAGL devices, though.

NiLuJe:
Okay, tested on my H2O, seems okay there too.

The only mystical bit is collision_test handling on Kindles... I'm not quite sure where the fuck those magic numbers come from, and the state of the kernel sources is inconsistent (the headers are up to date, but the driver source is... weird. There's bits of the updates in, and not others, and while there's a gigantic list of patches, some of which do include the update to this ioctl, it appears to be half-in (some defines are there, but the iocitl's signature doesn't match the header, and what's actually in the release binaries).

AFAICT, it relates to the id of the LUT in charge of a specific update, with 0 being we don't care about collision. But that doesn't explain the 1642888, unless that's INVALID_LUT (-1) after a number of weird-ass bit operations...

If shit goes wrong on Kindle, that's the first thing to go ;).

cramoisi:
I pushed the files minus the 3 last commit on the device.

poire-z:
Tested with up to https://github.com/koreader/koreader-base/pull/676/commits/7f77f480d0993c2525674ac13d3b858373302cf4 since last evening: no problem (and no change :) noticed on my GloHD.

cramoisi:
@NiLuJe : if you need some specific test, don't hesitate to point them out.

NiLuJe:
@cramoisi : Nothing specific, really, the only massive REAGL user is the reader itself, so, nothing jumping out when reading is pretty good news ;).

I should (hopefully) be done with the reset of this (w/ the uimanager-wip branch) soon, so once that other PR pops up, real reviewing & testing can begin ;).

NiLuJe:
There we go!

NiLuJe:
If everyone's happy with this, it's ready to go in (and then we can get [KO#3983](https://github.com/koreader/koreader/pull/3983) moving ;).

cramoisi:
happy here.

Frenzie:
Haven't tested but looks fine

PR #679, opened 2018-05-24
Adds cre.getLatestDomVersion() and cre.requestDomVersion()

poire-z:
The previous usage of `getIntProperty("crengine.dom.version")` and `setIntProperty("crengine.dom.version", version)` had some unwelcome side effects.

Let's wait for https://github.com/koreader/crengine/pull/192 to be merged to bump crengine with this PR too. (I cancelled the Travis checks to save a few trees)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -367,6 +367,17 @@ static int newDocView(lua_State *L) {
 	return 1;
 }
 
+static int getLatestDomVersion(lua_State *L) {
+    lua_pushnumber(L, gDOMVersionCurrent);
+    return 1;
+}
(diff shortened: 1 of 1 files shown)


NiLuJe:
Base CIs are broken anyway, I sneaked a fix in https://github.com/koreader/koreader-base/pull/676/commits/0d59a25155be493dd5ebdfbb7d5ce29b24843b53 ;).

poire-z:
Looks like I did that same fix in #678 :) (which is already merged into master, your branch may be lagging?)

PR #684, opened 2018-06-12
bump crengine: text formatting: fix leading spaces issues

poire-z:
Includes:
- text formatting: fix leading spaces issues https://github.com/koreader/crengine/pull/204
- new English US & GB hyph patterns https://github.com/koreader/crengine/pull/206

(small bump to have nightly builds with individual changes in case we need to bissect some side effects with the recent and coming rendering fixes)

### DIFF
--- a/thirdparty/lua-Spore/CMakeLists.txt
+++ b/thirdparty/lua-Spore/CMakeLists.txt
@@ -31,7 +31,7 @@ set(PATCH_CMD sh -c "mkdir -p doc && ${ISED} \"s| 'luasocket|--'luasocket|g\" ${
 
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME
-    https://github.com/fperrad/lua-Spore
+    https://framagit.org/fperrad/lua-Spore
     29193c1f48e2d7b71ba4121792e6cb7b45579d2d
     ${SOURCE_DIR}
 )


poire-z:
https://github.com/koreader/koreader-base/blob/340e809afdfd1354dec1a9577efe33e1ec6d7d94/thirdparty/lua-Spore/CMakeLists.txt#L32-L37
It looks like https://github.com/fperrad/lua-Spore has disappeared :|
What's to be done in such cases?

Frenzie:
I suspect the dev may have moved it over somewhere else due to the Microsoft takeover? If nothing turns up when looking around (sorry, no time today or tomorrow) then any one of us can (presumably) easily push a copy to a repo either here on GH or on GL but hopefully that won't be necessary.

Btw there are forks on GH though I don't know if they're up to date atm, e.g., https://github.com/LuaDist/lua-spore

Frenzie:
Actually that one's forked from https://github.com/vladfedin/lua-Spore

But yeah, weird.

poire-z:
Looks like it was moved to https://framagit.org/fperrad/lua-Spore (a french instance of gitlab it looks like).
(The other ones you linked haven't been updated for years, so better stick to that one).

Frenzie:
A little "we've moved" left behind would have been nice. :-)

That Framasoft place offers a pretty interesting bunch of free software services btw.

PR #688, opened 2018-06-17
Theoretical Kindle KOA2 & Kobo Mark7 eink framebuffer support

NiLuJe:
The Mark7 kernel(s?) appears to have kept a degree of backward compatibility in place (I'm guessing because pickel sucks and no-one wants to touch it with a ten-foot pole to update it).

The KOA2, on the other hand, massively broke backward compat, going so far as re-ordering the waveform update modes....

Fortunately, our code was already modular enough that handling these two cases of precious snowflake syndrome were fairly straightforward once the headers were updated.

The headers change were tested w/ FBInk first. There are no changes for Kindle support, for Kobo, we needed to change a few names, but that's it.

Tested on my Mark5 (H2O), and nothing appears to have been broken.

----

The KOA2 code is 'live', since support for this device is already broken because of missing HW information.

The Mark7 code is commented out, because those device do appear to work with the current codepath for now.

In any case, I'll need strace logs on both fronts to make sure it's correct.


NiLuJe:
Of course luacheck doesn't like unused functions, but, hey, that was the point ;).

Frenzie:
Just tell it to ignore. :-)

NiLuJe:
Good point, will do ;p.

NiLuJe:
Doing a pass on the FFI stuff, because it's a mess.

Please never write those manually, and always use ffi-cdecl.
Because I have to carefully diff the results to make sure no-one sneaked in stuff manually, and it's a PITA.
Plus, it makes it that much easier to handle updates quickly, as well as catching API/ABI issues.
And, also, it handles data types properly, so we don't use 64bits types on 32bits arches, and vice versa.

Frenzie:
Sorry. :blush:

NiLuJe:
Okay, I think I'm done :).

Quickly tested on my H2O, nothing blew up in my face (FM/ePub/DjVu/PDF/Reflow) ;).

poire-z:
I get a crash when I try to maximize the emulator:
```
06/26/18-16:49:06 DEBUG input event => type: 53, code: 6(C), value: cdata<struct SDL_WindowEvent &>: 0x40563860, time: 1530024546.740239
06/26/18-16:49:06 DEBUG input event => type: 53, code: 5(B), value: cdata<struct SDL_WindowEvent &>: 0x40564430, time: 1530024546.740709
./luajit: ./ffi/SDL2_0.lua:112: missing declaration for symbol 'SDL_DestroyTexture'
stack traceback:
        [C]: in function '__index'
        ./ffi/SDL2_0.lua:112: in function 'destroyTexture'
        ./ffi/framebuffer_SDL2_0.lua:38: in function 'resize'
        frontend/device/sdl/device.lua:118: in function 'waitEvent'
        frontend/ui/uimanager.lua:856: in function 'handleInput'
```
Looks like `SDL_DestroyTexture` was indeed lost in this cleanup work.

NiLuJe:
Ouch, nice catch!

I'll fold a fix into #692

PR #699, opened 2018-07-03
Bump zsync to git master

NiLuJe:
Plus a new patch to:
  * make it actually build
  * make it abort gracefully when encountering the HTTP frontend of an
    OpenStack container, instead of busy-looping for ever.

Addresses https://github.com/koreader/koreader/pull/4043/commits/0b448a0f0f7cc7f878f3717827a9838a32cd55ff from another angle.

### DIFF
--- a/thirdparty/zsync/CMakeLists.txt
+++ b/thirdparty/zsync/CMakeLists.txt
@@ -17,15 +17,18 @@ ep_get_binary_dir(BINARY_DIR)
 set(RECONF_CMD sh -c "cd c && autoreconf -fi")
 set(CFG_OPTS "-q --prefix=${BINARY_DIR} --host=\"${HOST}\"")
 set(CFG_CMD sh -c "CC=\"${CC}\" CFLAGS=\"$(CFLAGS) -I${SOURCE_DIR}/c\" LIBS=\"${LIBS}\" ${SOURCE_DIR}/c/configure ${CFG_OPTS}")
+# Pickup some build fixes from https://github.com/ahmedammar/zsync
+# Plus, abort if we hit an OpenStack container to avoid infinite loops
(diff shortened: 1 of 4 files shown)


NiLuJe:
Hmm, it's slightly trickier than what I initially diagnosed.

OpenStack does support Range requests... As long as there's only a *single* byte range being requested.
As soon as you ask more than one (and zsyncs asks for up to 20 at a time), it'll return bogus data, which makes zsync loop forever, re-downloading crap forever.

Patching zsync to only request a single byte range works (but it's of course over 20 times slower).
Plus, it's hard to properly detect this case (I have a working patch... when we fetch the zsync file directly via zsync. But we don't, we get it via LuaSocket first).

NiLuJe:
In short, patching zsync to only request 1 byte range is hard, and leaves us with a slow solution anyway.

So, I think just doing without OpenStack is the way to go, as much as it annoys me... :/

NiLuJe:
When I say slow: going from 1753 to 1756:

OpenStack, 1 byte range
```
verifying download...checksum matches OK
used 58812416 local, fetched 2282338
- Real: 54.70s User: 0.28s System: 0.09s Percent: 0% Cmd: ./zsync -i koreader.installed.tar -o koreader.new.tar -u   MajFault: 0 VolContSw: 1626 InvContSw: 0 -
```

NGINX, 20 byte ranges
```
verifying download...checksum matches OK
used 58812416 local, fetched 1995011
- Real: 0.86s User: 0.26s System: 0.05s Percent: 36% Cmd: ./zsync -i koreader.installed.tar -o koreader.new.tar -u   MajFault: 0 VolContSw: 1137 InvContSw: 0 -
```

So, err, 1min vs. <1s :D.

NiLuJe:
TL;DR: This patch will have to do :).

PR #700, opened 2018-07-04
Bump all the things!

NiLuJe:
\o/


NiLuJe:
A few random notes:

* The LuaJIT bump might seem scary, but the changelog actually looks fairly tame.
* Don't squash this, to make it more revert-friendly in case shit goes boom.
* I did not touch leptonica/tesseract, becaise I think @Frenzie has something in the works, and there a major version of tesseract incoming anyway.
* I did not touch any of the øMQ stuff, because we're using very old stuff from around v2.x... and it's now up to v4.x. So there a whole bunch of API changes, and this stuff is weird and tied into our main event loop, so, NOPE. :D.

NiLuJe:
Yeah, I say let's worry about it when we move to tesseract 4 ;).

PR #703, opened 2018-07-07
Bump some more third-party libs

NiLuJe:
Namely, LuaFileSystem, because we've been forgetting about it for three years since it's not handled in the thirdparty folder ;).

And FBInk, to pickup the out-of-Nickel fixes discussed in [KO#4043](https://github.com/koreader/koreader/pull/4043) ;).


Frenzie:
Huh, why is that still in the base root anyway? :-) You're quite right; thirdparty is where I look.

PR #698, opened 2018-07-02
Bundle FBInk for visual feedback in Kobo startup scripts

NiLuJe:
Letting the CI do the work to check if I did this right ;p.

PR #704, opened 2018-07-08
Unbreak debug builds

NiLuJe:
Oops.

c.f., #698


NiLuJe:
(My Makefiles also track the DEBUG env var to switch to Debug builds, which turns out to be neat in theory, except that in practice I forgot to handle it in CMake, and the binaries end up in a different folder...).

PR #705, opened 2018-07-11
Flip the switch on the Mark 7 codepath

### LINKED ISSUE #694: Fix Kobo Mark7 mxcfb handling
Creating a new issue, as a followup of #390 ;).

The Mark7 kernel exposes a new API, but still appears to support the old one (for now?).
There's currently code in to handle the new API, but it's commented out.

Long story short, much like in the issue mentioned earlier, I need a strace dump from Nickel, on a Mark7 device running the latest FW, to decide if we need to flip the switch to the new API on those devices ;).

Quoting the relevant instructions:

----

The binaries are available [here](http://files.ak-team.com/niluje/mrpub/Other/strace-kobo.tar.gz).

Basically, you need SSH access *under Nickel*, and with the attached binary, you trace nickel's ioctls, like so: `./strace -fitv -e trace=ioctl -p $(pidof nickel)`. Then you move around a bit, first in the Library, then in a book (where you make sure you do so long enough to trigger at least once a flashing update), moving around in a couple of menus while you're there, and then post the output here ;).

NiLuJe:
Fix #694

\o/

(Based on data from a Clara HD, but appears consistent with what we've seen from the other Mk7 kernels).

(I'll fold a bump to FBInk 1.0.0 in there once testing's done and it's actually tagged ;p).

### DIFF
--- a/thirdparty/fbink/CMakeLists.txt
+++ b/thirdparty/fbink/CMakeLists.txt
@@ -27,7 +27,7 @@ endif()
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME
     https://github.com/NiLuJe/FBInk.git
-    7c753adf8ad2c9dec0a4ef449abe9947704ab8de
+    v1.0.0
     ${SOURCE_DIR}
 )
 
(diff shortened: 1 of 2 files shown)


thirdparty/fbink/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ endif()
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME
     https://github.com/NiLuJe/FBInk.git
-    7c753adf8ad2c9dec0a4ef449abe9947704ab8de
+    v1.0.0
```
Frenzie: Prosaic. ;-)

NiLuJe: Cathartic, too ;).

PR #706, opened 2018-07-14
Minor dropbear build tweaks

NiLuJe:
Hopefully without breaking anything ;).

(Ground-work for minor changes to the plugin to make it easier to test without risking clobbered system files).

PR #710, opened 2018-07-23
bump crengine: css text-transform, various fixes

poire-z:
Includes:
- Avoid stylesheet reparsing, fix some cache mismatch https://github.com/koreader/crengine/pull/218 https://github.com/koreader/crengine/pull/219
- CSS: adds support for the 'text-transform:' property https://github.com/koreader/crengine/pull/220
- Fix table & lists when in centered/right aligned container https://github.com/koreader/crengine/pull/221


poire-z:
Merging, in spite of the clang build failure (see https://github.com/koreader/koreader-base/pull/706#issuecomment-406962296)

PR #711, opened 2018-07-23
Tone down Clang verbosity

NiLuJe:
To make Travis logs slightly less insane.
This implies some minor fixes to a few third-party builds, because
CMake. >_<".


NiLuJe:
That apparently shaved off ~5min from the Clang build, because everything is terrible, I guess?

PR #712, opened 2018-07-25
Add a truly awful hack to make scp work with our dropbear build...

NiLuJe:
... provided we're installed on the default "modern" location,
(i.e., /mnt/onboard/.adds/koreader).

And look for sftp-server there, too, even if we don't bundle one.

Re [KO#4014](https://github.com/koreader/koreader/issues/4104)


NiLuJe:
Can't do anything better than that, it's the shell that handles the actual command, so it's either an absolute path, or something in $PATH.

Don't do this at home, kids, that's the stupidest dirty patch I've ever done.

NiLuJe:
Switch to a non multicall dropbear build to save a few KBs, since we don't rightly care about the standalone keygen.

Frenzie:
Huh, interesting idea about sftp.

Frenzie:
@NiLuJe I'm getting all kinds of builds failures. But since the nightlies on Gitlab succeeded I guess I'll just do a grand cleanup sometime later and try again. (Not sure I want to heat up the CPU and the room atm…)

NiLuJe:
What kind? I don't think this should affect incremental builds, but you never know ;)

NiLuJe:
The tail end on #713 might, though ;)

Frenzie:
Errors like this:

screenshot_2018-07-28_06-45-50

Also stuff on Leptonica, not sure about LodePNG. I'll try a targeted deletion of those specific thirdparty libs to see if it helps.

Frenzie:
At this time of day at least my CPU is at 55° under load instead of while idling.

screenshot_2018-07-28_06-53-29

Targeted deletion of those from #713 plus djvulibre, dropbear, leptonica, and k2pdfopt made it compile successfully. Better than everything. ;-)

NiLuJe:
Yep, that was probably the LDFLAGS bits from #713 ;).

NiLuJe:
Do you happen to usually have CFLAGS/LDFLAGS set in your env, by chance?

Frenzie:
Nope. :-)

Btw, current idle temp on my CPU is a whopping 41°! D-:

(It might sound lowish if you're thinking stock, but thanks to my bequiet Shadow Rock Slim it's usually around 31/32ish.)

Frenzie:
The Benelux is the hottest part of Europe: https://m.nieuwsblad.be/cnt/dmf20180726_03632882//

NiLuJe:
Ouch.

Frenzie:
Also yesterday around 14:30 the A/C broke down at work. :-P

PR #713, opened 2018-07-26
Minor cblit tweaks

NiLuJe:
A div255 fix, and a few unused variables squished.


NiLuJe:
NOTE:

I'm not a huge fan of the branching inside nested for-loops (for performance reasons), but that would require *significant* refactoring (... for a target I'm essentially not using, so, meh :D).

(Mainly talking about the BB_GET_PIXEL macro here, also SET_ALPHA_FROM_A could also move some stuff outside of the loops).

PR #720, opened 2018-08-14
thirdparty: adds libharfbuzz.so.0 (HarfBuzz 1.8.7)

poire-z:
For use by crengine to provide a toggable alternative kerning method. See https://github.com/koreader/crengine/issues/224

For now, just checking if harfbuzz compiles on all platforms.


Frenzie:
You're so active about me basically just sharing my notes for future ideas publicly. ;-)

I restarted the build because obviously a failure to download tar is unrelated to anything.

NiLuJe:
We possibly need to rebuild FT w/ harfbuzz enabled after that, which is potentially goign to be fun...

(I'm thinking via touch'ed state files to keep make's dependency management and rebuild feature working).

poire-z:
Would we really need a harfbuzz-enabled freetype? We only ship ttf fonts for the UI, so from my understanding they won't benefit much of harfbuzz.

And there's another small thing I don't really like after testing harfbuzz kerning on my gloHD that does not happen with freetype only: the rendered glyph bitmaps must have some kind of grey, or more aggressive antialiasing. On my GloHD (same for the grey lines or grey toggles in the UI), grey is first drawn black, then whiten to grey, which make some kind of unpleasant animation... (must be GloHD specific, because the few times I mentionned that around here, nobody reacted :) Dunno (yet) if it's a harfbuzz thing, or just the incorrect way it is used by crengine... But I surely don't want more of that in the UI :)

NiLuJe:
But we do support third-party OTF files, so why not make the most of them ;o).

For a grey-on-grey antialising (i.e., no subpixel rendering), kerning shouldn't affect the actual rendering (if you look at a it on a glyph per glyph basis).

Tighter kerning *may* allow more ghosting to creep up, though, which is what I think you mean?

NiLuJe:
I have no idea what harfbuzz does by default/in this patch, though, but we probably do NOT want subpixel positioning (just like we don't want subpixel rendering from FT).

NiLuJe:
Dug up an old paper by the main HB guy,  reading it right now (http://goo.gl/yf3M7) :).

poire-z:
> But we do support third-party OTF files, and we default to the auto-hinter, so why not make the most of that ;o).

Ah, you're talking about crengine (was talking about koreader UI).
But talking about crengine, it's possible that linking freetype to harfbuzz will make the default freetype auto-hint not work as it does now :) Which is still my prefered rendering after a few minutes comparing. All the other combinations of hinting/kerning make words more condensed. Freetype original kerning + freetype auto hint is the sole combination that kind of put air into words (dunno why - or how it would be different from css letter-spacing...)
I wish to make the harfbuzz kerning code (the upstream PRs') a toggable, because for performance and taste reason, the old pure freetype code may have its supporters :)

> On an eInk screen, tighter kerning may allow more ghosting to creep up, though, which is what I think you mean?

Nope, I mean two steps drawing of grey stuff :) like if the harfbuzz glyps had more pixels in the 50% grey that make these 2 steps noticable. And like freetype-only had less grey pixels, or more in the 90% grey which wouldn't make these 2 steps less noticed.

Frenzie:
> Freetype original kerning + freetype auto hint is the sole combination that kind of put air into words

Does FreeType even do any kerning by itself? It looks to me like typical bad on-screen typewriter-like display. :-P

poire-z:
It does. But not much. The only font among the 5 or 6 i tested I managed to see some difference between no-kerning/freetype kerning was FreeSerif...
So, let's say I like "kerning off" :)
But it's not so much kerning than the differences in the hinting algo that are noticable. And their combinations with harfbuzz kerning.

NiLuJe:
@poire-z : Not a request BTW, just a "Would Be Nice To Have" braindump while I'm there ;).

@Frenzie: I *think* it had basic support for old-style kern tables, yeah.

----

Apparently, it's also pretty easy to fuck up using HB properly, and end up with mismatched FT loading flags between Ft & HB, which obviously leads to terrible results.

(Haven't had the time to check the code, though, and I've never actually used HB myself ;p).

Frenzie:
> So, let's say I like "kerning off" :)

Kerning off is of course preferable over bad kerning. Also, on lower DPIs (read: your present-day typical desktop 96-ish DPI) proper kerning may result in bad-looking kerning due to pixel boundaries, although on a 300 DPI device you shouldn't be affected by that.

Anyway, quick test with FreeSerif.

On the left, LibreOffice with pair kerning (on by default, but specifying just for clarity).
On the right, crengine the way it is right now.

screenshot_2018-08-15_07-23-17

Unfortunately I wasn't able to compare at more appropriate sizes because this happened:

screenshot_2018-08-15_07-21-36

My test document:
```html
<!DOCTYPE html>
<html><body>font</body></html>
```

poire-z:
> because this happened

Can't reproduce it...
Might be because of your super high dpi and you trigger some Int8 overflow. There is a commit upstream that changed some to Int16 https://github.com/buggins/coolreader/commit/195b53eeff1daca353d181dc1f67eb868637c743 (many stuff to look at in the PR this commit comes from).
Could you give it a quick try (might have some rejects, there are stuff related to what the harfbuzz PRs added)?

Frenzie:
My window size there is my more or less default `-s=kobo-aura-one`. It's only a couple of pixels more than my H2O or your own device. :-P

Frenzie:
Too late for quick tries, alas.

poire-z:
Tested with `-s=kobo-aura-one`, could only see 1/4 of the window on my laptop, but enough to see that this commit fixes this bug. I'll add it in next PR.

Frenzie:
That's a tiny resolution then. I'd expect at least half generally speaking, except on my by now pretty ancient (and barely used) netbook with 1024x600.

It's basically just a harddrive with an OS attached and a reasonably convenient way of testing on old bare metal.

poire-z:
> But talking about crengine, it's possible that linking freetype to harfbuzz will make the default freetype auto-hint not work as it does now 

I had a quick test recompiling freetype with linking to harfbuzz, and it looked to me like it didn't change anything to what I looked at (menu items, cre rendering with freetype kerning and freetype auto hinting with an otf font). Which means good :) so @NiLuJe , if you feel like _We possibly need to rebuild FT w/ harfbuzz enabled after that_, go ahead.

The small mods needed to rebuild:
```diff
diff --git a/Makefile.third b/Makefile.third
index e5d21bd..e867848 100644
--- a/Makefile.third
+++ b/Makefile.third
@@ -18,6 +18,7 @@ $(FREETYPE_LIB) $(FREETYPE_DIR)/include: $(THIRDPARTY_DIR)/freetype2/CMakeLists.
        cd $(FREETYPE_BUILD_DIR) && \
                $(CMAKE) -DCC="$(CC)" -DCXX="$(CXX)" -DCFLAGS="$(CFLAGS)"\
                -DCXXFLAGS="$(CXXFLAGS)" -DLDFLAGS="$(LDFLAGS)" -DCHOST=$(CHOST) \
+               -DHARFBUZZ_DIR="$(HARFBUZZ_DIR)" \
                $(CURDIR)/$(THIRDPARTY_DIR)/freetype2 && \
                $(MAKE)
        cp -fL $(FREETYPE_DIR)/$(if $(WIN32),bin,lib)/$(notdir $(FREETYPE_LIB)) $@
diff --git a/thirdparty/freetype2/CMakeLists.txt b/thirdparty/freetype2/CMakeLists.txt
index a60cc15..d233c80 100644
--- a/thirdparty/freetype2/CMakeLists.txt
+++ b/thirdparty/freetype2/CMakeLists.txt
@@ -10,12 +10,15 @@ enable_language(C CXX)
 assert_var_defined(CXX)
 assert_var_defined(CXXFLAGS)
 assert_var_defined(LDFLAGS)
+assert_var_defined(HARFBUZZ_DIR)

 ep_get_source_dir(SOURCE_DIR)
 ep_get_binary_dir(BINARY_DIR)

 set(CFG_ENV_VAR "CC=\"${CC}\" CXX=\"${CXX}\" CFLAGS=\"${CFLAGS}\" CXXFLAGS=\"${CXXFLAGS}\" LDFLAGS=\"${LDFLAGS}\"")
-set(CFG_OPTS "-q --prefix=${BINARY_DIR} --disable-static --enable-shared --with-zlib=no --with-bzip2=no --with-png=no --with-harfbuzz=no --host=\"${CHOST}\"")
+set(CFG_ENV_VAR "${CFG_ENV_VAR} HARFBUZZ_CFLAGS=\"-I${HARFBUZZ_DIR}/include/harfbuzz\"")
+set(CFG_ENV_VAR "${CFG_ENV_VAR} HARFBUZZ_LIBS=\"-L${HARFBUZZ_DIR}/lib -lharfbuzz\"")
+set(CFG_OPTS "-q --prefix=${BINARY_DIR} --disable-static --enable-shared --with-zlib=no --with-bzip2=no --with-png=no --with-harfbuzz=yes --host=\"${CHOST}\"")
 set(CFG_CMD sh -c "${CFG_ENV_VAR} ${SOURCE_DIR}/configure ${CFG_OPTS}")

 if($ENV{ANDROID})

```

PR #721, opened 2018-08-15
Bump OpenSSL & FBInk

NiLuJe:


Frenzie:
We could just add makedepend to the Docker image?

NiLuJe:
It's already in some of them, so, why not, but I think the issue with 1.0.2p is that they actually stopped using makedepend when it's not needed (i.e., if clang/gcc is not archaic), and we're possibly fucking up the Configure's autodetection ;).

PR #724, opened 2018-09-02
Support for Sony PRS-T*

v01d:
This PR adds support for Sony PRS-T readers, which run a custom-made Linux distribution (based on Debian Jessie) which I named [PRSTUX](https://gitlab.com/phreakuencies/prstux/). I have a PRS-T2 only so I only that model is supported but most likely the other two models aren't two different. I have my ereader fully working with PRSTUX and koreader.

After this PR get merged I will make a PR for the main repo, with the submodule correctly pointing to this repo's current commit.

Please note that I had to include a fix in `libjpeg-turbo/CMakeLists.txt` file, since it wouldn't detect my architecture as ARM (since the check was specialized for armv6) and also CXX flags where not correctly set (only C flags were).


ffi/mxcfb_sony_h.lua
```diff
@@ -0,0 +1,46 @@
...
+static const int UPDATE_MODE_FULL = 1;
+
+static const int WAVEFORM_MODE_INIT = 0;
+static const int WAVEFORM_MODE_DU = 1;
+static const int WAVEFORM_MODE_GC16 = 2;
+static const int WAVEFORM_MODE_GC4 = 3;
+static const int WAVEFORM_MODE_A2 = 4;
+static const int WAVEFORM_MODE_AUTO = 257;
+static const int TEMP_USE_AMBIENT = 4096;
+static const int TEMP_USE_AUTO = 4097; // this does not exist, simply use a value different than above
```
Frenzie: I'm not quite sure I follow?

v01d: You mean the comment? I meant there's no actual AUTO constant for temperature in the headers, but looking at the code, if it doesn't receive a AMBIENT setting it will use the temperature sensor. Thus, I simply used a different value for that constant.

v01d:
PS: I realized I should've added a CI entry for this. One I figure that out I can make a PR for that.

ffi/mxcfb_sony_h.lua
Frenzie: Okay, thanks! :-)

Frenzie:
Ideally you'd probably have to create a Docker image, see:

https://github.com/koreader/koreader-base/blob/22d94e931c9f272eca96fbaab389afca6aaff3dc/.travis.yml#L21-L23

Those are from https://github.com/koreader/virdevenv/tree/13c37dcc21d6e8acffe53e0d88e2a6e3562bd169/docker/ubuntu

In lieu of that it'd also work to use the pre-Docker situation, see https://github.com/koreader/koreader-base/commit/e4f185bee6fc36befeb7a3094e060a1a4b84313e#diff-354f30a63fb0907d4ad57269548329e3

However, the Docker images are used on GitLab to auto-build the nightlies as well.

poire-z:
I indeed would prefer to see "sony" in all that, whether followed by PRST or PRSTUX or PRSTux.
For the frontend code, `device:isSonyPRST` or `device:isSonyPRSTUX` or `device:isSonyAnythingYouLike`
But let's see what @Frenzie and @NiLuJe think.

Makefile.third
```diff
@@ -38,10 +38,13 @@ $(TURBOJPEG_LIB) $(JPEG_LIB): $(THIRDPARTY_DIR)/libjpeg-turbo/CMakeLists.txt
 	cd $(JPEG_BUILD_DIR) && \
 		$(CMAKE) \
 		-DCMAKE_VERBOSE_MAKEFILE:BOOL=OFF \
-		$(if $(findstring armv6, $(ARM_ARCH)),-DWITHOUT_SIMD:BOOL=ON,) \
+		$(if $(findstring arm, $(ARM_ARCH)),-DWITHOUT_SIMD:BOOL=ON,) \
```
NiLuJe: Nope, I want SIMD enabled on armv7 ;).

Makefile.third
```diff
@@ -38,10 +38,13 @@ $(TURBOJPEG_LIB) $(JPEG_LIB): $(THIRDPARTY_DIR)/libjpeg-turbo/CMakeLists.txt
...
 		-DCHOST="$(CHOST)" \
 		-DCFLAGS="$(CFLAGS)" \
 		-DLDFLAGS="$(LDFLAGS)" \
+		-DCMAKE_CXX_COMPILER="$(CMAKE_CXX_COMPILER)" \
```
NiLuJe: Nope, libjpeg-turbo is C only, we shouldn't need that.

If CMake complains, you have a CMake toolchain setup snafu somewhere earlier ;).

ffi/framebuffer_mxcfb.lua
```diff
@@ -502,6 +514,17 @@ function framebuffer:init()
         self.waveform_flashui = self.waveform_ui
         self.waveform_full = C.WAVEFORM_MODE_GC16
         self.waveform_partial = C.WAVEFORM_MODE_GC16
+    elseif self.device:isPRST() then
+        require("ffi/mxcfb_sony_h")
+
+        self.mech_refresh = refresh_prst
+        self.mech_wait_update_complete = prst_mxc_wait_for_update_complete
+
+        self.waveform_fast = C.WAVEFORM_MODE_DU
```
NiLuJe: Unless there's a platform-specific issue with A2 (like there is on REAGL), we usually go with A2 instead of DU for that ;).

ffi/framebuffer_mxcfb.lua
```diff
@@ -502,6 +514,17 @@ function framebuffer:init()
...
+        self.waveform_ui = C.WAVEFORM_MODE_AUTO
+        self.waveform_flashui = C.WAVEFORM_MODE_GC16
```
NiLuJe: In the same vein, unless there's a specific platform issue where AUTO doesn't always honor UPDATE_FULL, this should be AUTO too.

ffi/mxcfb_sony_h.lua
NiLuJe: Huh, that's interesting ;).

I'd add the matching 0x1001 define in the header like we do on Kobo/Kindle, just so that it doesn't get zapped if someone decides to re-run gcc-lua one day ;).

Frenzie:
I agree that something closer Sony PRST would be less mysterious at a glance.

Makefile.third
NiLuJe: If you had a build issue related to ARM detection, that's to be addressed in the crappy CMakeFile for libjpeg-turbo (there's a bunch of eInk & Android specific cases, it should be somewhat obvious, ping me if that looks too arcane ;)).

NiLuJe: Namly, you might need to add PRST the ginormous ENV(...) branch ;).

v01d:
Ok, I will then rename everything to sony_prstux or sony_prst. I will also look into reverting those changes to libjpeg-turbo cmakelist and see what is wrong with the toolchain that may have triggered that.

I'm building a Docker image for building the code, which could be later used for CI. Should I base it on the same base image used for those on the virdevenv repo? Or could I roll my own? I see you use gcc 4.8  and I was using stock ARM 4.9 (cross compilation) found on Ubuntu 16.04, since it matches that of ARMHF Debian Jessie.

v01d:
Ok, so here's the error I get if I revert the changes. I see a mix of i386 and arm compilation. I don't fully understand why or how it happens. What I do see is that CMakeLists grabs gcc for ARM but g++ for x86. Also, SIMD appears to mean SSE and not NEON. That's why I originally assumed I did not want SIMD for any arm arch.
What could be wrong ?

v01d:
@NiLuJe I think I traced it back to CMAKE_SYSTEM_NAME not being set for PRST. I added the check $ENV{PRST} but it does not seem to be set (I'm exporting it as you see in the Makefile).

v01d:
@NiLuJe I think I traced it back to CMAKE_SYSTEM_NAME not being set for PRST. I added the check $ENV{PRST} but it does not seem to be set (I'm exporting it as you see in the Makefile).

NiLuJe:
Yeah, that's what happens when CMAKE_SYSTEM_NAME/CMAKE_SYSTEM_PROCESSOR aren't set or mismatched w/ libjpeg-turbo

Adding an `OR ($ENV{PRST})` to that check *should* be enough, provided the CHOST also matches arm-*...

NiLuJe:
Try a new build from a clean tree? I've had some weird & unreproducible  issues when hot-swapping CMake configs...

NiLuJe:
I had the exact same issue when I initially started writing this using `DEFINED` (but then I was doing it wrong at the time :D), so I'm not that terribly surprised about it ;).

If it works for you and doesn't break the CI, go for it ;).

I'm on Gentoo here, so a fairly current CMake version (3.12.1, FWIW).

NiLuJe:
IIRC, there might be a similar issue with the CHOST expansion in the regex, I tweaked the syntax a few times until either I stopped doing it wrong and/or it happened to play nice with the various CIs.

v01d:
Hi, I addressed the aforementioned issues in my last commit. I also added a CI entry using my docker build image. This failed due to some dumb issue in the Docker file which I just fixed. If you want/can re-trigger the job I think it should run OK now.

Next step is the rename.

poire-z:
(prst build restarted)

poire-z:
In case you missed it: the 2nd build run failed too. Reason:
```
cleaning up crengine checkout...
[sudo] password for ko: 
No output has been received in the last 10m0s,
```

NiLuJe:
(I launched it a 3rd time and it failed the same way, so not a quirk ;)).

v01d:
Sorry, those changes were left on my reader when testing and did not made the commit. Just pushed that. I also fixed the docker image (supposeddly).

NiLuJe:
`autogen.sh: error: could not find libtool.  libtool is required to run autogen.sh.`

(I don't know if we have a real documentation for how to setup those Docker images, but in any case, feel free to jot down what worked/didn't work somewhere... ;)).

v01d:
Thanks! I saw the error as well. I think I just solved that also. I'm pushing here the rename and see if this time builds OK.

NiLuJe:
:+1: LGTM!

v01d:
Cool, thanks! After this is merged I will make the PR on koreader repo.

v01d:
Ok, will do soon.

PR #736, opened 2018-09-12
Sony PRSTUX USB/charging detection + small fix

v01d:
After using koreader some time on the Sony PRS-T2 I saw that A2 has problems: it starts rendering a lot of artifacts and what should be black appears gray sometimes. DU does not have these problems and is quite fast anyway.

I also added charging/discharging detection to generate fake input events. This allows me to get out of suspend when the cable is plugged in and also to inhibit autosuspend.


NiLuJe:
I don't remember anything jumping out at me, so let's say LGTM ;).

NiLuJe:
And yeah, my bad, too, I thought I'd already merged this ;/.

v01d:
No worries, thanks!

PR #740, opened 2018-10-08
cre: adds HTML and CSS extraction functions, alternate highlight method, bump crengine

poire-z:
Bump crengine with changes from https://github.com/koreader/crengine/pull/236
Adds a few function to cre.cpp to make use of that. See individual commit messages.

(I let the links & footnote stuff for some other base + frontend PRs after these ones are merged to keep things clearer).

(Needs the crengine merge and bump, making this PR early for easier review)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -1353,6 +1353,107 @@ static int getWordBoxesFromPositions(lua_State *L) {
 	return 1;
 }
 
+static int getDocumentFileContent(lua_State *L) {
+    CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+    const char* internalFilePath = luaL_checkstring(L, 2);
+
+    LVStreamRef stream = doc->text_view->getDocumentFileStream(lString16(internalFilePath));
+    if (!stream.isNull()) {
+        unsigned size = stream->GetSize();
+        lvsize_t read_size = 0;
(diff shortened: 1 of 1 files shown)


poire-z:
(Cancelled the build, that failed without the crengine bump)

cre.cpp
```diff
@@ -1311,6 +1353,107 @@ static int getWordBoxesFromPositions(lua_State *L) {
...
+    LVStreamRef stream = doc->text_view->getDocumentFileStream(lString16(internalFilePath));
+    if (!stream.isNull()) {
+        unsigned size = stream->GetSize();
+        lvsize_t read_size = 0;
+        void *buffer = (void *)malloc(size);
+        if (buffer != NULL) {
+            stream->Read(buffer, size, &read_size);
+            if (read_size == size) {
+                // (We can push a string containing NULL bytes with lua_pushlstring,
+                // so this function works with binary files like images.)
```
Frenzie: Are you just saying that as a theoretical possibility?

poire-z: No, it's a practical reality :) I tested that with embedded png (it failed first, until I found the right way to do it).

poire-z:
Thanks for the quick read (and not much worry, even for me, it's mostly added optional stuff).
We'll see now what clang thinks about it... :|

PR #741, opened 2018-10-09
cre: adds isLinkToFootnote() and highlightXPointer()

poire-z:
Footnote detection code is quite huge, and tweakable from Lua with the use of flags.
See frontend PR (coming soon) for details.

Also adds optional parameter to `getPageLinks()` to request only internal links.
Also have `getPageLinks()` add to each link the segments (rect for each line) it spans on the page, for a more accurate computation of distance from gesture position.

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -1456,6 +1456,10 @@ static int getHTMLFromXPointers(lua_State *L) {
 
 static int getPageLinks(lua_State *L) {
 	CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+	int internalLinksOnly = false;
+	if (lua_isboolean(L, 2)) {
+		internalLinksOnly = lua_toboolean(L, 2);
+	}
 
 	lua_newtable(L); // all links
 
@@ -1465,10 +1469,11 @@ static int getPageLinks(lua_State *L) {
 	doc->text_view->getCurrentPageLinks( links );
 	int linkCount = links.length();
 	if ( linkCount ) {
-		sel.clear();
+		// sel.clear();
 		lvRect margin = doc->text_view->getPageMargins();
 		int x_offset = margin.left;
 		int y_offset = doc->text_view->GetPos() - doc->text_view->getPageHeaderHeight() - margin.top;
+		int linkNum = 1;
 		for ( int i=0; i<linkCount; i++ ) {
 			lString16 txt = links[i]->getRangeText();
 			lString8 txt8 = UnicodeToLocal( txt );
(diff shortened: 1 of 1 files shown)


cre.cpp
```diff
@@ -1510,28 +1519,676 @@ static int getPageLinks(lua_State *L) {
...
+            // may not be the original numbering in the document.
+            // Howerver, this should not happen with documents using <LI id=...>
+            // as the container for each footnote (like Wikipedia EPUBs), as
+            // the ID= check will then prevent us from including multiple <LI>
+            // in our extended range: the root node will be the <LI>, that we
+            // will mask with CSS "body > li { list-style-type: none; }".
+        }
+    }
+
+    // Target text must be < 10000 chars
```
Frenzie: I wonder what the amount of characters is on some of those academic footnotes…

Also, I hope we're talking displayed text right? Because I've seen some `<span><span><span><span><span><span><span><span><span><span><span><span><span><span>tiny bit of text` type stuff around. ;-)

poire-z: > I hope we're talking displayed text right?

Yes we do :) That's the point of this function, otherwise I could have just used `#html` on the lua side.
This just counts text length text node by text node. I could have just used topNode:getText() to get the full text and get its length, but I just made this function so it stops counting when limit is reached, so we don't have a whole book text in memory before we decide its too much.

MuPDF can cope fine with huge text, but it's just to not block the use too long while MuPDF renders.
5000 is what kobo chose, I met one footnote of 6000, so I pushed it to 10000 just be be large.
But I think I'll make that SIZE_LIMIT passed from Lua, just so we can tweak it from there if need be.

cre.cpp
```diff
@@ -1510,28 +1519,676 @@ static int getPageLinks(lua_State *L) {
...
+    if (flags & 0x8000) {
+        int SIZE_LIMIT = 10000;
```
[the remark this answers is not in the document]
poire-z: Yes is can/should, but there are so many more in cre.cpp that could/should :) like `flags`, `source_xpointer` passed to the function, etc... (It's the kind of thing I never think about - guess it doesn't really matter in such called-once-on-user-demand function.

Anyway, I'll revisit this tomorrow to make it pass'able from Lua. (Just noticed I made flags optional with a default value of 0 - so I'll fix that too).

cre.cpp
Frenzie: 5000 seems rather small when I've seen a three page footnote in a minuscule footnote font.

(This was in a German book. Those can be something else.)

poire-z: Well, that's also the point of this test: when it's that long, it's probably more confortable to go read it in the book (false => follow link) instead of in that 1/3 screen.height footnote widget.

PR #746, opened 2018-10-22
cre text selection: grab word when just holding and not panning 

poire-z:
Fix text selection behaviour introduced in #742 when holding only and not panning.
Otherwise, when holding to just do a dictionary lookup:
- when holding on a space or a punctuation, this char would be and stay highlighted while the dict lookup would not happen because text selection would have been cleaned and become empty.
- when holding on a blank space (eg: after the last word on a non-full last line), no text would be selected here, but getWordFromPosition(), used as a fallback, would return the nearest word on above or below line, which would then be used as dict lookup input, with a few side effects: no highlighting done while looking up; if then Highlighted from the dict window button and bookmarked, the bookmark would have no text and no highlight box would be shown when revisiting the page...

Current behaviour, holding between `deux` and `Angoulême`, with side effects described above:
<kbd>select_bad</kbd>

Previous and restored behaviour, holding at the same place:
<kbd>select_good</kbd>
This is still unperfect, but visually consistent (this bad selection would he highlighted and the highlight boxe shown). And then, when panning, depending on the direction, only one of these 2 parts would be considered in the extended selection.

Also adds some code to blitbuffer.c to make it fully usable on 16bpp Kobo (build yet not enabled, as the performance gain is not obvious.)

### DIFF
--- a/blitbuffer.c
+++ b/blitbuffer.c
@@ -959,6 +959,69 @@ void BB_alpha_blit_from(BlitBuffer *dst, BlitBuffer *src,
             }
             o_x += 1;
         }
+    } else if (dbb_type == TYPE_BBRGB16 && sbb_type == TYPE_BB8A) {
+        o_x = offs_x;
+        Color8A *srcptr;
+        ColorRGB16 *dstptr;
+        for (d_x = dest_x; d_x < dest_x + w; d_x++) {
+            o_y = offs_y;
+            for (d_y = dest_y; d_y < dest_y + h; d_y++) {
+                BB_GET_PIXEL(dst, dbb_rotation, ColorRGB16, d_x, d_y, &dstptr);
+                BB_GET_PIXEL(src, sbb_rotation, Color8A, o_x, o_y, &srcptr);
+                dstptr->v = RGB_To_RGB16(srcptr->a, srcptr->a, srcptr->a);
+                o_y += 1;
+            }
+            o_x += 1;
+        }
+    } else if (dbb_type == TYPE_BBRGB16 && sbb_type == TYPE_BB8) {
+        o_x = offs_x;
+        Color8 *srcptr;
+        ColorRGB16 *dstptr;
+        for (d_x = dest_x; d_x < dest_x + w; d_x++) {
+            o_y = offs_y;
+            for (d_y = dest_y; d_y < dest_y + h; d_y++) {
+                BB_GET_PIXEL(dst, dbb_rotation, ColorRGB16, d_x, d_y, &dstptr);
+                BB_GET_PIXEL(src, sbb_rotation, Color8, o_x, o_y, &srcptr);
+                dstptr->v = RGB_To_RGB16(srcptr->a, srcptr->a, srcptr->a);
+                o_y += 1;
+            }
+            o_x += 1;
+        }
+    } else if (dbb_type == TYPE_BBRGB16 && sbb_type == TYPE_BBRGB24) {
+        o_x = offs_x;
+        ColorRGB24 *srcptr;
+        ColorRGB16 *dstptr;
+        uint8_t r, g, b;
+        for (d_x = dest_x; d_x < dest_x + w; d_x++) {
+            o_y = offs_y;
+            for (d_y = dest_y; d_y < dest_y + h; d_y++) {
+                BB_GET_PIXEL(dst, dbb_rotation, ColorRGB16, d_x, d_y, &dstptr);
+                BB_GET_PIXEL(src, sbb_rotation, ColorRGB24, o_x, o_y, &srcptr);
+                dstptr->v = RGB_To_RGB16(srcptr->r, srcptr->g, srcptr->b);
+                o_y += 1;
+            }
+            o_x += 1;
+        }
+    } else if (dbb_type == TYPE_BBRGB16 && sbb_type == TYPE_BBRGB32) {
+        o_x = offs_x;
+        ColorRGB32 *srcptr;
+        ColorRGB16 *dstptr;
+        uint8_t r, g, b;
+        for (d_x = dest_x; d_x < dest_x + w; d_x++) {
+            o_y = offs_y;
+            for (d_y = dest_y; d_y < dest_y + h; d_y++) {
+                BB_GET_PIXEL(dst, dbb_rotation, ColorRGB16, d_x, d_y, &dstptr);
+                BB_GET_PIXEL(src, sbb_rotation, ColorRGB32, o_x, o_y, &srcptr);
+                alpha = srcptr->alpha;
+                ainv = 0xFF - alpha;
+                r = DIV_255(ColorRGB16_GetR(dstptr->v) * ainv + srcptr->r * alpha);
+                g = DIV_255(ColorRGB16_GetG(dstptr->v) * ainv + srcptr->g * alpha);
+                b = DIV_255(ColorRGB16_GetB(dstptr->v) * ainv + srcptr->b * alpha);
+                dstptr->v = RGB_To_RGB16(r, g, b);
+                o_y += 1;
+            }
+            o_x += 1;
+        }
     } else {
         fprintf(stderr, "incompatible bb (dst: %d, src: %d) in file %s, line %d!\r\n",
                 dbb_type, sbb_type, __FILE__, __LINE__); exit(1);
--- a/cre.cpp
+++ b/cre.cpp
@@ -1136,13 +1136,14 @@ static int getTextFromPositions(lua_State *L) {
 			return 0;
 		r.sort();
 
-		// Only extend to include the whole word when we are actually
-		// holding inside a word. This allows selecting punctuation,
+		// When panning, only extend to include the whole word when we are
+		// actually holding inside a word. This allows selecting punctuation,
 		// quotes or parens at start or end of selection to have them
 		// included in the highlight.
-		if (r.getStart().isVisibleWordChar() && !r.getStart().isVisibleWordStart())
+		bool not_panning = r.getStart() == r.getEnd();
+		if ( (not_panning || r.getStart().isVisibleWordChar()) && !r.getStart().isVisibleWordStart())
 			r.getStart().prevVisibleWordStart();
-		if (r.getEnd().isVisibleWordChar() && !r.getEnd().isVisibleWordEnd())
+		if ( (not_panning || r.getEnd().isVisibleWordChar()) && !r.getEnd().isVisibleWordEnd())
 			r.getEnd().nextVisibleWordEnd();
 		if (r.isNull())
 			return 0;


blitbuffer.c
```diff
@@ -959,6 +959,69 @@ void BB_alpha_blit_from(BlitBuffer *dst, BlitBuffer *src,
...
+        o_x = offs_x;
+        Color8A *srcptr;
+        ColorRGB16 *dstptr;
+        for (d_x = dest_x; d_x < dest_x + w; d_x++) {
+            o_y = offs_y;
+            for (d_y = dest_y; d_y < dest_y + h; d_y++) {
+                BB_GET_PIXEL(dst, dbb_rotation, ColorRGB16, d_x, d_y, &dstptr);
+                BB_GET_PIXEL(src, sbb_rotation, Color8A, o_x, o_y, &srcptr);
+                dstptr->v = RGB_To_RGB16(srcptr->a, srcptr->a, srcptr->a);
+                o_y += 1;
```
Frenzie: I still haven't quite figured out if I prefer this or `o_y++`. ;)

Frenzie:
Btw, GH is/was having some technical issues so it might be prudent to wait a few more hours before merging.

poire-z:
About that 16bpp commit: I did some cheap timing of cre function calls, and found out that on my kobo, a crengine full page rendering takes 30ms, while the blit from the buffer we gave to crengine to draw on, to the device framebuffer, takes from 100ms to 200ms.
200ms with blitbuffer.lua, 100ms with libblitbuffer.so. But in some conditions, with some automated page turning hook, the blitbuffer.lua mean time could go down to 70ms... Still unsure if it's JIT. Didn't have the time to go on with that benchmarking and understanding.
So, not enabling cblitbuffer, just pushing that code so it is not lost and might be useful in some other contexts.

(I was looking at optimising that not because it feels slow - it does not - but just in a battery usage perspective, knowing now the amount of work crengine does when rendering and in other calls - but it is still amazingly fast at that.)

So, that pixel by pixel blitting is quite expensive.
We could still have a super fast optimisation in one obvious case:
```lua
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
 function BB_mt.__index:blitDefault(dest, dest_x, dest_y, offs_x, offs_y, width, height, setter, set_param)
+    print(dest_x.."-".. dest_y.."-".. offs_x.."-".. offs_y.."-".. width.."-".. height..
+    "/"..self:getWidth().."-"..self:getHeight().."("..self:getType()..")/"..dest:getWidth().."-"..dest:getHeight().."("..dest:getType()")")
+    -- 0-0-0-0-600-610/600-610(3)/600-610(5)
+    if dest_x == 0 and dest_y == 0 and offs_x == 0 and offs_y == 0 and width == self:getWidth() and
+            self:getWidth() == dest:getWidth() and self:getType() == dest:getType() and self:getType() ~= TYPE_BB4A then
+        local pixelsize = 1
+        if self:getType() == TYPE_BBRGB16 then pixelsize = 2
+        elseif self:getType() == TYPE_BBRGB24 then pixelsize = 3
+        elseif self:getType() == TYPE_BBRGB32 then pixelsize = 4
+        end
+        local size = width * height * pixelsize
+        ffi.copy(dest.data, self.data, size)
+        print("youpi")
+        return
+    end
     -- slow default variant:
     local hook, mask, count = debug.gethook()
     debug.sethook()
```
so a single memory segment copy is done instead of millions of pixel copies.
But this works only for crengine on the emulator (not widh PDF, when the MuPDF buffer is cropped before being blit to the device/sdl framebuffer).
And this does not work on my Kobo, because there is some BB rotation involved !
https://github.com/koreader/koreader-base/blob/3488223893d76a3ec045ed92003cdf4322474c43/ffi/framebuffer.lua#L11-L22
On my Kobo, cur_rotation_mode=0, blitbuffer_rotation_mode=0 but native_rotation_mode=3, so a direct buffer copy is all messed up. That rotation needs that pixel by pixel blit where the rotations is handled.
@NiLuJe , Master Of the Framebuffers: is that normal that I have that rotation ?

Frenzie:
See https://status.github.com/messages

They posted this a couple of minutes ago:

> 18:24 CEST
We've completed validation of data consistency and have enabled some background jobs. We're continuing to monitor as the system recovers and expect to resume delivering webhooks at 16:45UTC. 

Then on the other thing…

> We could still have a super fast optimisation in one obvious case:

I thought about it last year but decided against messing with it at the time. (My brain was telling me it'd be more fun than fixing some weird Android issue that I was working on. ;-) )

NiLuJe:
Will double check later (not at a computer rn), but yeah, Kobo starts in
landscape mode, hence the rotation ;). Since FW 4.something (a few releases
after the 32bpp switch, nickel itself puts things back on proper landscape
mode though ;).

NiLuJe:
Portrait... Nickel puts things back in portrait on startup. :D

But pickel drops things back to 16bpp landscape, which is how I tested
16bpp stuff on my device.

NiLuJe:
Yeah, LGTM ;).

We could arguably handle full-screen blits in a scanline-per-scanline manner, but I don't know how efficient that bit of manual unrolling vs. what the JIT does would be.

Plus, that stuff gets my brain all twisty. And when you add rotation to the mix... -_-".

Without unrolling anything, just writing a pixel at once, instead of blitting per-component generally netted me a nice speedup in FBInk, but that's hard to do on 16bpp, because RGB156 is terrible ;p.

NiLuJe:
TL;DR: There are many, many, many ways to make blitting faster, but all of them have a heavy readability cost (plus, err, if I can't read it, I certainly can't write it ;p).

You can take a look at SDL 2's blitting codepathS (with a very large S :D) for some mind-bending fun times...

(i.e., besides logic tricks, in the same vein as my 'write a full pixel', there's "write many pixels in //" w/ vector instructions, since nearly all of our targets support NEON).

poire-z:
I too would rather like not to have to look at such algorithms :)
But, if nickel or pickel is able to change the hardware(?) or its tied-to-memory/mmap'ed(?) framebuffer from one rotation to another, couldn't koreader do the same (tried a few `echo 0 or 1 or 2> /sys/class/graphics/fb0/rotate`, didn't insist) so all the 3 values mentionned above are 0 and we and the harware understand all buffers are full portrait mode, and we can take the fast ffi.copy() path?
(Not saying we should do that, and deal with other interaction problems with nickel or ksm... just curious)

Or is it that the hardware is hardwar'y landscape, and any rotation trick like fb0/rotate is just moving the rotation work to the kernel driver code, so it's of no real interest?

NiLuJe:
I utterly failed to get anything even resembling proper portrait rotation via fbset, but that could theoretically help (while indeed breaking everything else except Nickel in the process ;)).

poire-z:
I've been suggested by a friend that the most efficient way of doing this blitbuffer copying when rotation/conversion is involved is doing by tiles (small squares) instead of full-lines, to avoid CPU cache trashing.
Some links about that:
https://stackoverflow.com/questions/16699247/what-is-a-cache-friendly-code
https://stackoverflow.com/questions/9936132/why-does-the-order-of-the-loops-affect-performance-when-iterating-over-a-2d-arra

I'm not really sure, because even if doing it by square, for each line in this square, one has to jump a full line length to get to the bytes of that 2nd square line...

Anyway, just looking at our code, I see that there is may be something not well done in blitbuffer.c, which is used on the emulator and on android where there shouldn't be any rotation involved: we seem to do the loop in the non-natural/non-optimized way: x by the outer loop, y by the inner loop:
https://github.com/koreader/koreader-base/blob/a9a8f93fb416891826486301be939424124fdc09/blitbuffer.c#L647-L664


[Timing it](https://stackoverflow.com/questions/41929712/how-can-i-make-this-function-timing-macro-work-in-c/41930358#41930358) on the emulator when changing pages in an EPUB takes 8ms. With the x/y loops inverted, it takes only 6ms.
That's not much, but still :) (so many loops to update in that code...).

The ffi/blitbuffer.lua Lua code does it in the natural order (y by the outer, x by the inner), but on our devices where there are often some rotation involved, there's anyway one of the source or the destination that will be scanned in the non-natural/non-optimized order... so it doesn't matter much.

NiLuJe:
That's indeed a perfectly valid "free" optimization.

GCC *may* be doing some of it on its own when using aggressive optimizations (something which we currently don't do), so, yeah, +1 ;).

NiLuJe:
But, yeah, as I've said somewhere else, there are a lot of ways of doing blitting better, but most of all are severely less readable (and/or annoying to implement).

See SDL2's various blitting codepaths for braintwisty ideas...

Frenzie:
Note that this function also has a fair bit of impact:

https://github.com/koreader/koreader-base/blob/a9a8f93fb416891826486301be939424124fdc09/blitbuffer.c#L116-L127

[callgrind.out.7359.tar.gz](https://github.com/koreader/koreader-base/files/2803115/callgrind.out.7359.tar.gz)

You might be able to do something clever with memset on a square or rectangular block of pixels so that you only have one loop, maybe with some extra logic to deal with when `i` is odd or something. Not sure if it'd be worth the effort. :-)

poire-z:
> See SDL2's various blitting codepaths for braintwisty ideas...

You mean all the SDL_blit stuff in https://github.com/davidsiaw/SDL2/tree/master/src/video, like https://github.com/davidsiaw/SDL2/blob/master/src/video/SDL_blit_N.c, which uses stuff like https://en.wikipedia.org/wiki/Duff%27s_device in https://github.com/davidsiaw/SDL2/blob/master/src/video/SDL_blit.h?
Quite spooky indeed :)

That Duff's device stuff talks about loop unrolling. Do you think that Luajit (we tweaked some of its loop unroll thresholds some months ago) would do that optimization all alone behind the scenes?!

NiLuJe:
Fair warning: On modern processors & compilers, a Duff's device *may* actually be slower. IIRC, Xorg got rid opf them a few years ago, and that was a net gain.

NiLuJe:
I'd assume that, yes, the LuaJIT optimizer also does unrolling on its own.

You can also try to build blitbuffer.c w/ aggressive flags:

```-O3 -ffast-math -pipe -fomit-frame-pointer -frename-registers -fweb -ftree-vectorize -funroll-loops```

Frenzie:
@NiLuJe Btw, you might know if the hypothesis I voiced yesterday that gcc 4.9 might be crap at optimizing loops (compared to 8, not crap crap) is plausible at all?

The ancient gcc came to mind when thinking about ways in which Android is different.

NiLuJe:
Quite possibly, yeah ;). That said, I don't think it even *tries* to unroll @ -O2!

Frenzie:
Alright, so the impact should be fairly negligible in this case.

PS I'm talking about https://github.com/koreader/koreader/issues/4470#issuecomment-457864847 in case anyone's reading this in the future.

Frenzie:
It sounds innocent enough. :-)

PR #748, opened 2018-10-28
[CI] Update Docker images

Frenzie:


Frenzie:
Hm, libzmq has some changelogs about API changes, so a bump there might not be a good idea even if it builds. I'll just make updated Docker images instead, play it safe.

NiLuJe:
Yeah, we're like 3 major API version behind ;).

If you're in the "burning cpu cycles" it mood, I bumped koxtoolchain recently for the Kindle/Kobo TCs (I think cervented got the bump before it was merged, but not sure).

Frenzie:
You mean use this commit hash from cervantes…

https://github.com/koreader/virdevenv/blob/afb19d60c8b5f75ffd36ac4c26f9c9ef3a135bea/docker/ubuntu/kocervantes/build_cervantes_tc.sh#L6

Everywhere?

Will most definitely do.

Honestly I barely even notice if my CPU is compiling several things at once at 100% load, other than the fact that it reaches 57° or so whereas normal more idle-like temp is between 32-36.

By contrast, I definitely notice a negative impact on my Internet speed from uploading the darn images.

Frenzie:
> Yeah, we're like 3 major API version behind ;).

Ah yes, my quick bump made libzmq itself bump, but then you need to bump *all* of those mq things and last time I figured I'd give it a go I couldn't figure out czmq. (Or rather, I didn't want to spend hours figuring it out with the risk of still failing and decided to call it a loss sufficiently early in the process.)

NiLuJe:
On my phone right now, but yeah, b212bd2f61fcbe0a2bc2d3c4085d7b95a4f79a48 appears to be Head :+1:

Frenzie:
@NiLuJe While building the new Kindle TC it kept looping or something, so I wasn't able to finish that one yet.

NiLuJe:
@Frenzie : Do you happen to have the log laying around?

Frenzie:
@NiLuJe Never mind, my bad. It's the ct-ng-kindle-config, ct-ng-kindle5-config and ct-ng-kindlepw2-config. So it does something very close to the same thing several times in a row (especially as far as what it spits out on the console goes) and is supposed to take quite a bit longer than the equivalent Kobo stuff.

I don't keep a close eye on it obviously, so when I glanced over I was like "whaaaaat it's doing the same thing all over again?" :-)

NiLuJe:
Yeah, IIRC, that's roughly half-an-hour per TC over here ;).

Glad I didn't subtly break anything, then ^^.

Frenzie:
Seems to be about 20 minutes for me. (Those "seconds" at the end definitely aren't mere seconds.)

```
[INFO ]  Build completed at 20181028.210915
[INFO ]  (elapsed: 18:17.97)

[INFO ]  Build completed at 20181028.212824
[INFO ]  (elapsed: 18:53.85)


[INFO ]  Build completed at 20181028.214734
[INFO ]  (elapsed: 18:54.74)
[INFO ]  Finishing installation (may take a few seconds)...
```

That's with this:

```
$ lscpu
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              8
On-line CPU(s) list: 0-7
Thread(s) per core:  2
Core(s) per socket:  4
Socket(s):           1
NUMA node(s):        1
Vendor ID:           GenuineIntel
CPU family:          6
Model:               60
Model name:          Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz
Stepping:            3
CPU MHz:             3935.373
CPU max MHz:         4000.0000
CPU min MHz:         800.0000
BogoMIPS:            7195.46
Virtualization:      VT-x
L1d cache:           32K
L1i cache:           32K
L2 cache:            256K
L3 cache:            8192K
NUMA node0 CPU(s):   0-7
Flags:               fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm cpuid_fault epb invpcid_single pti ssbd ibrs ibpb stibp tpr_shadow vnmi flexpriority ept vpid fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid xsaveopt dtherm ida arat pln pts flush_l1d
```

PR #749, opened 2018-10-30
[fix] HarfBuzz Android build with newer libtool

Frenzie:
Cf. #521


NiLuJe:
Any idea why this one slipped through the cracks before the bump?

Frenzie:
Travis and the old Docker image are on 14.04 with old libtool and I guess no one's built their own Android binaries for a couple of months.

We should replace the ancient Android tc with the Docker container.

PR #750, opened 2018-11-01
[build] Drop Android NDK r11c; also test build on Docker `koandroid`

Frenzie:
This deprecates the outdated r11c to "it may not break" status while ensuring that our nightly builds will succeed.


Frenzie:
Since the Docker image is also r12b we can probably just remove it.

Sidenote: the r15c builds also work fine on Android 2.3+ (API level 10) even though officially it only supports API level 14 and higher.

Also see https://developer.android.com/ndk/guides/stable_apis and https://stackoverflow.com/questions/21888052/what-is-the-relation-between-app-platform-androidminsdkversion-and-androidtar/41079462#41079462

PR #753, opened 2018-11-09
[Feature] Implement getMetadata() function for DjVu documents

xelxebar:
## Overview

Pretty much what the title says. This implements the `getMetadata()` function for DjVu documents, following in the style of the same for the PDF implementation.

This just dumbly copies all the metadata key-value pairs into a lua table. The intent is to give the frontend everything we possibly can and let it figure out what to do with the info, an example of which can be found in the companion pull request koreader/koreader#4314.

## Notes

This is my first foray into koreader's code, not to mention anything lua-related. I tried to copy the surrounding style as much as I could, and I tested this out on my actual kindle. However, I'd not be surprised if I'm missing something obvious. Anyway, I hope this turns out to be helpful.

### DIFF
--- a/djvu.c
+++ b/djvu.c
@@ -162,6 +162,32 @@ static int setColorRendering(lua_State *L) {
 	return 0;
 }
 
+static int getMetadata(lua_State *L) {
+	DjvuDocument *doc = (DjvuDocument*) luaL_checkudata(L, 1, "djvudocument");
+	miniexp_t anno = ddjvu_document_get_anno(doc->doc_ref, 1);
+	miniexp_t *keys = ddjvu_anno_get_metadata_keys(anno);
+
+	// `keys` can be null if there's an error. In that case,
+	// we have the choice of returning either `nil` or
+	// the empty table back to lua.  Here we prefer the latter.
+	lua_newtable(L);
+	if (!keys) return 1;
+
+	int i;
+	for (i = 0; keys[i] != miniexp_nil; i++) {
+		const char *value = ddjvu_anno_get_metadata(anno, keys[i]);
+
+		if (value) {
+			lua_pushstring(L, miniexp_to_name(keys[i]));
+			lua_pushstring(L, value);
+			lua_settable(L, -3);
+		}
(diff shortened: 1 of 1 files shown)


poire-z:
Nice to see fixes on the base/libraries side :)

The gcc builds fail with:
```
djvu.c: In function 'getMetadata':
djvu.c:172:2: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
  for (int i = 0; keys[i] != miniexp_nil; i++) {
  ^
djvu.c:172:2: note: use option -std=c99, -std=gnu99, -std=c11 or -std=gnu11 to compile your code
```
or just move `int i;` outside (https://stackoverflow.com/questions/24881/how-do-i-fix-for-loop-initial-declaration-used-outside-c99-mode-gcc-error)

xelxebar:
Oh. Huh. Didn't get that on my machine. Extracted the declaration.

djvu.c
```diff
@@ -162,6 +162,32 @@ static int setColorRendering(lua_State *L) {
...
+static int getMetadata(lua_State *L) {
+	DjvuDocument *doc = (DjvuDocument*) luaL_checkudata(L, 1, "djvudocument");
+	miniexp_t anno = ddjvu_document_get_anno(doc->doc_ref, 1);
+	miniexp_t *keys = ddjvu_anno_get_metadata_keys(anno);
+
+	// `keys` can be null if there's an error. In that case,
+	// we have the choice of returning either `nil` or
+	// the empty table back to lua.  Here we prefer the latter.
+	lua_newtable(L);
+	if (!keys) return 1;
```
xelxebar: This was bothering me. The docs indicate that `keys` should never be null; however, looking at the source `malloc` can fail to allocate our array---which is quite obvious in retrospect.

To that effect, I ammended the original PR with this extra null check. Since we have a choice on how to handle it and the ordering with `lua_newtable` is important, I went ahead and threw in a comment.

Frenzie:
Thanks, looks good!

PR #755, opened 2018-11-10
Correctly detect Cervantes3 as "Cervantes 3"

### LINKED ISSUE #754: Cervantes 3 model detection is incorrect
In https://github.com/koreader/koreader-base/pull/745 @pazos implemented support for the bq Cervantes line of devices, with the Cervantes 4 being the test device.  I've had some time today to test out my Cervantes 3, and have found that `koreader` won't start.  There's a few separate bugs which I'll describe here; I can resolve one of them but I usually program further up the stack and I'm hitting the limits of my low-level powers.

The first issue is that `koreader` wouldn't start.  I tracked this down to ` frontend/device/cervantes/device.lua` in [koreader](https://github.com/koreader/koreader) and will open a PR there -- it appears that there isn't a default/fall-through for unimplemented methods like `hasNaturalLight`, so I added a `no` method (the inverse of `yes`, defined at the top of that file) and managed to get `koreader` starting.

Unfortunately, the screen never does anything.  `fbink` appears to work, and a bit of digging led me to the following:

```
11/10/18-09:41:21 DEBUG refresh: partial -1 -1 1074 1450
11/10/18-09:41:21 DEBUG mxc_update: 1072x1448 region @ (0, 0) with marker 43 (WFM: 257 & UPD: 0)
11/10/18-09:41:21 DEBUG MXCFB_SEND_UPDATE ioctl failed: Inappropriate ioctl for device
```

Google led me to `strace` and the following:

```
write(1, " mxc_update: 1072x1448 region @ "..., 73 mxc_update: 1072x1448 region @ (0, 0) with marker 43 (WFM: 257 & UPD: 0)) = 73
write(1, "\n", 1
)                       = 1
ioctl(4, 0x4040462e, 0x2c613458)        = -1 ENOTTY (Inappropriate ioctl for device)
```

I note that the second argument there, `0x4040462e` is the `MXCFB_SEND_UPDATE` for the *Pocketbook*.  The *Cervantes*, as per `ffi/framebuffer_mxcfb.lua` should be `0x4044462e`.

So -- it appears that this argument is wrong, but I'm stuck getting much further here.  The value in `mxcfb_cervantes_h.lua` is correct, and *I believe* that's getting imported in `framebuffer_mxcfb.lua` (EDIT -- ah, but that's based on some device detection and it checks if it `isPocketBook` before it gets to the Cervantes -- I'm digging here now).

If anyone can suggest what might've gone wrong, I'd appreciate it!  I'm happy to open a PR when we manage to dig the fix up :).

frio:
Resolves #754.

It'd be good to get @pazos to have a quick look -- should the `Cervantes 4` and `Cervantes 2013` also be fixed as per the device names in https://github.com/koreader/koreader/blob/1e69fae7bc3f65fce516a180ebfd0820651d8c04/frontend/device/cervantes/device.lua?

### DIFF
--- a/ffi/framebuffer_mxcfb.lua
+++ b/ffi/framebuffer_mxcfb.lua
@@ -574,7 +574,7 @@ function framebuffer:init()
 
         -- new devices
         local is_new = false
-        if self.device.model == "Cervantes2013" or self.device.model == "Cervantes3"
+        if self.device.model == "Cervantes2013" or self.device.model == "Cervantes 3"
             or self.device.model == "Cervantes4" then is_new = true end
 
         if is_new then


ffi/framebuffer_mxcfb.lua
```diff
@@ -574,7 +574,7 @@ function framebuffer:init()
 
         -- new devices
         local is_new = false
-        if self.device.model == "Cervantes2013" or self.device.model == "Cervantes3"
+        if self.device.model == "Cervantes2013" or self.device.model == "Cervantes 3"
             or self.device.model == "Cervantes4" then is_new = true end
```
Frenzie: In that case there's also something off about this line, isn't there?

https://github.com/koreader/koreader/blob/4ae43aee2c084e31aa1134a2c92d20f7681fdfcc/frontend/device/cervantes/device.lua#L55

frio: Yeah; that's what I was wondering.  I'm happy to modify the other two as well but wasn't sure if I was missing something.

NiLuJe:
We pick those names ourselves, and we generally don't keep any whitespace.

NiLuJe:
Was there an actual issue you were trying to fix by this?

NiLuJe:
Okay, caught up ;). So:

1. Fix device to set whitespace less model
2. Fix framebuffer_mxcfb to use an if ladder (mainly to match the Kobo syntax)
3. Fix whatever might be checking those model names to match

Frenzie:
Right, so the fix should be the reverse.

frio:
Righto.  It's pretty late in my corner of the world, but I'll raise a proper issue in this project and fix this up in the morning.  Cheers for having a look :).

pazos:
@frio: my bad, I tested that hardcoding Cervantes4 values and didn't check  if the conditional self.device.model stuff worked.

Please feel free to update this, I will submit a new PR on a couple of days to be on par with what we do on FBInk.

Also @NiLuJe: please ping me when you bump FBInk to something higher that https://github.com/NiLuJe/FBInk/commit/3ca7cde0aa31198f98c5d1118369adeb0428248c to add a proper FBInk build for Cervantes (or do it yourself, as you like ;))

NiLuJe:
I'll be doing that soon, waiting on your Makefile.defs PR, as well as this
now I guess ;).

frio:
No worries @pazos; I'm glad I got a chance to test it out and contribute something minor.  Thanks for fixing it up :).

PR #759, opened 2018-11-10
Properly disable ASM on targets where we can't use SIMD instructions

NiLuJe:
Hopefully unfucks the situation on Android...

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -463,8 +463,9 @@ ifneq (,$(findstring march=arm,$(CFLAGS)))
 	# We're on ARM
 	ifeq (,$(findstring fpu=neon,$(CFLAGS)))
 		# We don't support NEON
-		# NOTE: undefine is new to GNU make 3.8...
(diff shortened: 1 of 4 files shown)


Makefile.defs
```diff
@@ -463,7 +463,8 @@ ifneq (,$(findstring march=arm,$(CFLAGS)))
 	# We're on ARM
 	ifeq (,$(findstring fpu=neon,$(CFLAGS)))
 		# We don't support NEON
-		undefine WANT_SIMD
+		# NOTE: undefine is new to GNU make 3.8...
```
Frenzie: Umm where do we come across this make from, what, 2010? Older? >_>

Frenzie: Oh, the Android TC? Makes sense. They also use some shell scripting methods that were deprecated around the year I was born, so… :-P

NiLuJe: Android, apparently ;D

NiLuJe: That's should say gmake 3.82, but, still ;p.

PR #765, opened 2018-11-18
[build, fix] Support PocketBook TC in path

### LINKED ISSUE #764: PocketBook isn't equipped to deal with custom TC location
tl;dr It's defeating the purpose of the Docker image by downloading it from scratch.

Needs a little adjustment in the definitions here as well as in the kodev script in front.

Something with the path as well as the SYSROOT var…

```bash
[[ ":$PATH:" == *":pocketbook-toolchain:"* ]]
[ -z "${SYSROOT}" ]
```

Frenzie:
Fixes #764.


Frenzie:
Not pulling in that whole TC from scratch does save a minute, maybe even two.

I can probably slim down the Docker image another few hundred MB by removing all kinds of stuff we don't use (e.g. anything with libqt in it). The current version is down from 2.83GB to 2.27GB.

NiLuJe:
:+1:

PR #767, opened 2018-11-20
Bump a few thurd-party libs

NiLuJe:
* OpenSSL 1.0.2q
* HarfBuzz 2.1.3

### DIFF
--- a/thirdparty/harfbuzz/CMakeLists.txt
+++ b/thirdparty/harfbuzz/CMakeLists.txt
@@ -41,7 +41,7 @@ endif()
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME
     https://github.com/harfbuzz/harfbuzz.git
-    2.1.0
+    2.1.3
     ${SOURCE_DIR}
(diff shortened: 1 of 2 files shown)


Frenzie:
Is our beloved gcc 4.8 causing trouble again? >_>

NiLuJe:
Yup' ;).

Frenzie:
The official line is to stick to older HB, btw. 
```
https://github.com/pocketbook-free/SDK_481/issues/3#issuecomment-442148725
```

Frenzie:
Over in `https://github.com/buggins/coolreader/pull/59` it's said that newer HB is much faster. ;_;

NiLuJe:
I thought so too, which was why I was chasing the bump... ;p.

Next version *might* possibly build again (it breaks on old VS compilers, too).

Frenzie:
Yeah, but twice as fast? Whoa! So time to try 2.2? :-P

NiLuJe:
2.2 still won't build, AFAICT. Next one might?

PR #768, opened 2018-11-20
[fix] libjpeg-turbo crash on Android <=4.x

Frenzie:
As in the libjpeg-turbo building instructions:

https://github.com/libjpeg-turbo/libjpeg-turbo/blob/43ce78e0321da44fe359f40a847fe79d2de06d4c/BUILDING.md

Cf. https://github.com/koreader/koreader/issues/4254


NiLuJe:
Well, I'll be damned!

I have *no* idea what that would actually change in practice, but, hey, I'll take it ;).

PR #771, opened 2018-12-04
Bump libpng to 1.6.36

NiLuJe:


Frenzie:
Hm, @poire-z is right, Travis isn't working. Looks like we're still using some legacy GH services integration that may need to be replaced.

Unfortunately I shouldn't spend time on that right now, but feel free to fix it up if you want.

NiLuJe:
Ups, might be my fault: I disabled [third-party OAuth access to the org](https://github.com/organizations/koreader/settings/oauth_application_policy) last week, because when I added lgtm integration on my own repos, it was requesting access to the KO org by default, which seemed... stupid.

NiLuJe:
I'm seeing Travis in there now, though, which wasn't the case last week, IIRC?

NiLuJe:
CircleCI probably needs the same kind of nudge, but its UI is either far too arcane for me, or I don't have access to the KO org setup there.

poire-z:
Yep, CircleCI checks not triggered either since a few days ago.

Frenzie:
CircleCI isn't on legacy integration though. I'll try to have a look tomorrow.

Frenzie:
So GitHub says the "payload" (sounds more like a bomb to me) was delivered to CircleCI using a webhook as it should. And! CircleCI shows it ran the tests as it should, e.g.,

This: https://github.com/koreader/koreader/pull/4379

Is here: https://circleci.com/gh/koreader/koreader/2039

NB @poire-z there were some failures here https://circleci.com/gh/koreader/koreader/2038#tests/containers/0 for https://github.com/koreader/koreader/pull/4378

So the question is why it's not showing the test results on GitHub. According to the settings, it should same as always:

screenshot_2018-12-05_21-15-49

@NiLuJe You should also be able to go here? https://circleci.com/gh/koreader/koreader/edit

Frenzie:
@NiLuJe It's over in the top right.

screenshot_2018-12-05_21-28-48

Anyway, apparently I had to grant CircleCI access the same way as Travis because it was also telling me something about lacking permissions when trying to access the organization settings. So I think it might work now, but someone would need to submit a commit.

Frenzie:
Yup, working (forgot to tag it [Kindle]):

screenshot_2018-12-05_21-32-33

PR #776, opened 2018-12-13
[fix] Android nightmode landscape

Frenzie:
Forgot to set rotation on the second blitbuffer. Fixes https://github.com/koreader/koreader/issues/3945.


poire-z:
Nice!
Just curious, why did you not have the issue on [your phone](https://user-images.githubusercontent.com/202757/40071482-3024ffd4-5871-11e8-863a-b08612ee14cb.jpg) and on the android emulators? And how did you finally find it out?

Frenzie:
I must've accidentally run a customized version last time, probably one with this line or something functionally equivalent that I forgot to commit. >_>

PR #780, opened 2018-12-22
Fix Kobo Mk7 detection

NiLuJe:
I'll be in the corner, laughing hysterically.

*facepalm*


Frenzie:
That's… interesting.

NiLuJe:
:D

PR #781, opened 2018-12-22
Dust up setScreenMode's logic

NiLuJe:
Make it human readable (hopefully), heavily commented, simpler, and symmetrical, so we actually get easy access to Inverted Portrait.

Attempt to fix [KO#3002](https://github.com/koreader/koreader/issues/3002) in the process by making sure inversion trickery is only applied on explicit UI interaction (other end of that in [KO#4291](https://github.com/koreader/koreader/issues/4291).

AFAICT, that works as expected \o/.

### DIFF
--- a/ffi/framebuffer.lua
+++ b/ffi/framebuffer.lua
@@ -24,8 +24,8 @@ local fb = {
 --[[
 Codes for rotation modes:
 
-1 for no rotation,
-2 for landscape with bottom on the right side of screen, etc.
+0 for no rotation (i.e., Portrait),
+1 for landscape with bottom on the right side of screen, etc.
 
            2
    +--------------+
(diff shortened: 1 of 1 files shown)


NiLuJe:
I'm fairly confident in the rotation and bit trickery, what's a bit more shaky is the interaction with `interactive ` or not, and KO#3002

In my tests, whatever the state of DLANDSCAPE_CLOCKWISE_ROTATION, setting a default orientation in a book was respected, *even* if you swap DLANDSCAPE_CLOCKWISE_ROTATION  after the fact.

NiLuJe:
Setting a default orientation in a book properly sticks for the next book without existing setting, and you can still reset that global default by long-pressing the other orientation without messing with the current book's defaults.

So, I'm kind of amazed it does, but it appears to behave properly?

NiLuJe:
Yeah, which is why I went with the faster of the two, and lots of comments ;).

Frenzie:
Besides that saying Portrait and Landscape looks slightly odd to me (but I understand why) I can't say that I do. Looks good to me. :-)

NiLuJe:
Eh, I get that, I'd have probably gone with the linux/fb constants from the get go (UR, CW, UD, CCW), but I also understand the history behind the portrait/landscape choice (I'm fairly sure the original eink_fb EPDC did NOT support hardware rotation, and this was all probably originally devised at that time ;)).

PR #783, opened 2018-12-26
Rejig PB fb setup shenanigans

NiLuJe:
Attempt to make sense of PB's various epic fails in terms of framebuffer setup...

This either fixes a lot of weird PB segfaults, or massively braks PB
support \o/.


NiLuJe:
(Split off from #781, because it really didn't belong there ;)).

Frenzie:
This is pretty disturbing. :-P

PR #787, opened 2019-01-04
A few minor third-party libs bumps

NiLuJe:
* GNU Tar 1.31
* djvulibre
* luasocket
* NanoSVG

### DIFF
--- a/thirdparty/djvulibre/CMakeLists.txt
+++ b/thirdparty/djvulibre/CMakeLists.txt
@@ -44,7 +44,7 @@ endif()
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME
     https://gitlab.com/koreader/djvulibre.git
-    35e56567b200832fcf3c76551a22e3ac4070c636
+    91d383bc1d0c546176c27770ac5e13ae30c0eda5
     ${SOURCE_DIR}
 )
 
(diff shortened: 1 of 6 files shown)


NiLuJe:
Oh, joy, Android's bionic doesn't support glob -_-".

Frenzie:
glob is available on API levels 28 and higher, see https://github.com/android-ndk/ndk/issues/718

Not that this will help us in any way, since that's Android 9.

How did this not come up before? It doesn't really sound like this should be new:

http://lists.gnu.org/archive/html/info-gnu/2019-01/msg00001.html

One potential workaround from a quick search: https://groups.google.com/forum/#!topic/android-ndk/vSH6MWPD0Vk

NiLuJe:
Yeah, I'll have to diff the two tar versions, because that felt weird, too ;).

Might be missing some gnulib shenanigans...

Otherwise, will probably have to ship an internal copy (from a quick look around, if gnulib doesn't have one, a few large projects probably have one in place, nginx f.g.).

NiLuJe:
Yes, pulling in glob is [new](https://git.savannah.gnu.org/cgit/tar.git/log/lib/wordsplit.c) ;).

Frenzie:
I [see](https://git.savannah.gnu.org/cgit/tar.git/commit/lib/wordsplit.c?id=c7b3f0217fb9a02b43469404b36b357e38dff850).

We'll have to include our own then. /sigh

Like I said in my previous comment, this one is already known to work (but by now slightly older, insofar as it matters):
https://github.com/white-gecko/TokyoCabinet/blob/master/glob.c
https://github.com/white-gecko/TokyoCabinet/blob/master/glob.h

Frenzie:
> Allows us to hopefully ask gnulib to handle the glob mess on Android for
us automagically...

Huh, really? Tell me more! :-D

NiLuJe:
Don't have makeinfo either ;).

NiLuJe:
Of course, that would have been too easy. -_-".

* Missing yacc in the prstux image
* And possibly a stray glob.h polluting the x86 Android build that screws things up...

NiLuJe:
\o/

PR #788, opened 2019-01-06
More PB fb shenanigans!

NiLuJe:
Mandatory status dump for everyone, so we get useful crash.logs straight away ;).

Plus a tentative fix for https://github.com/koreader/koreader/issues/4447
But I'd *really* like to get logs, because depending on those, I might be able to simply do away with the whole ugly side of the branch ;).

### DIFF
--- a/ffi/framebuffer_linux.lua
+++ b/ffi/framebuffer_linux.lua
@@ -101,6 +101,14 @@ function framebuffer:init()
         end
     -- PocketBook eink framebuffer seems to have no finfo.id
     elseif string.byte(ffi.string(finfo.id, 16), 1, 1) == 0 then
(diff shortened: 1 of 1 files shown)


NiLuJe:
Just wanted to confirm that we don't have a prettier way than io.write to handle logging this early?

Frenzie:
Nothing other than calling the same thing in different ways (e.g., `io.stdout:write()`) and `print()`.

Also keep in mind io.write doesn't flush.

NiLuJe:
Eh, good enough for me, thanks!

Frenzie:
Close? No no, it just forces the buffered output to be written. Bad for performance, good for actually catching it in a log file in case of a crash. ;-)

NiLuJe:
I'll remember that for the next round of PB fun times ;).

PR #789, opened 2019-01-06
bump crengine: in-page footnotes fixes + CSS !important tweak

poire-z:
Includes https://github.com/koreader/crengine/pull/250:
- CSS: avoid publisher's !important from overriding ours
- In-page footnotes: fix vertical position when full status bar
- In-page footnotes: avoid duplicated footnotes on same page
- In-page footnotes: fix ignored link in nested tags
- In-page footnotes: gather links in table cells
- In-page footnotes: fix page splitting edge cases


poire-z:
Of course :)
(And awaiting your wording review of https://github.com/koreader/koreader/pull/4454 :)

PR #790, opened 2019-01-07
Even more PB fb setup fun!

NiLuJe:
Hopefully for good, this time...


Frenzie:
I hope so, because it's kind of a blocker. Admittedly I should also repackage those lzip dicts for better compatibility but that's not as breaking. :-P

PR #791, opened 2019-01-09
Fix undefined _dewarpBuildPageModel_ex symbol in k2pdfopt

NiLuJe:
Involves bumping Leptonica to v1.74.1, the one k2pdfopt 2.42 expects, as well as fixing our buildsystem to include k2pdfopt's new patches to Leptonica.

Re: https://github.com/koreader/koreader/issues/4448
Re: https://github.com/koreader/koreader/issues/4467


Frenzie:
Thanks!

PR #800, opened 2019-02-01
Honor dithering requests from UIManager

NiLuJe:
... on supported devices.

That is:

Kindle: KOA2+ (KOA2, PW4)
Kobo: Mk.7+ (Clara, Forma)

Possibly Cervantes 3 & 4, but the required "new" codepath isn't implemented in KO (pinging @pazos) ;).

Used by [KO#4541](https://github.com/koreader/koreader/pull/4541)

### DIFF
--- a/blitbuffer.c
+++ b/blitbuffer.c
@@ -39,24 +39,24 @@
 #define GET_BB_TYPE(bb) (((MASK_TYPE & bb->config) >> SHIFT_TYPE))
 
 #define ColorRGB32_To_Color8(color) \
-    (Color8){(4898*color->r + 9618*color->g + 1869*color->b) >> 14}
+    (Color8){(4898U*color->r + 9618U*color->g + 1869U*color->b) >> 14U}
 #define ColorRGB32_To_Color8A(color) \
-    (Color8A){(4898*color->r + 9618*color->g + 1869*color->b) >> 14, color->alpha}
+    (Color8A){(4898U*color->r + 9618U*color->g + 1869U*color->b) >> 14U, color->alpha}
 #define ColorRGB32_To_Color16(color) \
-    (ColorRGB16){((color->r & 0xF8) << 8) + ((color->g & 0xFC) << 3) + ((color->b >> 3))}
+    (ColorRGB16){((color->r & 0xF8) << 8U) + ((color->g & 0xFC) << 3U) + ((color->b >> 3U))}
 #define ColorRGB32_To_Color24(color) \
     (ColorRGB24){color->r, color->g, color->b}
 
-#define ColorRGB16_GetR(v) (((v >> 11) << 3) + ((v >> 11) >> 2))
-#define ColorRGB16_GetG(v) ((((v >> 5) & 0x3F) << 2) + (((v >> 5) & 0x3F) >> 4))
-#define ColorRGB16_GetB(v) (((v & 0x001F) << 3) + ((v & 0x001F) >> 2))
+#define ColorRGB16_GetR(v) (((v >> 11U) << 3U) + ((v >> 11U) >> 2U))
+#define ColorRGB16_GetG(v) ((((v >> 5U) & 0x3F) << 2U) + (((v >> 5U) & 0x3F) >> 4U))
+#define ColorRGB16_GetB(v) (((v & 0x001F) << 3U) + ((v & 0x001F) >> 2U))
 #define ColorRGB16_To_A(v) \
     ((39919*ColorRGB16_GetR(v) + \
       39185*ColorRGB16_GetG(v) + \
-      15220*ColorRGB16_GetB(v)) >> 14)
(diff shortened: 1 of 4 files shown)


NiLuJe:
If someone's looking for technical details, this is an EPDC V2 feature, which does quantization + ordered dithering, with a fairly nice pattern.

I'd already noticed that simple ordered dithering patterns were more than "good-enough", but this one is really pretty decent.

Even throwing a nasty testcase full of gradients at it leads to pretty good results!

And that's coming from a guy [that *really* cares about this stuff](https://www.mobileread.com/forums/showpost.php?p=3728291&postcount=17) ;).

pazos:
> Possibly Cervantes 3 & 4, but the required "new" codepath isn't implemented in KO 

@NiLuJe: I don't want to mess with my main reader, plus I remember I tried some dithering calls (x1, x4, ntx_d8) and they were awful.

NiLuJe:
@pazos : I was thinking of what amounts to the Mk. 7 codepath (which you implemented in [FBInk](https://github.com/NiLuJe/FBInk/blob/590c1dbec54871164b13ea044fd15e02a09d7c78/fbink.c#L1401-L1462)), not those weird and legacy flags which date back to the equally weird and esoteric Aura, IIRC ;).

Namely, the new update_data struct with the dither_mode/quant_bit fields (and the mxcfb_dithering_mode enum), not the old, quirky and indeed potentially horribly broken legacy EPDC_FLAG ones ;).

NiLuJe:
It's hidden behind the `-D` flag in FBInk, if you want to try it out on your device.

And here's my usual torture test, thanks to jackie_w on MR ;).
gradient_forma

Frenzie:
Reminds me of your standard printer test page type thing:

https://opensource.apple.com/source/cups/cups-136.9/cups/test/testfile.pdf

NiLuJe:
It has the added benefit of having a large magenta section, which is a nice color to pickup differences in greyscaling methods (it ends up looking fairly different depending on whether you gamma-correct grayscale, or do it in linear light).

Which was a bonus point, in my case ;).

I have a full archive of dithering experiments in Nickel [over here](http://files.ak-team.com/niluje/mrpub/Other/Kobo-Dithering-Test-4.12.tar.xz), especially as far as full-screen covers are concerned ;).

NiLuJe:
@pazos : Speaking of, given the age of the Cervantes 3, if that 2016 mention is accurate, I'd be inclined to think only the Cervantes 4 could actually support that codepath?

Frenzie:
If you want to take a break, have a look at this: https://www.youtube.com/watch?v=EySDyp1SQ-E

Frenzie:
@pazos Is https://github.com/koreader/android-luajit-launcher/blob/14f9c2491d181db34ed694e0b9a7d307762919be/src/org/koreader/device/rockchip/RK30xxEPDController.java#L47 the same thing? (If so, in the long term Android could tap into this as well.)

pazos:
> I was thinking of what amounts to the Mk. 7 codepath (which you implemented in FBInk), not those weird and legacy flags which date back to the equally weird and esoteric Aura, IIRC ;).

> Speaking of, given the age of the Cervantes 3, if that 2016 mention is accurate, I'd be inclined to think only the Cervantes 4 could actually support that codepath?

@NiLuJe: yeah, the code in FBInk is broken, it just works because it is never called (I bet you can find why), I will submit a patch when I have some time. So no, the new codepath won't be supported in current cervantes models.

@Frenzie: i didn't test the Dither mode on Android, but I think it will be broken like any other EPD mode in the controller. The only one that worked (other than FULL or black2white) was A2, which was nice for scrolling but a real pain for fast page turning. Probably things are better with recent devices.

NiLuJe:
@pazos : Even on a Cervantes 4?

Where did the mxcfb header come from then, Kobo?

Because, yeah, right now, I woudln't say that it's never called, unless I'm reading things wrong, it's kind of worse: the "classic" ioctl is called, but with a struct full of the new stuff.
Why that doesn't horribly blow up is beyond me ;).

pazos:
> So I just tried to clear that up in FBInk (bascially getting rid of the cervantes_new codepath, which should match KOReader's behavior), could you check that I didn't do anything stupid when you have the time? ;)

Look good to me. Again sorry for the mess :sob:

NiLuJe:
@pazos : :tada:

No worries, I remember how the Aura quirks threw me for a loop at the time, and this turned out to be fairly similar ;).

Frenzie:
@NiLuJe Were we waiting on anything to merge this one?

NiLuJe:
@Frenzie : Nope, it's good to go :).

ptrm:
How about Aura One? Too old for this ;) ?

NiLuJe:
Yep, Mk. 7 only ;).

PR #804, opened 2019-02-07
Switch to GNU basename in ffi.util

NiLuJe:
The POSIX variant is unavailable on Android (c.f., #4543).

Fix https://github.com/koreader/koreader/issues/4543

### DIFF
--- a/ffi-cdecl/posix_decl.c
+++ b/ffi-cdecl/posix_decl.c
@@ -67,7 +67,7 @@ cdecl_func(usleep)
 cdecl_func(statvfs)
 cdecl_func(gettimeofday)
 cdecl_func(realpath)
-cdecl_func(basename)
+cdecl_func(basename) // NOTE: We'll want the GNU one (c.f., https://github.com/koreader/koreader/issues/4543)
 cdecl_func(dirname)
 
 cdecl_func(malloc)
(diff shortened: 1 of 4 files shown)


NiLuJe:
Note that, even in C, it's potentially tricky to end up with the GNU version:

```
NOTES
       There are two different versions of basename() - the POSIX version described above, and the GNU version, which one gets after

               #define _GNU_SOURCE         /* See feature_test_macros(7) */
               #include <string.h>

       The GNU version never modifies its argument, and returns the empty string when path has a trailing slash, and in particular also when it is "/".  There is no GNU version of dirname().

       With glibc, one gets the POSIX version of basename() when <libgen.h> is included, and the GNU version otherwise.
```

(Even w/ `_GNU_SOURCE`, if you include `<libgen.h>`, which you need for `dirname`, you're screwed without a bit of linker magic)

ffi-cdecl/posix_decl.c
```diff
@@ -67,7 +67,7 @@ cdecl_func(usleep)
 cdecl_func(statvfs)
 cdecl_func(gettimeofday)
 cdecl_func(realpath)
-cdecl_func(basename)
+cdecl_func(basename) // NOTE: We'll want the GNU one (c.f., #4543)
```
Frenzie: Might want to use the URL or otherwise make it clear that it's not that one over here. (Admittedly, it'll be a little while until we reach it. But still, #804 already!)

NiLuJe:
Done.

Took the opportunity to switch lpeg to a mirror, as upstream still appears to be down.

Frenzie:
Does Gentoo also keep older versions around?

Frenzie:
Anyway, lgtm.

NiLuJe:
Yeah, the distfiles mirrors keep the files mirrored as long as there's an ebuild in the tree matching that version.

PR #807, opened 2019-02-11
decouple reader setting from framebuffer

houqp:


houqp:
close to rebase on a different branch

PR #808, opened 2019-02-11
 Add nextVisibleWordEnd and prevVisibleWordEnd methods

Galunid:
Add two methods that allow us to get the next/previous word from xpointer.


cre.cpp
```diff
@@ -1257,6 +1257,30 @@ void lua_pushSegmentsFromRange(lua_State *L, CreDocument *doc, ldomXRange *range
     }
 }
 
+static int nextVisibleWordEnd(lua_State *L){
+    CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+	const char* xp = luaL_checkstring(L, 2);
```
Frenzie: tab

cre.cpp
```diff
@@ -1257,6 +1257,30 @@ void lua_pushSegmentsFromRange(lua_State *L, CreDocument *doc, ldomXRange *range
...
+        return 0;
+    nodep.nextVisibleWordEnd();
+    lua_pushstring(L, UnicodeToLocal(nodep.toString()).c_str());
+
+    return 1;
+}
+
+static int prevVisibleWordEnd(lua_State *L){
+    CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+	const char* xp = luaL_checkstring(L, 2);
```
Frenzie: tab

cre.cpp
```diff
@@ -1257,6 +1257,30 @@ void lua_pushSegmentsFromRange(lua_State *L, CreDocument *doc, ldomXRange *range
     }
 }
 
+static int nextVisibleWordEnd(lua_State *L){
```
poire-z: **get**NextVisibleWordEnd() and **get**PrevVisibleWordEnd() , for reasons explained in https://github.com/koreader/koreader/issues/4555#issuecomment-462245085

poire-z:
Yeah, forget the tabs, stick to 4 spaces. There are still some functions with tabs, that we just haven't removed to not add unuseful commits.
And either keep the blank lines after the parameters checking and the one before the return parameters, or just drop the blank lines (for such little functions, you may as well remove these blank lines).

You may as well add now the prevVisibleWordStart and nextVisibleWordStart even if you don't use them yet.

cre.cpp
poire-z: (I see you've done that in credocument.lua. They deserve than name here too :)

cre.cpp
```diff
@@ -2591,6 +2640,10 @@ static const struct luaL_Reg credocument_meth[] = {
 	{"getHeaderHeight", getHeaderHeight},
 	{"getToc", getTableOfContent},
 	{"getVisiblePageCount", getVisiblePageCount},
+    {"getNextVisibleWordStart", getNextVisibleWordStart},
+    {"getNextVisibleWordEnd", getNextVisibleWordEnd},
```
poire-z: Ok, looks like we have tabs here :| Best to adjust to your surroundings and use a tab too.

Frenzie: Or update the whole thing :-p

Galunid: Well, that's strange, I'm using Ubuntu on VM with tab set as 4 spaces, I've probably copied this part from somewhere else. I've checked, and the whole thing uses tabs instead of spaces. I've updated the parts I've noticed(2 arrays where lua functions are defined).

poire-z: Allright :)

NiLuJe: Heathens! tabs 4 evah'!

:dagger:

/me goes back into his beautiful land of 8-characters wide tabs ;p.

cre.cpp
```diff
@@ -2603,106 +2603,106 @@ static int getImageDataFromPosition(lua_State *L) {
...
+    {"getCoverPageImageData", getCoverPageImageData},
+    {"gotoLink", gotoLink},
+    {"goBack", goBack},
+    {"goForward", goForward},
+    {"clearSelection", clearSelection},
+    {"readDefaults", readDefaults},
+    {"saveDefaults", saveDefaults},
+    {"close", closeDocument},
+    {"__gc", closeDocument},
 	{NULL, NULL}
```
Frenzie: This one stands out a bit. :-P

cre.cpp
Frenzie: I prefer tabs too but the house style is four spaces because, um, it is. xD

poire-z:
If you can fix that last tab before ` 	{NULL, NULL}`, we can merge this PR, so you can then bump base in your frontend PR.
(I described how to do that on the frontend side in https://github.com/koreader/koreader/pull/4314#discussion_r232264326 in case you're not familiar with submodules.)

poire-z:
Looks like you need to fix (in your other PR) the 4 functions. For each of them:
```c
     if (nodep.isNull())
         return 0;
-    nodep.nextVisibleWordEnd();
+    if (!nodep.nextVisibleWordEnd())
+        return 0;
     lua_pushstring(L, UnicodeToLocal(nodep.toString()).c_str());
     return 1;
```
Possibly only needed for the next* ones, but doing if for all is a good thing.
Because these nextVisibleWordEnd() return false when the move is not possible (while still moving the xpointer into a non-valid position - it becomes bad and corrupted when it return false, so it shouldn't be used for anything).

Noticable if you highlight some of the last text in a document, and move selection end ++ till you reach end of document. Next move will return false, but move the xpointer to the start of document, making moving back again impossible, highlighting all the document and koreader super slow.

PR #809, opened 2019-02-11
Log something useful if SDL2 fails to init

NiLuJe:
(Thanks to @ptrm ;)).

### DIFF
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -68,7 +68,7 @@ function S.open()
     if SDL.SDL_Init(bit.bor(SDL.SDL_INIT_VIDEO,
                             SDL.SDL_INIT_JOYSTICK,
                             SDL.SDL_INIT_GAMECONTROLLER)) ~= 0 then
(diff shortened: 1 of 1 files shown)


NiLuJe:
All in all, a great showing :D.

I'll fix that, thanks ;).

ffi/SDL2_0.lua
```diff
@@ -68,7 +68,7 @@ function S.open()
     if SDL.SDL_Init(bit.bor(SDL.SDL_INIT_VIDEO,
                             SDL.SDL_INIT_JOYSTICK,
                             SDL.SDL_INIT_GAMECONTROLLER)) ~= 0 then
-        error("Cannot initialize SDL.")
+        error("Cannot initialize SDL: " .. ffi.string(SDL.SDL_GetError()))
```
Frenzie: Fun fact: I added the FFI header for it but I don't recall what I debugged with it. (Definitely not this line, 'cause I wouldn't have cleaned that out.)

PR #811, opened 2019-02-13
Add getTextFromXPointers method

Galunid:
Adds missing getTextFromXPointers method. This is pretty much direct copy of getTextFromPositions.

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -1124,6 +1124,37 @@ static int getWordFromPosition(lua_State *L) {
 	return 1;
 }
 
+static int getTextFromXPointers(lua_State *L) {
+	CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+	const char* pos0 = luaL_checkstring(L, 2);
+	const char* pos1 = luaL_checkstring(L, 3);
+
+	ldomDocument *dv = doc->dom_doc;
+
+	ldomXPointer startp = dv->createXPointer(lString16(pos0));
+	ldomXPointer endp = dv->createXPointer(lString16(pos1));
+	if (!startp.isNull() && !endp.isNull()) {
+		ldomXRange r(startp, endp);
+		if (r.getStart().isNull() || r.getEnd().isNull())
+			return 0;
+		r.sort();
+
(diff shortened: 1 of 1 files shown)


cre.cpp
```diff
@@ -1124,6 +1124,37 @@ static int getWordFromPosition(lua_State *L) {
...
+			int offset = r.getEnd().getOffset();
+			if (offset < textLen - 1)
+				r.getEnd().setOffset(offset + 1);
+		}
+
+		lString16 selText = r.getRangeText( '\n', 8192 );
+		lua_pushstring(L, UnicodeToLocal(selText).c_str());
+	    return 1;
+	}
+    return 0;
```
poire-z: Some spaces before the return. As that block use tabs (no prob), make the 2 `return 0` and `return 1` use tabs too and be aligned how they should be.

cre.cpp
```diff
@@ -2644,6 +2675,7 @@ static const struct luaL_Reg credocument_meth[] = {
     {"getNextVisibleWordEnd", getNextVisibleWordEnd},
     {"getPrevVisibleWordStart", getPrevVisibleWordStart},
     {"getPrevVisibleWordEnd", getPrevVisibleWordEnd},
+	{"getTextFromXPointers", getTextFromXPointers},
```
poire-z: No tab here, as you converted them to spaces some days ago :)

poire-z:
I'll be adding to crengine `ldomXPointerEx::prevVisibleChar()` and `nextVisibleChar()` https://github.com/koreader/crengine/pull/260.
Would be nice if you could bump crengine (once it's merged there) in this PR, and make the 2 wrapping functions `getPrevVisibleChar()` and `getNextVisibleChar()` (similar to your 4 previous ones), so there are ready to be used (even if you don't yet use them in your frontend PR).

Similar to the base bump you did in front, but to be done in koreader-base/:
```
git submodule init thirdparty/kpvcrlib/crengine
git submodule update --remote thirdparty/kpvcrlib/crengine
git add thirdparty/kpvcrlib/crengine
```

poire-z:
OK, merging as other PRs are flowing. I'll fix and add the missing stuff in a coming crengine bump PR.

PR #817, opened 2019-02-16
A few minor buildsystem tweaks

NiLuJe:
* Avoid compat CFLAGS on kobo & kindle-legacy when using our TCs
* Disable ARM ASM in MµPDF, and honor debug builds

### DIFF
--- a/thirdparty/mupdf/CMakeLists.txt
+++ b/thirdparty/mupdf/CMakeLists.txt
@@ -10,7 +10,12 @@ enable_language(C)
 assert_var_defined(HOSTCFLAGS)
 assert_var_defined(HOSTCC)
 assert_var_defined(OS)
-set(BUILD_CMD_GENERATE sh -c "env CFLAGS=\"${HOSTCFLAGS}\" $(MAKE) -j${PARALLEL_JOBS} generate build=\"release\" CC=\"${HOSTCC}\" verbose=1")
+if($ENV{KODEBUG})
+    set(MUPDF_BUILD_TYPE "debug")
+else()
+    set(MUPDF_BUILD_TYPE "release")
+endif()
+set(BUILD_CMD_GENERATE sh -c "env CFLAGS=\"${HOSTCFLAGS}\" $(MAKE) -j${PARALLEL_JOBS} generate build=\"${MUPDF_BUILD_TYPE}\" CC=\"${HOSTCC}\" verbose=\"yes\"")
 
 assert_var_defined(LDFLAGS)
 assert_var_defined(XCFLAGS)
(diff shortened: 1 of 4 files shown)


NiLuJe:
AFAICT, it's really *ARM* asm, not NEON.
I couldn't really feel any performance impact in what little I tested, and that was in a debug build.

thirdparty/mupdf/CMakeLists.txt
```diff
@@ -71,6 +71,8 @@ set(PATCH_CMD3 sh -c "patch -N -p1 < ${CMAKE_CURRENT_SOURCE_DIR}/libjpeg_shared.
 # NOTE: There's also a few ARCH_UNALIGNED_OK checks, but we never did pass that define.
 #       (FWIW, we *could*, on anything not kindle-legacy).
 set(PATCH_CMD4 sh -c "patch -N -p1 < ${CMAKE_CURRENT_SOURCE_DIR}/no_arm_asm.patch || true")
+# Honor CFLAGS
+set(PATCH_CMD5 sh -c "patch -N -p1 < ${CMAKE_CURRENT_SOURCE_DIR}/honor_cflags.patch || true")
```
Frenzie: Getting pretty complex there, lol.

thirdparty/mupdf/CMakeLists.txt
```diff
@@ -10,7 +10,12 @@ enable_language(C)
 assert_var_defined(HOSTCFLAGS)
 assert_var_defined(HOSTCC)
 assert_var_defined(OS)
-set(BUILD_CMD_GENERATE sh -c "env CFLAGS=\"${HOSTCFLAGS}\" $(MAKE) -j${PARALLEL_JOBS} generate build=\"release\" CC=\"${HOSTCC}\" verbose=1")
+if($ENV{KODEBUG})
+    set(MUPDF_BUILD_TYPE "debug")
+else()
+    set(MUPDF_BUILD_TYPE "release")
+endif()
+set(BUILD_CMD_GENERATE sh -c "env CFLAGS=\"${HOSTCFLAGS}\" $(MAKE) -j${PARALLEL_JOBS} generate build=\"${MUPDF_BUILD_TYPE}\" CC=\"${HOSTCC}\" verbose=\"yes\"")
```
Frenzie: Is `1` and `yes` equivalent here or was the `1` not doing anything?

NiLuJe: The Makefile is explicitly doing `ifneq "$(verbose)" "yes"``, so, yeah, if the intent was to get verbose builds, that wasn't doing the trick ;).

Whether we actually *want* a verbose build here is debatable, though ;).
You know my position on the subject, but I'm perfectly okay with switching that back to "no" here ;).

NiLuJe: Except I remember this check being somewhat counter-intuitive...

I think we *want* this to no, because it's actually *more* silent when set to yes?

Frenzie: Verbose all the way as far as I'm concerned. It's just that too much of it can cause trouble with one of the CI services (definitely Travis and GitLab, possibly CircleCI as well).

Frenzie: I think at least one of them (not Travis, certainly not GitLab) might do something more akin to a terminal emulator (i.e., just a limited amount of scrollback), which sounds potentially a lot more useful to catch the error.

NiLuJe:
From what I've seen, it's literally nearly a decade old *ARM* inline asm (switching from thumb to ARM mode and everything!).

I'm fairly confident GCC should nowadays be doing something fairly non-stupid with the pure C versions.

NiLuJe:
(I'll bump base tomorrow, already too much new stuff in the next nightly ;)).

PR #825, opened 2019-02-20
Leave bitness-dependent types alone in FFI stuff

NiLuJe:
ffi-cdecl unfortunately converts them to their native type *based on the compiler used*, which may lead to hilariously annoying to debug memory corruption issues (c.f., #816 ...) given that we can run both on 32bit (arm) & 64bit (x86_64)....

The posix cdecl tried to handle that by forcing a cdecl_type on 'em, but that doesn't quite work, so, do the same arch-dance as pthread ;).


NiLuJe:
Note that in *most* cases, when that applies to a function argument, it's not dramatic... *unless it's a pointer*, because we'll have to ffi.new it.
But that would crash with a helpful "invalid cast" error.

If, on the other hand, you're allocating a struct with fields using one of those types, that's where hilarity ensues...

PR #830, opened 2019-02-21
Unbreak SDL2 FFI

NiLuJe:
Ooops.

Re https://github.com/koreader/koreader/pull/4631


Frenzie:
Merging since this doesn't affect anything currently tested anyway.

PR #831, opened 2019-02-21
Bump FBInk to v1.11.0

NiLuJe:
Relevant to our interests: a progressbar rounding fix.

Also build & ship fbdepth on Kobo, for the upcoming great 8bpp switcheroo!


poire-z:
> for the upcoming great 8bpp switcheroo!

Some spoilers please!
Will we be able to watch old B&W movies on them ?!

PR #832, opened 2019-02-21
Bump giflib to 5.1.9

NiLuJe:
They killed autoconf (and the people rejoiced!), so this may be quirky on some platforms.

Also, probably breaks win32 build for sure.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/832)


Frenzie:
> Also, probably breaks win32 build for sure.

On the plus side, as a whole it's already a bit broken, cf. https://github.com/koreader/koreader/issues/4512.

On the downside, you just made the potential of a potentially relatively simple quick fix more distant. ;-)

NiLuJe:
That'd take a customized Makefile, as upstream's only left a *nix Makefile ;).

NiLuJe:
Also, apparently macOS isn't *nix enough, because that's also not supported by the Makefile ;p.

Frenzie:
@NiLuJe I haven't yet been able to test if Android builds = Android actually works. Unless we want to live dangerously. :-P

NiLuJe:
:D. I don't recall anything android specific in the Makefile, so, yeah,
might be some surprises ;).

Frenzie:
You know, the Android x86 build actually should be able to run the unit tests without too much trouble. That might be something worth exploring…

On another note, Travis CI has apparently started firing employees, which doesn't sound like a good thing: https://blog.steve.fi/experimenting_with_github_actions.html

Frenzie:
@NiLuJe It's not very informative, but it's not working.

```
02-26 20:31:32.089  8186  8186 W KOReader:  cannot open document /storage/emulated/0/Download/Rotating_earth_(large).gif frontend/document/picdocument.lua:22: Failed to open image
02-26 20:31:32.189  8186  8186 V KOReader:    BG extraction done
```

Frenzie:
Oh, actually it doesn't seem to be working in normal KOReader either? This is supposed to load in PicDocument (i.e., just opening a GIF), isn't it?

NiLuJe:
@Frenzie : That broke on Android only, or everywhere?

Frenzie:
@NiLuJe Everywhere. By "normal" I mean, um, well, basically anything other than Android.

NiLuJe:
>_<".

I'll try to see what went wrong... sometime soon-ish ;).

Frenzie:
No functional difference for our purposes I think, but 5.1.7 came out in the meantime. :-P

NiLuJe:
Yeah, it's just a respin of the release tarball ;).

NiLuJe:
Jotting that down here, too, for posterity: 5.1.7 is just a respin of the release tarball, no functional changes.

Frenzie:
5.1.9 sounds more relevant: (although I think I saw you switch from c89 to c11 so it might independently have become less relevant?)

```
Version 5.1.9
=============

The documentation directory now includes an HTMlified version of the
GIF89 standard, and a more detailed description of how LZW compression
is applied to GIFs.

* Address SF bug #129: The latest version of giflib cannot be build on windows.

* Address SF bug #126: Cannot compile giflib using c89

Version 5.1.8
=============

* Address SF bug #119: MemorySanitizer: FPE on unknown address

* Address SF bug #125: 5.1.7: xmlto is still required for tarball

* Address SF bug #124: 5.1.7: ar invocation is not crosscompile compatible

* Address SF bug #122: 5.1.7 installs manpages to wrong directory

* Address SF bug #121: make: getversion: Command not found

* Address SF bug #120: 5.1.7 does not build a proper library - no 
```

NiLuJe:
It was building fine, IIRC, so, nothing new, I'm afraid :/.

Frenzie:
I dunno, "does not build a _proper_ library" may not be quite the same as "doesn't build". :-)

NiLuJe:
Oh, I missed that one :D.

I'll see how it fares when I get back home ;).

roshavagarga:
@NiLuJe Out of boredom, checked this one too, 2 newer small versions available:

```
Version 5.2.1
==============

This is the "Maybe I shouldn't have dome a release while imn surgical recovery" release.

* In gifbuild.c, avoid a core dump on no color map.

* Restore inadvertently removed library version numbers in Makefile.

Version 5.2.0
=============

The undocumented and deprecated GifQuantizeBuffer() entry point
has been moved to the util library to reduce libgif size and attack
surface. Applications needing this function are couraged to link the
util library or make their own copy.

The following obsolete utility programs are no longer installed:
gifecho, giffilter, gifinto, gifsponge. These were either installed in
error or have been obsolesced by modern image-transformmation tools
like ImageMagick convert. They may be removed entirely in a future
release.

* Address SourceForge issue #136: Stack-buffer-overflow in gifcolor.c:84

* Address SF bug #134: Giflib fails to slurp significant number of gifs

* Apply SPDX convention for license tagging.

NiLuJe:
Unfortunately still lacking an actual buildsystem allowing us to build it on non-Linux platforms ;).

jimman2003:
New patchfile:
`diff --git a/Makefile b/Makefile
index 767c73a..4485f97 100644
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,6 @@ UTILS = $(INSTALLABLE) \
 LDLIBS=libgif.a -lm

 all: libgif.so libgif.a $(UTILS)
-	$(MAKE) -C doc

 $(UTILS):: libgif.a

@@ -82,7 +81,7 @@ check: all

 # Installation/uninstallation

-install: all install-bin install-include install-lib install-man
+install: all install-bin install-include install-lib
 install-bin: $(INSTALLABLE)
 	$(INSTALL) -d "$(DESTDIR)$(BINDIR)"
 	$(INSTALL) $^ "$(DESTDIR)$(BINDIR)"`

benoit-pierre:
This can be closed, as #1750 includes giflib 5.2.2.

NiLuJe:
Fixed via #1750

PR #836, opened 2019-02-27
PocketBook: Fix hanging on waking up from sleep

tob1az:
The forked process based implementation of the PocketBook event thread has been replaced with a background thread based one. This should fix the hanging of Koreader process  on waking from suspension state every second time (koreader/koreader#4646). Koreader suspending can be achieved by:

- Pushing the _Home_ button => resume by selecting _koreader_ task in the Task Manager
- Pushing the _Power_ button => Push it again to resume
- Waiting until the _Auto-suspend_ timeout has passed => Push the _Power_ button to resume

The issue root cause is currently unknown, although it seems to be introduced by one of more recent firmware updates. Tested on _PB InkPad3 FW v.5.19.660_ and _PB Touch HD FW v.5.19.673_. Requires testing on more devices. It would be really nice to check if it doesn't break anything on some older firmwares/models.

### DIFF
--- a/input/input-pocketbook.h
+++ b/input/input-pocketbook.h
@@ -216,34 +216,29 @@ static int pb_event_handler(int type, int par1, int par2) {
     return 0;
 }
 
-static int forkInkViewMain(lua_State *L, const char *inputdevice) {
-    struct sigaction new_sa;
-    int childpid;
-    if ((childpid = fork()) == -1) {
-        return luaL_error(L, "cannot fork() emu event listener");
+static void *runInkViewThread(void *arg) {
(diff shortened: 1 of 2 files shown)


tob1az:
Now that it was tested on at least 5 devices, I guess it's more or less safe to merge it. There are no results from any older PB but I believe it is more important to ensure Koreader is usable on the newer devices. If something breaks indeed, we could deal with it later.

Frenzie:
Sounds good. If you submit a PR to bump base then I can merge that too, or otherwise that could take a little longer.

PR #838, opened 2019-03-01
cre: tweak footnote popup detection

poire-z:
- consider links with only roman numerals (eg: `xiv`) as link to a footnote (1 to **10** roman numerals digix as I don't know how big these roman numbers can reasonably get for numbers with 500 footnotes)
- when extending footnote, also stop when meeting `<a name="someId">` as `name=` works as well as `id=` when following links.

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -2046,19 +2046,27 @@ static bool _isLinkToFootnote(CreDocument *doc, const lString16 source_xpointer,
     // Source node text (punctuation stripped) is only numbers (3 digits max,
     // to avoid catching years ... but only years>1000)
     // Source node text (punctuation stripped) is 1 to 2 letters, with 0 to 2
-    // numbers (a, z, ab, 1a, B2)
+    // numbers (a, z, ab, 1a, B2) - or 1 to 10 roman numerals
     if ( (flags & 0x0400 || flags & 0x0800) && trusted_source_xpointer && !likelyFootnote) {
         lString16 sourceText = sourceNode->getText();
         int nbDigits = 0;
         int nbAlpha = 0;
         int nbOthers = 0;
+        int nbRomans = 0;
         for (int i=0 ; i<sourceText.length(); i++) {
             if ( !lStr_isWordSeparator(sourceText[i]) ) { // ignore space, punctuations...
                 int props = lGetCharProps(sourceText[i]);
                 if (props & CH_PROP_DIGIT)
                     nbDigits += 1;
-                else if (props & CH_PROP_ALPHA)
+                else if (props & CH_PROP_ALPHA) {
                     nbAlpha += 1;
+                    // also check for roman numerals (i v x l c d m)
+                    lChar16 c = sourceText[i];
+                    if (c == 'i' || c == 'v' || c == 'x' || c == 'l' || c == 'c' || c == 'd' || c == 'm' ||
+                        c == 'I' || c == 'V' || c == 'X' || c == 'L' || c == 'C' || c == 'D' || c == 'M' ) {
+                            nbRomans += 1;
+                    }
+                }
                 else
                     nbOthers += 1; // CJK, other alphabets...
             }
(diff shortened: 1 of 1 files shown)


Frenzie:
Speaking of footnotes, I have an EPUB here that uses `class="footnoteRef"`. It seems to be somewhat standard, also see https://github.com/koreader/koreader/issues/2107#issuecomment-390405450.

Frenzie:
Sorry, I described it incorrectly. It's `<span class="footnoteRef"><a href="#_doc_fragment_18">`. I have no idea why they're using a useless span, but it is what it is. :-P

poire-z:
And they are detected or not? From the 2107 link, there is the vertical-align in `a.footnoteRef { vertical-align: super; }` that would make them detected. Usually, there are other hints like this digit detection that helps. And currently no classname check.

If you're talking about the in-page footnotes, there's nothing to check for classes of the link origin. Only the footnote targeted. But chances are that if your link has a classname, the footnote too :)
If we meet other footnote target common classnames, we can add them in this style tweak (and the next one):
https://github.com/koreader/koreader/blob/d8845f8589123786d1244c231ef237e56bd1471f/frontend/ui/data/css_tweaks.lua#L487-L491

Frenzie:
No, they're not detected, hence why I mentioned it.

Let's see, the footnotes themselves are in: `<p class="footnote">` so you'd think they'd work.

Frenzie:
Could it be because each footnote is on an individual page?

poire-z:
footnote popups or in-page footnotes ?

Frenzie:
Neither seem to be working.

poire-z:
For in-page footnotes, a style tweak with `p.footnote > a {-cr-hint: footnote-inpage;}` may help - although not sure given what I wrote below (the in-page code may want the hint to be on something that is a block container AND holds the target id).

And I forgot there are also -cr-hint for popup footnotes, so you may make a similar one for the origin or the target with one of (ex: `span.footnoteRef > a {-cr-hint: noteref; }`)
```c
                            // For footnote popup detection:
        "noteref",          // link is to a footnote
        "noteref-ignore",   // link is not to a footnote (even if everything else indicates it is)
        "footnote",         // block is a footnote (must be a full footnote block container)
        "footnote-ignore",  // block is not a footnote (even if everything else indicates it is)
        "footnote-inpage",  // block is a footnote (must be a full footnote block container), and to be
                            // displayed at the bottom of all pages that contain a link to it.
```

poire-z:
> although not sure given what I wrote below (the in-page code may want the hint to be on something that is a block container AND holds the target id).

Nope, it should work, cf https://github.com/koreader/crengine/blob/e68e6b98fd5889a16d0ef57b30438424c6e7dc05/crengine/src/lvrend.cpp#L2751-L2763, `getFirstInnerAttributeValue()` was added just for that.

Frenzie:
Okay, what I posted previously was based on what I could see in debug view on my H2O. In reality it goes like this:

```xml
<span class="footnoteRef" id="fn01"><a href="footnote1.html">*</a></span>
```
(That's changed to `#_doc_fragment_12` by crengine.)

There's a file per footnote, containing (other than the regular XHTML) just this:
```xml
<body class="text" id="text">
<p class="footnote" id="footnote1"><span class="footnotePara"><a href="book_ch01.html#fn01">*</a>Footnote text.</span></p></body>
```
(That `body` is turned into `<DocFragment id="#_doc_fragment_12">`.)

Your suggested `span.footnoteRef > a {-cr-hint: noteref; }` works.

Based on your suggestion that it needs to be a block container with an ID, and `DocFragment {-cr-hint: footnote-inpage;}` does the trick.

This is a ('09, according to the imprint) HarperCollins ebook, so this is or was probably a somewhat common pattern.

poire-z:
> this is or was probably a somewhat common pattern.

Some remarks if you think about adding them into frontend/ui/data/css_tweaks.lua:

> Your suggested `span.footnoteRef > a {-cr-hint: noteref; }` works.

There is no "Popup footnotes hints>" yet, so if need be, this one could go in a new style tweak.
But it's quite possible it would work with just `Links> Show more links as footnotes` checked.

> `DocFragment {-cr-hint: footnote-inpage;}` does the trick.

This one is more tricky, as any linked chapter that would be itself an individual HTML file would be shown as a footnote. So it can't be added into `In-page classic classname footnotes` - it would need to be in another new tweak.

Frenzie:
> But it's quite possible it would work with just Links> Show more links as footnotes checked.

No, it doesn't. That's one of the first things I (double) checked. ;-)

> This one is more tricky, as any linked chapter that would be itself an individual HTML file would be shown as a footnote.

This doesn't actually seem to happen in this specific EPUB, and it's not like TOC items such as "about the publisher" are more than one or two pages.

Anyway, I'm not really planning anything atm. :-P

Frenzie:
Interestingly, I have another one here that says it's from 2007 that works out of the box. This one doesn't link straight to the file, but to an element with an ID in the file the way we want it.

screenshot_2019-03-03_20-07-36

screenshot_2019-03-03_20-06-37

Of course I have no idea how consistent they were about this sort of thing…

Frenzie:
Hold on, actually this one has some that work as in-page and some that don't. Gah, they probably did a bad job on the markup. :-P

Issue #828, opened 2019-02-21
Deprecate SDL 1.2


Frenzie:
SDL 1.2 has been deprecated since 2014 or so.

https://wiki.libsdl.org/Support
> SDL 1.2 is deprecated and not recommended for new projects. 

Qemu recently followed suit: https://patchwork.kernel.org/patch/10164537/

I haven't actually tested it in years, so the status of SDL 1.2 in KOReader is that it may not be broken.

I think that other than the base CI, nothing really uses it anymore.

https://github.com/koreader/koreader-base/blob/8e55e05e9309954891f95dd34f0e901f7ba4015b/.travis.yml#L35

I briefly entertained the idea of testing both SDL 1.2 as well as SDL 2 (cf. https://github.com/koreader/koreader/pull/4631#issuecomment-465916198) but the additional maintenance burden isn't worth it.

PR #843, opened 2019-03-05
simplify framebuffer dpi handling

houqp:
The dpi handling logic is a bit convoluted, looks like the original `screen_dpi_override` is not needed. So I removed it.

Anyone knows the story behind `dpi_scale = self.dpi / 167`? Why 167 instead of the default 160?

I feel like `if self.device and self.device.display_dpi ~= self.dpi then` can be further simplified to `if self.dpi ~= 167`?


Frenzie:
I imagine 167 must come from some old Kindle model?

And indeed, the basic Kindle models from the first all the way to 2014 seem to come with 800 x 600 @ 167 dpi:
https://wiki.mobileread.com/wiki/Amazon_Kindle_1
https://wiki.mobileread.com/wiki/Kindle_5
https://wiki.mobileread.com/wiki/Kindle_6

houqp:
I see, will document it. In that case, shouldn't we use 167 as default instead in getDPI method?

Frenzie:
Indeed, that's a curious inconsistency. It's probably just a typo in 3d1b071cd5de09dc0fbe0d6ad36a969c78418457 based on the fact that 160 is more logical.

I'm wondering if maybe we should switch the other one to 160 instead since it's basically just a bit of legacy. It's practically the same thing. I doubt you'll even notice a difference in pixels except at really high DPI, except 160 DPI is exactly one inch making for much easier (head) calculations.

spec/unit/framebuffer_spec.lua
```diff
@@ -0,0 +1,30 @@
+describe("Framebuffer unit tests", function()
+    local fb
+
+    setup(function()
+        fb = require("ffi/framebuffer_dummy"):new{
```
Frenzie: Oh, you forced-pushed this dummy in? There was probably overlap with #828 because iirc you used the  	framebuffer_SDL2_0 as opposed to framebuffer_SDL1_2.

I didn't investigate the switch/deprecation yet. It should be as simple as switching from libsdl1.2-dev to libsdl2-dev in the Travis setup).

That being said, a dummy might be a good idea regardless.

houqp: Sorry, could you expand a little bit more on what you mean by overlap with #828? What do you mean by I used framebuffer_SDL2_0?

The eventual goal of this dummy module is to remove the dummy code path from SDL2.

Frenzie: Iirc the initial version of this PR had a call to framebuffer_SDL2_0 which failed because that's not currently in Travis, only after which you added the dummy. Maybe I misremember? No matter, it's not important. :-)

houqp: Ha, I see. Yeah, I replaced the SDL2 dummy call with a proper dummy module :P

houqp:
@Frenzie I submitted the PR for the 160 dpi change because it's very risky and needs more testing.

.luacheckrc
```diff
@@ -4,9 +4,7 @@ std = "luajit"
 -- ignore implicit self
 self = false
 
-globals = {
-    "G_reader_settings",
-}
+globals = {}
```
Frenzie: Worth keeping around just in case?

PR #846, opened 2019-03-12
Add a paintInnerBorder method

NiLuJe:
That paints a border on the inside, instead of the outside.
Doesn't support radius, because maths (and, also, I didn't need it to :D).


NiLuJe:
Okay, Travis really doesn't like framagit today... ^^.

PR #847, opened 2019-03-12
[CI] Switch Travis to Ubuntu Xenial (16.04)

Frenzie:
This PR is a minor WIP to see about side effects, but should mostly be uneventful.


Frenzie:
I could fix easily up the Travis Android build, but I'm inclined to think deprecating it in favor of the Docker image will be easier going forward.

PR #851, opened 2019-03-12
[build] CMake: Introduce DownloadProject module

Frenzie:
This module is like FetchContent in CMake 3.11, but for the moment it may be more inviting to hackers not to require that yet.

The principle is explained here: https://crascit.com/2015/07/25/cmake-gtest/

It allows us to work with CMake in a more natural environment for CMake-submodules, which are otherwise nigh-uncontrollable in between escaping variables in CMake syntax and CMake being extremely bad at taking command-line arguments. (A fact I utterly despise as a regular user just downloading a project to compile & try.)

Upping the minimum CMake version over 3.3 should also allow us to get rid of a few braindead Android-related workarounds for sdcv, where we had to completely force-feed the `.a` suffix.

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -605,7 +605,7 @@ LIBFFI_BUILD_DIR=$(THIRDPARTY_DIR)/libffi/build/$(MACHINE)
 LIBFFI_DIR=$(CURDIR)/$(LIBFFI_BUILD_DIR)/libffi-prefix/src/libffi-build
 
 SDCV_BUILD_DIR=$(THIRDPARTY_DIR)/sdcv/build/$(MACHINE)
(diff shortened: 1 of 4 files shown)


Frenzie:
Drat, I hadn't noticed that the binary seems to be slightly larger. It's not as if 100-200 kB matters that much in the grand scheme of things, but still… I'd like to know why. *mumble, mumble, mumble*

Frenzie:
Oh, this is my own fault. I added a stub but this was never quite properly implemented:

https://github.com/koreader/koreader-base/blob/6d27a176d1bc6124aafe7bee9dbbcaea0651fe35/Makefile.defs#L26-L28

https://github.com/koreader/koreader-base/blob/6d27a176d1bc6124aafe7bee9dbbcaea0651fe35/Makefile.third#L151

PR #852, opened 2019-03-13
 Make the JIT loopunroll tweak mandatory when not using the C blitter

NiLuJe:
We've found more corner-cases not limited to 16bpp since the last
time...
Some of them barely even corner-y ;).

Re: https://github.com/koreader/koreader/pull/4782


poire-z:
On all platforms then? Not only Kobo? You experienced some on Kindle too?

NiLuJe:
No, but I don't see how this could only be Kobo-specific, so, eh. ;).

Frenzie:
Agreed, I think only Android is really unique.

PR #853, opened 2019-03-14
A round of blitbuffer tweaks & fixes

NiLuJe:
* Add fast paths to setPixelBlend (works around https://github.com/koreader/koreader/issues/4774)
* Add fast paths to the C blitter
* Add some color constants and make sure they're actually in the eInk palette
* Use an XOR instead of a subtraction to compute ainv when alpha blending
* Avoid a bunch of useless casts
* Make sure that if a getColor* method returns an alpha value, it returns a fully opaque one
* Fix setPixelBlend when targeting non RGB BB types, as these expect a Grayscale color as input for blend() (actually tackles the root issue behind https://github.com/koreader/koreader/issues/4774)


ffi/blitbuffer.lua
```diff
@@ -1387,14 +1418,18 @@ return a Color value resembling a given level of blackness/gray
...
-BB.COLOR_BLACK = Color8(0)
 BB.COLOR_WHITE = Color8(0xFF)
-BB.COLOR_GREY = Color8(0x80)
-BB.COLOR_LIGHT_GREY = Color8(0xD0)
+BB.COLOR_GRAY_E = Color8(0xEE)
+BB.COLOR_LIGHT_GRAY = Color8(0xCC)
+BB.COLOR_GRAY_A = Color8(0xAA)
+BB.COLOR_GRAY_9 = Color8(0x99)
+BB.COLOR_GRAY = Color8(0x88)
+BB.COLOR_GRAY_5 = Color8(0x55)
```
Frenzie: I think it might be clearer if we call 'em something like:

* `GRAY_5` → `DIM_GRAY`
* `GRAY` →  `DARK_GRAY`
* `GRAY_9` → `WEB_GRAY`
* `GRAY_A` → `GRAY`
* `LIGHT_GRAY`

I plucked those names from <https://cgit.freedesktop.org/xorg/app/rgb/tree/rgb.txt>. The actual shades are slightly different, but should be Close Enough™.

NiLuJe:
Will do, thanks!

poire-z:
Are you going to change all the frontend `BB.COLOR_GRAY` to `BB.COLOR_DARKGRAY`?
Or are we fine with having all our gray stuff changed from `0x80` (or `0x88` earlier in this PR) to now `0xAA` ? (which will make them lightier I guess.)

NiLuJe:
Yep, did the matching switch in https://github.com/koreader/koreader/pull/4789 ;).

PR #855, opened 2019-03-14
[build] Pull in Git submodules with depth 20

Frenzie:
I couldn't include this in #675 due to old Git (mainly in Travis) but it should now be possible.


Frenzie:
I guess there's still some relevant difference between git 2.20.1 on my local system and 2.7.4 on the Docker image in this matter.

Adding `ppa:git-core/ppa` should resolve that issue, but I still regard maintaining compatibility with regular Ubuntu 16.04-era systems out of the box as a soft goal.

PR #856, opened 2019-03-14
[build] Move libjpeg-turbo to DownloadProject

Frenzie:
The idea is that it should make things slightly easier, though that doesn't really seem to apply terribly much here.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/856)


Makefile.third
```diff
@@ -49,8 +49,8 @@ $(TURBOJPEG_LIB) $(JPEG_LIB): $(THIRDPARTY_DIR)/libjpeg-turbo/CMakeLists.txt
 		-DRANLIB="$(RANLIB)" \
```
Frenzie: Needs `-DCMAKE_BUILD_TYPE=$(CMAKE_BUILD_TYPE) \`.

NiLuJe:
Yeah, slightly less obtuse with a layer of indirection removed ;).

PR #861, opened 2019-03-15
[feat, build] Compatibility with other CMake generators

Frenzie:
Pass `USE_NINJA=1` or `CMAKE="cmake -G 'Generator Name'" CMAKE_MAKE_PROGRAM=generator-executable`.


Makefile.defs
```diff
@@ -34,6 +34,11 @@ export CMAKE_BUILD_TYPE
 
 # CMAKE=cmake --no-warn-unused-cli
 CMAKE=cmake
+KO_BUILD_SYSTEM=$(MAKE)
+ifdef USE_NINJA
+	CMAKE:=cmake -G Ninja
+	KO_BUILD_SYSTEM:=ninja
```
Frenzie: @NiLuJe Any feedback on the variable name?

thirdparty/cmake_modules/koreader_thirdparty_common.cmake
```diff
@@ -31,6 +31,15 @@ if(NOT DEFINED PARALLEL_JOBS)
     math(EXPR PARALLEL_JOBS "${PROCESSOR_COUNT}+1")
 endif()
 
+# $(MAKE) is for recursive make invocations, but evidently when using another
+# generator there's no recursion. For us that other generator is ninja, but
+# maybe one day also Visual Studio or Xcode…
+if(CMAKE_GENERATOR MATCHES Makefiles)
+    set(KO_MAKE_RECURSIVE "$(MAKE)")
```
Frenzie: @NiLuJe And also this one.

The system works slightly faster on first build, and definitely provide a speed-up in front on rebuilds, so I pretty much intend to switch over.

NiLuJe: No issues with that one ;).

Makefile.defs
NiLuJe: Not a huge fan, but I'm at a loss for anything better right now...

NiLuJe: (It's mainly the loss of "MAKE" in there that bothers me, as it's recognizable at a glance)

Frenzie: Yeah, I'd actually called it something like `KOMAKE` at first (not sure if that was it; never committed that) but then I looked at `KOMAKE:=ninja` and changed the name. :-P

Frenzie: It could also be called something like `(KO)MAKE_BACKEND`, although that makes me think more of GNU Make vs BSD Make or whatever. But maybe `KO_BUILD_BACKEND` sounds better than system?

Frenzie: Or just something like `KO_CMAKE_GENERATOR` since it is in fact intended for CMake, with that `KO_` prefix to differentiate it from CMake's built-in `CMAKE_GENERATOR` thingy? Or possibly not even that because CMake doesn't care about environment variables like that, although that might be confusing.

Frenzie: Well, except CMake uses names like `Ninja` and `Unix Makefiles` for that one. They *do* use `MAKE` in this one:

```
//Program used to build from build.ninja files.
CMAKE_MAKE_PROGRAM:FILEPATH=/usr/bin/ninja
```

So… `(KO_)CMAKE_MAKE_PROGRAM`?

NiLuJe: Yep, I like that one ;)!

PR #863, opened 2019-03-16
More alpha-blending fixes

NiLuJe:
Mainly add a method to properly handle MuPDF's premultiplied alpha


Frenzie:
lgtm, but maybe we could slightly expand on the tests? Seems kinda easy to break when modding otherwise. :-)

https://github.com/koreader/koreader-base/blob/e2ea51dad6b46cedde8ac1dcc9e6120d802466c3/spec/unit/blitbuffer_spec.lua#L306-L321

NiLuJe:
That can be arranged ;). Any specific ideas besides a simple pmulblend test?

Frenzie:
Not really beyond that atm, no.

NiLuJe:
I'll add a few when I get back home ;).

PR #866, opened 2019-03-17
[build] Unified CMake thirdparty/build dir

Frenzie:
Just a passing thought, putting it up in a PR on GH for public archival/inspiration purposes or possibly to extend upon. This is a proof of concept of building in thirdparty/build/$(MACHINE) instead of thirdparty/build/$DEPENDENCY/$(MACHINE).

That way instead of using the Makefile.third with a lot of complex variable passing to CMake etc. you'd keep the logic almost entirely in CMake (not implemented in this proof of concept), so you should be able to write things like `add_dependencies(mupdf zlib)`.

Also the dist-clean command could be simplified to removing `THIRDPARTY_BUILD_DIR`.

Or to put it another way, if you're using CMake you might as well actually use it, so that errors like https://github.com/koreader/koreader-base/pull/865 can't occur.

### DIFF
--- a/thirdparty/zlib/CMakeLists.txt
+++ b/thirdparty/zlib/CMakeLists.txt
@@ -1,4 +1,4 @@
-PROJECT(zlib)
+PROJECT(zlib-download)
 cmake_minimum_required(VERSION 3.5.1)
 
 SET(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_LIST_DIR}/../cmake_modules")
@@ -11,6 +11,10 @@ assert_var_defined(CFLAGS)
 assert_var_defined(LDFLAGS)
 
 ep_get_source_dir(SOURCE_DIR)
+
+if ($ENV{ANDROID})
+    set(LDFLAGS "${LDFLAGS} -shared -Wl,-soname,libz.so.1")
+endif()
 
 if($ENV{WIN32})
     # configure
@@ -44,3 +48,7 @@ ExternalProject_Add(
     BUILD_COMMAND ${BUILD_CMD}
     INSTALL_COMMAND ${INSTALL_CMD}
 )
+
+set(ZLIB_INCLUDE_DIR "${SOURCE_DIR}/include" PARENT_SCOPE)
+set(ZLIB_LIB_DIR "${SOURCE_DIR}/lib" PARENT_SCOPE)
+set(ZLIB_STATIC "${SOURCE_DIR}/lib/libz.${CMAKE_STATIC_LIBRARY_SUFFIX}" PARENT_SCOPE)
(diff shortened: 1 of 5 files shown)


thirdparty/zlib/CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ set(ZLIB_VER "1.2.11")
 set(ZLIB_MD5 "1c9f62f0778697a09d36121ead88e08e")
 ExternalProject_Add(
     ${PROJECT_NAME}
-    DOWNLOAD_DIR ${KO_DOWNLOAD_DIR}
+    DOWNLOAD_DIR @KO_DOWNLOAD_DIR@
```
Frenzie: To clarify, `configure_file()` copies a file and replaces all variables. By using `configure_file()` with the `@ONLY` option regular variables remain intact and are processed as they would normally, but only `@VARIABLE@` is replaced.

thirdparty/zlib/CMakeLists.txt
```diff
@@ -36,11 +40,15 @@ set(ZLIB_VER "1.2.11")
...
     URL_MD5 ${ZLIB_MD5}
     BUILD_IN_SOURCE 1
     ${optional_cfg_cmd}
     BUILD_COMMAND ${BUILD_CMD}
     INSTALL_COMMAND ${INSTALL_CMD}
 )
+
+set(ZLIB_INCLUDE_DIR "${SOURCE_DIR}/include" PARENT_SCOPE)
+set(ZLIB_LIB_DIR "${SOURCE_DIR}/lib" PARENT_SCOPE)
+set(ZLIB_STATIC "${SOURCE_DIR}/lib/libz.${CMAKE_STATIC_LIBRARY_SUFFIX}" PARENT_SCOPE)
```
Frenzie: Probably also needs `ZLIB`

The idea being not to do all this stuff quite so excessively:

https://github.com/koreader/koreader-base/blob/6582561d074cf332ba1e9df9a459fe09b0ee4b26/Makefile.third#L130-L157

Frenzie:
@NiLuJe Do you think the concept as it stands now seems sound?

NiLuJe:
In theory, yes.

In practice, I guess it'll take very very very carefully migrating the myriad of package-specific crap we pass from Makefile.third to each individual CMakefile ;).

NiLuJe:
For instance, all the various paths shenanigan might be *less* obvious and/or easy to deal with?

Frenzie:
As it turns out I was mistaken about being able to define dependencies in CMake.

> CMake Error at CMakeLists.txt:24 (add_dependencies):
  Cannot add target-level dependencies to non-existent target "sdcv".
>
>  The add_dependencies works for top-level logical targets created by the
  add_executable, add_library, or add_custom_target commands.  If you want to
  add file-level dependencies see the DEPENDS option of the add_custom_target
  and add_custom_command commands.

Because this is basically just a very bad Makefile PHONY type thing (I mean, Makefile syntax can be a bit strange but CMake syntax for escaping commands and such is incomprehensible gibberish):
https://cmake.org/cmake/help/v3.14/command/add_custom_target.html

There would be zero advantage to that.

Perhaps there might be something to the build directory structure reorganization I experimented with here though. That doesn't need CMake, just a few clever variable changes.

PR #868, opened 2019-03-17
[chore] CMake: deprecate custom processor count logic

Frenzie:
CMake <ins>2.8.5</ins> shipped with a built-in module.

Also quote `KO_DOWNLOAD_DIR`.

### DIFF
--- a/thirdparty/cmake_modules/koreader_thirdparty_common.cmake
+++ b/thirdparty/cmake_modules/koreader_thirdparty_common.cmake
@@ -1,9 +1,8 @@
 if(NOT DEFINED PROCESSOR_COUNT)
     include(ProcessorCount)
     ProcessorCount(N)
-    if(NOT N EQUAL 0)
-        set(PROCESSOR_COUNT ${N})
-    endif()
+    # 0 if unknown
+    set(PROCESSOR_COUNT ${N})
 endif()
 
 if(APPLE)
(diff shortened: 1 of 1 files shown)


Frenzie:
Today it looks like this: https://github.com/Kitware/CMake/blob/283330976f2379eec214bfe1d51a27c29cc1f8fc/Modules/ProcessorCount.cmake

For 3.5 it looks like this: https://github.com/Kitware/CMake/blob/64130a7e793483e24c1d68bdd234f81d5edb2d51/Modules/ProcessorCount.cmake

Either way, same idea but more complete with "automatic" updates & improvements.

thirdparty/cmake_modules/koreader_thirdparty_common.cmake
```diff
@@ -1,24 +1,8 @@
...
-  endif()
-
-  # Windows:
-  if(WIN32)
-    set(PROCESSOR_COUNT "$ENV{NUMBER_OF_PROCESSORS}")
-  endif()
+    include(ProcessorCount)
+    ProcessorCount(N)
+    # 0 if unknown
+    set(PROCESSOR_COUNT ${N})
```
NiLuJe: I'm assuming N falls-back to 0, right? ;).

NiLuJe: Yes it does (reading the links provided helps. :D).

PR #873, opened 2019-03-20
[fix] Add util.ffiLoadCandidates to try multiple SDL2 library names

Frenzie:
Fixes <https://github.com/koreader/koreader/issues/4826#issuecomment-475043234>.

A bit like the idea penned down [here](https://github.com/koreader/koreader/issues/3108#issuecomment-377548634).

### DIFF
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -19,7 +19,11 @@ local dummy = require("ffi/linux_input_h")
 
 -----------------------------------------------------------------
 
-local SDL = ffi.load("SDL2")
+local SDL = util.ffiLoadCandidates{
+    "SDL2",
+    -- this unfortunately needs to be written in full due to the . in the name
+    "libSDL2-2.0.so",
+}
 
 -- for frontend SDL event handling
 local EV_SDL = 53 -- ASCII code for S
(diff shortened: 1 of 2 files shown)


xelxebar:
Unfortunately, even with this fix, I ran into the missing libSDL.so issue. I'm running a Void Linux chroot, and the problem is that it ships libraries with version strings in the filename: *e.g.* `libSDL2-2.0.so.0.9.0`. Maintaining a list of all possible filename permutations seems like the Cthulhu solution, though.

Frenzie:
No symlinks of any sort? On my Debian `libSDL2-2.0.so` is a symlink to `libSDL2-2.0.so.0`, which is a symlink to `libSDL2-2.0.so.0.9.0` (the real deal). 

On older versions it's the same, except on Ubuntu 16.04, for example, the actual library is `libSDL2-2.0.so.0.4.0`.

`libSDL2.so` is also a symlink to `libSDL2-2.0.so.0.9.0`, but apparently that only happens if you have the `libsdl2-dev` package installed.

xelxebar:
Looks like I've just got `/usr/lib/libSDL2-2.0.so.0 -> libSDL2-2.0.so.0.9.0*`, without the `libSDL2-2.0.so` one.

Frenzie:
Alright, so then we can add `libSDL2-2.0.so.0` to the list without having to check for the whole list of `libSDL2-2.0.so.0.?.0` possibilities. :-)

PR #876, opened 2019-03-21
[chore] Remove luajit-clean target

Frenzie:
Deprecated by https://github.com/koreader/koreader-base/pull/567.

Originally added in https://github.com/koreader/koreader-base/pull/531 and https://github.com/koreader/koreader-base/pull/533.


Frenzie:
@NiLuJe Unless there's a reason it should stay?

NiLuJe:
None that I can see OTOH ;).

PR #878, opened 2019-03-22
Faster blitting for BB8 & BBRGB32 when no processing is needed

NiLuJe:
This basically attempts to use memcpy for simple blitting of BB8-to-BB8 and BBRGB32-to-BBRG32 if possible, as well as using memset for paintRect.

"no processing" implies: no invert, no blending (of any kind), and *no rotation*. That means non-inverted Portrait from our POV, but *also* from the fb's POV. This means Kobos are out unless started from Nickel on post-32bpp FWs (except on the Forma, where we enforce a proper orientation in every case).

Leaving this for review, it *should* be essentially done and non-crashy, but I still have some debug prints to kill before it's ready to merge ;).

Tested @ 8bpp & 32bpp, both with the Lua BB and the C BB.


NiLuJe:
Also fixes a stupid mistake from #863 that broke setPixelBlend on >= 16bpp (noticed it thanks to dimRect).

NiLuJe:
Also has a fairly unrelated LuaSocket bump & cleanup, because, err, it was in my tree when I started :D.

NiLuJe:
And a few minor cosmetic/linting cleanups (mainly around BitOp usage, and odd/even checks).

NiLuJe:
And an actual fix for the various *Rect to not try to paint OOB pixels.

poire-z:
Trusting you on all the maths and loops :)

Just a few questions and remarks:
```
We're only using fill with BB COLOR constants, which are grayscale.
So we can pretty much always use filling instead of a pixel loop ;).
```
We are, but should the function really be limited by that fact? (some might be using our code to do colored stuff on specific colored platforms). Isn't it worth a quick check the color is really grayscale to take that shortcut? (or is this function already just used in pure 8bpp calls?)

Regarding the `%` (modulo) changed to `bit.band`, is it really faster? (I assume % arithmetic in lua is still about floats as numbers are floats internatlly I think - but `bit.band()` is an additional function call.)

> This means Kobos are out unless started from Nickel 

:( We're really not able to do what [Nps]ickel do? With a `fbdepth++` maybe ? :)

Frenzie:
Like I said on Gitter, this is basically the better version of something that seemed like a fairly obvious potential improvement.

NiLuJe:
The C version was already enforcing grayscale, so that kind of sealed the
deal for me ;).

Yeah, I could technically try to set a sane rotation via fbdepth, but I'm
definitely gonna get it wrong on a few devices at first ;).

NiLuJe:
I'm assuming Lua isn't being a special snowflake here, so, yeah, AND should
be much faster than a modulo.

The JIT should be handling BitOps pretty effectively.

NiLuJe:
Okay, I have some fbdepth shenanigans waiting in the wings to enforce Portrait.

Might not make it into this PR, but they're ready ;).

poire-z:
Wow :) (I was half assuming it was impossible :)
Just to be certain I understand what that will do: it will solve all the rotation mess I mentionned at bottom of https://github.com/koreader/koreader-base/pull/746#issuecomment-431884723 ?

NiLuJe:
@poire-z : Yeah, the whole thing is essentially a souped-up version of that idea, and, yes, enforcing a Portrait orientation on the fb's side makes it viable codepath on Kobo (indeed, for CRe page blits, in a single memcpy).

We still have `isAlwaysPortrait` enabled, but being in a native Portrait otientation ensures this branch is never taken, so our screen BB is unrotated ;).

https://github.com/koreader/koreader-base/blob/a0088f5803fd05851704c792b3add0893b47b945/ffi/framebuffer.lua#L77-L89

NiLuJe:
As usual, the main question mark will be the H2O² which might end up upside down until I get some logs, but it should be okay on other devices.

NiLuJe:
The real solution is of course to handle rotation at the fb's level (like Plato), but that's much more intrusive in our case, as we support a number of platforms where it's plain impossible (legacy Kindle), utterly quirky (Kobo), or plain broken (I'm going to assume PocketBook :D), and that's only speaking for Linux fbdev platforms.

NiLuJe:
FWIW, fbdepth updates are in FBInk master, and the startup script update in my matching https://github.com/NiLuJe/koreader/commits/blitcpy frontend branch ;).

poire-z:
Had a quick try, with just replacing ffi/blitbuffer.lua from this PR, koreader.sh from your branch, and fbdepth recompiled for kobo with your https://github.com/NiLuJe/FBInk/commit/d0c85988eca6c1eee78f9418e00c60d5a43eff9e.
Not going too well :)
- It started with portrait mode upside down ("Opening ... blah.epub", the menu, Book status, History).
- Footer was displayed (inverted) at top, but the cre book (in portrait mode) was unreadable: like scanlines too short or too long, making some diagonal noise.
- Switched to a epub I know I had in 2-pages-landscape mode, and it displayed fine! (possibly inverted, dunno, but touch did not match book orientation)!
- Switched to a pdf, and it displayed ok, but inverted. Tapping bottom right (so tap , on top left) must have switched to some page flipping mode: and it no more displayed fine: same diagonal stuff as for cre. Tapping again to get out of this mode made the PDF displayed ok again. Actually, it seems only "swith to content width is fine. two fingers pinch or spread makes bad except when spread horizontal.
- Touch was not inverted, so now it's easier to get out of that with gestures if you set one for exit :)
- KSM (07) was fine when exiting koreader
- Same after a reboot

poire-z:
Logs for a first start after a reboot:
```
Original fb rotation is set @ 0
Original fb bitdepth is set @ 16bpp
Switching fb bitdepth to 8bpp & rotation to Portrait
Device's expected Portrait orientation should be: 1 (Clockwise, 90Â°)!

Variable fb info: 1448x1072 (1472x2304), 16bpp @ rotation: 0 (Upright, 0Â°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 6782976 bytes & line length: 2944 bytes

Switching fb to 8bpp @ rotation 1 . . .
Setting bitdepth to 8bpp
Setting grayscale to 1
Setting rotate to 1 (Clockwise, 90Â°)
Bitdepth is now 8bpp (grayscale: 1) @ rotate: 1 (Clockwise, 90Â°)

Variable fb info: 1072x1448 (1088x3072), 8bpp @ rotation: 1 (Clockwise, 90Â°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 6782976 bytes & line length: 1088 bytes
---------------------------------------------
                launching...
SDL2 not loaded:    ./setupkoenv.lua:33: Not able to load dynamic library: SDL2
SDL not loaded: ./setupkoenv.lua:33: Not able to load dynamic library: SDL
Tweaking LuaJIT's max loop unroll factor (15 -> 45)
FB: computed size: 3342336
FB: id           : mxc_epdc_fb
FB: smem_len     : 6782976
FB: line_length  : 1088
FB: xres         : 1072
FB: xres_virtual : 1088
FB: yres         : 1448
FB: yres_virtual : 3072
FB: bpp          : 8
03/23/19-00:33:33 INFO  initializing for device Kobo_alyssum
03/23/19-00:33:33 INFO  framebuffer resolution: {
    ["h"] = 1448,
    ["w"] = 1072
}
[ko-input] Forked off fake event generator(pid:1011).
BB8_mt.__index.blitTo8: cdata<struct BlitBuffer8>: 0x2bf08198   2   12  0   0   29  31  function: 0x2ac13d18    Color8A(0, 255)
Can't do a fast blit from BB8 to BB8! setter    function: 0x2ac13d18    src rota    0   dest rota   0   src invert  0   dest invert 0
blitDefault cdata<struct BlitBuffer8>: 0x2bf021f8   cdata<struct BlitBuffer8>: 0x2bf08198   2   12  0   0   29  31  function: 0x2ac13d18    Color8A(0, 255)
BB8_mt.__index.blitTo8: cdata<struct BlitBuffer8>: 0x2bf08198   37  19  0   0   22  35  function: 0x2ac13d18    Color8A(0, 255)
Can't do a fast blit from BB8 to BB8! setter    function: 0x2ac13d18    src rota    0   dest rota   0   src invert  0   dest invert 0
paintRect   104 544 864 360 Color8(255) nil
paintRect checked dimensions    104 544 864 360
Scanline fill paintRect
paintRect   104 544 864 3   Color8(0)   nil
paintRect checked dimensions    104 544 864 3
Scanline fill paintRect
paintRect   104 901 864 3   Color8(0)   nil
paintRect checked dimensions    104 901 864 3
Scanline fill paintRect
paintRect   104 547 3   354 Color8(0)   nil
paintRect checked dimensions    104 547 3   354
Scanline fill paintRect
paintRect   965 547 3   354 Color8(0)   nil
paintRect checked dimensions    965 547 3   354
Scanline fill paintRect
blitDefault cdata<struct BlitBufferRGB24>: 0x2befa6a8   cdata<struct BlitBuffer8>: 0x2ac36130   116 670 0   0   1
BB8_mt.__index.blitTo8: cdata<struct BlitBuffer8>: 0x2ac36130   241 556 0   0   714 336 function: 0x2ac139f0    n
BB8 to BB8 scanline copy
03/23/19-00:33:34 INFO  opening file /mnt/onboard/...
[...]
Restoring original fb bitdepth @ 16bpp & rotation @ 0
Variable fb info: 1072x1448 (1088x3072), 8bpp @ rotation: 1 (Clockwise, 90Â°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 6782976 bytes & line length: 1088 bytes

Switching fb to 16bpp @ rotation 0 . . .
Setting bitdepth to 16bpp
Setting grayscale to 0
Setting rotate to 0 (Upright, 0Â°)
Bitdepth is now 16bpp (grayscale: 0) @ rotate: 0 (Upright, 0Â°)

Variable fb info: 1448x1072 (1472x2304), 16bpp @ rotation: 0 (Upright, 0Â°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 6782976 bytes & line length: 2944 bytes
[Back to KSM OK]
```

NiLuJe:
@poire-z : Yeah, that's because fbdepth is doing something stupid, which ends up coputing a bogus Portrait orientation (... which happened to be the right one for an H2O, by chance) :D.

NiLuJe:
Okay, current master should have a less buggy fbdepth ;). Sorry about that -_-".

poire-z:
Just had a try with https://github.com/NiLuJe/FBInk/commit/c2e913213e4b2517795bbe24d02cde30a9884f54.
KOReader orientation is now allright, and widgets display fine.
But this still happens:
- the cre book (in portrait mode) was unreadable: like scanlines too short or too long, making some diagonal noise.
- Switched to a epub I know I had in 2-pages-landscape mode, and it displayed fine! Touch orientation was correct.
- Switched to a pdf, and ~~it displayed ok, but inverted. Tapping bottom right (so tap , on top left) must have switched to some page flipping mode: and it no more displayed fine: same diagonal stuff as for cre. Tapping again to get out of this mode made the PDF displayed ok again. Actually, it seems only "swith to content width is fine. two fingers pinch or spread makes bad except when spread horizontal.~~ displayed bad with some diagonal noise in all zoom mode now - couldn't find one that displayed ok.

So, possibly some issue with this PR fast blit code? A thing with line_length or xres_virtual > xres?

```
Original fb rotation is set @ 0
Original fb bitdepth is set @ 16bpp
Switching fb bitdepth to 8bpp & rotation to Portrait
[FBInk] Detected a Kobo Glo HD (371)
Variable fb info: 1448x1072 (1472x2304), 16bpp @ rotation: 0 (Upright, 0Â°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 6782976 bytes & line length: 2944 bytes
Device's expected Portrait orientation should be: 3 (Counter Clockwise, 270Â°)!

Switching fb to 8bpp @ rotation 3 . . .
Setting bitdepth to 8bpp
Setting grayscale to 1
Setting rotate to 3 (Counter Clockwise, 270Â°)
Bitdepth is now 8bpp (grayscale: 1) @ rotate: 3 (Counter Clockwise, 270Â°)

Variable fb info: 1072x1448 (1088x3072), 8bpp @ rotation: 3 (Counter Clockwise, 270Â°)
Fixed fb info: ID is "mxc_epdc_fb", length of fb mem: 6782976 bytes & line length: 1088 bytes
---------------------------------------------
                launching...
 [*] Current time: 03/23/19-07:54:02
 [*] Version: v2019.03.1-33-g83da4d1_2019-03-18

SDL2 not loaded:    ./setupkoenv.lua:33: Not able to load dynamic library: SDL2
SDL not loaded: ./setupkoenv.lua:33: Not able to load dynamic library: SDL
Tweaking LuaJIT's max loop unroll factor (15 -> 45)
FB: computed size: 3342336
FB: id           : mxc_epdc_fb
FB: smem_len     : 6782976
FB: line_length  : 1088
FB: xres         : 1072
FB: xres_virtual : 1088
FB: yres         : 1448
FB: yres_virtual : 3072
FB: bpp          : 8
03/23/19-07:54:02 INFO  initializing for device Kobo_alyssum
03/23/19-07:54:02 INFO  framebuffer resolution: {
    ["h"] = 1448,
    ["w"] = 1072
}
[ko-input] Forked off fake event generator(pid:1015).
BB8_mt.__index.blitTo8: cdata<struct BlitBuffer8>: 0x2bf48208   2   12  0   0   29  31  function: 0x2ad88c60    Color8A(0, 255)
Can't do a fast blit from BB8 to BB8! setter    function: 0x2ad88c60    src rota    0   dest rota   0   src invert  0   dest invert 0
blitDefault cdata<struct BlitBuffer8>: 0x2bf40490   cdata<struct BlitBuffer8>: 0x2bf48208   2   12  0   0   29  31  function: 0x2ad88c60    Color8A(0, 255)
BB8_mt.__index.blitTo8: cdata<struct BlitBuffer8>: 0x2bf48208   37  19  0   0   22  35  function: 0x2ad88c60    Color8A(0, 255)
Can't do a fast blit from BB8 to BB8! setter    function: 0x2ad88c60    src rota    0   dest rota   0   src invert  0   dest invert 0
blitDefault cdata<struct BlitBuffer8>: 0x2bf41880   cdata<struct BlitBuffer8>: 0x2bf48208   37  19  0   0   22  35  function: 0x2ad88c60    Color8A(0, 255)
paintRect   104 572 864 304 Color8(255) nil
paintRect checked dimensions    104 572 864 304
Scanline fill paintRect
paintRect   104 572 864 3   Color8(0)   nil
paintRect checked dimensions    104 572 864 3
Scanline fill paintRect
paintRect   104 873 864 3   Color8(0)   nil
paintRect checked dimensions    104 873 864 3
Scanline fill paintRect
paintRect   104 575 3   298 Color8(0)   nil
paintRect checked dimensions    104 575 3   298
Scanline fill paintRect
paintRect   965 575 3   298 Color8(0)   nil
paintRect checked dimensions    965 575 3   298
Scanline fill paintRect
blitDefault cdata<struct BlitBufferRGB24>: 0x2bf38f98   cdata<struct BlitBuffer8>: 0x2ad63298   116 670 0   0   107 107 function: 0x2ad88938    nil
BB8_mt.__index.blitTo8: cdata<struct BlitBuffer8>: 0x2ad63298   241 584 0   0   714 280 function: 0x2ad88938    nil
BB8 to BB8 scanline copy

03/23/19-07:54:03 INFO  opening file /mnt/onboard/.LIBRARY/Wikipedia/Gilgamesh.FR.epub
BB8_mt.__index.blitTo8: cdata<struct BlitBuffer8>: 0x2ad63298   0   0   0   0   1072    1448    function: 0x2ad88938    nil
BB8 to BB8 full copy
paintRect   0   1420    1072    28  Color8(255) nil
paintRect checked dimensions    0   1420    1072    28
Single fill paintRect
paintRect   22  1426    433 4   Color8(255) nil
paintRect checked dimensions    22  1426    433 4
Scanline fill paintRect
```

NiLuJe:
Okay, so full copies are broken, probably because xres is not
xres_virtual...

Hmm, this one'll take a bit of thinking...

Thanks for testing :).

NiLuJe:
(of course, I tested this on the Forma, where xres is equal to xres_virtual, so this works...).

But the actual FB scanlines are _virtual sized, so, yeah, basically that width check needs to coloar the against that, not getWidth.

Which means no single copy on a lot of devices out there, as those with an aligned xres are the exception...

NiLuJe:
A viewport wouldn't help, as it's using the same storage as the real BB...

And somehow I don't see asking CRe to pad its working buffer by virtual - x px just for us ;).

poire-z:
What's the reason for some devices having xres != xres_virtual ? A bigger framebuffer than the screen? The eInk controller working only with "some" sizes? Just some software stuff ?

> And somehow I don't see asking CRe to pad its working buffer by virtual - x px just for us as a viable (or easy) solution ;).

Mhhh, that might work :) Give it a blitbuffer of size W2xH and keep telling it the render view is W1xH, and it might just do that :)
But well, everything would become quite ugly fast if we were to do that :)

But why the UI is ok while mupdf & cre are not ? Because we only draw smaller-than-screen stuff, so not taking the fast path?

poire-z:
Render view https://github.com/koreader/koreader/blob/54b0e85bddda37be23013f4e4063d50738710d9b/frontend/document/credocument.lua#L119
vs 
blitting (so, possibly to be taken care of in readerview.lua, which is callign that) https://github.com/koreader/koreader/blob/54b0e85bddda37be23013f4e4063d50738710d9b/frontend/document/credocument.lua#L355-L374

Looking at my comment, cre can do RGB32 - there's possibly just some color shift to solve.

poire-z:
> Which means no single copy on a lot of devices out there

Well, line by line must already be quite a boost.

NiLuJe:
@poire-z : The discrepancy between xres & xres_virtual is because the pixel processor from the eInk controller has specific alignment requirements, so the kernel needs to have some extra scratch space to correct for that in the driver. Hence xres_virtual being aligned to 32 bits boundaries.

c.f., https://github.com/koreader/koreader-base/blob/a0088f5803fd05851704c792b3add0893b47b945/ffi/framebuffer_linux.lua#L129-L135

But xres is limited by the actual panel size ;). On the Forma, both happen to match.

----

Yeah, messing with CRe like that seems like a massive kludge ;).

----

The UI is okay exactly because of that, nothing else is actually doing blits start start at the beginning of a line, and with a width exactly matching the fb line lenght ;).
So that goes through the scanline-by-scanline path, because it can handle x offsets.
And as we don't cross scanline boundaries, padding is irrelevant there.

NiLuJe:
I have a few ideas, and a few notes, but I probably won't be able to get to it until Monday ;).

poire-z:
> Plus, you'd have to be able to convey that the extra padding on the side shouldn't actually be rendered to, as it's off-screen ;). (i.e., make a viewport).

(with my 2 credocument code links above, I meant I think all that is already there: the first link sets the viewport/canvas - the second link create a blitbuffer, which may be of a different size, than crengine will draw its viewport/canvas on - so the extra width on the right of the blitbuffer would be untouched)

NiLuJe:
@poire-z: I couldn't manage to untangle where the visible_area was coming from, but a quick'n dirty `if rect.w == 1080 then rect.w = 1088 end` yielded unwelcome results: the buffer was properly sized (and I could get single copy blits by fixing the backend), but CRe was also rendering to a larger canvas, so centering was slightly off...

NiLuJe:
Note for potential testers that this now requires a rebuild not to crash, as I updated the BlitBuffer* structs, which are used by our C shims.

NiLuJe:
Okay, things should be saner on devices with padded scanlines.

I'll check if I didn't utterly break it on the Forma tomorrow ;).

Here's a [release build](http://files.ak-team.com/niluje/kopub/koreader-kobo-arm-kobo-linux-gnueabihf-v2019.03.1-50-g60577e9f_2019-03-25.zip), because building custom branches is annoying.

NiLuJe:
TL;DR; on my various edits re CRe: AFAICT, no easy silver bullet, short of enforcing a right margin larger than usual by `phys_w - w`....

poire-z:
Regarding your https://github.com/NiLuJe/koreader/commit/60577e9f0b14cd55e88eea3b361db9cbaa4b9de2 , I think it's a small check I forgot in https://github.com/koreader/koreader/pull/4805.
You must have long-press on portrait|landscape and made a global default on your H20 and not on your other devices :)
Anyway, I think we shouldn't have these `values` and `default_value` because that Orientation toggle is different from others, being dynamic. The patch should be in what I added in that PR, this seems enough:
```diff
--- a/frontend/ui/widget/configdialog.lua
+++ b/frontend/ui/widget/configdialog.lua
@@ -356,7 +356,7 @@ function ConfigOption:init()
                 end
                 local default_option_name = self.config.config_options.prefix.."_"..self.options[c].name
                 local default_value = G_reader_settings:readSetting(default_option_name)
-                if default_value then
+                if default_value and self.options[c].values then
                     local val = default_value
                     local min_diff
                     if type(val) == "table" then

```

poire-z:
Had a go with your build (you forgot to bump fbink :) so I took the one I built some days ago - without your fbink 2 latest commits - they shouldn't matter).

Quite ok! and I think I notice the speedup ("think", because temperature, Earth's electromagnetism, global economy... all have an impact on eInk... but it is quite noticable :)

With CRE, mostly only `Scanline fill paintRect` `BB8 to BB8 scanline copy`, rare `Single fill paintRect` (possibly for the clear/full white paint of the bb).
With a PDF, I get a few more `Single fill paintRect` and `BB8 to BB8 full copy`, but mostly scanline too, and a few `Old-style paintRect pixel loop`. I feel that speedup is even more noticable on PDF.
No issue so far :)

poire-z:
> but a quick'n dirty if rect.w == 1080 then rect.w = 1088 end at the top of CreDocument:drawCurrentView yielded unwelcome results
EDIT: Ah, drawCurrentPage in cre.cpp is resizing to bb->wxbb->h, that's probably it...
EDIT²: And for good reason, it also needs to know the actual length of a scanline in the buffer it's drawing to

I don't think it needs to. I think there can't be any buffer full line filling or drawing in crengine, as there are always the margins we set that should prevent crengine from taking any shortcut when drawing.

I can get `BB8 to BB8 full copy` with CRE with these changes:

Assuming this is done correctly on init and rotation changes elsewhere - so no need to check/resize/render (this is only a check, no resize or render if w and h match what's currently set) when drawing, allowing oversized BB:
```diff
--- a/cre.cpp
+++ b/cre.cpp
@@ -2499,8 +2499,8 @@ static int drawCurrentPage(lua_State *L) {
        int drawn_images_count;
        int drawn_images_surface;

-       doc->text_view->Resize(w, h);
-       doc->text_view->Render();
+//     doc->text_view->Resize(w, h);
+//     doc->text_view->Render();
        if (color) {
                /* Use Color buffer - caller should have provided us with a
                  * Blitbuffer.TYPE_BBRGB16, see CreDocument:drawCurrentView
```

Having checkBounds() checks for the physical width so to not reduce our excess:
```diff
--- blitbuffer.lua
+++ blitbuffer.lua
@@ -984,9 +982,9 @@
 function BB_mt.__index:blitFrom(source, dest_x, dest_y, offs_x, offs_y, width, height, setter, set_param)
     width, height = width or source:getWidth(), height or source:getHeight()
     -- NOTE: If we ever manage to convince CRe to render to a padded buffer (to match phys_w and allow us single-copy blitting),
     --       change the self:get* calls to self:getPhysical* ones ;).
-    width, dest_x, offs_x = BB.checkBounds(width, dest_x or 0, offs_x or 0, self:getWidth(), source:getWidth())
+    width, dest_x, offs_x = BB.checkBounds(width, dest_x or 0, offs_x or 0, self:getPhysicalWidth(), source:getWidth())
     height, dest_y, offs_y = BB.checkBounds(height, dest_y or 0, offs_y or 0, self:getHeight(), source:getHeight())
     if width <= 0 or height <= 0 then return end

     if not setter then setter = self.setPixel end
```
And some hardcoding in credocument.lua (haven't checked that rect.copy() is really needed - just assumed there would be less side effect with that):
```diff
 function CreDocument:drawCurrentView(target, x, y, rect, pos)
+    rect = rect:copy()
+    if rect.w == 1072 then
+      rect.w = 1088
+      print("switched bb from 1072 to 1088")
+    end
     if self.buffer and (self.buffer.w ~= rect.w or self.buffer.h ~= rect.h) then
         self.buffer:free()
         self.buffer = nil
     end
     if not self.buffer then
         -- Note about color rendering:
         -- If we use TYPE_BBRGB32 (and LVColorDrawBuf drawBuf(..., 32) in cre.cpp),
         -- we get inverted Red and Blue in the blitbuffer (could be that
         -- crengine/src/lvdrawbuf.cpp treats our 32bits not as RGBA).
         -- But it is all fine if we use TYPE_BBRGB16.
         self.buffer = Blitbuffer.new(rect.w, rect.h, self.render_color and Blitbuffer.TYPE_BBRGB16 or nil)
+        logger.warn("created buffer of w", rect.w)
     end
     -- TODO: self.buffer could be re-used when no page/layout/highlights
     -- change has been made, to avoid having crengine redraw the exact
     -- same buffer.
     -- And it could only change when some other methods from here are called
     self._drawn_images_count, self._drawn_images_surface_ratio = self._document:drawCurrentPage(self.buffer, self.render_color)
+        logger.warn("buffer have w", self.buffer.w, rect.w, "target has w", target.w)
     target:blitFrom(self.buffer, x, y, 0, 0, rect.w, rect.h)
 end
```

(But no additional speedup felt :)

NiLuJe:
@poire-z : Ooh, hadn't checked PDFs, I was wondering how you could get full blits there, but I think it's the pre-render, nice :].

I'll double-check you full-blit trick, thanks ;).

NiLuJe:
Huh. I smell a crappy memcpy implementation in the libc :D.

Indeed, switching to full blits has a negligible impact on my H2O (it might even be a tiny bit slower!).

```diff
diff --git a/frontend/document/credocument.lua b/frontend/document/credocument.lua
index d16e9f93..fe0b512e 100644
--- a/frontend/document/credocument.lua
+++ b/frontend/document/credocument.lua
@@ -353,6 +353,10 @@ function CreDocument:getNextVisibleChar(xp)
 end
 
 function CreDocument:drawCurrentView(target, x, y, rect, pos)
+    if rect.w == target.w and rect.h == target.h then
+        rect.w = target.phys_w
+        print("Switched bb rect to phys_w")
+    end
     if self.buffer and (self.buffer.w ~= rect.w or self.buffer.h ~= rect.h) then
         self.buffer:free()
         self.buffer = nil
@@ -369,8 +373,12 @@ function CreDocument:drawCurrentView(target, x, y, rect, pos)
     -- change has been made, to avoid having crengine redraw the exact
     -- same buffer.
     -- And it could only change when some other methods from here are called
+    local start_clock = os.clock()
     self._drawn_images_count, self._drawn_images_surface_ratio = self._document:drawCurrentPage(self.buffer, self.render_color)
+    print("CreDocument:drawCurrentView: Rendering took", os.clock() - start_clock)
+    start_clock = os.clock()
     target:blitFrom(self.buffer, x, y, 0, 0, rect.w, rect.h)
+    print("CreDocument:drawCurrentView: Blitting took", os.clock() - start_clock)
 end
 
 function CreDocument:drawCurrentViewByPos(target, x, y, rect, pos)
```

I generally get rendering times between `0.05` and `0.06`, scanline blitting averages around `0.0065`, full blitting averages around `0.00725`, classic blitting averages around `0.0375` *with a primed optimizer*, on a cold boot, or when the JIT is getting its panties in a twist, it averages at around `0.242`.

NiLuJe:
@poire-z : So, I guess we're agreed it's not worth it trying to pile on a few hacks around CRe just to get full blits where they might not actually help all that much? ;).

NiLuJe:
And, FWIW, similar results with the C blitter (although both variants are much closer together, and on the faster end of the spectrum).

Average timing are roughly similar to the Lua codepaths, which is neat, and makes sense, given that much of it is no longer really Lua ;).

NiLuJe:
Usual warnings that I'm using `os.clock()` for simplicity's sake, so timings are a bit iffy, depend on the platform and the clock speed and all that ;).

NiLuJe:
Here's a [test binary](http://files.ak-team.com/niluje/kopub/koreader-kobo-arm-kobo-linux-gnueabihf-v2019.03.1-54-g70542770_2019-03-25.zip), with the right FBInk build, and the cre.cpp tweak applied so the full blit hacks from https://github.com/koreader/koreader-base/pull/878#issuecomment-476312508 can be used ;).

poire-z:
> switching to full blits has a negligible impact on my H2O (it might even be a tiny bit slower!).

I had that feeling too after I made full blits work with cre this morning... :) I put that on the device getting warmer for being plugged - but same feeling now after cold resume.

> So, I guess we're agreed it's not worth it trying to pile on a few hacks around CRe just to get full blits where they might not actually help all that much? ;).

Yes, I agree. Keep all that simple (or hidden in ffi/blitbuffer :)

NiLuJe:
Good news: didn't break it on the Forma :D.

Also, full blits appear fairly efficient there, as I'm getting an average around 3.5ms (... while rotated it averages between 350 and 400ms. Ouch.).

So, this *should* be ready (I just need to tag an FBInk release and kill the debug), and even my H2O² worries have been somewhat assuaged thanks to @BrainDamage's logs in  https://github.com/koreader/koreader/issues/4834 ;).

poire-z:
Tested your binary: work as fine and feels faster than with my previous full blit cre hacks.

Btw (possible I asked you that before): why is the default orientation 0° not the natural one?
So, fbdepth switches it to 270°. Do you think it's possible the hardware is doing that additional translation/rotation (that one we don't have to do anymore, and HW will do that faster than us in software) that it didn't have to with the natural rotation?

Still no clue about some kobo freezes with 8bpp, that disappear when back to 16bpp?
(Just thought it could be some wrong/excessive x_res_virtual that would have the code poke into some other memory area that is used by something else that did not like what we wrote there.)

NiLuJe:
@poire-z : Because most of those panels are actually landscape, so, their native orientation is either Landscape or Inverted Landscape. Since we default to using them in Portrait, poof, rotation ;).

There's a *faint* possibility this will help with those weird freezes, since paintRect could indeed try to plot pixels with completely bogus coordinates without any bounds checking being done, but that's a fairly fait hope. It's more likely an actual kernel bug.

NiLuJe:
@poire-z : Okay, someone is indeed handling the pixelformat wrong somewhere as far as CRe is concerned...

I can convince it to spit out stuff in the right order, but it's not especially pretty...

```diff
diff --git a/crengine/include/lvdrawbuf.h b/crengine/include/lvdrawbuf.h
index f331d08e..24e1fd4b 100644
--- a/crengine/include/lvdrawbuf.h
+++ b/crengine/include/lvdrawbuf.h
@@ -195,10 +195,10 @@ public:
     /// draws text string
     /*
     virtual void DrawTextString( int x, int y, LVFont * pfont,
-                       const lChar16 * text, int len, 
+                       const lChar16 * text, int len,
                        lChar16 def_char, lUInt32 * palette, bool addHyphen=false ) = 0;
-    */                       
-                      
+    */
+
 /*
     /// draws formatted text
     virtual void DrawFormattedText( formatted_text_fragment_t * text, int x, int y ) = 0;
@@ -258,8 +258,8 @@ public:
     /// draws text string
     /*
     virtual void DrawTextString( int x, int y, LVFont * pfont,
-                       const lChar16 * text, int len, 
-                       lChar16 def_char, 
+                       const lChar16 * text, int len,
+                       lChar16 def_char,
                        lUInt32 * palette, bool addHyphen=false );
     */
     /// draws formatted text
@@ -269,7 +269,7 @@ public:
     int getDrawnImagesCount() { return _drawnImagesCount; }
     /// Get surface of images drawn on buffer
     int getDrawnImagesSurface() { return _drawnImagesSurface; }
-    
+
     LVBaseDrawBuf() : _dx(0), _dy(0), _rowsize(0), _data(NULL), _hidePartialGlyphs(true),
                         _drawnImagesCount(0), _drawnImagesSurface(0) { }
     virtual ~LVBaseDrawBuf() { }
@@ -389,7 +389,7 @@ public:
 
 inline lUInt32 RevRGB( lUInt32 cl )
 {
-    return ((cl>>16)&255) | ((cl<<16)&0xFF0000) | (cl&0x00FF00);
+    return ((cl>>16)&0x0000FF) | ((cl<<16)&0xFF0000) | (cl&0x00FF00);
 }
 
 inline lUInt32 rgb565to888(lUInt32 cl ) {
diff --git a/crengine/src/lvdrawbuf.cpp b/crengine/src/lvdrawbuf.cpp
index 58a702dc..19a6d0a1 100644
--- a/crengine/src/lvdrawbuf.cpp
+++ b/crengine/src/lvdrawbuf.cpp
@@ -536,12 +536,12 @@ public:
                     if ( xx<clip.left || xx>=clip.right || alpha==0xFF )
                         continue;
                     if ( !alpha )
-                        row[ x ] = cl;
+                        row[ x ] = RevRGB(cl);
                     else {
                        if ((row[x] & 0xFF000000) == 0xFF000000)
-                            row[ x ] = cl; // copy as is if buffer pixel is transparent
+                            row[ x ] = RevRGB(cl); // copy as is if buffer pixel is transparent
                        else
-                               ApplyAlphaRGB( row[x], cl, alpha );
+                               ApplyAlphaRGB( row[x], RevRGB(cl), alpha );
                     }
                 }
             }
@@ -1324,7 +1324,7 @@ void LVColorDrawBuf::Clear( lUInt32 color )
             lUInt32 * line = (lUInt32 *)GetScanLine(y);
             for (int x=0; x<_dx; x++)
             {
-                line[x] = color;
+                line[x] = RevRGB(color);
             }
         }
     }
@@ -1502,9 +1502,9 @@ void LVColorDrawBuf::FillRect( int x0, int y0, int x1, int y1, lUInt32 color )
             for (int x=x0; x<x1; x++)
             {
                 if (alpha)
-                    ApplyAlphaRGB(line[x], color, alpha);
+                    ApplyAlphaRGB(line[x], RevRGB(color), alpha);
                 else
-                    line[x] = color;
+                    line[x] = RevRGB(color);
             }
         }
     }
@@ -1539,8 +1539,8 @@ void LVColorDrawBuf::DrawLine(int x0,int y0,int x1,int y1,lUInt32 color0 ,int le
             lUInt32 * line = (lUInt32 *)GetScanLine(y);
             for (int x=x0; x<x1; x++)
             {
-                if (direction==0 &&x%(length1+length2)<length1)line[x] = color0;
-                if (direction==1 &&y%(length1+length2)<length1)line[x] = color0;
+                if (direction==0 &&x%(length1+length2)<length1)line[x] = RevRGB(color0);
+                if (direction==1 &&y%(length1+length2)<length1)line[x] = RevRGB(color0);
             }
         }
     }
@@ -1579,7 +1579,7 @@ void LVColorDrawBuf::FillRectPattern( int x0, int y0, int x1, int y1, lUInt32 co
             for (int x=x0; x<x1; x++)
             {
                 lUInt8 patternBit = (patternMask << (x&7)) & 0x80;
-                line[x] = patternBit ? color1 : color0;
+                line[x] = patternBit ? RevRGB(color1) : RevRGB(color0);
             }
         }
     }
@@ -1740,6 +1740,8 @@ void LVColorDrawBuf::Draw( int x, int y, const lUInt8 * bitmap, int width, int h
     } else {
 
 
+        lUInt32 bmpcl32 = RevRGB(bmpcl);
+
         lUInt32 * dst;
         lUInt32 * dstline;
 
@@ -1754,11 +1756,11 @@ void LVColorDrawBuf::Draw( int x, int y, const lUInt8 * bitmap, int width, int h
             {
                 lUInt32 opaque = ((*(src++))>>1)&0x7F;
                 if ( opaque>=0x78 )
-                    *dst = bmpcl;
+                    *dst = bmpcl32;
                 else if ( opaque>0 ) {
                     lUInt32 alpha = 0x7F-opaque;
-                    lUInt32 cl1 = ((alpha*((*dst)&0xFF00FF) + opaque*(bmpcl&0xFF00FF))>>7) & 0xFF00FF;
-                    lUInt32 cl2 = ((alpha*((*dst)&0x00FF00) + opaque*(bmpcl&0x00FF00))>>7) & 0x00FF00;
+                    lUInt32 cl1 = ((alpha*((*dst)&0xFF00FF) + opaque*(bmpcl32&0xFF00FF))>>7) & 0xFF00FF;
+                    lUInt32 cl2 = ((alpha*((*dst)&0x00FF00) + opaque*(bmpcl32&0x00FF00))>>7) & 0x00FF00;
                     *dst = cl1 | cl2;
                 }
                 /* next pixel */
@@ -2418,7 +2420,7 @@ void LVColorDrawBuf::DrawTo( LVDrawBuf * buf, int x, int y, int options, lUInt32
                     lUInt32 * dst = ((lUInt32 *)buf->GetScanLine(y + yy)) + x;
                     for (int xx = 0; xx < _dx; xx++) {
                         if (x+xx >= clip.left && x + xx < clip.right) {
-                            *dst = *src;
+                            *dst = RevRGB(*src);
                         }
                         dst++;
                         src++;
@@ -2561,7 +2563,7 @@ void LVColorDrawBuf::DrawOnTop( LVDrawBuf * buf, int x, int y)
                     lUInt32 * dst = ((lUInt32 *)buf->GetScanLine(y + yy)) + x;
                     for (int xx = 0; xx < _dx; xx++) {
                         if (x+xx >= clip.left && x + xx < clip.right) {
-                            if(*src!=0) *dst = *src;
+                            if(*src!=0) *dst = RevRGB(*src);
                         }
                         dst++;
                         src++;
@@ -2732,7 +2734,7 @@ void LVColorDrawBuf::DrawRescaled(LVDrawBuf * src, int x, int y, int dx, int dy,
                                                        dst[x + xx] = rgb888to565(cl);
                                                } else {
                                                        lUInt32 * dst = (lUInt32 *)GetScanLine(y + yy);
-                                                       dst[x + xx] = cl;
+                                                       dst[x + xx] = RevRGB(cl);
                                                }
                                        }
                                }
@@ -2751,7 +2753,7 @@ void LVColorDrawBuf::DrawRescaled(LVDrawBuf * src, int x, int y, int dx, int dy,
                                                        dst[x + xx] = rgb888to565(cl);
                                                } else {
                                                        lUInt32 * dst = (lUInt32 *)GetScanLine(y + yy);
-                                                       dst[x + xx] = cl;
+                                                       dst[x + xx] = RevRGB(cl);
                                                }
                                        }
                                }
```

I'm also not entirely convinced it's the right approach...

NiLuJe:
Yeah, I have a slightly updated version in my [CRe tree](https://github.com/NiLuJe/crengine), so, if that doesn't look completely insane to you, @poire-z, I can PR that.

My other train of though was creating a BGRA32 BB type, and use that where appropriate (which would be pretty much everything that's currently using RGB32, as it is, in fact, wrong ;)).
MuPDF stuff can probably be handled sanely w/ the colorspace setting...
If it works, we can get rid of the bgr kludge we currently have, too.

poire-z:
Looks allright (small clear changes) to me.

> lvimg.cpp is definitely building BGR pixels every time, very specifically _on purpose_

_on purpose_: why is that?
So `RevRGB()` always get BGR as input and actually makes correct RGB ?

> I fear the crappy "leave that alone, and reverse 'em again in lvdrawbuf" approach will be the easiest and least intrusive one

I wondered about why that, while reading your changes :) But OK, I agree with simple :)

(Btw, when making changes to crengine, I tried when possible to keep the original behaviour (when not wrong and not a bug) as the default, and have our changes applied with optional parameter or stuff - just in case upstream might pick our changes (we might very well be past that possibility :). So, if they were expecting BGR in their frontend code, that would break them. So, if it is possible without too much overhead, could be nice to have that toggable, with a `#define` or a global variable, or an additional parameter to `LVColorDrawBuf drawBuf()` creation.)

> My other train of though was creating a BGRA32 BB type

Drop that thought :)

NiLuJe:
@poire-z: I can ifdef RevRGB directly, as it was only used once in an obscure Win32 codepath... ;).

As for BGR, beats me, but:
JPG: https://github.com/koreader/koreader/issues/1375#issuecomment-423413385 (despite there being a perfectly fine JCS_EXT_BGR colorspace to ask libjpeg-turbo to do it).
PNG: https://github.com/koreader/crengine/blob/6eac0d8f8705338770c35215041b3ec6d645fa03/crengine/src/lvimg.cpp#L929
GIF: https://github.com/koreader/crengine/blob/6eac0d8f8705338770c35215041b3ec6d645fa03/crengine/src/lvimg.cpp#L1102
SVG: https://github.com/koreader/crengine/blob/6eac0d8f8705338770c35215041b3ec6d645fa03/crengine/src/lvimg.cpp#L1773

Hell, even the CSS color constants are in BGR: https://github.com/koreader/crengine/blob/6eac0d8f8705338770c35215041b3ec6d645fa03/crengine/src/lvstsheet.cpp#L486

Because we're agreed that in RGB, while blue's color code is indeed `#0000FF`, you have to remember that if you want to store that in an uint32_t manually, you have to take endianess into account, so that becomes `0xFF0000`. Otherwise, `0x0000FF`  means red in RGB order... but blue in BGR order. >_<".

In lvdrawbuf, everything assumes BGR input, RevRGB, the RGB565 conversions, the Grayscale conversions, ..., everything.

So, while I'd ideally like to avoid doing an essentially useless double conversion, it'd be somewhat more intrusive to fix, especially if you want to ifdef it ;).

NiLuJe:
I'm using unions in FBInk because this is very easy to fuck up.

```C #define _GNU_SOURCE

#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>

typedef union
{
	uint32_t p;
	struct
	{
		uint8_t r;
		uint8_t g;
		uint8_t b;
		uint8_t a;
	} color;
} FBInkPixelRGBA;

typedef union
{
	uint32_t p;
	struct
	{
		uint8_t b;
		uint8_t g;
		uint8_t r;
		uint8_t a;
	} color;
} FBInkPixelBGRA;


int main(void) {
	FBInkPixelRGBA rgba;
	rgba.color.r = 0xFF;
	rgba.color.g = 0xCC;
	rgba.color.b = 0x00;
	rgba.color.a = 0x88;
	printf("RGBA: #%02X%02X%02X%02X -> 0x%X (R: %X vs. [0]: %X / G: %X vs. [1]: %X / B: %X vs. [2]: %X / A: %X vs. [3]: %X)\n", rgba.color.r, rgba.color.g, rgba.color.b, rgba.color.a, rgba.p, rgba.color.r, ((uint8_t*)&rgba.p)[0], rgba.color.g, ((uint8_t*)&rgba.p)[1], rgba.color.b, ((uint8_t*)&rgba.p)[2], rgba.color.a, ((uint8_t*)&rgba.p)[3]);

	FBInkPixelBGRA bgra;
	bgra.color.r = 0xFF;
	bgra.color.g = 0xCC;
	bgra.color.b = 0x00;
	bgra.color.a = 0x88;
	printf("BGRA: #%02X%02X%02X%02X -> 0x%X (B: %X vs. [0]: %X / G: %X vs. [1]: %X / R: %X vs. [2]: %X / A: %X vs. [3]: %X)\n", bgra.color.r, bgra.color.g, bgra.color.b, bgra.color.a, bgra.p, bgra.color.b, ((uint8_t*)&bgra.p)[0], bgra.color.g, ((uint8_t*)&bgra.p)[1], bgra.color.r, ((uint8_t*)&bgra.p)[2], bgra.color.a, ((uint8_t*)&bgra.p)[3]);

	return EXIT_SUCCESS;
}
```

```
RGBA: #FFCC0088 -> 0x8800CCFF (R: FF vs. [0]: FF / G: CC vs. [1]: CC / B: 0 vs. [2]: 0 / A: 88 vs. [3]: 88)
BGRA: #FFCC0088 -> 0x88FFCC00 (B: 0 vs. [0]: 0 / G: CC vs. [1]: CC / R: FF vs. [2]: FF / A: 88 vs. [3]: 88)
```

poire-z:
> So, while I'd ideally like to avoid doing an essentially useless double conversion, it'd be somewhat more intrusive to fix, especially if you want to ifdef it ;).
>  I can ifdef RevRGB directly

So let's go with simple :)
(If you can add a bit of the above as comments in the code, like _In lvdrawbuf, everything assumes BGR input_ - so one doesn't have to wonder anymore when reading the code :)

NiLuJe:
I added some stuff around the ifdef in the include, let me know what you
think ;).

poire-z:
I think it's fine :) Thanks.

PR #882, opened 2019-03-28
Fix a couple bugs from #878

NiLuJe:
* Fix Android, because the framebuffer can also have a physical stride larger than the pixel width ;).
* Fix RGB32-to-RGB32 fast blitting w/ the Lua blitter, which I broke late in #878 because of a stupid typo.
* Allow disabling the C blitter, to make testing easier.
* Bump FBInk to v1.13.0 (unbreak legacy Kindle handling)
* Possibly unbreak some PB corner-cases (koreader/koreader#4851)

### DIFF
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -689,7 +689,6 @@ function BB_mt.__index:setPixel(x, y, color)
     local px, py = self:getPhysicalCoordinates(x, y)
     -- NOTE: The cbb exemption is because it's assuming you're using an inverted bb copy to handle nightmode (c.f., Android & SDL2),
     --       and setPixel can be used from outside blitFrom, unlike the other setters.
-    -- NOTE: Speaking of the other setters, some of the alpha = 0xFF fastpaths *might* be skipping the invert?
     if not use_cblitbuffer and self:getInverse() == 1 then color = color:invert() end
     self:getPixelP(px, py)[0]:set(color)
 end
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -140,8 +140,9 @@ void free(void *ptr);
 ]]
 
 -- NOTE: Try the C blitter, unless it was disabled by the user
+local no_cbb_flag = os.getenv("KO_NO_CBB")
 local use_cblitbuffer, cblitbuffer
-if G_reader_settings:nilOrFalse("dev_no_c_blitter") then
+if not no_cbb_flag or no_cbb_flag == "false" then
     use_cblitbuffer, cblitbuffer = pcall(ffi.load, 'blitbuffer')
 end
 -- NOTE: This works-around a number of corner-cases which may end up with LuaJIT's optimizer blacklisting this very codepath,
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -140,8 +140,9 @@ void free(void *ptr);
 ]]
 
 -- NOTE: Try the C blitter, unless it was disabled by the user
+local use_cblitbuffer, cblitbuffer
 if G_reader_settings:nilOrFalse("dev_no_c_blitter") then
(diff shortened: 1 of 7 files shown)


ffi/blitbuffer.lua
```diff
@@ -139,7 +139,10 @@ void *malloc(int size);
 void free(void *ptr);
 ]]
 
-local use_cblitbuffer, cblitbuffer = pcall(ffi.load, 'blitbuffer')
+-- NOTE: Try the C blitter, unless it was disabled by the user
+if G_reader_settings:nilOrFalse("dev_no_c_blitter") then
+    local use_cblitbuffer, cblitbuffer = pcall(ffi.load, 'blitbuffer')
+end
```
poire-z: I really don't mind it. But there seems to be no access to G_reader_settings in base yet.
Not sure if that's a rule, or just luck :) @Frenzie ?

Frenzie: That is indeed intentional, also see https://github.com/koreader/koreader-base/commit/7fb42187478cc4c05756ab744a02b3369f140f4a

NiLuJe: So, what to do? Not going to lie, not a massive fan of the seven layers of indirection for something like that... ;).

poire-z: I think the real problem is, if our base/ffi/ modules are to be used for other things, `G_reader_settings` possibly not being set, and accessing `G_reader_settings:nilOrFalse` cause a "can't index a nil value" crash.
So, set an env variable from frontend (dunno if that's possible), or koreader.sh (i know it's ugly, but you grep for dev_startup_no_fbdepth), and use `os.getenv()` (commonly used in base) ? Or just set a plain boolean global variable with no risk of crash?

NiLuJe: Which, incidentally, is exactly what happens in the testsuite ;).

And, no, there's no setenv in Lua because it's POSIX, not ANSI C89 :/.

NiLuJe: Ohkay, after cursing at Linux's putenv that doesn't take a const, switched to setenv over FFI, as some stuff was already using it on Android & Windows ;).

Frenzie: Yeah, functionally speaking if g and g:bla() would be fine. Architecturally speaking you probably wouldn't want that to be the only way to play with it.

NiLuJe: env var it is, then ;). Thanks!

poire-z:
Don't the prevous CI runs get cancelled when a new commit is pushed, and a new CI run queued for the same PR?
I think I saw that happening previously, may be in koreader or crengine (CircleCI only?)
Anyway, one can manually cancel them on the Travis page, if you want the new one to not wait for the previous no-more-interesting ones to free the queue.

poire-z:
Confirming that commit `Oops. Fix RGB32-to-RGB32 blitting in the Lua codepath` solves the issue I mentionned at https://github.com/koreader/koreader/pull/4847#issuecomment-477641175:

> Wanted to check that on the emulator with blitbuffer.lua, without libblitbuffer.so. But if I move it out, the emulator is stuck on displaying only the widgets: the rendering of the book below stays blank, but the footer page number changes!

NiLuJe:
@poire-z : CircleCI indeed handles it properly every time. I *think* Travis is supposed to do it, too, but, err, sometimes it doesn't ;p.

Yeah, I don't really know how the RGB32 stuff slipped in, because it was amazingly (and stupidly broken), and just plain segfaulted on *start* when I double-checked it. And yet I remember testing it after that change... :D.

Frenzie:
I think CircleCI cancels them and Travis only *skips* (even though it seems like it should cancel). Unless something accidentally changed about the settings while switching from ye olde integrations to webhooks.

So push a commit, it starts building. Push another, it's queued. Push a third one, then the second one will be skipped (canceled).

PR #887, opened 2019-04-13
simplify luajit_lib build logic

houqp:


houqp:
This should be fine. Because OS X already builds LUAJIT_LIB. Given `LUAJIT_LIB` is ahead of `libs` in `all` target, it should not have any impact. Even if `LUAJIT_LIB` is after `libs`, it would only change the build order, which should be harmless :)

Makefile.third
```diff
@@ -169,17 +169,10 @@ $(LUAJIT) $(LUAJIT_LIB) $(LUAJIT_JIT): $(THIRDPARTY_DIR)/luajit/CMakeLists.txt
...
-ifdef WIN32
-	cp -fL $(LUAJIT_DIR)/src/$(notdir $(LUAJIT_LIB)) $(LUAJIT_LIB)
-endif
-ifdef DARWIN
+ifdef USE_LUAJIT_LIB
 	cp -fL $(LUAJIT_DIR)/src/$(notdir $(LUAJIT_LIB)) $(LUAJIT_LIB)
 endif
-ifdef ANDROID
-	cp -fL $(LUAJIT_DIR)/src/$(notdir $(LUAJIT_LIB)) $(LUAJIT_LIB)
-else
```
Frenzie: What about this else?

houqp: this one already includes `ifdef DARWIN` right?

Frenzie: But it said

```sh
if ANDROID
cp file x
else
cp file y
```

Now it's saying

```sh
if USE_LUAJIT_LIB
cp file x
endif
#ALWAYS
copy file y
```

Unless I'm misreading it, the functional equivalent should be something like `ifndef ANDROID`.

houqp: Oh, sorry, good catch. You are right, I was only lookig at `DARWIN` :P

Will think about how to update to test to catch this going forward.

.circleci/config.yml
```diff
@@ -0,0 +1,181 @@
+version: 2
+
+workflows:
```
Frenzie: This is so nice looking, but one reason I didn't really touch Travis here (all the while keeping an eye on GitHub CI), is that I'm a bit afraid of CircleCI's free limits. I believe there's a maximum of 4 jobs for our organization, plus even just for front I've at least once received a warning that we were more or less at some monthly computation time limit.

Makefile.defs
```diff
@@ -167,6 +167,8 @@ endif
 
 USE_LUAJIT_LIB=$(or $(DARWIN),$(ANDROID),$(WIN32))
 
+SKIP_LUAJIT_BIN=$(or $(ANDROID),)
```
Frenzie: Nice and clear. :+1:

.circleci/config.yml
houqp: yes, max concurrency for circleci is 4. do you know what's the computation time limit for travis ci?

Frenzie: I have no idea, but I've never seen any warnings and it's likely always used up more than CircleCI. :-)

Frenzie: Do you have any idea why CircleCI is a few minutes faster btw?

.circleci/config.yml
```diff
@@ -0,0 +1,181 @@
...
+      - checkout
+      - run:
+          name: install
+          command: |
+            source .ci/before_install.sh
+            source .ci/emulator_install.sh
+
+      - restore_cache:
+          keys:
+            - x86-{{ .Environment.CIRCLE_JOB }}-
```
Frenzie:
```suggestion
            - {{ arch }}-{{ .Environment.CIRCLE_JOB }}-
```
See https://github.com/koreader/koreader/pull/3329

houqp: i actually didn't intend to put the actual arch here, just a prefix to distinguish between `xcompile-` jobs. since we are only using linux runners here, `arch` should be the same for all jobs right?

i changed the prefix to `emu` so it's less confusing :)

.circleci/config.yml
houqp: OK, we can keep using travis since my android test change should also work for travis ci build.

Given that travis ci has recently been acquired and the new parent company is firing engineers left and right, I think eventually we will have to move off it to something else. I am going to keep the circleci config and just disable circleci build from their UI for now.

I think CircleCI build is faster because they have beefier machine?

houqp: oh wow, travis ci is very slow. the full circle ci build is almost done. travis ci is still in queuing state and hasn't started even a single job.

NiLuJe: Ah, so, not just on my PR ;p. It was starting fine à few hours ago, FWIW, but, err, I've kinda been hammering jobs because of those cmake shenanigans that I can't reproduce locally...

Frenzie: @houqp Usually Travis takes a little longer to get going but I mean at most tens of seconds.

CircleCI does occasionally also end up with long queues when it's busy.

> I think eventually we will have to move off it to something else.

I'm keeping an eye on GitHub Actions.

https://medium.com/selleo/experimenting-with-github-actions-beta-760e61fab0e1

@NiLuJe CMake 3.6 seemed like a sensible version to support, but we can always go for newer if necessary.

Frenzie:
@houqp We could merge, my memory deceived me!

https://circleci.com/open-source/

> We support the open source community. Every open source repository gets 4 free containers. No usage limits.

At least if I'm interpreting this correctly it means base shouldn't interfere with front, and no usage limits is self-explanatory.

But I'd swear that when I looked into it a couple of years ago I found that there's be organization-wide limits. :-)

Frenzie:
Most suspicious indeed:

Screenshot_2019-04-17_14-40-41

Frenzie:
Yup, so "every open source repository" is a big fat lie:

Screenshot_2019-04-17_14-44-50

Frenzie:
Maybe something's not set up correctly? It does say this:

> Build on Linux
>
> CircleCI offers 4 Linux containers with no build limits ($2400 annual value) to every open source project in an organization. Simply keeping your repo public will enable this for you. 

Screenshot_2019-04-17_14-47-56

They also advertise 500 Mac OS build minutes for open source. Assuming a Mac OS build takes 20 minutes and one build per day, that's tight. Is there a way to only do a Mac OS build every once every few days or some such? It would reduce the value of CI, but at least it wouldn't take months to find out you accidentally broke something…

Screenshot_2019-04-17_14-49-08

Frenzie:
Regarding Mac on Travis, I tried that once and it went nowhere: https://github.com/koreader/koreader-base/pull/564#issuecomment-339693544

NiLuJe:
Ubuntu whatever is still stuck on 3.5.1 unfortunately ;/.

Frenzie:
@NiLuJe Yes, but that hardly matters. It slightly lowers the barrier to entry, that's all.

NiLuJe:
But the docker images don't have anything newer, AFAICT.

Frenzie:
Well yes… but *I* make the Docker images. ;-)

Frenzie:
Basically, something like this in the Dockerfile or alternatively some PPA:

```
wget -qO- "https://cmake.org/files/v3.14/cmake-3.14.1-Linux-x86_64.tar.gz" | tar --strip-components=1 -xz -C /usr/local
```

Although I figure we should still strive to support either 3.7 (Debian Stretch) or 3.10 (Ubuntu 18.04) if reasonably possible, so potential contributors don't need to do that.

NiLuJe:
3.10 behaves the same way, apparently, so, yeah, let's forget about it, I
have a few more things to try anyway...

Frenzie:
@NiLuJe Alternatively, at least CMake 3.11 for FetchContent. What was the problem you were facing with "old" CMake again?

NiLuJe:
We'll see if it behaves with the latets tweaks, but it seems to handle the
resolution of to-be cached values from a toolchain file differently.

(i.e., CMAKE_AR gets lost when expanded in a generator expression, while it
isn't on my own setup).

houqp:
> I'd swear that when I looked into it a couple of years ago, I found that there were organization-wide limits. :-) 

That's my impression as well. We ran into the same issue with Linkedin's github org, projects were contenting for concurrency.

CircleCI supports scheduled builds, so we can use that feature to scheduled mac builds in a fixed cadence.

Another thing we can do is to reduce concurrency contention is to enable some of the redundant builds only for master merge through branch filter. For example, it doesn't make sense to build so many variants of emulator build/test for each PR commit. We can just build ninja-gcc-debug for PRs and restrict the full builds only to master.

Frenzie:
@houqp CircleCI seems to handle it less problematically than Travis used to in any case, the way you've set it up right now.

> CircleCI supports scheduled builds, so we can use that feature to scheduled mac builds in a fixed cadence.

Interesting. I guess someone should contact them about it. :-)

> We can just build ninja-gcc-debug for PRs and restrict the full builds only to master.

With at least one caveat. It probably doesn't matter for debug vs optimized, nor has there been any indication that it matters for Make vs Ninja. But I normally depend on the CI to tell me if I just accidentally broke the Clang build.

PR #892, opened 2019-04-19
Fix MuPDF vendoring

NiLuJe:
We were failing to properly instruct it to use our own libjpeg-turbo, meaning it was building & linking against its own bundled copy (of jpeg 9c, so, not the same API).
How that never blew up before is beyond me ;).
Now that we build harfbuzz, make it use our own harfbuzz, too :).

Also, a few assorted fixes to make LTO builds possible.


NiLuJe:
The idea was testing things a bit first, to see how it performed ;).

Also, I can only vouch for it with GCC 6+, but ideally, 7+ (i.e.,
koxtoolchain platforms ;)).

NiLuJe:
(also, it takes roughly twice as long to build ;)).

houqp:
Do we get a noticeable performance boost with LTO?

Frenzie:
Does Clang have LTO as well?

thirdparty/lua-serialize/CMakeLists.txt
```diff
@@ -12,7 +12,9 @@ assert_var_defined(OUTPUT_DIR)
 
 ep_get_source_dir(SOURCE_DIR)
 
-set(BUILD_CMD sh -c "${KO_MAKE_RECURSIVE} -j${PARALLEL_JOBS} CC=\"${CC}\" LDFLAGS=\"${LDFLAGS}\" OUTPUT_DIR=${OUTPUT_DIR}")
+set(PATCH_CMD sh -c "patch -N -p1 < ${CMAKE_CURRENT_SOURCE_DIR}/lua-serialize-lto.patch || true")
```
Frenzie: Btw, not directly related to this PR but I've been wondering why let the patches fail semi-quietly? That is, you won't notice without studying the logs.

NiLuJe: Wasn't that to handle rebuilds? I have no idea if that still makes sense with the CMake system, though :D.

Frenzie: Afaik CMake handles when to patch or not to patch with an uncharacteristic lack of stupidity.

NiLuJe: Then yeah, it doesn't make sense anymore, and that's just a bad habit carried over from the old makefile ;)

Frenzie: In that case I'll submit a PR to get rid of the load of 'em. It's tripped me up more than once.

Frenzie: Most recently https://github.com/koreader/koreader-base/pull/860

Frenzie: Oh, you can see I already got rid of the `|| true` there, so you'd have noticed if that were somehow trying to apply itself multiple times. ;-)

PR #894, opened 2019-04-19
[fix, Android] Versioned OpenSSL libs

Frenzie:
Unversioned is a proven recipe for disaster. #thanksandroid

But perhaps more to the point, this isn't just a personal fancy. It crashes on load.


NiLuJe:
OpenSSL's build system automatically generates those on Android targets, so it may take some extra shenanigans to handle the rename earlier in the CMake file?

Frenzie:
There's no rename involved. I'm not really sure why you used the unversioned one?

Frenzie:
Screenshot_2019-04-19_15-47-32

NiLuJe:
I vaguely thought there was only one and the [copy](https://github.com/koreader/koreader-base/blob/master/Makefile.third#L449) would then fail ;).

libjpeg-turbo did this, so I went the proactive way *during* a build without checking ;).

Frenzie:
Yes, OpenSSL is either purposefully going against Android's beyond idiotic policy as it very well should, or it's sufficiently up-to-date that it's ready for Android 8+. The latter decided to reverse this particular tidbit of broken by design idiocy.

It doesn't exactly help that the Android linker is total piece of trash that functions and reads like something I could write. When that is the case in something this low-level it's either incredibly elegant or rather shitty.

PR #895, opened 2019-04-19
[fix, build] Don't fail quietly on patching failure

Frenzie:
It's a vestigial property from before we switched to CMake.

See <https://github.com/koreader/koreader-base/pull/892#discussion_r276926952> for discussion.


Frenzie:
@NiLuJe Hm, apparently I was wrongish. I'll look into it a little deeper in any case, because this `|| true` is freaking terrible.

```
-- Build files have been written to: /home/frans/src/kobo/koreader/base/thirdparty/luajit/build/x86_64-linux-gnu-debug
[1/5] Performing patch step for 'luajit'
FAILED: luajit-prefix/src/luajit-stamp/luajit-patch 
cd /home/frans/src/kobo/koreader/base/thirdparty/luajit/build/x86_64-linux-gnu-debug/luajit-prefix/src/luajit && sh -c "patch -N -p1 < /home/frans/src/kobo/koreader/base/thirdparty/luajit/koreader-luajit-makefile-tweaks.patch" && /usr/bin/cmake -E touch /home/frans/src/kobo/koreader/base/thirdparty/luajit/build/x86_64-linux-gnu-debug/luajit-prefix/src/luajit-stamp/luajit-patch
patching file src/Makefile
Reversed (or previously applied) patch detected!  Skipping patch.
4 out of 4 hunks ignored -- saving rejects to file src/Makefile.rej
```

NiLuJe:
Hide it away in a utility function in our cake bag of tricks?

Would have to take a list for projects with multiple patches and we'd just have to feed the filename?

(Similar to how Gentoo ebuilds use a PATCHES bash array).

Frenzie:
It looks like it might be really easy to handle. The error code for previously applied is 1, and for actual errors it's greater than 1.

NiLuJe:
Oh, wait, that wouldn't solve the "silent failure" issue, though -_-".

Frenzie:
I think there's an easier solution: `--merge`.

NiLuJe:
Not supported on macOS ;).

```
patch: unrecognized option `--merge'
```

Frenzie:
@NiLuJe Could you check the error codes?

NiLuJe:
In which instances?

(Right now, it return 2 for unsupported option ;)).

Frenzie:
Like I said above, `1` is previously applied, greater than `1` is other (actual) errors.

NiLuJe:
Yep, that holds true on macOS, too.

Frenzie:
Alright, in that case I'll write a simple wrapper script.

Frenzie:
@NiLuJe Something like this work on Mac OS?

```sh
#!/bin/sh

# get input from stdin
read -r patchfile

patch -p1 -N --dry-run --silent <"${patchfile}" 2>/dev/null

exit_status=$?

if [ $exit_status -eq 0 ]; then
    # No issues, apply
    patch -p1 -N <"${patchfile}"
elif [ $exit_status -eq 1 ]; then
    # Previously applied, ignore
    exit 0
fi

exit 1
```

NiLuJe:
That *looks* sane, but I'll double-check when I'm back on the laptop later ;)

Frenzie:
I'll update the PR to include that. The basic strategy seems right even if some minor details in the script itself could need tweaking for Mac OS. This way it seems more elegant to me, but the same thing can always be done without the dry-run part.

NiLuJe:
CMake being helpful, as always :D.

Frenzie:
https://github.com/Kitware/CMake/blob/0793464d95c3dbbe9ed87bb5882d29efb7a8e3c3/Modules/ExternalProject.cmake#L352-L358
````
    ``PATCH_COMMAND <cmd>...``
      Specifies a custom command to patch the sources after an update. By
      default, no patch command is defined. Note that it can be quite difficult
      to define an appropriate patch command that performs robustly, especially
      for download methods such as git where changing the ``GIT_TAG`` will not
      discard changes from a previous patch, but the patch command will be
called again after updating to the new tag.
````

**Comment by Frenzie** [comment]
@NiLuJe Not 100 % sure, but I think the problem may be localized to `BUILD_IN_SOURCE 1`. (luajit, mupdf, openssl)

**Comment by Frenzie** [comment]
@NiLuJe In other news, introducing a failure (on purpose here for testing) still returns an exit status of 0 for the OpenSSL configure script? O_o

```
-- Build files have been written to: /home/frans/src/kobo/koreader-base/thirdparty/openssl/build/x86_64-linux-gnu
[1/4] Performing configure step for 'openssl'
Operating system: x86_64-whatever-linux2

Failure!  build file wasn't produced.
Please read INSTALL and associated NOTES files.  You may also have to look over
your available compiler tool chain or change your configuration.

***** Unsupported options: no-r5
0
[2/4] Performing build step for 'openssl'
make[1]: Entering directory '/home/frans/src/kobo/koreader-base/thirdparty/openssl/build/x86_64-linux-gnu/openssl-prefix/src/openssl'
make[1]: Leaving directory '/home/frans/src/kobo/koreader-base/thirdparty/openssl/build/x86_64-linux-gnu/openssl-prefix/src/openssl'
make[1]: Entering directory '/home/frans/src/kobo/koreader-base/thirdparty/openssl/build/x86_64-linux-gnu/openssl-prefix/src/openssl'
make _build_libs
make[2]: Entering directory '/home/frans/src/kobo/koreader-base/thirdparty/openssl/build/x86_64-linux-gnu/openssl-prefix/src/openssl'
make[2]: Nothing to be done for '_build_libs'.
make[2]: Leaving directory '/home/frans/src/kobo/koreader-base/thirdparty/openssl/build/x86_64-linux-gnu/openssl-prefix/src/openssl'
make[1]: Leaving directory '/home/frans/src/kobo/koreader-base/thirdparty/openssl/build/x86_64-linux-gnu/openssl-prefix/src/openssl'
[4/4] Completed 'openssl'
```

**Comment by Frenzie** [comment]
Darn, you also get an exit code of 1 for hunks failed to apply. @NiLuJe Does Mac OS have the `-b` flag?

**Comment by Frenzie** [comment]
@NiLuJe I think this updated script is pretty fool-proof.

```sh
#!/bin/sh

if [ $# -eq 0 ]; then
    echo "Patch file required as argument."
    exit 1
fi

PATCH_FILE=$1

# Reverse patch will succeed if the patch is already applied.
# In case of failure, it means we should try to apply the patch.
if ! patch -R -p1 -N --dry-run <"${PATCH_FILE}" >/dev/null 2>&1; then
    # Now patch for real.
    patch -p1 -N <"${PATCH_FILE}"
    exit $?
fi
```

So if the reverse patch *doesn't* succeed, try to patch, and return the exit status. If the reverse patch succeeds, it means the patch was already successfully applied previously.

NiLuJe:
Yeah, the reverse test should do the job.

I'd check what happens with patches that add completely new files though,
because I have to a vague memory of trying something like that somewhere
and this specific case being weird...

houqp:
nice work on the patch wrapper :+1:

PR #897, opened 2019-04-20
[fix] Detect Clang

Frenzie:
This will result in horrible failures, but it'll hopefully be easier to get an overview through the CI.


Frenzie:
Observations while the tests are running: Travis now uses Clang again while CircleCI doesn't. (Good thing we kept Travis around for the moment.)

NiLuJe:
I should have just ditched that naive check and used the real Clang check that I added a few lines below that ;).

NiLuJe:
In slightly related concerns (verbosity), I just checked a local build, without VERBOSE in my env, yields a 15232 lines long log (2761890 bytes, so, a bit over half the allowed size).

So I'm not quite sure what Android is doing to have logs that are at least twice as large :?

NiLuJe:
Oh. the Android-only stuff (libiconv, gettext), is *insanely* verbose.

NiLuJe:
So, yeah, libiconv & gettext *probably* deserve their `--silent` in the BUILD & INSTALL commands ;).

Frenzie:
I say "horrible failures", the CI says "WHEEEEEE".

> I should have just ditched that naive check and used the real Clang check that I added a few lines below that ;).

Ah, so I should just move that up a few lines? Sometime after dinner/tomorrow, that is. ;-P

NiLuJe:
There's a bit of a catch-22 in that the test definitely should be using the $(CHOST) prefixed CC when that makes sense, and that block is the one that sets CC to CHOST-CC,, but, yeah, at least it should pretty solidly devices it it's actually Clang instead of something masquerading as ca clang binary ;).

Frenzie:
@NiLuJe I think this should do the trick then?

Frenzie:
Hm, it's a little unclear to me what the purpose is here:

```
##############################################################################
#############################  COMPILER OPTIONS  #############################
##############################################################################
# These options mainly affect the speed of the JIT compiler itself, not the
# speed of the JIT-compiled code. Turn any of the optional settings on by
# removing the '#' in front of them. Make sure you force a full recompile
# with "make clean", followed by "make" if you change any options.
#
DEFAULT_CC = gcc
#
# LuaJIT builds as a native 32 or 64 bit binary by default.
CC= $(DEFAULT_CC)
#
# Use this if you want to force a 32 bit build on a 64 bit multilib OS.
#CC= $(DEFAULT_CC) -m32
```

Of course on Mac OS gcc is simply a symlink to Clang. I could just add an `update-alternatives --install /usr/bin/gcc gcc /usr/bin/clang-7 380`.

NiLuJe:
> @NiLuJe I think this should do the trick then?

Nope, see my "Catch-22" comment. That'd make it check CC instead of CHOST-CC (which is what we do want to check) in cases where CHOST is set (which is nearly everywhere except emulator builds). ;).

NiLuJe:
IIRC, we should be enforcing pretty much everything when building LuaJIT, I don't *think* this can be an issue, as we *do* enforce CC (& HOST_CC & TARGET_CC) when building it ;).

Frenzie:
@NiLuJe Well, I'm not really following the Catch 22. It seems to be setting it based on CC, so then what's there to check afterward?

NiLuJe:
Except this high up, CC is stil CC, not yet CHOST-CC. We do want to check CGOST-CC? not CC ;).

Frenzie:
Right, thanks.

NiLuJe:
I don't really see a cleaner solution than either setting an initial

```
ifdef CHOST
    CC:=$(CHOST)-$(CC)
endif
```

right above it, and then let the "real" one set that to CHOST-gcc or CHOST-clang ;).

Frenzie:
What about keeping things more or less as they were, but using `$(findstring)` instead of a stupid "misses clang-x" `ifeq ($(CC), clang)`.

NiLuJe:
Yeah, that might end up being less troublesome ;).

NiLuJe:
I don't remember where I wondered about that, but I don't think we actually need the strip CCACHE shenanigans *anywhere*, as we no longer enforce CC:=CCACHE CC anywhere ;).

NiLuJe:
Oh, wait, it's worse. I'm *really* stupid. I was reading those as "strip CCACHE from CC", while they're in fact doing "strip whitespace from CCACHE CC".

i.e., I'm doing the very thing I just said we shouldn't do: prepend CCACHE to CC.

*facepalm*

Frenzie:
lol, I'll remove them all then

NiLuJe:
Thanks ;).

Hopefully CMake will make everything pick it up, otherwise we might actually need to keep one, when we set CC for the final time ;).

Frenzie:
Now we have a failing Clang test in Circle CI, but at least it's using Clang. The Travis Clang test isn't failing, so I'll go ahead and merge this for the CCACHE CC stuff. (Hopefully that'll fix Android.)

PR #893, opened 2019-04-19
reduce circleci PR build

houqp:

PR #898, opened 2019-04-21
Re-enforce silent builds for iconv & gettext

NiLuJe:
We only build them on Android, but they are *extremely* verbose,
which the CIs don't appreciate ;).


Frenzie:
Linking my comment about making it depend on the environment: https://github.com/koreader/koreader-base/pull/893#issuecomment-484797148

(Not too important here.)

I *think* GitLab doesn't really care about the log size, but that it just annoyingly only shows you the first lines instead of the last lines. Which is not a great conclusion if true. ;-) It's only Travis that actively punishes log size.

NiLuJe:
Yeah, in this instance I don't have much qualms about enforcing it to silent, since it's all pretty boring stuff anyway ;).

Frenzie:
(Doesn't need testing.)

PR #909, opened 2019-04-24
Pass arbitrary DjVu text annotations to frontend

xelxebar:
## Overview

Partially addresses koreader/koreader#2943.

This just transparently translates the annotation tree into a Lua table. The frontend needs to be prepared to handle this more flexible structure. Fortunately, almost all DjVu files in practice adhere to the rigid `page -> line -> word` hierarchy, so this patch should transparently work without any noticeable difference in practice.

## Comments

The final product here is my original implementation, which departs from the prevailing style of `djvu.c`, so I made an attempt at orthogonalizing style choices into the second commit.

Currently, `djvu.c` has a lot of magic numbers and repeated code, so my goal was to give these understandable names and make them reusable. Arguably, this should be it's own pull request, and be  carried out file-wide, but I figured that the comparison here might help review, so am including it.

### DIFF
--- a/djvu.c
+++ b/djvu.c
@@ -394,26 +394,55 @@ static int getPageInfo(lua_State *L) {
 	return 5;
 }
 
-/*
- * Return a table like following:
- * {
- *    -- a line entry
- *    1 = {
- *       1 = {word="This", x0=377, y0=4857, x1=2427, y1=5089},
- *       2 = {word="is", x0=377, y0=4857, x1=2427, y1=5089},
- *       3 = {word="Word", x0=377, y0=4857, x1=2427, y1=5089},
- *       4 = {word="List", x0=377, y0=4857, x1=2427, y1=5089},
- *       x0 = 377, y0 = 4857, x1 = 2427, y1 = 5089,
- *    },
+/** Fill Lua table with DjVu text annotations
  *
- *    -- an other line entry
- *    2 = {
- *       1 = {word="This", x0=377, y0=4857, x1=2427, y1=5089},
- *       2 = {word="is", x0=377, y0=4857, x1=2427, y1=5089},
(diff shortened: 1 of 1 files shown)


Frenzie:
Btw, in principle the coding style of the project is with 4 spaces instead of tabs, but I suppose we should just keep this file in tabs.

But part of me wonders if we should migrate "organically". ;-)

xelxebar:
@Frenzie 
Roger on the spaces vs. tabs. I have half a mind to do some housecleaning on `djvu.c` anyway, and if I do that might be the time to switch. Or did you mean migrate everything else the other way? lol

Thanks for the props! I put a Doxygen-style comment header on `lua_settable_djvu_anno`, but is there any project convention on this too? I haven't noticed any from my little bit of poking around.

xelxebar:
Strange that `spec/unit/md5_spec.lua` would be failing on this.

Frenzie:
> Thanks for the props! I put a Doxygen-style comment header on lua_settable_djvu_anno, but is there any project convention on this too? I haven't noticed any from my little bit of poking around.

The project is mainly Lua. We don't really have any formal C conventions. However, if it doesn't matter for Doxygen then it might be best to stick closer to the luadoc style for project-wide consistency, in this case with an `@` instead of a `\`: https://github.com/koreader/koreader/blob/c0ba6ae48ea3ee9b693448b43e2c5f30db5991b7/frontend/util.lua#L70-L77

> Strange that spec/unit/md5_spec.lua would be failing on this.

It's not related to this PR. It's some issue with the Docker Clang container I made, but I can't reproduce it locally (in the exact same Docker container) and the CircleCI SSH login didn't work for me yesterday.

I have no idea why it's failing.

xelxebar:
Push change to comment. Doxygen accepts `@` as well.
Thanks for the quick feedback.

PR #912, opened 2019-04-30
Build the C BB on Kindle, Kobo & PB

NiLuJe:
Now that it's feature-complete for BB8 targets, and generally faster and
less finicky.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,7 @@ $(OUTPUT_DIR)/data:
 
 libs: \
 	$(if $(or $(SDL),$(ANDROID)),,$(OUTPUT_DIR)/libs/libkoreader-input.so) \
-	$(if $(or $(SDL),$(ANDROID)),$(OUTPUT_DIR)/libs/libblitbuffer.so,) \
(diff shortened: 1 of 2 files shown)


NiLuJe:
AFAICT, we're free from truncation-induced SIGBUS crashes when an OTA includes C BB changes, at the very least of Kobo & PB, as tar handles replacing an existing file "right" (unlike scp).

Kindle might be another kettle of fish entirely, if the fuse layer starts fucking with us. That said, it's currently half-broken already when the launch script gets updated, and a "crash" on Kindle is fairly harmless, as the background system is still running, just paused, so the annoyance factor is fairly minor.
It'd only happen in case there's actual changes in the binary layout, which should be pretty rare in any case.

NiLuJe:
I'm leaving Cervantes out, because I *think* those default to 16bpp?

Might be worth investigating getting fbdepth to behave properly there, @pazos... ;).

NiLuJe:
OOP certainly helps making it slightly less insane to read, so this is not meant to break the 1:1 parity between the two ;).

NiLuJe:
Oops. I knew I'd forgotten something : blacklisting the C BB on 4 & 24bpp devices... (i.e., legacy Kindles, and the color PB).

PR #915, opened 2019-05-07
thirdparty: adds libutf8proc.so.2 (utf8proc 2.3.0)

poire-z:
https://github.com/JuliaStrings/utf8proc a clean C library for processing UTF-8 Unicode data 
For use by crengine, for now mostly for more complete and accurate text-transform: uppercase/lowercase/capitalize.
See https://github.com/koreader/koreader/issues/4958#issuecomment-486385893 .

(Added an unusual `chmod 755 $@` because their original Makefile does `libutf8proc.so.$(MAJOR).$(MINOR).$(PATCH): [...] chmod a-x $@`, dunno why.
I didn't like that .so to have a different color than the others in `ls` my terminal :)

Also bump crengine for:
- Ensure text decoration (underline) is continued over word gaps https://github.com/koreader/crengine/pull/286
- Invalidate TOC page numbers on rendering change
- Ensure reproducible cache files when same rendering settings
- Use utf8proc for string uppercase/lowercase/capitalize https://github.com/koreader/crengine/pull/287
- LVBlockWriteStream: workaround to exclude fatal error


poire-z:
It adds a 300 Kb (arm) or 337 Kb (x64) libutf8proc.so.2.
I don't understand much of all that it allows and what else we would do with it, but it allows fixing the polish uppercase issue with just:
```c
@@ -2613,6 +2617,9 @@ void lStr_uppercase( lChar16 * str, int len )
 {
     for ( int i=0; i<len; i++ ) {
         lChar16 ch = str[i];
+#if (USE_UTF8PROC==1)
+        str[i] = utf8proc_toupper(ch);
+#else
         if ( ch>='a' && ch<='z' ) {
             str[i] = ch - 0x20;
         } else if ( ch>=0xE0 && ch<=0xFF ) {

```
Also, adding a styletweak `* { text-transform: uppercase;} ` to my timing reference book does not increase the load/render times of 17s+4s: same times with utf8proc or with the original code.

thirdparty/kpvcrlib/CMakeLists.txt
```diff
@@ -70,20 +73,23 @@ if(DEFINED ENV{WIN32})
     find_library(PNG_LIB NAMES "png16-16" PATHS $ENV{LIBS_DIR})
     find_library(FREETYPE_LIB NAMES "freetype-6" PATHS $ENV{LIBS_DIR})
     find_library(HARFBUZZ_LIB NAMES "harfbuzz-0" PATHS $ENV{LIBS_DIR})
+    find_library(UTF8PROC_LIB NAMES "harfbuzz-2" PATHS $ENV{LIBS_DIR})
```
NiLuJe: harfbuzz-2?

NiLuJe: Should be `utf8proc-2` instead, right? (I hate CMake's `find_` stuff, so take that with a grain of salt).

poire-z: Oups :/ Indeed. Cutiand pasting CMhinese is hard...
(It was still linked properly by libcrengine.so, as ldd showed.)

Frenzie: oops, lol, thanks for close reading ;-)

poire-z:
@Frenzie: how do you ETA the next stable? Should I wait before bumping that and the other crengine cache stuff? Or do we have a few days to validate all that?

Frenzie:
@poire-z I haven't had time to evaluate if there's anything important still open.

NiLuJe:
@Frenzie : Of the stuff in the milestone, I'd say only #4902 is a hard blocker (and also highly annoying in practice ;p).

Frenzie:
Alright, then I'll apply the blah mitigation.

poire-z:
What's with the **ci/circleci: emu_clang_ninja** that seems to fail on all PRs with:
```
[  FAILED  ] ./spec/base/unit/md5_spec.lua @ 10: MD5 module should calculate correct MD5 hashes
[  FAILED  ] ./spec/base/unit/md5_spec.lua @ 23: MD5 module should calculate MD5 sum of a file
```

Frenzie:
I have no idea. I can't reproduce it locally with the same Docker image.

PR #917, opened 2019-05-25
bump harfbuzz to 2.6.0

Frenzie:
> This release does not include much functional changes, but includes major internal
code-base changes. We now require C++11. Support for gcc 4.8 and earlier has been
dropped.

https://github.com/harfbuzz/harfbuzz/releases/tag/2.5.0

### DIFF
--- a/thirdparty/harfbuzz/CMakeLists.txt
+++ b/thirdparty/harfbuzz/CMakeLists.txt
@@ -32,7 +32,7 @@ endif()
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME
     https://github.com/harfbuzz/harfbuzz.git
-    2.4.0
+    2.5.0
     ${SOURCE_DIR}
 )
 
--- a/thirdparty/harfbuzz/CMakeLists.txt
+++ b/thirdparty/harfbuzz/CMakeLists.txt
@@ -29,10 +29,19 @@ if($ENV{ANDROID})
     set(CFG_CMD "${CFG_CMD} && ${ISED} 's|soname_spec=.*|soname_spec=\"\\\\\$libname\\\\\$release\\\\\$shared_ext\\\\\$major\"|' libtool")
 endif()
 
+if($ENV{POCKETBOOK})
+    # The PocketBook toolchain is stuck with gcc 4.8, support of which was dropped in HB 2.5.
+    # In the short term we deal with this by keeping PB at a lower version.
+    # See https://github.com/koreader/koreader-base/pull/917#issuecomment-495940413
+    set(HB_VER 2.4.0)
+else()
+    set(HB_VER 2.6.0)
+endif()
+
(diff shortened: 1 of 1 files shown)


Frenzie:
@NiLuJe Oh, I just remembered… PocketBook is on 4.8, isn't it. /)_-)

NiLuJe:
Yep ^^.

NiLuJe:
They have released an updated [SDK](https://github.com/pocketbook/SDK_6.3.0), but it probably drops support for an assload of old devices/FW (AFAICT, 5.19 targets Glibc 2.23, for instance).

Frenzie:
https://github.com/harfbuzz/harfbuzz/releases/tag/2.6.0 just came out (glares at PB)

poire-z:
(Just reminding @NiLuJe we still have that _FT rebuilt against HB_ thing on standby at https://github.com/koreader/crengine/pull/230#issuecomment-450857765 - dunno what improvment that would bring and if it would not make things slower)

NiLuJe:
@poire-z: That would only affect the auto-hinter, and fonts with fancy-ish OpenType features (meaning not much for Latin scripts) ;).

NiLuJe:
@Frenzie: How convoluted would it be to peg PB to an older version given the current CMake shenanigans?

Is it as easy as moving the version to a variable that get flipped depending on ENV${PB} (or whetver the actual insane syntax actually is ;p)?

Frenzie:
@NiLuJe Sure, for a single exception it could hardly be any simpler. I'm just afraid of starting a maintenance nightmare in the long run.

Dramatization:
"Oh no, my fancy Arabic ligature doesn't look right."
"Odd, that should've already been fixed in HB 6.5."
"I forgot to mention, I use PocketBook."
"Damn you HarfBuzz 2.4!"

NiLuJe:
@Frenzie: Eh, we already have a metric ton of weird crap that's broken/weird on PB, what's one more? :D.

Frenzie:
You've... got a point there. I'll update this PR accordingly.

poire-z:
> https://github.com/harfbuzz/harfbuzz/releases/tag/2.5.0
Support for gcc 4.8 and earlier has been dropped.

In case you missed it, in https://github.com/harfbuzz/harfbuzz/releases for 2.5.2:
> Fix build on gcc 4.8. That's supported again.

Some talks about dropping, less about resupporting it again in ```https://github.com/harfbuzz/harfbuzz/issues/1724``` , relevant commit is ```https://github.com/harfbuzz/harfbuzz/commit/e4e518f33d933a02058bad86a6aae714e59814db```

Frenzie:
> In case you missed it

I definitely did. So that workaround wasn't necessary? /sigh

Frenzie:
As an aside, https://github.com/koreader/koreader/pull/5268 was a strategic part of the larger strategy toward moving away from Android's gcc 4.9.

PR #921, opened 2019-06-10
[fix] DjVu gamma correction

Frenzie:
The existing implementation only worked in 8-bit mode. By using the `ddjvu_format_set_gamma()` function that comes with djvulibre it suddenly and attractively becomes Not Our Problem™.

I noticed this after doing a quick experiment with a custom power function `#ifdef __ANDROID_API__` like in https://stackoverflow.com/a/19488271/2470572

(For example, if Android define a custom power function, else `#define power(x, y) pow(x, y)`.)

I believe the proper way to go about it would probably be to ignore document-specific implementations and to do it in the blitbuffer, but this is quicker to implement.

As a side effect, this fixes https://github.com/koreader/koreader/issues/3493.

### DIFF
--- a/djvu.c
+++ b/djvu.c
@@ -662,7 +662,7 @@ static int drawPage(lua_State *L) {
 	if (gamma == 2) {
 		// default
 		gamma = 2.2;
-	}	else if (gamma < 2) {
+	} else if (gamma < 2) {
 		// with this function, 0.8 = 5, 2 = 2.2
 		gamma = 6.86666 - 2.33333 * gamma;
 	} else if (gamma > 2) {
--- a/djvu.c
+++ b/djvu.c
@@ -32,6 +32,8 @@
 #include "koptcrop.h"
 #include "djvu.h"
 
+#define ABS(x) ((x<0)?(-x):(x))
+
 #define MIN(a, b)      ((a) < (b) ? (a) : (b))
 #define MAX(a, b)      ((a) > (b) ? (a) : (b))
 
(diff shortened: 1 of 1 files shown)


Frenzie:
Illustration of the problem (because pixels are 24 bits here instead of 8):

Screenshot_2019-06-10_12-24-14

djvu.c
```diff
@@ -654,6 +656,20 @@ static int drawPage(lua_State *L) {
 	BlitBuffer *bb = (BlitBuffer*) lua_topointer(L, 3);
 	ddjvu_render_mode_t djvu_render_mode = (int) luaL_checkint(L, 6);
 	ddjvu_rect_t pagerect, renderrect;
+	// map KOReader gamma to djvulibre gamma
+	// djvulibre goes from 0.5 to 5.0
```
poire-z: What happens when you give it < 0.5 or > 5.0 ? which can happen if one hacked koptoptions.lua to extend boundary values.

Which it seems I once did, I have that in my own patch:
```lua
                 name_text = S.CONTRAST,
                 buttonprogress = true,
-                values = {1/0.8, 1/1.0, 1/1.5, 1/2.0, 1/3.0, 1/4.0, 1/6.0, 1/9.0},
+                -- for pdf reflowing mode (kopt_contrast):
+                values = {1/0.8, 1/1.0, 1/1.5, 1/2.0, 1/4.0, 1/6.0, 1/10.0, 1/50.0},
                 default_pos = 2,
                 default_value = DKOPTREADER_CONFIG_CONTRAST,
                 event = "GammaUpdate",
-                args = {0.8, 1.0, 1.5, 2.0, 3.0, 4.0, 6.0, 9.0},
-                labels = {0.8, 1.0, 1.5, 2.0, 3.0, 4.0, 6.0, 9.0},
+                -- for pdf non-reflowing mode (mupdf):
+                args =   {0.8, 1.0, 1.5, 2.0, 4.0, 6.0, 10.0, 50.0},
+                labels = {0.8, 1.0, 1.5, 2.0, 4.0, 6.0, 10.0, 50.0},
+                -- see https://github.com/koreader/koreader/issues/1299#issuecomment-65183895
                 name_text_hold_callback = optionsutil.showValues,
```

poire-z: (I mean: I don't mind having my higher values limited and causing no change from previous values - just want to be sure there could be no crash or strange things.)

Frenzie: djvulibre ignores out of bound values.

https://gitlab.com/koreader/djvulibre/blob/9658b01431cd7ff6344d7787f855179e73fe81a7/libdjvu/ddjvuapi.cpp#L2166

(One wonders if it wouldn't make more sense to cap them instead of to ignore them completely, but that aside.)

poire-z: > it wouldn't make more sense to cap them instead of to ignore them

I think you should cap them, after your `gamma = A - B*gamma` s, as they could give out of bound values, so better to have them capped that suddenly getting back with no gamma change.

PR #926, opened 2019-06-25
Better nightmode handling on the KOA2 & PW4

NiLuJe:
By utilizing *both* of the nightmode waveform modes, as they ought to, instead of blindly picking one ;).

Thanks to ilovejedd on MR for the strace logs ;).


NiLuJe:
This will probably make https://github.com/koreader/koreader/issues/5076 *worse*, but I'd like confirmation of that before fixing it the dumb way ;).

PR #936, opened 2019-07-16
Switch back to if ladders for inner loops

NiLuJe:
Potentially very slightly faster on decent CPUs (branch prediciton?).


NiLuJe:
Only noticeable difference in inverted Portrait on my Forma, with a ~8% gain. No averse effect on Landscape. And no effect on Portrait, since that's a fast-path anyway.

NiLuJe:
It's apparently very close, since code size is identical, but, yeah, not bit-for-bit exact.

Way too lazy to fire up objdump to see what's what, though ;p.

PR #944, opened 2019-08-07
[CI] Switch to Clang 8

Frenzie:
With any luck the odd MD5 problem will go away…

See <https://github.com/koreader/virdevenv/pull/42>.


Frenzie:
Annd of course it didn't. :-)

PR #946, opened 2019-08-07
[chore] Remove useless utf8proc CXX vars

Frenzie:
The program only uses C.


Frenzie:
It's harmless for the most part, but it makes it ever so slightly harder to maintain.

PR #960, opened 2019-08-29
FFI: Add MuPDF fz_is_document_reflowable

Frenzie:
This will be queried in front to dynamically set font size.

Cf. <https://github.com/koreader/koreader/issues/4368>.

### DIFF
--- a/ffi/mupdf.lua
+++ b/ffi/mupdf.lua
@@ -173,7 +173,7 @@ end
 
 function document_mt.__index:isDocumentReflowable(width, height, em)
     if self.is_reflowable then return self.is_reflowable end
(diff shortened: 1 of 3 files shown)


ffi/mupdf.lua
```diff
@@ -171,6 +171,12 @@ function document_mt.__index:getPages()
     return pages
 end
 
+function document_mt.__index:isDocumentReflowable(width, height, em)
+    if self.is_reflowable then return self.is_reflowable end
+    self.is_reflowable = M.fz_is_document_reflowable(context(), self.doc)
```
Frenzie: Actually I should turn it into a boolean; this is just 0 or 1.

PR #966, opened 2019-09-05
Build FBInk with OpenType support

NiLuJe:
Might be using it for fun later ;p.


NiLuJe:
(Increase in binary size is minimal, ~66KB to ~125KB)

PR #968, opened 2019-09-07
Add some more rtc stuff

NiLuJe:
(re: https://github.com/koreader/koreader/pull/5335)


Frenzie:
Is that nothrow stuff the regular ffi-cdecl output?

NiLuJe:
Yeah, possibly because that's from my native compiler/glibc.

Should be fairly harmless (i.e., I'm sure it's actually mostly ignored in practice).

PR #969, opened 2019-09-08
[feat] Add FFI RTC interface

Frenzie:
This will allow for scheduling wakeups.

Prerequisite for <https://github.com/koreader/koreader/pull/5335>.

### DIFF
--- a/ffi/rtc.lua
+++ b/ffi/rtc.lua
@@ -75,29 +75,28 @@ function RTC:setWakeupAlarm(epoch, enabled)
     if rtc0 == -1 then
         err = ffi.string(C.strerror(ffi.errno()))
         print("setWakeupAlarm open /dev/rtc0", rtc0, err)
+        return nil, re, err
     end
     local re = C.ioctl(rtc0, C.RTC_WKALM_SET, wake)
     if re == -1 then
         err = ffi.string(C.strerror(ffi.errno()))
         print("setWakeupAlarm ioctl RTC_WKALM_SET", re, err)
+        return nil, re, err
     end
     re = C.close(rtc0)
     if re == -1 then
         err = ffi.string(C.strerror(ffi.errno()))
         print("setWakeupAlarm close /dev/rtc0", re, err)
-    end
-
-    if re == 0 then
-        if enabled then
-            self._wakeup_scheduled = true
-        else
-            self._wakeup_scheduled = false
-            self._wakeup_scheduled_ptm = nil
-        end
-        return true
-    else
         return nil, re, err
     end
+
+    if enabled then
+        self._wakeup_scheduled = true
+    else
+        self._wakeup_scheduled = false
+        self._wakeup_scheduled_ptm = nil
+    end
+    return true
 end
 
 --[[--
@@ -133,16 +132,19 @@ function RTC:getWakeupAlarmSys()
     if rtc0 == -1 then
         err = ffi.string(C.strerror(ffi.errno()))
         print("getWakeupAlarm open /dev/rtc0", rtc0, err)
+        return nil, re, err
     end
     re = C.ioctl(rtc0, C.RTC_WKALM_RD, wake)
     if re == -1 then
         err = ffi.string(C.strerror(ffi.errno()))
         print("getWakeupAlarm ioctl RTC_WKALM_RD", re, err)
+        return nil, re, err
     end
     re = C.close(rtc0)
     if re == -1 then
         err = ffi.string(C.strerror(ffi.errno()))
         print("getWakeupAlarm close /dev/rtc0", re, err)
+        return nil, re, err
     end
 
     if wake ~= -1 then
--- a/ffi/rtc.lua
+++ b/ffi/rtc.lua
@@ -1,0 +1,165 @@
+--[[--
+Module for interfacing with the RTC (real time clock).
+
+This module provides the ability to schedule wakeups through RTC.
+
+@module ffi.rtc
+]]
+
+local ffi = require("ffi")
+local bor = bit.bor
+local C = ffi.C
+
+local dummy = require("ffi/posix_h")
+local dummy = require("ffi/rtc_h")
+
+-----------------------------------------------------------------
+
+local RTC = {
+    _wakeup_scheduled = false,
+    _wakeup_scheduled_ptm = nil,
+}
+
+function RTC:secondsFromNowToEpoch(seconds_from_now)
+    local t = ffi.new("time_t[1]")
+    t[0] = C.time(nil)
+    t[0] = t[0] + seconds_from_now
+    local epoch = C.mktime(C.localtime(t))
+    return epoch
+end
+
+function RTC:setWakeupAlarm(epoch, enabled)
+    enabled = (enabled ~= nil) and enabled or true
+
+    local ptm = C.gmtime(ffi.new("int[1]", epoch))
+    self._wakeup_scheduled_ptm = ptm
+
+    local wake = ffi.new("struct rtc_wkalrm")
+    wake.time.tm_sec = ptm.tm_sec
+    wake.time.tm_min = ptm.tm_min
+    wake.time.tm_hour = ptm.tm_hour
+    wake.time.tm_mday = ptm.tm_mday
+    wake.time.tm_mon = ptm.tm_mon
+    wake.time.tm_year = ptm.tm_year
+    -- wday, yday, and isdst fields are unused by Linux
+    wake.time.tm_wday = -1
+    wake.time.tm_yday = -1
+    wake.time.tm_isdst = -1
+
+    wake.enabled = enabled and 1 or 0
+
+    local err
+    local rtc0 = C.open("/dev/rtc0", bor(C.O_RDONLY, C.O_NONBLOCK))
+    if rtc0 == -1 then
+        err = ffi.string(C.strerror(ffi.errno()))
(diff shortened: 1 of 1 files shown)


ffi/rtc.lua
```diff
@@ -0,0 +1,165 @@
...
+    wake.time.tm_year = ptm.tm_year
+    -- wday, yday, and isdst fields are unused by Linux
+    wake.time.tm_wday = -1
+    wake.time.tm_yday = -1
+    wake.time.tm_isdst = -1
+
+    wake.enabled = enabled and 1 or 0
+
+    local err
+    local rtc0 = C.open("/dev/rtc0", bor(C.O_RDONLY, C.O_NONBLOCK))
```
Frenzie: Whoops, I forgot to make these configurable. Tbh it may just be a waste of time to.

ffi/rtc.lua
```diff
@@ -0,0 +1,196 @@
...
+
+@module ffi.rtc
+]]
+
+local ffi = require("ffi")
+local bor = bit.bor
+local C = ffi.C
+
+local dummy = require("ffi/posix_h")
+local dummy = require("ffi/rtc_h")
```
poire-z: (remind me) what does these do, are they needed?

ffi/rtc.lua
```diff
@@ -0,0 +1,196 @@
...
+
+-----------------------------------------------------------------
+
+local RTC = {
+    _wakeup_scheduled = false,
+    _wakeup_scheduled_ptm = nil,
```
poire-z: ptm = ? (I guess it's a pointer to a "tm" :) but some comment might help getting into all that (I know nothing about that, so... discovering/learning)

ffi/rtc.lua
```diff
@@ -0,0 +1,196 @@
...
+        print("setWakeupAlarm ioctl RTC_WKALM_SET", re, err)
+    end
+    re = C.close(rtc0)
+    if re == -1 then
+        err = ffi.string(C.strerror(ffi.errno()))
+        print("setWakeupAlarm close /dev/rtc0", re, err)
+    end
+
+    if re == 0 then
+        if enabled then
```
poire-z: you go on doing things above even if the first one failed? expecting all the others will fail so you still get `re==-1` in the end?
I would expect a `return nil, re, err` in each of the `ifs`

ffi/rtc.lua
```diff
@@ -0,0 +1,196 @@
...
+    wake.time.tm_hour = ptm.tm_hour
+    wake.time.tm_mday = ptm.tm_mday
+    wake.time.tm_mon = ptm.tm_mon
+    wake.time.tm_year = ptm.tm_year
+    -- wday, yday, and isdst fields are unused by Linux
+    wake.time.tm_wday = -1
+    wake.time.tm_yday = -1
+    wake.time.tm_isdst = -1
+
+    wake.enabled = enabled and 1 or 0
```
poire-z: If you want to disable it, you can provide any time value, including 0? (= it doesn't matter?)

ffi/rtc.lua
```diff
@@ -0,0 +1,196 @@
...
+
+--[[--
+Get wakealarm as set by us.
+--]]
+function RTC:getWakeupAlarm()
+    return self._wakeup_scheduled_ptm
+end
+
+--[[--
+Get RTC wakealarm from system.
```
poire-z: What does that mean?
Is it an alternative/different alarm?
Or just to get a previously set alarm, by another process or a previous run of KOReader?
And do we need to clear any alarm set on exit/crash?

ffi/rtc.lua
NiLuJe: The rtc path, you mean?

Yeah, probably not worth it ;p.

ffi/rtc.lua
NiLuJe: We need them to add the requisite functions & constants to the ffi C. namespace ;).

ffi/rtc.lua
NiLuJe: RTC stands for Real Time Clock. It's the physical crystal-based, battery-powered ticking clock responsible for a good part of the time-keeping (as far as "wall-clock", i.e., real word time) is concerned ;).

ffi/rtc.lua
```diff
@@ -0,0 +1,196 @@
...
+    alarm_epoch = C.mktime(alarm)
+    alarm_sys_epoch = C.mktime(alarm_sys)
+
+    print("validateWakeupAlarmByProximity", task_alarm_epoch, alarm_epoch, alarm_sys_epoch, current_time_epoch)
+
+    -- If our stored alarm and the system alarm don't match, we didn't set it.
+    if not (alarm_epoch == alarm_sys_epoch) then return end
+
+    -- If our stored alarm and the provided task alarm don't match,
+    -- we're not talking about the same task. This should never happen.
```
poire-z: Unless 2 features/plugins set each an alarm. This will happen to the oldest plugin that did, no?
(if task_alarm_epoch should be the same used to set the alarm, and that the plugins should store so they can provide it to this function later?)

ffi/rtc.lua
poire-z: Thanks, I knew the thing :)
I was wondering about the notion of "system alarm" (the one we fetch from rtc0) vs "our alarm" (that we store as an attribute) here :) which I ended up understanding after reading the whole file.

ffi/rtc.lua
poire-z: Might be worth a comment (a quick check at base/ffi/* shows no comment around these, so your sentence deserves to be at least in one file :)

ffi/rtc.lua
Frenzie: Good point.

ffi/rtc.lua
Frenzie: Will do. Btw, it's clearer if you look at rtc_h.lua.

ffi/rtc.lua
Frenzie: I'll add some explanation in the function description (+ internal references).

ffi/rtc.lua
Frenzie: Yeah, setting it to a time in the past would effectively disable it, and I don't think I actually tried what happens when you combine disabled with a time in the future. It can probably be removed.

Also see http://man7.org/linux/man-pages/man4/rtc.4.html

ffi/rtc.lua
Frenzie: Not true :-P

https://github.com/koreader/koreader-base/blob/4327fd094cef238eed00d9b897650b3c0f8fc144/ffi/kobolight.lua#L9-L10

ffi/rtc.lua
Frenzie: It could happen if a rogue plugin avoided the WakeupMgr:addTask/removeTask mechanism, in which case the check could prevent WakeupMgr from running its top task. (However, I haven't currently implemented that in front.)

ffi/rtc.lua
Frenzie: Although technically the close doesn't mean it's not scheduled.

ffi/rtc.lua
poire-z: @NiLuJe sentence above is just perfect and more precise.

Frenzie: @poire-z Something like this?

poire-z: yep, fine, thanks

Frenzie:
I'll merge this now since this syscall stuff should be fairly stable, plus I've gone over several nights of sleep since first writing it. @NiLuJe also directly assisted with the implementation of the first iteration, from which it hasn't substantially changed.

Then if you could please direct your critical eye to koreader/koreader#5335 I'd be much obliged. :-)

ffi/rtc.lua
```diff
@@ -0,0 +1,210 @@
...
+@enabled bool Whether the call enables or disables the alarm. Defaults to true.
+
+@treturn bool Success.
+@treturn re Error code (if any).
+@treturn err Error string (if any).
+--]]
+function RTC:setWakeupAlarm(epoch, enabled)
+    enabled = (enabled ~= nil) and enabled or true
+
+    local ptm = C.gmtime(ffi.new("int[1]", epoch))
```
Frenzie: @NiLuJe This was my new simplified approach (without a whole time = time_t etc.) but apparently I accidentally made the module ARM-specific because on my computer it errors saying it should be uint_64. :man_facepalming: 

Anyway, not an issue for the initial Kobo implementation but I figured I'd leave a note. I'll (hopefully) look into recomplexifying it after the finishing touches on the frontend implementation. Which I expected to have finished already but I had to rush that MuPDF patch and stuff. :-P

NiLuJe:
@Frenzie: That's because time_t is actually a long int (c.f., `echo | gcc -E -xc -include 'time.h' - | grep time_t`), so, 32bit on ARM, 64bit on x64 ;).

What you're doing here should work just fine with `ffi.new("time_t[1]", epoch)` instead.

Frenzie:
@NiLuJe I think there might've been TZ complications. Either way, I won't look at it today.

NiLuJe:
I don't see how that could be the case, there's no magic involved, time_t is just a typedef,
```lua
local ptm = C.gmtime(ffi.new("time_t[1]", epoch))
```
is essentially doing
```C
time_t foo = epoch;
struct tm* ptm = gmtime(&echo);
```

PR #971, opened 2019-09-09
[fix] MuPDF Android font dir patch update

Frenzie:
Cf. <https://github.com/koreader/koreader/issues/5347#issuecomment-529480686>.

### DIFF
--- a/thirdparty/mupdf/external_fonts.patch
+++ b/thirdparty/mupdf/external_fonts.patch
@@ -68,7 +68,7 @@ index f6951ba2..b1fdd981 100644
  /*
  	Base 14 PDF fonts from URW.
  	Noto fonts from Google.
-@@ -367,3 +371,185 @@ fz_lookup_noto_emoji_font(fz_context *ctx, int *size)
+@@ -367,3 +371,188 @@ fz_lookup_noto_emoji_font(fz_context *ctx, int *size)
  	return *size = 0, NULL;
  #endif
  }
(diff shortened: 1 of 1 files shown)


thirdparty/mupdf/external_fonts.patch
```diff
@@ -93,6 +94,22 @@ index f6951ba2..b1fdd981 100644
 +	return filename;
 +}
 +
++char *
++get_font_file_android(char *name)
++{
++	char *fontdir;
```
NiLuJe:
```C
	const char fontdir[] = "/system/fonts";
	size_t len = sizeof(fontdir) + strlen(name) + 1;	// sizeof on a string literal includes the NUL
	char *filename = malloc(len);
	if (filename == NULL) {
		return NULL;
	}
```

(Fairly minor, this works ;)).

Frenzie: I'm pretty sure I tested with what I thought of as a stress test (EPUB) but I guess that just used the Times (or whatever) that's bundled with MuPDF. In any case I didn't have any HTML dictionaries installed on the emulator.

Frenzie: You're fine with the basic principle besides this nit? :-P

NiLuJe: Yep!

thirdparty/mupdf/external_fonts.patch
```diff
@@ -68,7 +68,7 @@ index f6951ba2..b1fdd981 100644
  /*
  	Base 14 PDF fonts from URW.
  	Noto fonts from Google.
-@@ -367,3 +371,185 @@ fz_lookup_noto_emoji_font(fz_context *ctx, int *size)
+@@ -367,3 +371,188 @@ fz_lookup_noto_emoji_font(fz_context *ctx, int *size)
```
Frenzie: @NiLuJe Apparently I thought your suggestion took off more lines than it did? :-P Oh well, luckily there's the CI.

PR #972, opened 2019-09-09
[chore] thirdparty/glib Get rid of unrecognized options

Frenzie:
These must've existed in the past.

```
configure: WARNING: unrecognized options: --without-included-gettext, --with-gettext
```


NiLuJe:
:+1:

Reminder that newer glib versions have moved to meson, and that it's somewhat unfun to deal with in our context.

Frenzie:
Well… drat.

NiLuJe:
We talked about it a while back on gitter, but basically, the gist of it is:

* It needs a TC config file (àla CMake).
* it doesn't support environment variables.

Which means that the current accepted workaround is to wrap everything in something that can dynamically build a proper toolchain file (via, oh, wait, environment variables... -_-").

c.f. [how I did it](http://trac.ak-team.com/trac/browser/niluje/Configs/trunk/Kindle/Misc/x-compile.sh?rev=16458#L3774), which turned out to be fairly similar to how [Gentoo handles it](https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/meson.eclass#n123).

NiLuJe:
There's also a sneaky new "feature" in [0.51](https://mesonbuild.com/Release-notes-for-0-51-0.html#sanity-checking-compilers-with-user-flags) where *FLAGS form the env are used to sanity check the *cross* compiler.

The reasoning for that bit of insanity is buried in an offhand comment in a bit of the docs: the environment should *only* apply to the *native* compiler.
Everything regarding the *cross* toolchain should come from the cross file.

In practice, I'm fairly sure I saw some stuff bleed through (mostly around linker flags), but, eh.

NiLuJe:
Once you have something that fits into the strict design rules, it's fairly neat, though.

PR #974, opened 2019-09-11
Minor mxcfb refresh cleanups

NiLuJe:
* Get rid of the commented out collision stuff, as it's unlikely we'll ever be able to get it right anyway, so ti's just confusing noise.
* Use a slightly less cute marker range to skip a comparison.
* Add some more comments about our wait_for_update_complete usage, after having tried a few things (with essentially similar results, hence the slightly functionally smaller diff than I originally anticipated ;p).
* Avoid waiting for completion of the same marker twice (could probably only happen on REAGL Kindles).

### DIFF
--- a/ffi/framebuffer_mxcfb.lua
+++ b/ffi/framebuffer_mxcfb.lua
@@ -107,19 +107,19 @@ end
 
 -- Kindle's MXCFB_WAIT_FOR_UPDATE_COMPLETE_PEARL == 0x4004462f
 local function kindle_pearl_mxc_wait_for_update_complete(fb, marker)
-    -- Wait for the previous update to be completed
+    -- Wait for a specific update to be completed
     return C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE_PEARL, ffi.new("uint32_t[1]", marker))
 end
 
 -- Kobo's MXCFB_WAIT_FOR_UPDATE_COMPLETE_V1 == 0x4004462f
 local function kobo_mxc_wait_for_update_complete(fb, marker)
-    -- Wait for the previous update to be completed
+    -- Wait for a specific update to be completed
     return C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_COMPLETE_V1, ffi.new("uint32_t[1]", marker))
 end
 
 -- Kobo's Mk7 MXCFB_WAIT_FOR_UPDATE_COMPLETE_V3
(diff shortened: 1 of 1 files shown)


ffi/framebuffer_mxcfb.lua
```diff
@@ -185,13 +183,13 @@ local function mxc_update(fb, update_ioctl, refarea, refresh_type, waveform_mode
...
     if w == 0 or h == 0 then
-        fb.debug("got a 0 size (height and/or width) refresh request, ignoring it.")
+        fb.debug("got an empty (no height and/or width) refresh request, ignoring it.")
         return
     end
 
     if w == 1 and h == 1 then
-        -- avoid system freeze when updating 1x1 pixel block on KPW2 and KV,
-        -- see koreader/koreader#1299 and koreader/koreader#1486
+        -- Avoid a kernel deadlock when updating 1x1 pixel block on KPW2 and KV,
```
Frenzie: Huh, I don't think I noticed this particular comment before. But 1 pixel updates could be used to cool effect! ;_;

poire-z: Wouldn't it be better to extend it to 2x2, instead of just dropping it? That standalone pixel . might be important   . :)

NiLuJe: `1xn` *ought* to work, but `1x1` kills some kernels ^^.

NiLuJe: The issue is, 2x2 also kills some kernels IIRC :D.

NiLuJe: Hmm, double-checked, I'm using `(region.width <= 1 || region.height <= 1)` in FBInk.

poire-z: If i remember correctly, in frontend, we always add 1 on each side - so, from KOReader, we should always have larger than 2x2 here. (?)

Frenzie:
@NiLuJe See https://github.com/koreader/koreader/blob/0dad707e2ed9f4f5e9444e2dbd0934e8b8bc2797/frontend/device/kobo/device.lua#L277

Should be the same on all device implementations.

NiLuJe:
@Frenzie: Huh. It appears to properly point to logger.dbg, but I'm not seeing it log anything o_O.

NiLuJe:
@Frenzie: Ah, caught it. reader.lua requires device *before* setting debug mode since the canvascontext stuff.

PR #978, opened 2019-09-14
Update to FBInk v1.20.0

NiLuJe:
Proper Kobo Libra H2O support


NiLuJe:
Basically detecting it as a Mk. 7, so it uses the right ioctls, instead of the legacy compat layer ;).

Frenzie:
@NiLuJe I was going to click merge but it looks like something went wrong. ;-)

PR #979, opened 2019-09-14
SDL: add rumble support

Frenzie:
Cf. <https://github.com/koreader/koreader/issues/5374>.


ffi/SDL2_0_h.lua
```diff
@@ -785,6 +785,7 @@ SDL_bool SDL_GameControllerGetAttached(SDL_GameController *) __attribute__((visi
 SDL_Joystick *SDL_GameControllerGetJoystick(SDL_GameController *) __attribute__((visibility("default")));
 int SDL_GameControllerEventState(int) __attribute__((visibility("default")));
 void SDL_GameControllerUpdate(void) __attribute__((visibility("default")));
+int SDL_GameControllerRumble(SDL_GameController *, Uint16, Uint16, Uint32) __attribute__((visibility("default")));
```
Frenzie: PS This is only available starting with 2.0.9.

PR #980, opened 2019-09-15
bump crengine: support RTL/bidi text, adds thirdparty/fribidi

poire-z:
Adds to thirdparty: https://github.com/fribidi/fribidi

Bump crengine https://github.com/koreader/crengine/pull/309 :
- Fix possible jumps in list items numbering
- lvfntman.cpp: reformating for better readability
- lvtextfm.cpp: upgrade m_flags from lUInt8 to lUInt16
- Fonts: use Freetype embolden API for fake bold
- Fonts: have fallback font be italic/bold as the main font
- Fonts: handle RTL, rework Harfbuzz measuring and drawing
- Text formatting: use fribidi for RTL/bidi text layout
- Update XPointers/pt translation methods to handle bidi/rtl words
- EPUB: forward dir and lang attributes from the html element
- RTL: fix in-page footnotes order
- Fix a few clang-tidy warnings


thirdparty/fribidi/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+assert_var_defined(LDFLAGS)
+
+ep_get_source_dir(SOURCE_DIR)
+ep_get_binary_dir(BINARY_DIR)
+
+set(CFG_ENV_VAR "CC=\"${CC}\" CXX=\"${CXX}\" CPPFLAGS=\"${CPPFLAGS}\" LDFLAGS=\"${LDFLAGS}\" ")
+set(CFG_OPTS "-q --prefix=${BINARY_DIR} --disable-static --enable-shared --host=\"${CHOST}\"")
+set(CFG_CMD sh -c "${CFG_ENV_VAR} ${SOURCE_DIR}/configure ${CFG_OPTS}")
+
+# Disable building docs, as it requires c2man
```
Frenzie: And reducing built time by reducing unnecessary stuff is a good idea regardless. ;-)

thirdparty/fribidi/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+    set(CFG_CMD "${CFG_CMD} && ${ISED} 's|need_lib_prefix=no|need_lib_prefix=yes|' libtool")
+    set(CFG_CMD "${CFG_CMD} && ${ISED} 's|need_version=no|need_version=yes|' libtool")
+    set(CFG_CMD "${CFG_CMD} && ${ISED} 's|library_names_spec=.*|library_names_spec=\"\\\\\$libname\\\\\$release\\\\\$shared_ext\\\\\$versuffix \\\\\$libname\\\\\$release\\\\\$shared_ext\\\\\$major \\\\\$libname\\\\\$shared_ext\"|' libtool")
+    set(CFG_CMD "${CFG_CMD} && ${ISED} 's|soname_spec=.*|soname_spec=\"\\\\\$libname\\\\\$release\\\\\$shared_ext\\\\\$major\"|' libtool")
+endif()
+
+ko_write_gitclone_script(
+    GIT_CLONE_SCRIPT_FILENAME
+    https://github.com/fribidi/fribidi.git
+    v1.0.5
```
Frenzie: I've taken to using:
```
tags/v1.0.5
```
Because I figured it might be clearer to onlookers. Also you might have a branch and a tag of the same name but that's generally unlikely to be a concern (because even if it is… the 1.0.5 branch would probably be pretty similar to the tag).

thirdparty/fribidi/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+    ${SOURCE_DIR}
+)
+
+include(ExternalProject)
+ExternalProject_Add(
+    ${PROJECT_NAME}
+    DOWNLOAD_COMMAND ${CMAKE_COMMAND} -P ${GIT_CLONE_SCRIPT_FILENAME}
+    PATCH_COMMAND NOCONFIGURE=1 ./autogen.sh
```
Frenzie: I think using CFG_CMD1 and CFG_CMD2 would be clearer.

Like this, except the example below refers to actual patches:
https://github.com/koreader/koreader-base/blob/780cb9aad8fccd79bcc4600330a3d076d4c99bd4/thirdparty/mupdf/CMakeLists.txt#L99

Frenzie: Actually since they depend on each other we should just use the && method.

poire-z:
Any idea why the android builds fail with:
```
[5/8] Performing configure step for 'fribidi'
configure: WARNING: using cross tools not prefixed with host triplet
[6/8] Performing build step for 'fribidi'
FAILED: cd /home/ko/base/thirdparty/fribidi/build/arm-linux-androideabi/fribidi-prefix/src/fribidi && make -j3 && /usr/bin/cmake -E touch /home/ko/base/thirdparty/fribidi/build/arm-linux-androideabi/fribidi-prefix/src/fribidi-stamp/fribidi-build
  ^=========== is that an issue?
make[1]: Entering directory `/home/ko/base/thirdparty/fribidi/build/arm-linux-androideabi/fribidi-prefix/src/fribidi'
(cd lib && make  fribidi-unicode-version.h)
git.mk: Generating .gitignore
make[2]: Entering directory `/home/ko/base/thirdparty/fribidi/build/arm-linux-androideabi/fribidi-prefix/src/fribidi/lib'
make[3]: Entering directory `/home/ko/base/thirdparty/fribidi/build/arm-linux-androideabi/fribidi-prefix/src/fribidi/gen.tab'
  CC       gen-unicode-version.o
  CCLD     gen-unicode-version
/usr/bin/ld: unrecognized option '--fix-cortex-a8'
  ^=========== or/and is that an issue?
/usr/bin/ld: use the --help option for usage information
collect2: error: ld returned 1 exit status
make[3]: *** [gen-unicode-version] Error 1
make[3]: Leaving directory `/home/ko/base/thirdparty/fribidi/build/arm-linux-androideabi/fribidi-prefix/src/fribidi/gen.tab'
make[2]: *** [fribidi-unicode-version.h] Error 2
make[2]: Leaving directory `/home/ko/base/thirdparty/fribidi/build/arm-linux-androideabi/fribidi-prefix/src/fribidi/lib'
make[1]: *** [lib/fribidi-unicode-version.h] Error 2
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory `/home/ko/base/thirdparty/fribidi/build/arm-linux-androideabi/fribidi-prefix/src/fribidi'
Makefile.third:46: recipe for target 'build/arm-linux-androideabi/libs/libfribidi.so.0' failed
make: *** [build/arm-linux-androideabi/libs/libfribidi.so.0] Error 1
```
while no other package complains about `--fix-cortex-a8` ?

The pocketbook builds have been failing for some time because:
`The job exceeded the maximum log length, and has been terminated.`

Frenzie:
> The pocketbook builds have been failing for some time because:

Only on Travis, which we can probably remove by now.

> while no other package complains about --fix-cortex-a8 ?

It looks like you're passing it as a flag to the linker, but it's only for the compiler.

Frenzie:
I can look at the Android build on my local environment this afternoon if you like. Trying to fix it on Travis would just take an eternity.

Frenzie:
Actually hold on, is Android even supposed to use /usb/bin/ld? I think it should use the one from the Android TC.

pazos:
https://github.com/koreader/koreader-base/blob/6b6497871ba269e25ac9b2be69e1c77a106ac4d8/Makefile.defs#L460

Frenzie:
@pazos Yeah, so it's using the wrong linker. The flags are fine. My bad. :-)

poire-z:
Had a random test by changing CMAKE_CURRENT_SOURCE_DIR to CMAKE_CURRENT_LIST_DIR, as some other package have it - but it doesn't help.
Could be something in fribidi sources that we would need patching, but I don't see anything related to using LD or ld... no hardcoded /usr/bin/ld...

> I can look at the Android build on my local environment this afternoon if you like

Yes, please, thanks.

Frenzie:
@poire-z Am I doing something wrong here btw? I'm getting a bunch of question marks on Hebrew characters:

Screenshot_2019-09-15_13-00-08

poire-z:
Dunno, do you have a font with these glyphs, as main or fallback font? FreeSerif or DejaVuSans should have them.
Works for me with FreeSerif:
image

Frenzie:
Oh, indeed. It was set to Noto Serif. I didn't realize Hebrew wasn't in there.

Frenzie:
@NiLuJe Could you also take a look? :blush: Forcing ld with a CCLD environment variable fails due to DEFS from the configure script, and in any case shouldn't gcc take care of using the correct linker?

In case it's of use:

```
$ ./configure --help
`configure' configures GNU FriBidi 1.0.5 to adapt to many kinds of systems.

Usage: ./configure [OPTION]... [VAR=VALUE]...

To assign environment variables (e.g., CC, CFLAGS...), specify them as
VAR=VALUE.  See below for descriptions of some of the useful variables.

Defaults for the options are specified in brackets.

Configuration:
  -h, --help              display this help and exit
      --help=short        display options specific to this package
      --help=recursive    display the short help of all the included packages
  -V, --version           display version information and exit
  -q, --quiet, --silent   do not print `checking ...' messages
      --cache-file=FILE   cache test results in FILE [disabled]
  -C, --config-cache      alias for `--cache-file=config.cache'
  -n, --no-create         do not create output files
      --srcdir=DIR        find the sources in DIR [configure dir or `..']

Installation directories:
  --prefix=PREFIX         install architecture-independent files in PREFIX
                          [/usr/local]
  --exec-prefix=EPREFIX   install architecture-dependent files in EPREFIX
                          [PREFIX]

By default, `make install' will install all the files in
`/usr/local/bin', `/usr/local/lib' etc.  You can specify
an installation prefix other than `/usr/local' using `--prefix',
for instance `--prefix=$HOME'.

For better control, use the options below.

Fine tuning of the installation directories:
  --bindir=DIR            user executables [EPREFIX/bin]
  --sbindir=DIR           system admin executables [EPREFIX/sbin]
  --libexecdir=DIR        program executables [EPREFIX/libexec]
  --sysconfdir=DIR        read-only single-machine data [PREFIX/etc]
  --sharedstatedir=DIR    modifiable architecture-independent data [PREFIX/com]
  --localstatedir=DIR     modifiable single-machine data [PREFIX/var]
  --runstatedir=DIR       modifiable per-process data [LOCALSTATEDIR/run]
  --libdir=DIR            object code libraries [EPREFIX/lib]
  --includedir=DIR        C header files [PREFIX/include]
  --oldincludedir=DIR     C header files for non-gcc [/usr/include]
  --datarootdir=DIR       read-only arch.-independent data root [PREFIX/share]
  --datadir=DIR           read-only architecture-independent data [DATAROOTDIR]
  --infodir=DIR           info documentation [DATAROOTDIR/info]
  --localedir=DIR         locale-dependent data [DATAROOTDIR/locale]
  --mandir=DIR            man documentation [DATAROOTDIR/man]
  --docdir=DIR            documentation root [DATAROOTDIR/doc/fribidi]
  --htmldir=DIR           html documentation [DOCDIR]
  --dvidir=DIR            dvi documentation [DOCDIR]
  --pdfdir=DIR            pdf documentation [DOCDIR]
  --psdir=DIR             ps documentation [DOCDIR]

Program names:
  --program-prefix=PREFIX            prepend PREFIX to installed program names
  --program-suffix=SUFFIX            append SUFFIX to installed program names
  --program-transform-name=PROGRAM   run sed PROGRAM on installed program names

System types:
  --build=BUILD     configure for building on BUILD [guessed]
  --host=HOST       cross-compile to build programs to run on HOST [BUILD]

Optional Features:
  --disable-option-checking  ignore unrecognized --enable/--with options
  --disable-FEATURE       do not include FEATURE (same as --enable-FEATURE=no)
  --enable-FEATURE[=ARG]  include FEATURE [ARG=yes]
  --enable-silent-rules   less verbose build output (undo: "make V=1")
  --disable-silent-rules  verbose build output (undo: "make V=0")
  --enable-dependency-tracking
                          do not reject slow dependency extractors
  --disable-dependency-tracking
                          speeds up one-time build
  --enable-static[=PKGS]  build static libraries [default=no]
  --enable-shared[=PKGS]  build shared libraries [default=yes]
  --enable-fast-install[=PKGS]
                          optimize for fast installation [default=yes]
  --disable-libtool-lock  avoid locking (might break parallel builds)
  --enable-debug          turn on debugging [default=no]
  --disable-deprecated    do not build deprecated functionality

Optional Packages:
  --with-PACKAGE[=ARG]    use PACKAGE [ARG=yes]
  --without-PACKAGE       do not use PACKAGE (same as --with-PACKAGE=no)
  --with-pic[=PKGS]       try to use only PIC/non-PIC objects [default=use
                          both]
  --with-aix-soname=aix|svr4|both
                          shared library versioning (aka "SONAME") variant to
                          provide on AIX, [default=aix].
  --with-gnu-ld           assume the C compiler uses GNU ld [default=no]
  --with-sysroot[=DIR]    Search for dependent libraries within DIR (or the
                          compiler's sysroot if not specified).

Some influential environment variables:
  CC          C compiler command
  CFLAGS      C compiler flags
  LDFLAGS     linker flags, e.g. -L<lib dir> if you have libraries in a
              nonstandard directory <lib dir>
  LIBS        libraries to pass to the linker, e.g. -l<library>
  CPPFLAGS    (Objective) C/C++ preprocessor flags, e.g. -I<include dir> if
              you have headers in a nonstandard directory <include dir>
  LT_SYS_LIBRARY_PATH
              User-defined run-time library search path.
  CPP         C preprocessor
  PKG_CONFIG  path to pkg-config utility
  PKG_CONFIG_PATH
              directories to add to pkg-config's search path
  PKG_CONFIG_LIBDIR
              path overriding pkg-config's built-in search path

Use these variables to override the choices made by `configure' or to help
it to find libraries and programs with nonstandard names/locations.

Report bugs to <https://github.com/fribidi/fribidi/issues/new>.
GNU FriBidi home page: <http://fribidi.org/>.
General help using GNU software: <http://www.gnu.org/gethelp/>.
```

NiLuJe:
CCLD is *supposed* to be CC, not LD.

Not quite sure what could be going especially wrong, as it appears to be bog standard autotools + libtools. :?

Frenzie:
> CCLD is supposed to be CC, not LD.

I figured as much but forgot to edit the message. Different failure. :-)

Frenzie:
Like so:
```
make[3]: Entering directory `/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi/lib'
make[4]: Entering directory `/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi/gen.tab'
  CC       gen-unicode-version.o
  CCLD     gen-unicode-version
/home/frans/src/kobo/koreader/base/toolchain/android-toolchain-x86/bin/../lib/gcc/i686-linux-android/4.9.x/../../../../i686-linux-android/bin/ld: error: gen-unicode-version.o: incompatible target
/home/frans/src/kobo/koreader/base/toolchain/android-toolchain-x86/bin/../sysroot/usr/lib/crtbegin_dynamic.o:crtbegin.c:function _start: error: undefined reference to 'main'
collect2: error: ld returned 1 exit status
make[4]: *** [gen-unicode-version] Error 1
```

NiLuJe:
Means something is picking up the wrong (i.E., native) compiler/linker somewhere, still.

Frenzie:
Missing something or other still I guess:

```
(cd lib && make  fribidi-unicode-version.h)
make[3]: Entering directory `/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi/lib'
make[4]: Entering directory `/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi/gen.tab'
  CC       gen-unicode-version.o
  CCLD     gen-unicode-version
  GEN      fribidi-unicode-version.h
/bin/bash: ./gen-unicode-version: No such file or directory
```

Frenzie:
Verbose:
```
make[2]: Entering directory `/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi'
(cd lib && make  fribidi-unicode-version.h)
make[3]: Entering directory `/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi/lib'
make[4]: Entering directory `/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi/gen.tab'
./gen-unicode-version unidata/ReadMe.txt unidata/BidiMirroring.txt gen-unicode-version > fribidi-unicode-version.h \
	|| (rm -f fribidi-unicode-version.h && false)
/bin/bash: ./gen-unicode-version: No such file or directory
make[4]: *** [fribidi-unicode-version.h] Error 1
```

poire-z:
Regarding the first error:
`FAILED: cd /home/ko/base/thirdparty/fribidi/build/arm-linux-androideabi/fribidi-prefix/src/fribidi && make -j3 && /usr/bin/cmake -E touch /home/ko/base/thirdparty/fribidi/build/arm-linux-androideabi/fribidi-prefix/src/fribidi-stamp/fribidi-build`

On an android build, it could be it/Cmake has created another subdir than `arm-linux-androideabi`. Do you see some other directory name around there? Like `x86_64-linux-gnu-debug` which could confirm CMake is a bit confused about the arch, and might use the native ld?

Frenzie:
@poire-z I think that's just CMake 3.5 giving off a confusing error message. Looks something like this on slightly more modern CMake:

```
[5/7] Performing build step for 'fribidi'
FAILED: fribidi-prefix/src/fribidi-stamp/fribidi-build 
cd /home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi && sh -c "make V=1 CC=\"i686-linux-android-gcc -static-libstdc++\" CCLD=\"i686-linux-android-gcc -static-libstdc++\" CFLAGS=\"-Og -g -pipe -fno-omit-frame-pointer  -std=gnu11  --sysroot /home/frans/src/kobo/koreader/base/toolchain/android-ndk-r15c/platforms/android-14/arch-x86 -static-libstdc++ -fPIC\" LDFLAGS=\"-Wl,-O1 -Wl,--as-needed -static-libstdc++ -no-canonical-prefixes -Wl,--fix-cortex-a8 -static-libstdc++\" LD=\"i686-linux-android-ld\" AR=\"\" RANLIB=\"\"" && /usr/bin/cmake -E touch /home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi-stamp/fribidi-build
make[2]: Entering directory `/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi'
(cd lib && make  fribidi-unicode-version.h)
make[3]: Entering directory `/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi/lib'
make[4]: Entering directory `/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi/gen.tab'
i686-linux-android-gcc -static-libstdc++ -DHAVE_CONFIG_H -I. -I..  -I../lib -I../lib   -Og -g -pipe -fno-omit-frame-pointer  -std=gnu11  --sysroot /home/frans/src/kobo/koreader/base/toolchain/android-ndk-r15c/platforms/android-14/arch-x86 -static-libstdc++ -fPIC -MT gen-unicode-version.o -MD -MP -MF .deps/gen-unicode-version.Tpo -c -o gen-unicode-version.o gen-unicode-version.c
mv -f .deps/gen-unicode-version.Tpo .deps/gen-unicode-version.Po
/bin/bash ../libtool  --tag=CC   --mode=link i686-linux-android-gcc -static-libstdc++  -Og -g -pipe -fno-omit-frame-pointer  -std=gnu11  --sysroot /home/frans/src/kobo/koreader/base/toolchain/android-ndk-r15c/platforms/android-14/arch-x86 -static-libstdc++ -fPIC  -Wl,-O1 -Wl,--as-needed -static-libstdc++ -no-canonical-prefixes -Wl,--fix-cortex-a8 -static-libstdc++ -o gen-unicode-version gen-unicode-version.o  
libtool: link: i686-linux-android-gcc -static-libstdc++ -Og -g -pipe -fno-omit-frame-pointer -std=gnu11 --sysroot /home/frans/src/kobo/koreader/base/toolchain/android-ndk-r15c/platforms/android-14/arch-x86 -static-libstdc++ -fPIC -Wl,-O1 -Wl,--as-needed -static-libstdc++ -no-canonical-prefixes -Wl,--fix-cortex-a8 -static-libstdc++ -o gen-unicode-version gen-unicode-version.o 
./gen-unicode-version unidata/ReadMe.txt unidata/BidiMirroring.txt gen-unicode-version > fribidi-unicode-version.h \
	|| (rm -f fribidi-unicode-version.h && false)
/bin/bash: ./gen-unicode-version: No such file or directory
make[4]: *** [fribidi-unicode-version.h] Error 1
make[4]: Leaving directory `/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi/gen.tab'
make[3]: *** [fribidi-unicode-version.h] Error 2
make[3]: Leaving directory `/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi/lib'
make[2]: *** [lib/fribidi-unicode-version.h] Error 2
make[2]: Leaving directory `/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi'
make[1]: *** [build/i686-linux-android-debug/libs/libfribidi.so.0] Error 1
make[1]: Leaving directory `/home/frans/src/kobo/koreader/base'
make: *** [Makefile:69: all] Error 2
```

Frenzie:
There's something horribly wrong with the `gen-unicode-version` executable.

Regular build:
```
frans@frenzie-desktop:~/src/kobo/koreader/base/thirdparty/fribidi/build/x86_64-linux-gnu-debug/fribidi-prefix/src/fribidi/gen.tab$ ./gen-unicode-version 
gen-unicode-version: usage:
  gen-unicode-version /path/to/a/UCD/file [/path/to/more/UCD/files ...]
```

Current Android build:
```
frans@frenzie-desktop:~/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi/gen.tab$ ./gen-unicode-version 
bash: ./gen-unicode-version: No such file or directory
```

NiLuJe:
Well, especially given what's in it:

```
/* fribidi-unicode-version.h
 * generated by gen-unicode-version (GNU FriBidi 1.0.5)
 * from the file ReadMe.txt */

#define FRIBIDI_UNICODE_VERSION "11.0.0"
#define FRIBIDI_UNICODE_MAJOR_VERSION 11
#define FRIBIDI_UNICODE_MINOR_VERSION 0
#define FRIBIDI_UNICODE_MICRO_VERSION 0

/* End of generated fribidi-unicode-version.h */
```

I'd switch to a release tarball, which hopefully should ship this pre-generated?

Because dealing with this kind of stuff when x-compiling is extremely unfun...

poire-z:
For testing, you might try to build with their latest commit, instead of the year old v1.0.5.
There might be one issue related to something_unicode_version `https://github.com/fribidi/fribidi/pull/110#issuecomment-495896952`

Frenzie:
@poire-z Way ahead of you. ;-D This is the messy diff of my current experiment:

```diff
diff --git a/Makefile.third b/Makefile.third
index 5ede639..a02712c 100644
--- a/Makefile.third
+++ b/Makefile.third
@@ -45,8 +45,8 @@ $(UTF8PROC_LIB) $(UTF8PROC_DIR): $(THIRDPARTY_DIR)/utf8proc/CMakeLists.txt
 $(FRIBIDI_LIB) $(FRIBIDI_DIR): $(THIRDPARTY_DIR)/fribidi/CMakeLists.txt
        install -d $(FRIBIDI_BUILD_DIR)
        cd $(FRIBIDI_BUILD_DIR) && \
-               $(CMAKE) $(CMAKE_FLAGS) -DCC="$(CC)" -DCFLAGS="$(CFLAGS)"\
-               -DLDFLAGS="$(LDFLAGS)" -DCHOST=$(CHOST) \
+               $(CMAKE) $(CMAKE_FLAGS) -DCC="$(CC)" -DCXX="$(CXX)" -DCFLAGS="$(CFLAGS)" -DCXXFLAGS="$(CXXFLAGS)" -DLD="$(LD)"\
+               -DLDFLAGS="$(LDFLAGS)" -DCHOST=$(CHOST) -DCC_FOR_BUILD=$(HOST_CC) \
                $(CURDIR)/$(THIRDPARTY_DIR)/fribidi && \
                $(CMAKE_MAKE_PROGRAM) $(CMAKE_MAKE_PROGRAM_FLAGS)
        cp -fL $(FRIBIDI_DIR)/lib/$(notdir $(FRIBIDI_LIB)) $@
diff --git a/thirdparty/fribidi/CMakeLists.txt b/thirdparty/fribidi/CMakeLists.txt
index b1c0af7..ddb10dc 100644
--- a/thirdparty/fribidi/CMakeLists.txt
+++ b/thirdparty/fribidi/CMakeLists.txt
@@ -1,7 +1,7 @@
 project(fribidi)
 cmake_minimum_required(VERSION 3.5.1)
 
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_LIST_DIR}/../cmake_modules")
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/../cmake_modules")
 include("koreader_thirdparty_common")
 include("koreader_thirdparty_git")
 
@@ -13,7 +13,7 @@ assert_var_defined(LDFLAGS)
 ep_get_source_dir(SOURCE_DIR)
 ep_get_binary_dir(BINARY_DIR)
 
-set(CFG_ENV_VAR "CC=\"${CC}\" CXX=\"${CXX}\" CPPFLAGS=\"${CPPFLAGS}\" LDFLAGS=\"${LDFLAGS}\" ")
+set(CFG_ENV_VAR "CC=\"${CC}\" CC_FOR_BUILD=\"${CC_FOR_BUILD}\" CCLD=\"${CC}\" CXX=\"${CXX}\" CFLAGS=\"${CFLAGS}\" LDFLAGS=\"${LDFLAGS}\" ")
 set(CFG_OPTS "-q --prefix=${BINARY_DIR} --disable-static --enable-shared --host=\"${CHOST}\"")
 set(CFG_CMD sh -c "${CFG_ENV_VAR} ${SOURCE_DIR}/configure ${CFG_OPTS}")
 
@@ -28,10 +28,12 @@ if($ENV{ANDROID})
     set(CFG_CMD "${CFG_CMD} && ${ISED} 's|soname_spec=.*|soname_spec=\"\\\\\$libname\\\\\$release\\\\\$shared_ext\\\\\$major\"|' libtool")
 endif()
 
+set(MAKE_CMD sh -c "${KO_MAKE_RECURSIVE} V=1 CC=\"${CC}\" CC_FOR_BUILD=\"${CC_FOR_BUILD}\" CCLD=\"${CC}\" CFLAGS=\"${CFLAGS}\" LDFLAGS=\"${LDFLAGS}\" LD=\"${LD}\" AR=\"${AR}\" RANLIB=\"${RANLIB}\"")
+
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME
     https://github.com/fribidi/fribidi.git
-    tags/v1.0.5
+    0f849e344d446934b4ecdbe9edc32abd29029731
     ${SOURCE_DIR}
 )
 
@@ -42,6 +44,7 @@ ExternalProject_Add(
     PATCH_COMMAND NOCONFIGURE=1 ./autogen.sh
     BUILD_IN_SOURCE 1
     CONFIGURE_COMMAND COMMAND ${CFG_CMD}
-    BUILD_COMMAND ${KO_MAKE_RECURSIVE} -j${PARALLEL_JOBS}
+    #BUILD_COMMAND ${KO_MAKE_RECURSIVE} -j${PARALLEL_JOBS} CC=${CC} LDFLAGS="${LDFLAGS}"  CCLD=${CC}
+    BUILD_COMMAND ${MAKE_CMD}
     INSTALL_COMMAND ${KO_MAKE_RECURSIVE} -j${PARALLEL_JOBS} install
 )
```

Frenzie:
@NiLuJe I think they may have deprecated release tarballs?

Frenzie:
Minor sidenote, CC_FOR_BUILD seems to be set intelligently by Autotools. In the end I guess that means that `--fix-cortex-a8` is the problem after all.

Frenzie:
Yup, removing that takes care of the problem. We've got a working Android binary:
```
frans@frenzie-desktop:~/src/kobo/koreader/base/thirdparty/fribidi/build/i686-linux-android-debug/fribidi-prefix/src/fribidi-build/bin$ readelf -a fribidi 
ELF Header:
  Magic:   7f 45 4c 46 01 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF32
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              EXEC (Executable file)
  Machine:                           Intel 80386
  Version:                           0x1
  Entry point address:               0x8048b80
  Start of program headers:          52 (bytes into file)
  Start of section headers:          39952 (bytes into file)
  Flags:                             0x0
  Size of this header:               52 (bytes)
  Size of program headers:           32 (bytes)
  Number of program headers:         9
  Size of section headers:           40 (bytes)
  Number of section headers:         35
  Section header string table index: 34

[…]

Displaying notes found in: .note.gnu.gold-version
  Owner                 Data size	Description
  GNU                  0x00000009	NT_GNU_GOLD_VERSION (gold version)
    Version: gold 1.11
```

Now the only question that remains is how to use two different sets of CFLAGS, because even if `--fix-cortex-a8` can be deprecated we'd still get a major hiccup on `-march=armv7-a`.

Frenzie:
PS I should probably point out I know it's a proper Android binary due to the machine. Here's the desktop one with `Advanced Micro Devices X86-64`.

```
frans@frenzie-desktop:~/src/kobo/koreader/base/thirdparty/fribidi/build/x86_64-linux-gnu-debug/fribidi-prefix/src/fribidi-build/bin$ readelf -a fribidi 
ELF Header:
  Magic:   7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF64
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           Advanced Micro Devices X86-64
  Version:                           0x1
  Entry point address:               0x21d0
  Start of program headers:          64 (bytes into file)
  Start of section headers:          47096 (bytes into file)
  Flags:                             0x0
  Size of this header:               64 (bytes)
  Size of program headers:           56 (bytes)
  Number of program headers:         11
  Size of section headers:           64 (bytes)
  Number of section headers:         37
  Section header string table index: 36
```

poire-z:
There are a few notes in `gen.tab/meson.build meson.build m4/m4_ax_prog_cc_for_build.m4 TODO` which look like they say that gen.tab stuff should be compiled with the host compiler and not the cross compiler - a bit like luajit ?
So, may be we should just patch some files in gen.tab with hardcoded stuff, if they made that work only with meson and forgot about the otherbuild  tools.
(But why does it work for kobo then?)

Frenzie:
Yes, that's what CC_FOR_BUILD, CFLAGS_FOR_BUILD and LDFLAGS_FOR_BUILD are for afaict.

It works for Kobo because there's no weird stuff in CFLAGS/LDFLAGS. So I guess we can make it work by getting rid of that particular flag (if necessary even just here) but that doesn't seem right at all. ;-)

Frenzie:
It seems to me that the Makefile actually isn't using the correct LDFLAGS?

```
CCLD = $(CC)
LINK = $(LIBTOOL) $(AM_V_lt) --tag=CC $(AM_LIBTOOLFLAGS) \
	$(LIBTOOLFLAGS) --mode=link $(CCLD) $(AM_CFLAGS) $(CFLAGS) \
	$(AM_LDFLAGS) $(LDFLAGS) -o $@
AM_V_CCLD = $(am__v_CCLD_$(V))
am__v_CCLD_ = $(am__v_CCLD_$(AM_DEFAULT_VERBOSITY))
am__v_CCLD_0 = @echo "  CCLD    " $@;
am__v_CCLD_1 = 
```

```
LDFLAGS = -Wl,-O1 -Wl,--as-needed -static-libstdc++ -no-canonical-prefixes -Wl,--fix-cortex-a8 -static-libstdc++
LDFLAGS_FOR_BUILD = -Og -g -pipe -fno-omit-frame-pointer
```

Which would be traced over here: https://github.com/fribidi/fribidi/blob/0f849e344d446934b4ecdbe9edc32abd29029731/gen.tab/Makefile.am#L26-L29

poire-z:
(Dunno how much we will wait for upstream merge, so just had a try patching here ... spent too much time with sed and trying to include `$(...)` and a `\n` and gave up, so here LDFLAGS is set to be empty.)

Frenzie:
@poire-z You could also include a simple patch file (but this solution is fine).

Btw, the CMakeLists.txt still needs some touch-ups. It's not passing through the CFLAGS, there's not enough checking if vars are defined (which would show that no CPPFLAGS are passed through, assuming that's even necessary), and some minor stuff like that. :-)

poire-z:
> and some minor stuff like that

Please make suggestions! that's not my natural language and culture at all :)
For now, I'd add that:
```diff
--- a/thirdparty/fribidi/CMakeLists.txt
+++ b/thirdparty/fribidi/CMakeLists.txt
@@ -8,6 +8,7 @@ include("koreader_thirdparty_git")
 enable_language(C)

 assert_var_defined(CC)
+assert_var_defined(CFLAGS)
 assert_var_defined(LDFLAGS)

 ep_get_source_dir(SOURCE_DIR)
@@ -16,7 +17,7 @@ ep_get_binary_dir(BINARY_DIR)
 # Temporary, while waiting for https://github.com/fribidi/fribidi/pull/115 to be merged
 set(PATCH_CMD sh -c "${ISED} \"s|^CFLAGS = |LDFLAGS = \\nCFLAGS = |\"  gen.tab/Makefile.am")

-set(CFG_ENV_VAR "CC=\"${CC}\" CXX=\"${CXX}\" CPPFLAGS=\"${CPPFLAGS}\" LDFLAGS=\"${LDFLAGS}\" ")
+set(CFG_ENV_VAR "CC=\"${CC}\" CFLAGS=\"${CFLAGS}\" LDFLAGS=\"${LDFLAGS}\" ")
 set(CFG_OPTS "-q --prefix=${BINARY_DIR} --disable-static --enable-shared --host=\"${CHOST}\"")
 set(CFG_CMD sh -c "${CFG_ENV_VAR} ${SOURCE_DIR}/configure ${CFG_OPTS}")

```

thirdparty/fribidi/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+cmake_minimum_required(VERSION 3.5.1)
+
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/../cmake_modules")
+include("koreader_thirdparty_common")
+include("koreader_thirdparty_git")
+
+enable_language(C)
+
+assert_var_defined(CC)
+assert_var_defined(LDFLAGS)
```
Frenzie: Should also check for CHOST.

Frenzie:
@poire-z That plus checking for CHOST should do the trick I think.

The `assert_var_defined` stuff is mostly to double-check ourselves when changing stuff in Makefile.defs and/or Makefile.third.

PR #983, opened 2019-09-26
Unify open flags for ioctl-only fds

NiLuJe:
Shouldn't have any visible effect, except for no longer leaking fds on
exec ;).

(c.f., man 3 open for the rationale behind this set of flags for ioctl handles).


NiLuJe:
No need to worry about CLOEXEC being unsupported on some platforms, this code is platform specific, and the only platform where CLOEXEC isn't implemented in a subset of kindle-legacy (FW 2.x).

PR #995, opened 2019-10-19
Use LIB_SUFFIX to fix building on opensuse

samcv:
[Why]
opensuse uses lib64 instead of lib. This causes the build to fail even
when using the docker images.

[How]
In Makefile.third use LIB_SUFFIX after `lib` so that the directory will
point to the correct place on opensuse or on other distros.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/995)

### DIFF
--- a/Makefile.third
+++ b/Makefile.third
@@ -20,7 +20,7 @@ $(FREETYPE_LIB) $(FREETYPE_DIR)/include: $(THIRDPARTY_DIR)/freetype2/CMakeLists.
 		-DCXXFLAGS="$(CXXFLAGS)" -DLDFLAGS="$(LDFLAGS)" -DCHOST=$(CHOST) \
 		$(CURDIR)/$(THIRDPARTY_DIR)/freetype2 && \
 		$(CMAKE_MAKE_PROGRAM) $(CMAKE_MAKE_PROGRAM_FLAGS)
-	cp -fL $(FREETYPE_DIR)/$(if $(WIN32),bin,lib)/$(notdir $(FREETYPE_LIB)) $@
+	cp -fL $(FREETYPE_DIR)/$(if $(WIN32),bin,lib$(LIB_SUFFIX))/$(notdir $(FREETYPE_LIB)) $@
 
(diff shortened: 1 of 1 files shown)


NiLuJe:
I assume it's distro-provided CMake overrides setting `LIB_SUFFIX`?

From a (very) quick glance, that seems to be an accepted workaround unless stuff is using `GNUInstallDirs` (in which case it could potentially backfire?).

samcv:
My mistake, this also needs `make LIB_SUFFIX=$(LIB_SUFFIX)` in `Makefile` when calling `Makefile.third` and then you have to set LIB_SUFFIX=64 in the environment before calling Makefile.

NiLuJe:
Okay, *now* that makes sense, thanks ;).

In which case, would you mind drafting a matching PR for the [docs](https://github.com/koreader/koreader/blob/master/doc/Building.md)?

NiLuJe:
Note that, ideally, I'd prefer taking the opposite approach: bludgeoning CMake into using plain `lib` everywhere.

(We're never going to merge/install anything into system-paths, so not honoring the distro's preferred libdir isn't an issue).

I vaguely recall attempting that somewhere (probably libjpeg-turbo), but CMake being CMake, I perfectly understand that it might be easier said than done ^^.

NiLuJe:
Yeah, the quick'n dirty approach for libjpeg-trubo looked like [that](https://github.com/koreader/koreader-base/blob/2700c5315d0307221a0c49c8bbb2508bdf9432f0/thirdparty/libjpeg-turbo/CMakeLists.txt#L19-L21).

benoit-pierre:
This can be closed, as #1750 includes giflib 5.2.2.

benoit-pierre:
I can compile #1750 fine in docker (image: [opensuse/leap](https://hub.docker.com/layers/opensuse/leap/latest/images/sha256-57958e44b5aa446b1466e06f6c74150e54ec5e44347a1242ef40d054b2ef8ebc?context=explore)).

PR #996, opened 2019-10-21
macOS build fixes

NiLuJe:
Possibly not enough for a full build, untested ;).


NiLuJe:
c.f., the last few days of discussions on gitter ;).

PR #999, opened 2019-10-26
Fix LuaJIT patchset on Darwin

NiLuJe:
strip arguments are target-sensitive, so enforcing GNU flags everywhere
was a bad call.


NiLuJe:
I thought I'd skipped that bit of the patch in the KOReader patchset -_-".

PR #1000, opened 2019-10-28
Darwin awards

NiLuJe:
Attempt to appease the insanity currently happening w/ XCode 11

Relies on obscure GNU make features to defeat parallelism issues, hopefully the CI won't implode.
I have a Mojave build running in // to double-check.

Said insanity being completely broken dependencies in dynamic libraries: insane random @rpath, or brew libraries pulled instead of our own, or just plain broken paths. There's also often a self-reference in the deps (why is it here in the first place?) that's also broke 72.4% of the time.

I'm not actually familiar with the intricacies of the Mach-O format (nor with the relevant tools to mangle it), so, here be dragons.


NiLuJe:
ONE THOUSAND§§!

(You couldn't have planned this one better.)

NiLuJe:
Note that, of course, this might, to a certain extent, depend on what exactly you happen to have installed via brew...

(Say, if you happened to have a matching tesseract or leptonica installed, something could potentially fuck those up, too).

NiLuJe:
Also kind of forgot to patch my macOS tree, so, one build in the wind :D.

NiLuJe:
Random remark: the latest SNAFU seems to confirm that the library's id is what gets used as a dependency in other libs that link to it, so fixing the id isn't as useless as I would have thought ;).

NiLuJe:
Okay, fresh build is OK on my macOS box, so this should be good to go if the CI is clear ;).

PR #1003, opened 2019-10-29
luarocks_dir compatibility with luarocks 3.x

yparitcher:
luarocks 3.x supports one luarocks install with multiple lua versions.
archlinux already uses 3.x, debian and friends still use 2.x

this will change the directory structure in the release (not sure if that is important)

as documented in:
https://github.com/luarocks/luarocks/blob/master/CHANGELOG.md

> Breaking change: The support for deprecated unversioned paths (e.g. /usr/local/lib/luarocks/rocks/ and /etc/luarocks/config.lua) was removed, LuaRocks will now only create and use paths versioned to the specific Lua version in use (e.g. /usr/local/lib/luarocks/rocks-5.3/ and /etc/luarocks/config-5.3.lua).

PR #1004, opened 2019-10-31
[fix] LuaRocks 2 and 3 compat

Frenzie:
Cf. <https://github.com/koreader/koreader-base/pull/1003#issuecomment-548260873>.


NiLuJe:
I'll get back to you as soon as I reinstall a bootloader, because Windows ate mine ;p.

poire-z:
(Works fine for me.)

PR #1008, opened 2019-11-11
[fix] Don't crash on SDL < 2.9

Frenzie:
Otherwise we'll get an `undefined symbol` error, but this function is too convenient to bother with the old-fashioned way of rumbling.

Fixes <https://github.com/koreader/koreader/issues/5570>.


Frenzie:
Drat, this doesn't actually work. The extracted AppImage picked up on my system's SDL 2.9. :man_facepalming:

PR #1010, opened 2019-11-14
Adds libkoreader-xtext.so: enhanced text shaping helper

poire-z:
New Lua C module to be used by frontend TextWidget and TextBoxWidget, as an alternative text layout method, using Harfbuzz, FriBiDi and libunibreak.
This is needed to correctly shape Arabic scripts (cursive glyphs), propertly draw RTL text (Hebrew,
Arabic) and bidi text (arabic/hebrew filenames in english UI).
Early description of how I imagined it in https://github.com/koreader/koreader/issues/5359#issuecomment-533211988.
Screenshots and discussion in https://github.com/koreader/koreader/issues/5359.
See individual commit message for (not much) details.

@NiLuJe:
I used ffi-cdecl for the added Freetype ffi stuff, so it should be fine :)
About libunibreak, I don't know if it needs any of the `ifdef DARWIN` you added to others recently (I don't really get why you do it for some and not for others).
I added some (cut and pasted from others) for libkoreader-xtext.so - dunno if that's ok.

Not sure about the load/hint flags to use (in both freetype.lua and xtext.cpp): I [somehow get](http://chanae.walon.org/pub/ttf/ttf_glyphs.htm) that we get integer rounded metrics (advance/offsets) only when we use hinting - and non rounded ones when not using hinting, like I do here. And I indeed did not get rounded metrics, and got some small glitches using `>>6` (like we do in crengine), so I had to do some real TRUNC(ROUND()) to get nice drawing. Is my understanding and what I'm doing ok?
And should I do that too in crengine?

If you're having a quick look at my C, you can skip the Harfbuzz text/glyphs iterations - it's quite the same as in crengine (with some substantial differences: we're not making words, and I don't need to copy chars in some temporary buffers - and some other small differences).
And you can skip the Lua stack magic, unless you're ready for some headaches :) It's quite some specific knowledge, although quite simpler and clever (and probably fast) once you dig into it.

Checked with cppcheck, only some "scope can be reduced" warnings, quite proud of myself! :)
Fixed some (inconsistend and delayed) segfaults or invalid "free" stuff crash, with the help of valgrind (which is my new friend, even if it is a super slow friend...), always me writting outside malloc'ed arrays in edge cases :| But it should be quite fine now.

Probably needs to be read with the frontend code. I'll push a PR tomorrow.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,6 +89,7 @@ libs: \
 	$(OUTPUT_DIR)/libs/libkoreader-lfs.so \
 	$(OUTPUT_DIR)/libs/libkoreader-djvu.so \
 	$(OUTPUT_DIR)/libs/libkoreader-cre.so \
+	$(OUTPUT_DIR)/libs/libkoreader-xtext.so \
 	$(OUTPUT_DIR)/libs/libwrap-mupdf.so
 
 $(OUTPUT_DIR)/libs/libkoreader-input.so: input/*.c input/*.h $(if $(KINDLE),$(POPEN_NOSHELL_LIB),)
@@ -147,6 +148,29 @@ ifdef DARWIN
 	install_name_tool -change \
 		`otool -L "$@" | grep "$(notdir $(CRENGINE_LIB)) " | awk '{print $$1}'` \
 		libs/$(notdir $(CRENGINE_LIB)) \
+		$@
+endif
+
+$(OUTPUT_DIR)/libs/libkoreader-xtext.so: xtext.cpp \
+			$(if $(USE_LUAJIT_LIB),$(LUAJIT_LIB),) \
+			$(FREETYPE_LIB) $(HARFBUZZ_LIB) $(FRIBIDI_LIB) $(LIBUNIBREAK_LIB)
+	$(CXX) -I$(FREETYPE_DIR)/include/freetype2 \
+	-I$(HARFBUZZ_DIR)/include/harfbuzz \
+	-I$(FRIBIDI_DIR)/include \
+	-I$(LIBUNIBREAK_DIR)/include \
+	$(DYNLIB_CXXFLAGS) -Wall -o $@ $^
+ifdef DARWIN
+	install_name_tool -change \
+		`otool -L "$@" | grep "libluajit" | awk '{print $$1}'` \
+		libs/$(notdir $(LUAJIT_LIB)) \
+		$@
+	install_name_tool -change \
+		`otool -L "$@" | grep "$(notdir $(HARFBUZZ_LIB)) " | awk '{print $$1}'` \
+		libs/$(notdir $(HARFBUZZ_LIB)) \
+		$@
+	install_name_tool -change \
+		`otool -L "$@" | grep "$(notdir $(FRIBIDI_LIB)) " | awk '{print $$1}'` \
+		libs/$(notdir $(FRIBIDI_LIB)) \
 		$@
 endif
 
--- a/Makefile.third
+++ b/Makefile.third
@@ -57,6 +57,16 @@ $(FRIBIDI_LIB) $(FRIBIDI_DIR): $(THIRDPARTY_DIR)/fribidi/CMakeLists.txt
 	cp -fL $(FRIBIDI_DIR)/lib/$(notdir $(FRIBIDI_LIB)) $@
 	chmod 755 $@
 
+$(LIBUNIBREAK_LIB) $(LIBUNIBREAK_DIR): $(THIRDPARTY_DIR)/libunibreak/CMakeLists.txt
+	install -d $(LIBUNIBREAK_BUILD_DIR)
+	cd $(LIBUNIBREAK_BUILD_DIR) && \
+		$(CMAKE) $(CMAKE_FLAGS) -DCC="$(CC)" -DCFLAGS="$(CFLAGS)"\
+		-DLDFLAGS="$(LDFLAGS)" -DCHOST=$(CHOST) \
+		$(CURDIR)/$(THIRDPARTY_DIR)/libunibreak && \
+		$(CMAKE_MAKE_PROGRAM) $(CMAKE_MAKE_PROGRAM_FLAGS)
+	cp -fL $(LIBUNIBREAK_DIR)/lib/$(notdir $(LIBUNIBREAK_LIB)) $@
+	chmod 755 $@
+
 # libjpeg-turbo and libjepg
 $(TURBOJPEG_LIB) $(JPEG_LIB): $(THIRDPARTY_DIR)/libjpeg-turbo/CMakeLists.txt
 	install -d $(JPEG_BUILD_DIR)
--- a/ffi/freetype_h.lua
+++ b/ffi/freetype_h.lua
@@ -2,6 +2,7 @@ local ffi = require("ffi")
 
 ffi.cdef[[
 typedef int FT_Error;
+typedef long int FT_Pos;
 struct FT_Generic_ {
   void *data;
   void (*finalizer)(void *);
@@ -19,18 +20,18 @@ struct FT_ListRec_ {
   FT_ListNode tail;
 };
 struct FT_BBox_ {
-  long int xMin;
-  long int yMin;
-  long int xMax;
-  long int yMax;
+  FT_Pos xMin;
+  FT_Pos yMin;
+  FT_Pos xMax;
+  FT_Pos yMax;
 };
 typedef struct FT_BBox_ FT_BBox;
 struct FT_Bitmap_Size_ {
   short int height;
   short int width;
-  long int size;
-  long int x_ppem;
-  long int y_ppem;
+  FT_Pos size;
+  FT_Pos x_ppem;
+  FT_Pos y_ppem;
 };
 typedef struct FT_Bitmap_Size_ FT_Bitmap_Size;
 struct FT_Bitmap_ {
@@ -45,8 +46,8 @@ struct FT_Bitmap_ {
 };
 typedef struct FT_Bitmap_ FT_Bitmap;
 struct FT_Vector_ {
-  long int x;
-  long int y;
+  FT_Pos x;
+  FT_Pos y;
 };
 typedef struct FT_Vector_ FT_Vector;
 struct FT_Outline_ {
@@ -96,14 +97,14 @@ typedef struct FT_LibraryRec_ *FT_Library;
 typedef struct FT_FaceRec_ *FT_Face;
 typedef struct FT_Glyph_Metrics_ FT_Glyph_Metrics;
 struct FT_Glyph_Metrics_ {
-  long int width;
-  long int height;
-  long int horiBearingX;
-  long int horiBearingY;
-  long int horiAdvance;
-  long int vertBearingX;
-  long int vertBearingY;
-  long int vertAdvance;
+  FT_Pos width;
+  FT_Pos height;
+  FT_Pos horiBearingX;
+  FT_Pos horiBearingY;
+  FT_Pos horiAdvance;
+  FT_Pos vertBearingX;
+  FT_Pos vertBearingY;
+  FT_Pos vertAdvance;
 };
 struct FT_SubGlyphRec_;
 typedef struct FT_SubGlyphRec_ *FT_SubGlyph;
@@ -129,8 +130,8 @@ struct FT_GlyphSlotRec_ {
   FT_SubGlyph subglyphs;
   void *control_data;
   long int control_len;
-  long int lsb_delta;
-  long int rsb_delta;
+  FT_Pos lsb_delta;
+  FT_Pos rsb_delta;
   void *other;
   FT_Slot_Internal internal;
 };
@@ -139,10 +140,10 @@ struct FT_Size_Metrics_ {
   short unsigned int y_ppem;
   long int x_scale;
   long int y_scale;
-  long int ascender;
-  long int descender;
-  long int height;
-  long int max_advance;
+  FT_Pos ascender;
+  FT_Pos descender;
+  FT_Pos height;
+  FT_Pos max_advance;
 };
 typedef struct FT_Size_Metrics_ FT_Size_Metrics;
 typedef struct FT_Size_InternalRec_ *FT_Size_Internal;
@@ -230,7 +231,26 @@ FT_Error FT_Load_Char(FT_Face, long unsigned int, int);
 FT_Error FT_Get_Kerning(FT_Face, unsigned int, unsigned int, unsigned int, FT_Vector *);
 void FT_GlyphSlot_Embolden(FT_GlyphSlot);
 void FT_GlyphSlot_Oblique(FT_GlyphSlot);
+enum FT_Render_Mode_ {
+  FT_RENDER_MODE_NORMAL = 0,
+  FT_RENDER_MODE_LIGHT = 1,
+  FT_RENDER_MODE_MONO = 2,
+  FT_RENDER_MODE_LCD = 3,
+  FT_RENDER_MODE_LCD_V = 4,
+  FT_RENDER_MODE_MAX = 5,
+};
+typedef enum FT_Render_Mode_ FT_Render_Mode;
+FT_Error FT_Load_Glyph(FT_Face, unsigned int, int);
+FT_Error FT_Render_Glyph(FT_GlyphSlot, FT_Render_Mode);
+FT_Error FT_Outline_Embolden(FT_Outline *, FT_Pos);
+void FT_Outline_Translate(const FT_Outline *, FT_Pos, FT_Pos);
+long int FT_MulFix(long int, long int);
 static const int FT_LOAD_RENDER = 4;
+static const int FT_LOAD_DEFAULT = 0;
+static const int FT_LOAD_TARGET_LIGHT = 65536;
+static const int FT_LOAD_NO_AUTOHINT = 32768;
+static const int FT_LOAD_NO_HINTING = 2;
+static const int FT_LOAD_FORCE_AUTOHINT = 32;
 static const int FT_FACE_FLAG_KERNING = 64;
 static const int FT_KERNING_DEFAULT = 0;
 ]]
--- a/ffi/pic.lua
+++ b/ffi/pic.lua
@@ -272,6 +272,9 @@ function Pic.openPNGDocument(filename)
     elseif re.ncomp == 3 then bbtype = BB.TYPE_BBRGB24
     elseif re.ncomp == 4 then bbtype = BB.TYPE_BBRGB32
     else
+        if re.data then
+            ffi.C.free(re.data)
+        end
         error("unsupported number of color components")
     end
 
--- a/thirdparty/libunibreak/CMakeLists.txt
+++ b/thirdparty/libunibreak/CMakeLists.txt
@@ -1,0 +1,45 @@
+project(libunibreak)
+cmake_minimum_required(VERSION 3.5.1)
+
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/../cmake_modules")
+include("koreader_thirdparty_common")
+include("koreader_thirdparty_git")
+
+enable_language(C)
+
+assert_var_defined(CC)
+assert_var_defined(CFLAGS)
+assert_var_defined(CHOST)
+assert_var_defined(LDFLAGS)
+
+ep_get_source_dir(SOURCE_DIR)
+ep_get_binary_dir(BINARY_DIR)
+
+set(CFG_ENV_VAR "CC=\"${CC}\" CFLAGS=\"${CFLAGS}\" LDFLAGS=\"${LDFLAGS}\" ")
+set(CFG_OPTS "--prefix=${BINARY_DIR} --disable-static --enable-shared --host=\"${CHOST}\"")
+set(CFG_CMD sh -c "${CFG_ENV_VAR} ${SOURCE_DIR}/configure ${CFG_OPTS}")
+
+if($ENV{ANDROID})
+    set(CFG_CMD "${CFG_CMD} && ${ISED} 's|version_type=none|version_type=linux|' libtool")
+    set(CFG_CMD "${CFG_CMD} && ${ISED} 's|need_lib_prefix=no|need_lib_prefix=yes|' libtool")
+    set(CFG_CMD "${CFG_CMD} && ${ISED} 's|need_version=no|need_version=yes|' libtool")
+    set(CFG_CMD "${CFG_CMD} && ${ISED} 's|library_names_spec=.*|library_names_spec=\"\\\\\$libname\\\\\$release\\\\\$shared_ext\\\\\$versuffix \\\\\$libname\\\\\$release\\\\\$shared_ext\\\\\$major \\\\\$libname\\\\\$shared_ext\"|' libtool")
+    set(CFG_CMD "${CFG_CMD} && ${ISED} 's|soname_spec=.*|soname_spec=\"\\\\\$libname\\\\\$release\\\\\$shared_ext\\\\\$major\"|' libtool")
+endif()
+
+ko_write_gitclone_script(
+    GIT_CLONE_SCRIPT_FILENAME
+    https://github.com/adah1972/libunibreak.git
+    tags/libunibreak_4_2-1
+    ${SOURCE_DIR}
+)
+
+include(ExternalProject)
+ExternalProject_Add(
+    ${PROJECT_NAME}
+    DOWNLOAD_COMMAND ${CMAKE_COMMAND} -P ${GIT_CLONE_SCRIPT_FILENAME}
+    PATCH_COMMAND NOCONFIGURE=1 ./autogen.sh
+    CONFIGURE_COMMAND ${CFG_CMD}
+    BUILD_COMMAND ${KO_MAKE_RECURSIVE} -j${PARALLEL_JOBS}
+    INSTALL_COMMAND ${KO_MAKE_RECURSIVE} -j${PARALLEL_JOBS} install
+)
--- a/xtext.cpp
+++ b/xtext.cpp
@@ -1,0 +1,2186 @@
+// xtext.cpp
+// Lua interface to wrap a utf8 string into a XText object
+// that provides various text shaping and layout methods
+// with the help of Fribidi, Harfbuzz and libunibreak.
+
+// We do many things similarly to how they are done in crengine,
+// and took and adapted much code from it.
+// For many links and notes about the concepts and libraries used,
+// see: https://github.com/koreader/crengine/issues/307
+
+extern "C"
+{
+#include <lua.h>
+#include <lauxlib.h>
+#include <lualib.h>
+#include "xtext.h"
+}
+
+// Harfbuzz
+#include <hb.h>
+#include <hb-ft.h>
+
+// FriBiDi
+#include <fribidi/fribidi.h>
+
+// libunibreak
+#include <wordbreak.h>
+#include <linebreak.h>
+    // linebreakdef.h is not wrapped by this, unlike linebreak.h
+    // (not wrapping it results in "undefined symbol" with the
+    // original function name kinda obfuscated)
+    #ifdef __cplusplus
+    extern "C" {
+    #endif
+#include <linebreakdef.h>
+    #ifdef __cplusplus
+    }
+    #endif
+
+// luajit doesn't provide lua_set/getuservalue (unless compiled
+// with LUAJIT_ENABLE_LUA52COMPAT) but these are equivalents
+// for our purpose
+#if LUA_VERSION_NUM < 502
+#define lua_setuservalue lua_setfenv
+#define lua_getuservalue lua_getfenv
+#endif
+
+// Some names, as they should be known to Lua
+#define XTEXT_LIBNAME "xtext"
+#define XTEXT_METATABLE_NAME "luaL_XText"
+#define XTEXT_HB_FONT_DATA_METATABLE_NAME "luaL_XText_HB_Font_Data"
+#define XTEXT_LUA_HB_FONT_DATA_TABLE_KEY_NAME "_hb_font_data"
+#define XTEXT_LUA_FONT_GETFONT_CALLBACK_NAME "getFallbackFont"
+
+// Max unicode chars per shaped (visual) line
+#define MAX_LINE_CHARS 4096
+// Max returned glyphs per line (usually less glyphs than chars,
+// but allow for more just in case some fonts don't come with many
+// glyphs and combine many diacritics to form a unicode char).
+#define MAX_LINE_GLYPHS 2*MAX_LINE_CHARS
+
+// Max number of fonts (main + fallbacks)
+// (main + 15 fallback fonts should be enough)
+#define MAX_FONT_NUM 16
+
+#define NOT_MEASURED INT_MIN
+#define REPLACEMENT_CHAR 0xFFFD
+#define ELLIPSIS_CHAR 0x2026
+
+// Helpers with font metrics (units are 1/64 px)
+#define FONT_METRIC_FLOOR(x)    ((x) & -64)
+#define FONT_METRIC_CEIL(x)     (((x)+63) & -64)
+#define FONT_METRIC_ROUND(x)    (((x)+32) & -64)
+#define FONT_METRIC_TRUNC(x)    ((x) >> 6)
+
+// Uncomment for debugging text measurement and line shaping:
+// #define DEBUG_MEASURE_TEXT
+// #define DEBUG_SHAPE_LINE
+
+// ==============================================
+// Utility functions
+
+inline bool is_unicodepoint_rtl(uint32_t c) {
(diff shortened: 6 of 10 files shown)


Makefile
```diff
@@ -150,6 +151,21 @@ ifdef DARWIN
...
 
+$(OUTPUT_DIR)/libs/libkoreader-xtext.so: xtext.cpp \
+			$(if $(USE_LUAJIT_LIB),$(LUAJIT_LIB),) \
+			$(FREETYPE_LIB) $(HARFBUZZ_LIB) $(FRIBIDI_LIB) $(LIBUNIBREAK_LIB)
+	$(CXX) -I$(FREETYPE_DIR)/include/freetype2 \
+	-I$(HARFBUZZ_DIR)/include/harfbuzz \
+	-I$(FRIBIDI_DIR)/include \
+	-I$(LIBUNIBREAK_DIR)/include \
+	$(DYNLIB_CXXFLAGS) -Wall -o $@ $^
+ifdef DARWIN
```
NiLuJe: Yep, those are probably a safe bet here ;).

NiLuJe: It'll probably need the same luajit one as our other bindings, though.

Makefile.third
```diff
@@ -57,6 +57,16 @@ $(FRIBIDI_LIB) $(FRIBIDI_DIR): $(THIRDPARTY_DIR)/fribidi/CMakeLists.txt
...
+$(LIBUNIBREAK_LIB) $(LIBUNIBREAK_DIR): $(THIRDPARTY_DIR)/libunibreak/CMakeLists.txt
+	install -d $(LIBUNIBREAK_BUILD_DIR)
+	cd $(LIBUNIBREAK_BUILD_DIR) && \
+		$(CMAKE) $(CMAKE_FLAGS) -DCC="$(CC)" -DCFLAGS="$(CFLAGS)"\
+		-DLDFLAGS="$(LDFLAGS)" -DCHOST=$(CHOST) \
+		$(CURDIR)/$(THIRDPARTY_DIR)/libunibreak && \
+		$(CMAKE_MAKE_PROGRAM) $(CMAKE_MAKE_PROGRAM_FLAGS)
+	cp -fL $(LIBUNIBREAK_DIR)/lib/$(notdir $(LIBUNIBREAK_LIB)) $@
+	chmod 755 $@
+
```
NiLuJe: Time will tell if XCode does funky shit with that one on the DARWIN front, so don't bother for now ;).

ffi/freetype_h.lua
```diff
@@ -2,6 +2,7 @@ local ffi = require("ffi")
 
 ffi.cdef[[
 typedef int FT_Error;
+typedef long int FT_Pos;
```
NiLuJe: Did ffi-cdecl handle everything okay, or did you have to manually post-process some stuff here?

ffi/pic.lua
```diff
@@ -272,6 +272,9 @@ function Pic.openPNGDocument(filename)
     elseif re.ncomp == 3 then bbtype = BB.TYPE_BBRGB24
     elseif re.ncomp == 4 then bbtype = BB.TYPE_BBRGB32
     else
+        if re.data then
```
NiLuJe: Probably a good idea, as unlikely as this should be ;).

thirdparty/libunibreak/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+assert_var_defined(CC)
+assert_var_defined(CFLAGS)
+assert_var_defined(CHOST)
+assert_var_defined(LDFLAGS)
+
+ep_get_source_dir(SOURCE_DIR)
+ep_get_binary_dir(BINARY_DIR)
+
+set(CFG_ENV_VAR "CC=\"${CC}\" CFLAGS=\"${CFLAGS}\" LDFLAGS=\"${LDFLAGS}\" ")
+set(CFG_OPTS "-q --prefix=${BINARY_DIR} --disable-static --enable-shared --host=\"${CHOST}\"")
```
NiLuJe: Hi, me and my crusade against quiet stuff would like to get rid of the quiet flag ;p.

NiLuJe:
Stupid question is stupid: does XText actually need to be C++? Neither of the dependencies are C++, and at an initial quick glance, I didn't see any C++ constructs in there ;).

(Plain C > C++, because no STD to resolve/load).

ffi/freetype_h.lua
poire-z: It did. No stdout/stderr output when running it that way (with full paths stripped out):
```
CPPFLAGS=
CPPFLAGS+=-I./base/thirdparty/freetype2/build/git_checkout/freetype2/include
CPPFLAGS+=' '
CPPFLAGS+=-I./base/thirdparty/harfbuzz/build/git_checkout/harfbuzz/src
CPPFLAGS+=' '
CPPFLAGS+=-I./base/thirdparty/fribidi/build/git_checkout/fribidi/lib
export CPPFLAGS
./ffi-cdecl/ffi-cdecl gcc ./base/ffi-cdecl/freetype2_decl.c ./base/ffi/freetype_h.lua
```
and everything worked with the results.
(May be uneeded `-I` stuff, I had some experiments with these other libs before going with a Lua module.)

xtext.cpp
```diff
@@ -0,0 +1,2157 @@
...
+// Max returned glyphs per line (usually less glyphs than chars,
+// but allow for more just in case some fonts don't come with many
+// glyphs and combine many diacritics to form a unicode char).
+#define MAX_LINE_GLYPHS 2*MAX_LINE_CHARS
+
+// Max number of fonts (main + fallbacks)
+// (main + 15 fallback fonts should be enough)
+#define MAX_FONT_NUM 16
+
+#define NOT_MEASURED INT_MIN
```
NiLuJe: INT32_MIN?

xtext.cpp
```diff
@@ -0,0 +1,2157 @@
...
+// - then with a non-null dst (malloc'ed to the previously obtained size)
+//   to decode and fill it with the Unicode chars.
+
+// adapted from crengine/src/lvstring.cpp
+#define HEAD_CHECK(mask, expect) ((s[0] & mask) == expect)
+#define HEAD_BYTE(mask, shift) (((uint32_t)(s[0]) & mask) << shift)
+#define CONT_BYTE(index, shift) (((uint32_t)(s[index]) & 0x3F) << shift)
+#define HAS_FOLLOWUP(n) (s+n < ends)
+#define IS_FOLLOWING(index) ((s[index] & 0xC0) == 0x80)
+int Utf8ToUnicode(const char * src,  int srclen, uint32_t * dst, int dstlen, bool &is_valid, bool &has_rtl)
```
NiLuJe: I'd much rather use a fast & accurate [UTF-8 decoder](https://github.com/NiLuJe/FBInk/blob/master/cutef8/dfa.c) (both here and in CRe), but that would obviously throw WTF-8 out the window ;).

poire-z:
> Stupid question is stupid: does XText actually need to be C++?

Well, stupid answer: I dunno :) I use some `class` XText, so I figure it needs C++. I initially played with std:vector stuff, and ended up not needing it. But as I'm just used to C++ in crengine, may be there's stuff that are specific to C++, and would need some rewrite/rethink if it had to be done in C. I really don't know :)
Dunno if it's as simple as replacing it with a `struct XText`, but well... not wishing to rework the whole thing :) But it you feel like it can be quickly made to C, feel free to have a quick try at compiling it :)

xtext.cpp
```diff
@@ -0,0 +1,2157 @@
...
+
+// Info, after measure(), about each m_text char
+typedef struct {
+    unsigned short flags;
+    signed short   width;
+} xtext_charinfo_t;
+
+// Glyph info when shaping a line (to be returned to Lua as a table of tables)
+// (16 bytes, making our static s_shape_result[MAX_LINE_GLYPHS] a 128Kb buffer)
+typedef struct {
```
NiLuJe: Random comment about packing rules if you want to micro-optimize memory usage.

Ordering the struct by data type's width (i.e., wider to narrower, or vice-versa) usually works well enough.

May hurt readability (and/or actual performance, depending on what's used, how and when).

C is fun :D.

xtext.cpp
poire-z: Is it different than INT_MIN ? Or it just can be, on some platforms?

xtext.cpp
NiLuJe: And I just noticed the bitfields, which will by definition cause padding anyway ^^.

NiLuJe: Speaking of the bitfields, any specific reason for the weird 1 + 1 + explicit padding instead of using two bools?

xtext.cpp
poire-z: I thought about asking you to provide me one :) but did not want to get stuck too long with discussion about it :)
But indeed, WTF-8 + a way to plug in RTL char detection is a requirement. And 3 invalid bytes > 3 replacement chars, and non failure (I may have some latin1 in TextEditor :) or invalid stuff from translators) and don't skip the whole invalid but try with next byte a personal preference.
But when comparing with the ones from Harfbuzz and FriBidi, it's always the same logic. Inline the macros and we get the same code :)

xtext.cpp
NiLuJe: My naive assumption would be that it'll match the target's bitness (i.e., will be INT64 on 64bit), while I assume you always want an INT32.

ffi/freetype_h.lua
```diff
@@ -19,18 +20,18 @@ struct FT_ListRec_ {
   FT_ListNode tail;
 };
 struct FT_BBox_ {
-  long int xMin;
-  long int yMin;
-  long int xMax;
-  long int yMax;
+  FT_Pos xMin;
```
Frenzie: This is because of the ffi-cdecl last month?

xtext.cpp
```diff
@@ -0,0 +1,2157 @@
...
+    hb_font_t *    hb_font;
+    hb_buffer_t *  hb_buffer;
+    hb_feature_t * hb_features;
+    int            hb_features_nb;
+} xtext_hb_font_data;
+
+// Global direction and language
+static bool default_para_direction_rtl = false;
+static char * default_lang = NULL;
```
NiLuJe: Random comment before I forget: Don't know where/how you use that (and other plain `char`s yet), but I assume you'd want `unsigned char`s pretty much everywhere in this context, to avoid surprises on platforms where char is signed.

NiLuJe:
> > Stupid question is stupid: does XText actually need to be C++?
> 
> Well, stupid answer: I dunno :) I use some `class` XText, so I figure it needs C++. I initially played with std:vector stuff, and ended up not needing it. But as I'm just used to C++ in crengine, may be there's stuff that are specific to C++, and would need some rewrite/rethink if it had to be done in C. I really don't know :)
> Dunno if it's as simple as replacing it with a `struct XText`, but well... not wishing to rework the whole thing :) But it you feel like it can be quickly made to C, feel free to have a quick try at compiling it :)

I had obviously missed the Class :D.

thirdparty/libunibreak/CMakeLists.txt
Frenzie: I agree, let's avoid quiet unless it causes some stupid "log size too big" or whatever issue on one of the CI platforms.

ffi/freetype_h.lua
poire-z: No, I think it's because I added FT_Pos. Before, it could just masquerade as long int without annoucing its real name :)

xtext.cpp
```diff
@@ -0,0 +1,2157 @@
...
+        if (m_bidi_ctypes) { free(m_bidi_ctypes); m_bidi_ctypes = NULL; }
+        if (m_bidi_btypes) { free(m_bidi_btypes); m_bidi_btypes = NULL; }
+        if (m_bidi_levels) { free(m_bidi_levels); m_bidi_levels = NULL; }
+        if (m_lang)        { delete m_lang;       m_lang = NULL; }
+        m_no_more_usable = true;
+    }
+
+    void setLanguage(const char * lang) {
+        m_lang = new char[strlen(lang)+1];
+        strcpy(m_lang, lang);
```
NiLuJe: Ah, now I get why `char` ;).

thirdparty/libunibreak/CMakeLists.txt
```diff
@@ -0,0 +1,45 @@
...
+    set(CFG_CMD "${CFG_CMD} && ${ISED} 's|need_lib_prefix=no|need_lib_prefix=yes|' libtool")
+    set(CFG_CMD "${CFG_CMD} && ${ISED} 's|need_version=no|need_version=yes|' libtool")
+    set(CFG_CMD "${CFG_CMD} && ${ISED} 's|library_names_spec=.*|library_names_spec=\"\\\\\$libname\\\\\$release\\\\\$shared_ext\\\\\$versuffix \\\\\$libname\\\\\$release\\\\\$shared_ext\\\\\$major \\\\\$libname\\\\\$shared_ext\"|' libtool")
+    set(CFG_CMD "${CFG_CMD} && ${ISED} 's|soname_spec=.*|soname_spec=\"\\\\\$libname\\\\\$release\\\\\$shared_ext\\\\\$major\"|' libtool")
+endif()
+
+ko_write_gitclone_script(
+    GIT_CLONE_SCRIPT_FILENAME
+    https://github.com/adah1972/libunibreak.git
+    libunibreak_4_2-1
```
Frenzie: Not really necessary, but I've been moving toward more explicit checkouts like this in these files. (Partially for clarity, partially because there *could* theoretically be a libunibreak_4_2-1 branch.)
```suggestion
    tags/libunibreak_4_2-1
```

xtext.cpp
poire-z: I thought that was when I was using it to store numbers (so issue, with signed and unsigned when I stored a negative number). There, I just store some string ("fr", "zh_CN"..)
Harfbuzz and FriBidi all have just `const char *` in their function declaration. Using something different, I would need to cast it, or something.

xtext.cpp
```diff
@@ -0,0 +1,2157 @@
...
+    #ifdef __cplusplus
+    extern "C" {
+    #endif
+#include <linebreakdef.h>
+    #ifdef __cplusplus
+    }
+    #endif
+
+// luajit doesn't provide lua_set/getuservalue (unless compiled
+// with LUAJIT_ENABLE_LUA52COMPAT) but these are equivalents
```
Frenzie: Well, there's probably no real harm in enabling that. Will LUA_VERSION_NUM be 502 if that's enabled?

xtext.cpp
poire-z: Limiting the size of the whole (a bit is 1 bit, a bool is 8 bits) - and trying to have the big ones aligned. I rewrote that one a few times, each time printing its sizeof :)

```
__________0
    int text_index;    // original index in m_text
__________4
    uint16_t glyph;    // glyph index in font
    unsigned char font_num;
    unsigned is_rtl:1;
    unsigned can_extend:1;
    unsigned _unused:6;
__________8
    signed short  x_advance;  // aligned
    signed short  x_offset;
__________12
    signed short  y_offset; // aligned
    unsigned char cluster_len;
    unsigned is_cluster_start:1; // could have been moved with the others
                   // but I like it near cluster_len, and I have the room
__________16
```

xtext.cpp
```diff
@@ -0,0 +1,2157 @@
...
+                        m_has_multiple_scripts = true;
+                    }
+                    prev_script = script;
+                }
+                // Note: as we have here guessed the script of what's to be measured
+                // next, we could store it in m_charinfo (an additional int32...),
+                // so we can pass it to getHbFontData() (or use it and check
+                // ourselves here), to skip fallback fonts that do not support
+                // this script - if fonts announce the scripts they support (I think
+                // I have seen that in some font tables, may be OTF only?).
```
NiLuJe: Yeah, that vaguely sounds like an OpenType feature ;).

xtext.cpp
poire-z: Dunno, but not taking any risk. That's mostly just some rename from the Lua5.2 people who wanted to get rid of the word "environment".
Really no need to take a risk. It enables quite a few more risk/incompatible features.

xtext.cpp
```diff
@@ -0,0 +1,2157 @@
...
+                int brk = lb_process_next_char(&lbCtx, ch);
+                // This tells us about a break between previous char and this 'ch'.
+                // printf("between <%c%c>: brk %d\n", m_text[i-1], m_text[i], brk);
+                // Note: LINEBREAK_ALLOWBREAK is set on the last space in a sequence
+                // of multiple consecutive spaces.
+                if ( brk == LINEBREAK_ALLOWBREAK ) {
+                    // Happens between a space (at i-1) and its following non-space
+                    // char, or after each CJK char.
+                    m_charinfo[i-1].flags |= CHAR_CAN_WRAP_AFTER;
+                    if ( m_text[i-1] == ' ' ) { // todo: also other kind of spaces?
```
NiLuJe: utf8proc *might* have a table of valid space codepoints.

Or something does. That kind of stuff sounds familiar, but I can't quite place it RN ;).

xtext.cpp
```diff
@@ -0,0 +1,2157 @@
...
+            // Line breaking and wrapping
+            bool line_break = false;
+            if ( i == 0 ) {
+                lb_init_break_context(&lbCtx, m_text[i], m_lang ? m_lang : default_lang);
+            }
+            else {
+                // When at end of m_text, add a letter ('Z') so a trailing \n can be
+                // flagged as CHAR_MUST_BREAK_AFTER, so we can show an empty line
+                // and allow the cursor to be positionned after that last \n.
+                int ch = i < m_length ? m_text[i] : 'Z';
```
NiLuJe: Can't you just shortcut the whole thing and assume brk == CHAR_MUST_BREAK_AFTER?

xtext.cpp
```diff
@@ -0,0 +1,2157 @@
...
+        m_is_measured = true;
+        m_width = final_width;
+    }
+
+    // Based on crengine/src/lvfntman.cpp measureText() with _kerningMode == KERNING_MODE_HARFBUZZ
+    // Changes:
+    // - we work on the full m_text/m_charinfo, with absolute indices start and end (end excluded)
+    // - we don't use cumulative widths: we store individual char widths (to store them in 16 bits
+    //   in m_charinfo, instead of needing a full 32 bits int for each char)
+    int measureSegment(int font_num, int start, int end, int hints) {
```
NiLuJe: Ah, now I get where the INT_MIN came from ;).

xtext.cpp
poire-z: Nope. If last char of file/m_text is "." or "a", I don't want `brk == CHAR_MUST_BREAK_AFTER`. I want it only if the file ends with a `\n` - or one of the rare other unicode chars libunibreak considers the same (like one named "paragraph break").
In the first case, I don't add an empty line in TextEditor at end of file. In the later case, I do.

xtext.cpp
NiLuJe: Okay, I get it now, see measureSegment's signature.

NiLuJe: Goddamnit, clicked the wrong button again. :D.

poire-z: I'm a bit confused with these `#define SOME_NUMBER`: they come with no type/length - and nothing to say in what context they should be used.
Can I assume they'll work as expected in any context?
Using `short i = INT_MIN64`, will it be cast to a short and be the min short possible?

xtext.cpp
```diff
@@ -0,0 +1,2157 @@
...
+                                printf("%sMSHB ### measured past failures > W= %d\n%s[...]",
+                                                                indent, fallback_width, indent);
+                            #endif
+
+                            t_notdef_start = -1;
+                            notdef_width = 0;
+                            // And go on with the found glyph now that we fixed what was before
+                        }
+                        // Glyph found in this font
+                        advance = FONT_METRIC_TRUNC(FONT_METRIC_ROUND(glyph_pos[hg].x_advance));
```
NiLuJe: I'm not *quite* sure why you need to TRUNC?

ROUND does make sense, AFAIU (which, granted, isn't much ;D).

xtext.cpp
```diff
@@ -0,0 +1,2157 @@
...
+        for ( int t = start; t < end; t++ ) {
+            #ifdef DEBUG_MEASURE_TEXT
+                printf("%sMSHB t%d (=%x) ", indent, t, m_text[t]);
+            #endif
+            // Grab all glyphs that do not belong to a cluster greater that our char position
+            int cur_width = 0; // current cluster width
+            while ( hg < glyph_count ) {
+                hcl = glyph_info[hg].cluster;
+                if ( hcl <= t ) { // glyph still part of a previous cluster
+                    int advance = 0;
```
NiLuJe: Sidebar: I'd keep that in a matching data type (hb_position_t, AFAIK, which points to an int32_t, FWIW).

xtext.cpp
Frenzie: I believe I compiled with that and it worked fine; if you check my pull requests in front there was one tiny fix I committed after testing it.

Frenzie: Anyway, I guess I'm asking if it's future-compatible with enabling that flag.

poire-z:
> I had obviously missed the Class :D.

xtext.cpp
```diff
@@ -0,0 +1,2157 @@
...
+            lua_pushinteger(m_L, s->x_offset);
+            lua_settable(m_L, -3);
+
+            lua_pushstring(m_L, "y_offset");
+            lua_pushinteger(m_L, s->y_offset);
+            lua_settable(m_L, -3);
+
+            lua_pushstring(m_L, "is_rtl");
+            lua_pushboolean(m_L, s->is_rtl);
+
```
NiLuJe: Okay, that appears to answer my earlier TRUNC question ;).

xtext.cpp
NiLuJe: Okay, see `shapeSegments` ;).

xtext.cpp
NiLuJe: s/work/word/ ;)

NiLuJe: s/it it/it is/

NiLuJe: s/indice/index/

(index is singular, indices is plural. indexes is accepted, but I hate it ;p).

NiLuJe: (Random comment: that's why `i` or `idx` is often used in variable names that represent an index ;)).

xtext.cpp
poire-z: I guess this PR would work with that flag enabled. But haven't checked what LUA_VERSION_NUM would be. Anyway, it would be a simple fix of removing these #define.
But if you'd like to enable it, better do it later, after all that RTL UI/text/keyboard stuff is done, to not add uneeded noise and errors :)

ffi/freetype_h.lua
NiLuJe: Cool, thanks!

xtext.cpp
Frenzie: Oh, here it is: koreader/koreader#3090

I guess I left it somewhat implied, or maybe it was also related to a newer LuaJIT version tout court? I don't recall the exact specifics, it's been a little while.

xtext.cpp
NiLuJe: Unless you use a specific format identifier for literals, that'll be an `int` (which in turn will be an int32_t or an int64_t depending on arch ^^).

AFAIK, there's no identifier for "short" or "char". You can just say "unsigned" via `u` or `U`, "long" via `L` (can be specified multiple times for long long), or float via `f` (a decimal literal defaults to double).

NiLuJe: As for your short example, GCC will happily shout at you that it just will not do:

```
warning: overflow in conversion from ‘int’ to ‘short int’ changes value from ‘-2147483648’ to ‘0’ [-Woverflow]
  326 |   short s_min = INT_MIN;
```

NiLuJe: Which nicely illustrates that a literal is an `int` by default, so you get an implicit cast, which overflows ;).

NiLuJe: In any case, here, as it is, int makes sense, as that matches `measureSegment`'s signature.

xtext.cpp
poire-z: GitHub looks a bit drunk (I assume it's github and not you, because it also does me strange things), but all your last comments points to that bit of code about "is_rtl" :) But I'll find out where they belong, no worry :)

xtext.cpp
NiLuJe: Yep, glib and utf8proc use a similar approach, IIRC ;).

xtext.cpp
NiLuJe: Err, `sizeof(bool) == 1`?

NiLuJe: Argh, that bool is 8 bits thingy appear to be a C++ shenanigan. >_<".

poire-z: Yes. It's not `0.125` :)
```c
typedef struct {
    int text_index;    // original index in m_text
    uint16_t glyph;    // glyph index in font
    signed short  x_advance;
    signed short  x_offset;
    signed short  y_offset;
    unsigned char font_num;
    unsigned char cluster_len;
    bool is_cluster_start;
    bool is_rtl;
    bool can_extend; // one byte too much, these last 3 are NOT merged into one byte
} xtext_shapeinfo_t;
```
ùakes its sizeof() = 20.
My previous with :bits kept it 16.

NiLuJe: Oh, yeah, duh, the bitfields essentially allows you to fit two bools in one byte.

The larger inconsistency in your last example is probably because of enforced alignment to the wider type (int == 4).

xtext.cpp
NiLuJe: Yup, made sense once I saw how it was used ;).

xtext.cpp
NiLuJe: Okay, makes sense ;).

xtext.cpp
NiLuJe: Oh, yeah, the recap in the "Conversation" panel inlines completely unrelated code (or, well, at least that certainly wasn't where I put 'em ;p).

NiLuJe: @poire-z: Appears to point to the right place in the "Changes" panel on my end ;).

ffi/pic.lua
poire-z: Well, valgring kept reporting this:
```
1,440,000 bytes in 1 blocks are possibly lost in loss record 3,972 of 3,980
   at 0x483577F: malloc (in .../vgpreload_memcheck-amd64-linux.so)
   by 0x10FB445A: lodepng_malloc (lodepng.c:78)
   by 0x10FBCA6F: lodepng_decode (lodepng.c:4899)
   by 0x12BCF5: lj_vm_ffi_call (buildvm_x86.dasc:2575)
   by 0x149F19: lj_ccall_func (lj_ccall.c:1161)
   by 0x127E8D: lj_cf_ffi_meta___call (lib_ffi.c:230)
   by 0x129D56: lj_BC_FUNCC (buildvm_x86.dasc:809)
   by 0x11A03F: lua_call (lj_api.c:1111)
```
Looking at (your) ffi code, it seems we do things right by passing that data to BlitBuffer and using setAllocated(1) - so it should properly be free. Dunno why valgrind kept telling that one - and mostly only that one that feels serious (1.4 Mb!)

NiLuJe: Any idea on which pixmap that was happening?

Nothing should ever return anything than [1..4) in ncomp. If it does, that could arguably mean there's a real bug somewhere ;).

xtext.cpp
poire-z: Yes, and HB too, see https://github.com/koreader/crengine/blob/cb3b2da13b70d8bea86ef70bf959703b02f58962/crengine/src/lvtextfm.cpp#L1038-L1051.
But as there are [not many kind of space](http://jkorpela.fi/chars/spaces.html), we might as well just hardcode them here

xtext.cpp
poire-z: Well, thought about that (adding the non >>6 advances, even returning them to frontend and have it do the >>6 with the cumulative widths to not accumulate rounding errors).
But no: with my recursive calls to measureSegments, I need to have consistent stuff with individual segment full advance = the sum of the individual x_advance I return for each glyph. So, better to end up doing the >>6 (well, trunc(round()) for better looking results) here.

ffi/pic.lua
poire-z: Nope. Can't have valgrind report that again now... although I think I got them even after that little fix.

poire-z: Mhhh, got it again. It may be logged at the end of exiting koreader/valgring, so dunno where it comes from. I did "Refresh cached book information" on a EPUB with a cover, although this should use MuPDF for that.
I added that:
```diff
--- a/ffi/pic.lua
+++ b/ffi/pic.lua
@@ -262,18 +262,20 @@ function Pic.openPNGDocument(filename)
         -- NOTE: LodePNG will NOT do RGB -> Grayscale conversions for us, for design reasons (multiple ways to do it, lossy).
         --       So we can only *ask* to keep grayscale PNGs as-is, but we might actually be getting fed back a RGB24 one ;).
         req_n = 1
     end

+    print("pngopen: "..filename)
     local ok, re = Png.decodeFromFile(filename, req_n)
     if not ok then error(re) end

     if re.ncomp == 1 then bbtype = BB.TYPE_BB8
     elseif re.ncomp == 2 then bbtype = BB.TYPE_BB8A
     elseif re.ncomp == 3 then bbtype = BB.TYPE_BBRGB24
     elseif re.ncomp == 4 then bbtype = BB.TYPE_BBRGB32
     else
+        print("pngfailfree: "..filename)
         if re.data then
             ffi.C.free(re.data)
         end
         error("unsupported number of color components")
     end
```
which looks to me the only place that could call `lodepng_decode`, and I didn't see any of my print printed once !
It's from ffi, so not from thirdparty/ stuff that could use lodepng themselves... Quite puzzled.

xtext.cpp
poire-z: Ok, adressed that one with hardcoded char codes, and added another flag CAN_EXTEND_WIDTH_FALLBACK to mark CJK chars (actually, all allow_break chars), to allow text justification when there is not a single space.

poire-z:
@NiLuJe : while we're on maths... any insight about my question in the 3rd paragraph of top post, about hinting, rounded or not advances/offsets... and if we'd need to use trunc(round()) tricks instead of >>6 in crengine?

NiLuJe:
@poire-z: Would you be so kind as to point me to one such place in CRe that wouldn't be too gnarly to grok?

NiLuJe:
@poire-z: That said, if you do want integer pixels (which you probably do, as I don't think we do subpixel positioning), rounding before trunc (>> 6) does make sense, otherwise there's a rather large window in which there'll be a "rounding" mistake after the trunc.

That's both napkin maths and not having actually thought about it that much, so, here be be dragon ;p.

But ROUND essentially does what it says on the tin (align to the closest multiple of 64), so that trunc ( / 64 ) will return an integer with no loss.

poire-z:
https://github.com/koreader/crengine/blob/master/crengine/src/lvfntman.cpp
Mostly all the places where we have `>>6` or `>> 6` , which is a floor().
I didn't mind that because I've always seen multiple of 64 when printing these metrics, so >>6 feels allright.
But when debugging this PR, i've seen some 68 or 77 where I used to see 64 :) so I'm puzzled to why.
Reading in http://chanae.walon.org/pub/ttf/ttf_glyphs.htm:
> The advance is always grid-fitted when the glyph was hinted

So, may be I've been doing my printf in a crengine with hinting native or auto. But with off, doing >>6 might be wrong, and a round might be prefered. Haven't investigated much, thought I'd ask you before.

NiLuJe:
Remember that advances are now essentially unhinted with `RENDER_LIGHT `, which might explain why you're seeing new numbers ;).

NiLuJe:
Sidebar: `>> 6` is [TRUNC](https://git.savannah.gnu.org/cgit/freetype/freetype2-demos.git/tree/src/ftcommon.c#n320).

I'm assuming this was a thinko in your message, since you've got it right in xtext ;).

poire-z:
> Sidebar: >> 6 is TRUNC.

Yes, but without anything else, it's doing a floor, getting rid of the part in the 6 lower bits, just like as TRUNC(FLOOR()) , no?

NiLuJe:
See my edit ;).

For positive values, trunc should pretty much match floor, yeah. Things might get dicier with negative values, which can definitely happen in this context (kerning?).

Frenzie:
Btw, since there were some complaints about the review system this could be of interest?

https://reviewable.io/

NiLuJe:
Eh, worth a try.

Not having everything drowned in a sea of green and red can only be an improvement ;).

Frenzie:
Anyway, I was satisfied by my tests, so I enabled it for all current & future KOReader repos. It then proceeded to warn me that I was out of GitHub API quota and that other admins should enable it to spread the load, so it could potentially be helpful if you did that.

That being said, it's probably just because it was doing it all at once. I doubt regular PR activity should be an issue.

I don't think GH PR reviews are *that* bad for the most part, but the fact that last week it simply hid half a dozen line comments on me so I didn't see them rubbed me the wrong way. ;-) I noticed the "hidden comments" thing, but I thought it was just hiding a bunch of old junk I'd already seen… nope, brand new suggestions.

PR #1011, opened 2019-11-17
xtext: link against stdc++, fix some clang-tidy warning

poire-z:
Linking with -static-libstdc++ might be needed on Kindle and Pocketbook to load (the simple fact of linking against the STL implies ABI compat checks). https://github.com/koreader/koreader/pull/5598#issuecomment-554764338

Also fix some clang-tidy warning about malloc'ing 0 bytes by handling the empty input case apart. https://github.com/koreader/koreader/pull/5598#issuecomment-554753824 - even if that's not really an issue:

```
$ man malloc
       The malloc() function allocates size bytes and returns a pointer to the
       allocated memory.  The memory is not initialized.  If size is  0,  then
       malloc()  returns either NULL, or a unique pointer value that can later
       be successfully passed to free().
```

Let's wait a bit for broken on Kindle reports to validate it's really needed - and in case we find the solution for https://github.com/koreader/koreader/issues/5602 ?
Or merge and bump in case it magically solves it? :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,7 +158,7 @@ $(OUTPUT_DIR)/libs/libkoreader-xtext.so: xtext.cpp \
 	-I$(HARFBUZZ_DIR)/include/harfbuzz \
 	-I$(FRIBIDI_DIR)/include \
 	-I$(LIBUNIBREAK_DIR)/include \
-	$(DYNLIB_CXXFLAGS) -Wall -o $@ $^
+	$(DYNLIB_CXXFLAGS) -static-libstdc++ -Wall -o $@ $^
 ifdef DARWIN
 	install_name_tool -change \
 		`otool -L "$@" | grep "libluajit" | awk '{print $$1}'` \
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -408,6 +408,7 @@ function S.setWindowIcon(icon)
 
     local BB = require("ffi/blitbuffer")
     local icon_bb = BB.new(re.width, re.height, BB.TYPE_BBRGB32, re.data)
+    icon_bb:setAllocated(1) -- free re.data when bb is freed
     local icon_bit_depth = 32
     local surface = SDL.SDL_CreateRGBSurfaceWithFormatFrom(icon_bb.data,
                                                            icon_bb:getWidth(), icon_bb:getHeight(),
--- a/xtext.cpp
+++ b/xtext.cpp
@@ -600,8 +600,15 @@ public:
     }
 
     void measure() {
-        if (m_is_measured)
+        if ( m_is_measured )
             return;
+
+        if ( m_length == 0 ) {
+            // Nothing to allocate nor measure
+            m_width = 0;
+            m_is_measured = true;
+            return;
+        }
 
         allocate();
         checkBidi();
@@ -741,8 +748,8 @@ public:
             last_bidi_level = new_bidi_level;
         }
 
-        m_is_measured = true;
         m_width = final_width;
+        m_is_measured = true;
     }
 
     // Based on crengine/src/lvfntman.cpp measureText() with _kerningMode == KERNING_MODE_HARFBUZZ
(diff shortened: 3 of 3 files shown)


poire-z:
Ignoring the other clang-tidy report, as it looks strlcpy is not that good (and needs libbsd), and we get these strings from `luaL_checkstring` which makes correct 0-terminated strings.
_Call to function 'strcpy' is insecure as it does not provide bounding of the memory buffer. Replace unbounded copy functions with analogous functions that support length arguments such as 'strlcpy'. CWE-119_

NiLuJe:
`strncpy` would work, but cppcheck will also complain about it.

The current usage is harmless anyway.

poire-z:
Last commit fix the last serious memory leak reported by valgrind. See https://github.com/koreader/koreader-base/pull/1010#discussion_r346679596.

The remaining one is in crengine:
```
==20837== 64 bytes in 1 blocks are definitely lost in loss record 2,258 of 4,031
==20837==    at 0x4835DEF: operator new(unsigned long) (in valgrind/vgpreload_memcheck-amd64-linux.so)
==20837==    by 0x13733AB1: tinyNodeCollection::allocTinyElement(ldomNode*, unsigned short, unsigned short) (lvtinydom.cpp:12186)
==20837==    by 0x13758809: ldomDocument::ldomDocument() (lvtinydom.cpp:3374)
==20837==    by 0x1380648E: LVDocView::createEmptyDocument() (lvdocview.cpp:4366)
==20837==    by 0x1380DE7A: LVDocView::LoadDocument(LVFastRef<LVStream>, bool) (lvdocview.cpp:4003)
==20837==    by 0x13811905: LVDocView::LoadDocument(wchar_t const*, bool) (lvdocview.cpp:3784)
==20837==    by 0x13811E3B: LVDocView::LoadDocument(char const*, bool) (lvdocview.cpp:4726)
==20837==    by 0x1361B144: loadDocument(lua_State*) (cre.cpp:629)
==20837==    by 0x129D56: lj_BC_FUNCC (buildvm_x86.dasc:809)
==20837==    by 0x12A905: lj_ff_coroutine_resume (buildvm_x86.dasc:1640)
==20837==    by 0x11A09F: lua_pcall (lj_api.c:1129)
==20837==    by 0x110943: docall (luajit.c:121)
```
But for 64 bytes... I'll just add a comment:
```
@@ -3372,6 +3372,14 @@ ldomDocument::ldomDocument()
 , lists(100)
 {
     allocTinyElement(NULL, 0, 0);
+    // Note: valgrind reports (sometimes, when some document is opened or closed,
+    // with metadataOnly or not) a memory leak (64 bytes in 1 blocks are definitely
+    // lost), about this, created in allocTinyElement():
+    //    tinyElement * elem = new tinyElement(...)
+    // possibly because it's not anchored anywhere.
+    // Attempt at anchoring into a _nullNode, and calling ->detroy()
+    // in ~ldomDocument(), did not prevent this report, and caused other ones...
```

Frenzie:
Serious memory leak? But, um, the icon is displayed as long as the program exists and then forgotten about. It might technically "leak" in the sense that it's not closed properly on exit, but it just lives & dies with the program. :-P

(Unless I'm missing something, it'd only be a leak when changing the icon dynamically.)

poire-z:
Nope, it was the png data pointer that was not free'd.
We read the png data, make a blitbuffer with it, pass it to some SDL texture stuff, which does its thing, and then we free() the blitbuffer as we (and SDL) don't need it anymore. But we forgot to free the png data, that was not done neither by us explicitely, and neither by the blitbuffer:free(). Now, we have the blitbuffer:free() do it (
Serious in the sense that it was 1.4 MB taken in the emulator process for nothing - only once, but that was still a waste :)

Frenzie:
Sure, sure. It's just that when you say leak I think more "1.4 MB here, 1.4 MB there", even if that only happens over the course of hours or days. But it's 1.4 MB period. (Would it fit on a floppy?) :-)

PR #1012, opened 2019-11-18
[chore] thirdparty patch file dependencies

Frenzie:
I'm going bonkers when a patch file is updated and nothing happens because only CMakeLists.txt is a dependency. This reruns CMake when any file in `thirdparty/<dependency>` changes. In practice that mostly means `*.patch` in addition to CMakeLists.txt. There's a `cr3.css` in kpvclib and that's probably about it, but I'm trying to be forward-thinking here.

### DIFF
--- a/Makefile.third
+++ b/Makefile.third
@@ -13,7 +13,7 @@ fetchthirdparty:
 		&& (cd thirdparty/kpvcrlib/crengine; git checkout .) \
 		|| echo warn: crengine folder not found
 
-$(FREETYPE_LIB) $(FREETYPE_DIR)/include: $(wildcard $(THIRDPARTY_DIR)/freetype2/*.*)
+$(FREETYPE_LIB) $(FREETYPE_DIR)/include: $(THIRDPARTY_DIR)/freetype2/*.*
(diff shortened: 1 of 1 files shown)


Frenzie:
Just `*` will include the directory `thirdparty/freetype/build`.

The manual doesn't seem to indicate any other way to exclude directories.

https://www.gnu.org/software/make/manual/html_node/Wildcard-Function.html
https://www.gnu.org/software/make/manual/html_node/Wildcard-Examples.html

NiLuJe:
Do we actually need the wildcard function for a rule prerequisite?

The doc seems to hint that wildcards are expanded automatically in rule prereqs.

Frenzie:
To my mind I was more or less following the example from the manual:

```
objects := $(patsubst %.c,%.o,$(wildcard *.c))

foo : $(objects)
        cc -o foo $(objects)
```

Which example/explanation are you referring to precisely?

NiLuJe:
The "print" one from https://www.gnu.org/software/make/manual/html_node/Wildcard-Examples.html

Frenzie:
D'oh, I must've overlooked that.

PR #1014, opened 2019-11-22
bump crengine: round FT metrics, split text drawing by script

poire-z:
Includes https://github.com/koreader/crengine/pull/318
- Fonts: round FT metrics instead of floor'ing them
- Fonts: switch to no hinting when native hinting fails
- Fonts: fix issue with Harfbuzz fallback font drawing
- Text: split measuring and word drawing by unicode script
- Page splitting: fix small memory leak
- Fix "background-color: black" ignored on inline elements
- Fix decoding of recent MOBI files
- Hardcoded elements list: add <font>

xtext.cpp: small cleanup (no longer "no more" :), no logic change, adapted to use the same macro used in 1st crengine commit.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1014)

### DIFF
--- a/xtext.cpp
+++ b/xtext.cpp
@@ -68,10 +68,11 @@ extern "C"
 #define ELLIPSIS_CHAR 0x2026
 
 // Helpers with font metrics (units are 1/64 px)
-#define FONT_METRIC_FLOOR(x)    ((x) & -64)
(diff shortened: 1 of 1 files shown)


NiLuJe:
The cervantes failures are unrelated (FBInk shenanigans discussed in another PR ;)).

PR #1017, opened 2019-11-23
thirdparty/libffi v3.3

Frenzie:
https://github.com/libffi/libffi/releases/tag/v3.3

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1017)

### DIFF
--- a/thirdparty/libffi/CMakeLists.txt
+++ b/thirdparty/libffi/CMakeLists.txt
@@ -20,12 +20,12 @@ if(DEFINED SYSROOT)
 endif()
 
 include(ExternalProject)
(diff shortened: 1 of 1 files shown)


NiLuJe:
Shouldn't cause any issue, I've been using the 3.3 RCs since the beginning ;).

poire-z:
What this is used for/by? Looks like only glib is built against it. luajit does its own ffi stuff by itself, right?

NiLuJe:
Yep, glib. LuaJIT indeed has its own heavily customized variant.

PR #1021, opened 2019-12-09
Add util.fsyncOpenedFile() and util.fsyncDirectory()

poire-z:
To have some tools to try to do better with our config files getting corrupted in case of device hang/crash.
More details in https://github.com/koreader/koreader/pull/3625#issuecomment-561838044.
Of interest:
http://blog.httrack.com/blog/2013/11/15/everything-you-always-wanted-to-know-about-fsync/
https://stackoverflow.com/questions/37288453/calling-fsync2-after-close2

Also bump crengine:
-  Open (as text) small or empty files https://github.com/koreader/crengine/pull/322

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1021)

### DIFF
--- a/ffi/util.lua
+++ b/ffi/util.lua
@@ -410,6 +410,72 @@ function util.readAllFromFD(fd)
     return table.concat(data)
 end
 
+--- Ensure content written to lua file or fd is flushed to the storage device.
+--
+-- Accepts a low-level file descriptor, or a higher level lua file object,
+-- which must still be opened (call :close() only after having called this).
+-- If optional parameter sync_metadata is true, use fsync() to also flush
+-- file metadata (timestamps...), otherwise use fdatasync() to only flush
+-- file content and file size.
+-- Returns true if syscall successful
+-- See https://stackoverflow.com/questions/37288453/calling-fsync2-after-close2
+function util.fsyncOpenedFile(fd_or_luafile, sync_metadata)
+    local fileno
+    if type(fd_or_luafile) == "number" then -- low-level fd
+        fileno = fd_or_luafile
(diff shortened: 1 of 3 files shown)


ffi/util.lua
```diff
@@ -410,6 +410,73 @@ function util.readAllFromFD(fd)
...
+    end
+    if attributes.mode ~= "directory" then
+        -- file, symlink...: get its parent directory
+        path = util.dirname(path)
+        attributes, err = lfs.attributes(path)
+        if not attributes or err ~= nil or attributes.mode ~= "directory" then
+            return false, err
+        end
+    end
+    local dirname = ffi.cast("uint8_t *", path)
```
NiLuJe: Technically `char *`, because it may be signed or unsigned depending on platform ;).

NiLuJe: I'd also simply inline that in the `open()` call, that's what we usually do for casts ;).

poire-z: Done. Ok with the principles?

NiLuJe: Yep, that looks like the perfect way to do it as right as we can, given how quirky this can be ;). :+1:

NiLuJe:
I'd *possibly* comment on the `open` quirk needed to get it to open a directory (i.e., `RDONLY`), but just mentioning it here is good enough for me, so you can simply xref this PR ;).

(And, FWIW, I do prefer doing it this way instead of opendir+dirfd ;)).

poire-z:
There's an url above that function that explains all that (so not-worth wasting energy with re-circleCI with an inline comment :)

PR #1022, opened 2019-12-16
bump crengine & fribidi, tweak xtext.makeLine()

poire-z:
- bump crengine: (Upstream) DocX: fix paragraphs with single hyperlink https://github.com/koreader/crengine/pull/323
- bump fribidi to 1.0.8 (for nested isolates fix, https://github.com/koreader/koreader/issues/5359#issuecomment-565001148)
- xtext.makeLine(): return `no_allowed_break_met=true` when that happens, so we know the line breaks inside a word. https://github.com/koreader/koreader/pull/5598#issuecomment-565836669

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1022)

### DIFF
--- a/thirdparty/fribidi/CMakeLists.txt
+++ b/thirdparty/fribidi/CMakeLists.txt
@@ -36,7 +36,7 @@ endif()
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME
     https://github.com/fribidi/fribidi.git
-    tags/v1.0.7
+    tags/v1.0.8
     ${SOURCE_DIR}
 )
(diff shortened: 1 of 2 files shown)


poire-z:
@Frenzie : when do you plan to release 2019.12?
Can I still bump this (and a few coverbrowser long titles overlap fixes)?

Frenzie:
@poire-z Sure, let's get it in. I haven't really been able to test it — well, it seems fine on my H2O and on Android but I normally prefer to do a tiny bit more manual testing than just the most basic.

PR #1023, opened 2019-12-19
[RFC] Remarkable port

tcrs:
I have a remarkable (https://remarkable.com/). The builtin PDF reader was too slow so I did a quick port of koreader. This is mostly based on the Sony PRS port.

I've never worked on koreader before so would appreciate any comments etc.

I display stuff could be improved. I think the remarkable has a modern eink display with reagl and stuff so could probably use better paths for partial updates etc (xochitl, the interface the remarkable ships with, does a good job of low-latency not too flashy updates). I don't really know much about the eink driver stuff so I've just left it alone now things seem to be working...

There's also a pull request open over in koreader/koreader for the other bits and bobs.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1023)


input/input-remarkable.h
```diff
@@ -0,0 +1,26 @@
...
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef _KO_INPUT_REMARKABLE_H
+#define _KO_INPUT_REMARKABLE_H
+
+static void generateFakeEvent(int pipefd[2]) {
+	// TODO anything required here? Sony PRS port monitors battery charge etc
```
pazos: You can generate fake events that will be forwarded to the frontend as normal input events. There are a few defined in https://github.com/koreader/koreader-base/blob/master/input/input.c#L36-L41 but you can define more.

Normally is used to notify the program hardware changes, and we use libue to get that info from udev.

Once you have your fake events ready you can react to them in UIManager, but first do note that names are slightly different in the frontend. See how they're named in https://github.com/koreader/koreader/blob/f1f75c5cb0fa063c35f49a84b2e025f1d292ea01/frontend/device/input.lua#L176

Makefile.defs
```diff
@@ -151,6 +151,10 @@ else ifeq ($(TARGET), sony-prstux)
 	CHOST?=arm-linux-gnueabihf
 	export SONY_PRSTUX=1
 	export USE_LJ_WPACLIENT=1
+else ifeq ($(TARGET), remarkable)
+	CHOST?=arm-kobo-linux-gnueabihf
```
pazos: Are you using the kobo toolchain from https://github.com/koreader/koxtoolchain?

Frenzie:
For reference, the other PR is https://github.com/koreader/koreader/pull/5697.

Makefile.third
```diff
@@ -682,7 +682,7 @@ $(LUA_SPORE_ROCK): $(THIRDPARTY_DIR)/lua-Spore/*.*
 		$(CMAKE) $(CMAKE_FLAGS) -DOUTPUT_DIR="$(CURDIR)/$(OUTPUT_DIR)" \
 		-DLUA_SPORE_VER=$(LUA_SPORE_VER) -DLD="$(CC)" \
 		-DCC="$(CC)" -DCFLAGS="$(CFLAGS) -I$(LUAJIT_DIR)/src" \
-		-DLUAROCKS=$(if $(DARWINHOST),"luarocks --lua-dir=/usr/local/opt/lua@5.1",luarocks) \
+		-DLUAROCKS=$(if $(DARWINHOST),"luarocks --lua-dir=/usr/local/opt/lua@5.1","luarocks --lua-version=5.1") \
```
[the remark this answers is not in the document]
NiLuJe: That's 3.x only, IIRC.

(For ref., I still patch my local copy w/ http://trac.ak-team.com/trac/browser/niluje/Configs/trunk/Kindle/Misc/koreader-luarocks-3.patch because I'm way too lazy to figure out what's got its panties in a twist ;)).

NiLuJe:
IIRC, there's some fancy undocumented (on their side) flags for their specific tweaks to waveform shenanigans.

Most of it is detailed in the libremarkable sources, if memory serves ;).

In any case, for a start, that should be perfectly fine enough. I don't actually recall the extent of their tweaks and if they'd actually be useful anyway ;).
(I seem to remember a fancier A2 which we probably won't ever care for, for instance. If, on the other hand, there's a fancier REAGL and/or making it behave requires some hoop-jumping like on some platforms (hi, Kindle), that'd be something worth looking into for later ;)).

Makefile.defs
tcrs: Ah yea I had forgotten about that, I did use the kobo toolchain from koxtoolchain. I started using the toolchain provided by remarkable but it doesn't have a static libstdc++ library (which the koreader build seems to require).

I will take a look at creating a custom toolchain, I think I compared the kobo/remarkable SoCs and decided that there wasn't much difference in the hardware really, but the remarkable is running a much newer version of the Linux kernel so possibly it's worth building a custom toolchain for that reason.

pazos: [remarkable's cpuinfo](https://remarkablewiki.com/tech/specs) is exactly the same as the [Kobo Forma](https://www.mobileread.com/forums/showpost.php?p=3793588&postcount=14). Not sure about kernel/libc versions in these devices and if is possible to make a new target for both and/or if makes sense. Pinging @NiLuJe

NiLuJe: Oh, then you're in luck ;).

It's not pluggedd in the koxtoolchain frontend, but the 1.24 branch of my ct-ng tree has a [kobomk7](https://github.com/NiLuJe/crosstool-ng/blob/1.24-kindle/samples/arm-kobomk7-linux-gnueabihf/crosstool.config) sample, which should be *much* closer to that ;).

NiLuJe: You'll just have to tweak the minimum kernel version (or not, as the Mk7 one is lower and still 4.1.x), and double-check the glibc version, as I'm not sure what poky 2.1.3 was using.

Probably something along those lines, since that's similar the the current Kindle (lab126) TC, IIRC.

NiLuJe: Probably best to take it as inspiration only, because it's using mainline GCC instead of the final GCC release, and GCC 8/9/10 is still suffering from severe neon performance regressions.

tcrs:
I've added event generation for USB plug/unplug and battery charging/discharging. Looks like there is a delay of up to a minute or two for the battery status to change after plug/unplug (watching the uevents with `udevadm monitor` shows the same behaviour so I'm pretty sure it's not something wrong in the input code - don't know if this is normal on other devices or if it's specific to the remarkable.

Frenzie:
You mean like this? I never really looked into the raw stuff.
https://github.com/koreader/koreader-base/blob/26bdb0b4604a5eb760e2ff70010cd705c1963aef/input/input-kobo.h#L44-L56

https://github.com/koreader/koreader-base/commit/36771d52db9121dc2b12228b5c0e17194409cfb9

NiLuJe:
If the plug/unplug events themselves are on time, I wouldn't worry about it too much ;).

tcrs:
I generate the plug/unplug events from the charger, which seem to be instantaneous. The charging/not-charging comes from the battery device, and its state seems slow to change. So the plugged and charging status are separate. The state usually goes something like:
```
plugged
...delay..
charging
...
unplugged
...delay...
not-charging
```
But you could get plugged+not-charging if the battery is full probably.

Looks like kindle & kobo have quite different device setup for usb plug/unplug.

I also noticed that the Sony PRS port has a custom input generation function in `input-sony-prstux.h` (which I started from to make the remarkable one), but never opens the `fake_event` input in its frontend device lua file so I don't think it will ever actually be used (in koreader at least).

pazos:
> I also noticed that the Sony PRS port has a custom input generation function in `input-sony-prstux.h` (which I started from to make the remarkable one), but never opens the `fake_event` input in its frontend device lua file so I don't think it will ever actually be used (in koreader at least).

It is in https://github.com/koreader/koreader/pull/4567

Having support for the SonyPRSTux platform is great. But the support wasn't intended for other people outside its contributor. We can safely assume the number of users are <= 1. Maybe I'm wrong, but we lack:

- toolchain definition
- platform instructions to install the software over OEM.

So the only person who is able to give feedback about its state is @v01d and we're not in a hurry to merge PRs because we're sure the author of the port is able to figure out and fix things, even when the fixes never reach this repo.

ffi-cdecl/include/mxcfb-remarkable.h
```diff
@@ -0,0 +1,211 @@
+/*
```
NiLuJe: Now that I've finished bothering you with my pesky tests, I'd recommend updating to the latest FBInk copy ;).

ffi-cdecl/mxcfb_remarkable_decl.c
```diff
@@ -0,0 +1,45 @@
+// standard Linux framebuffer headers
```
NiLuJe: And then update & refresh that accordingly ;).

ffi/framebuffer_mxcfb.lua
```diff
@@ -441,6 +452,12 @@ local function refresh_pocketbook(fb, refreshtype, waveform_mode, x, y, w, h)
     return mxc_update(fb, C.MXCFB_SEND_UPDATE, refarea, refreshtype, waveform_mode, x, y, w, h)
 end
 
+local function refresh_remarkable(fb, refreshtype, waveform_mode, x, y, w, h)
+    local refarea = ffi.new("struct mxcfb_update_data[1]")
+    refarea[0].temp = C.TEMP_USE_AMBIENT
```
NiLuJe: To match xochitl:

```suggestion
    if waveform_mode == C.WAVEFORM_MODE_DU then
        refarea[0].temp = C.TEMP_USE_REMARKABLE
    else
        refarea[0].temp = C.TEMP_USE_AMBIENT
    end
```

ffi/framebuffer_mxcfb.lua
```diff
@@ -667,6 +684,20 @@ function framebuffer:init()
         self.waveform_night = C.WAVEFORM_MODE_GC16
         self.waveform_flashnight = self.waveform_night
         self.night_is_reagl = false
+    elseif self.device:isRemarkable() then
+        require("ffi/mxcfb_remarkable_h")
+
+        self.mech_refresh = refresh_remarkable
+        self.mech_wait_update_complete = remarkable_mxc_wait_for_update_complete
+
+        self.waveform_fast = C.WAVEFORM_MODE_DU
```
NiLuJe: To match xochitl:

```suggestion
        self.waveform_fast = C.WAVEFORM_MODE_DU
        self.waveform_ui = C.WAVEFORM_MODE_GL16
        self.waveform_flashui = C.WAVEFORM_MODE_GC16
        self.waveform_full = C.WAVEFORM_MODE_GC16
        self.waveform_partial = C.WAVEFORM_MODE_GL16
        self.waveform_night = C.WAVEFORM_MODE_GC16
        self.waveform_flashnight = self.waveform_night
        self.night_is_reagl = false
```

ffi/framebuffer_mxcfb.lua
```diff
@@ -667,6 +684,20 @@ function framebuffer:init()
...
+        self.waveform_ui = C.WAVEFORM_MODE_AUTO
+        self.waveform_flashui = self.waveform_ui
+        self.waveform_full = C.WAVEFORM_MODE_GC16
+        self.waveform_partial = C.WAVEFORM_MODE_AUTO
+        self.waveform_night = C.WAVEFORM_MODE_GC16
+        self.waveform_flashnight = self.waveform_night
+        self.night_is_reagl = false
```
NiLuJe: To match xochitl:

```suggestion
        self.waveform_fast = C.WAVEFORM_MODE_DU
        self.waveform_ui = C.WAVEFORM_MODE_GL16
        self.waveform_flashui = C.WAVEFORM_MODE_GC16
        self.waveform_full = C.WAVEFORM_MODE_GC16
        self.waveform_partial = C.WAVEFORM_MODE_GL16
        self.waveform_night = C.WAVEFORM_MODE_GC16
        self.waveform_flashnight = self.waveform_night
        self.night_is_reagl = false
```

ffi/mxcfb_remarkable_h.lua
```diff
@@ -0,0 +1,63 @@
+local ffi = require("ffi")
```
NiLuJe: To be regenerated ;).

ffi-cdecl/include/mxcfb-remarkable.h
tcrs: Good plan, done

ffi-cdecl/mxcfb_remarkable_decl.c
tcrs: Done

ffi/framebuffer_mxcfb.lua
tcrs: Done

ffi/framebuffer_mxcfb.lua
tcrs: Done

ffi/mxcfb_remarkable_h.lua
tcrs: Done

tcrs:
Thanks @NiLuJe for sorting out the waveform mapping & related stuff.

I can remove the `luarocks --lua-version` change as it sounds like that isn't suitable for the build setup expected (I'll keep it locally for my machine). Is it OK to rebase the branch this pull request is from?

Frenzie:
Should be fine, here at least. :-)

tcrs:
Rebased on latest master, removed `--lua-version` build change

Makefile
```diff
@@ -95,7 +95,7 @@ libs: \
 $(OUTPUT_DIR)/libs/libkoreader-input.so: input/*.c input/*.h $(if $(KINDLE),$(POPEN_NOSHELL_LIB),)
 	@echo "Building koreader input module..."
 	$(CC) $(DYNLIB_CFLAGS) -I$(POPEN_NOSHELL_DIR) -I./input \
-		$(if $(KOBO),-DKOBO,) $(if $(KINDLE),-DKINDLE,) $(if $(POCKETBOOK),-DPOCKETBOOK,) $(if $(SONY_PRSTUX),-DSONY_PRSTUX,) $(if $(CERVANTES),-DCERVANTES,)\
+		$(if $(KOBO),-DKOBO,) $(if $(KINDLE),-DKINDLE,) $(if $(POCKETBOOK),-DPOCKETBOOK,) $(if $(SONY_PRSTUX),-DSONY_PRSTUX,) $(if $(REMARKABLE),-DREMARKABLE,) $(if $(CERVANTES),-DCERVANTES,)\
```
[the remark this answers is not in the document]
Frenzie: Oh wait, I didn't realize that in the rebasing process you didn't integrate the tab/order commits into the older ones.

Makefile
```diff
@@ -29,7 +29,8 @@ all: $(OUTPUT_DIR)/libs $(if $(ANDROID),,$(LUAJIT)) \
 		$(if $(or $(DARWIN),$(WIN32),$(ANDROID),$(UBUNTUTOUCH),$(APPIMAGE)),,$(OUTPUT_DIR)/dropbear) \
 		$(if $(or $(KINDLE),$(KOBO),$(CERVANTES)),$(OUTPUT_DIR)/sftp-server,) \
 		$(if $(or $(DARWIN),$(WIN32)),,$(OUTPUT_DIR)/tar) \
-		$(if $(or $(KINDLE),$(KOBO),$(CERVANTES)),$(OUTPUT_DIR)/fbink,) \
+		$(if $(or $(KINDLE),$(KOBO),$(CERVANTES),$(REMARKABLE)),$(OUTPUT_DIR)/fbink,) \
```
Frenzie: That being said, my order comment applies more broadly.

Makefile
```diff
@@ -95,7 +97,7 @@ libs: \
 $(OUTPUT_DIR)/libs/libkoreader-input.so: input/*.c input/*.h $(if $(KINDLE),$(POPEN_NOSHELL_LIB),)
 	@echo "Building koreader input module..."
 	$(CC) $(DYNLIB_CFLAGS) -I$(POPEN_NOSHELL_DIR) -I./input \
-		$(if $(KOBO),-DKOBO,) $(if $(KINDLE),-DKINDLE,) $(if $(POCKETBOOK),-DPOCKETBOOK,) $(if $(SONY_PRSTUX),-DSONY_PRSTUX,) $(if $(CERVANTES),-DCERVANTES,)\
+		$(if $(KOBO),-DKOBO,) $(if $(KINDLE),-DKINDLE,) $(if $(POCKETBOOK),-DPOCKETBOOK,) $(if $(REMARKABLE),-DREMARKABLE,) $(if $(SONY_PRSTUX),-DSONY_PRSTUX,) $(if $(CERVANTES),-DCERVANTES,)\
```
Frenzie: Also here

thirdparty/fbink/CMakeLists.txt
```diff
@@ -23,33 +23,36 @@ elseif($ENV{CERVANTES})
...
+    if(UTILS_TARGET)
         set(INSTALL_CMD3 ${CMAKE_COMMAND} -E rename ${SOURCE_DIR}/Release/fbdepth ${BINARY_DIR}/fbdepth)
     endif()
 endif()
 
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME
     https://github.com/NiLuJe/FBInk.git
-    tags/v1.20.3
+    origin/master
```
Frenzie: We really do prefer tags or specific commits.

Frenzie: master is a moving target. In this case we can prod @NiLuJe to create a new tag, generally speaking just use a hash like b06bd399611875fe98b52e5fdadd6525872dde5c.

NiLuJe:
Yeah, I'll tag today or tomorrow ;).

thirdparty/fbink/CMakeLists.txt
NiLuJe: I've just tagged v1.21.0 ;).

tcrs: Thanks, updated to use v1.21.0

Makefile
tcrs: Reordered

Makefile
tcrs: Reordered

NiLuJe:
And we're cooking with gas! ;).

Thanks!

PR #1024, opened 2020-01-01
[CI] Switch to Clang 9.0.1

Frenzie:
See <https://github.com/koreader/virdevenv/pull/45>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1024)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -117,11 +117,11 @@ jobs:
   emu_clang_ninja:
     <<: *EMU_TPL
     docker:
-      - image: koreader/kobase-clang:0.1.2
+      - image: koreader/kobase-clang:0.1.3
     environment:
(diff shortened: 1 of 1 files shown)


Frenzie:
@NiLuJe Well, looks like the Clang built was hit by that ragel thing. I can add it to the Docker image of course, but I don't understand how this can be a sudden and not necessarily dependable issue. O-o

NiLuJe:
I've had it happen intermittently before, IIRC.

Since it's entirely a Make dependency thing, I'm guessing timestamp shenanigans (or shitty timekeeping in some of the CI bots)?

Frenzie:
Well, the md5 test still isn't working. Nothing too surprising there but at least the compilation worked out. :-P

PR #896, opened 2019-04-19
Bump a few third party libraries

NiLuJe:
• HarfBuzz 2.4.0
• k2pfopt to pickup https://github.com/koreader/libk2pdfopt/pull/35
• libpng 1.6.37
• FBInk 1.15.0

PR #1025, opened 2020-01-03
xtext: fix truncation ellipsis position in bidi text

poire-z:
Cut & pasted from https://github.com/koreader/koreader/issues/5359#issuecomment-570490169:

> It looks like it's also the reason for my ellipsis when text is truncated to be on the wrong side:

A correlated reason for that is that, in xtext.cpp, when substituting the text char with an ellipsis, I keep the original chat bidi type and level. So, if the ellipsis is replacing a letter it would stay near the previous letter - but if it is replacing a space, it may be sent away to start or end of string by the Unicode bidi rules L1 (or I1, I2, L1 to L4...).
And when adding an ellipsis to a TextBoxWidget, we always add it on a breakable character (so, a space), so the issue is always there - unlike for TextWidget where we drop the libunibreak rules and can add the ellipsis at any place.

For this text string, starting in logical order with that hebrew word, we'd either get: when replacing a letter:
image
but when replacing a space:
image
Same in RTL UI:
image
image

With this PR (getting the right bidi type of the ellipsis, and assigning to it the bidi level of the previous char), we get:
image
image

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1025)

### DIFF
--- a/xtext.cpp
+++ b/xtext.cpp
@@ -1166,8 +1166,11 @@ public:
         // Temporarily substitute a char from text with an ellipsis if requested.
         // We'll restore the original item from m_text and m_charinfo when done.
         bool orig_idx_substituted = false;
-        uint32_t orig_idx_char; // Backups, to restore later
+        // Backups, to be restored when done
+        uint32_t orig_idx_char;
         unsigned short orig_idx_charinfo_flags; // .width is not used, no need to update it
+        FriBidiCharType orig_idx_bidi_ctype = 0;
+        FriBidiLevel orig_idx_bidi_level = 0;
         if ( idx_to_substitute_with_ellipsis >= 0 && idx_to_substitute_with_ellipsis < m_length ) {
             orig_idx_substituted = true;
             orig_idx_char = m_text[idx_to_substitute_with_ellipsis];
@@ -1176,7 +1179,26 @@ public:
             m_charinfo[idx_to_substitute_with_ellipsis].flags &= ~CHAR_CAN_EXTEND_WIDTH;
             m_charinfo[idx_to_substitute_with_ellipsis].flags &= ~CHAR_CAN_EXTEND_WIDTH_FALLBACK;
             m_charinfo[idx_to_substitute_with_ellipsis].flags &= ~CHAR_IS_CLUSTER_TAIL;
-                // Looks like we can keep all other flags as is
+            m_charinfo[idx_to_substitute_with_ellipsis].flags &= ~CHAR_SCRIPT_CHANGE; // ellipsis is script neutral
+                                                // Looks like we can keep all other flags as is
+            if ( m_has_bidi ) {
+                // Be sure UAX#9 rules I1, I2, L1 to L4 don't move the ellipsis away
+                // from its neighbours
+                orig_idx_bidi_ctype = m_bidi_ctypes[idx_to_substitute_with_ellipsis];
+                orig_idx_bidi_level = m_bidi_levels[idx_to_substitute_with_ellipsis];
+                // If we're substituting start or end (which is what we do in frontend),
+                // set the bidi level of our ellipsis to be start+1 or end-1, so
+                // it's not moved away from next or previous char.
+                if ( idx_to_substitute_with_ellipsis == end-1 && end-1 >= start)
+                    m_bidi_levels[idx_to_substitute_with_ellipsis] = m_bidi_levels[end-1];
+                else if ( idx_to_substitute_with_ellipsis == start && start+1 < end)
+                    m_bidi_levels[idx_to_substitute_with_ellipsis] = m_bidi_levels[start+1];
+                // Also get the real bidi type of our ellipsis (if it is replacing a space,
+                // we don't want it to be FRIBIDI_MASK_WS as fribidi_reorder_line() could
+                // then move it at start or end in visual order)
+                fribidi_get_bidi_types((const FriBidiChar*)(m_text+idx_to_substitute_with_ellipsis), 1,
+                                         (FriBidiCharType*)(m_bidi_ctypes+idx_to_substitute_with_ellipsis));
+            }
         }
         // If m_has_bidi, we need the help of fribidi to visually reorder
         // the text, before feeding segments (of possible different
@@ -1364,6 +1386,10 @@ public:
         if ( orig_idx_substituted ) {
             m_text[idx_to_substitute_with_ellipsis] = orig_idx_char;
             m_charinfo[idx_to_substitute_with_ellipsis].flags = orig_idx_charinfo_flags;
+            if ( m_has_bidi ) {
+                m_bidi_ctypes[idx_to_substitute_with_ellipsis] = orig_idx_bidi_ctype;
+                m_bidi_levels[idx_to_substitute_with_ellipsis] = orig_idx_bidi_level;
+            }
         }
 
         // Convert out s_shape_result to a Lua array.


xtext.cpp
```diff
@@ -1176,7 +1179,26 @@ class XText {
...
+            if ( m_has_bidi ) {
+                // Be sure UAX#9 rules I1, I2, L1 to L4 don't move the ellipsis away
+                // from its neighbours
+                orig_idx_bidi_ctype = m_bidi_ctypes[idx_to_substitute_with_ellipsis];
+                orig_idx_bidi_level = m_bidi_levels[idx_to_substitute_with_ellipsis];
+                // If we're substituting start or end (which is what we do in frontend),
+                // set the bidi level of our ellipsis to be start+1 or end-1, so
+                // it's not moved away from next or previous char.
+                if ( idx_to_substitute_with_ellipsis == end-1 && end-1 >= start)
+                    m_bidi_levels[idx_to_substitute_with_ellipsis] = m_bidi_levels[end-1];
```
Frenzie: I tend to prefer explicit `{}` so that you can't accidentally introduce mistakes and can easily add something without having to worry about adding that stuff.

poire-z: Yes, me too. But sometimes, in some context, for small things, I prefer without, to make these small things span as few lines as possible :)
Like when I could have used `value = condition ? value_yes : value no` but the items are too long to fit on a line, I'm fine with a if/else without `{ }` - rather than making the `? : ` multi-lines, which is less readable.

Frenzie: Sure, I like ternary operators too. :-)

poire-z:
Still some occasional failures with that ragel missing...
Note that (at some point? in some context?), harfbuzz had some fake ragel binary made:
```
https://github.com/harfbuzz/harfbuzz/pull/1638
run: echo "#""!""/bin/true" > /usr/bin/ragel && chmod +x /usr/bin/ragel
```
I don't know anything about ragel, if it's really needed or if we could use that trick...
They are also talking about timestamps and including ragel generated files or not in git. Not much interested trying to understand all that :|
Might be the generated files are there, but their similar timestamps occasionally (for some reason) make the build tolls noticed the ragel file is newer than the generated one and try to recompile some of them?
```
ls -la hb-buffer-deserialize-json* hb-buffer-deserialize-text* hb-number-parser* *machine.*
-rw-r--r-- 1  17293 Sep 27 22:23 hb-buffer-deserialize-json.hh
-rw-r--r-- 1   3651 Sep 27 22:23 hb-buffer-deserialize-json.rl
-rw-r--r-- 1  14639 Sep 27 22:23 hb-buffer-deserialize-text.hh
-rw-r--r-- 1   3287 Sep 27 22:23 hb-buffer-deserialize-text.rl
-rw-r--r-- 1   5367 Oct  5 15:58 hb-number-parser.hh
-rw-r--r-- 1   3193 Oct  5 15:58 hb-number-parser.rl
-rw-r--r-- 1  17888 Sep 27 22:23 hb-ot-shape-complex-indic-machine.hh
-rw-r--r-- 1   3591 Sep 27 22:23 hb-ot-shape-complex-indic-machine.rl
-rw-r--r-- 1  10458 Sep 27 22:23 hb-ot-shape-complex-khmer-machine.hh
-rw-r--r-- 1   2999 Sep 27 22:23 hb-ot-shape-complex-khmer-machine.rl
-rw-r--r-- 1  13050 Sep 27 22:23 hb-ot-shape-complex-myanmar-machine.hh
-rw-r--r-- 1   3331 Sep 27 22:23 hb-ot-shape-complex-myanmar-machine.rl
-rw-r--r-- 1  16718 Sep 27 22:23 hb-ot-shape-complex-use-machine.hh
-rw-r--r-- 1   5732 Sep 27 22:23 hb-ot-shape-complex-use-machine.rl

```
and if some `for f in *.rl ; do touch filename_with_rl_replaced_with_hh ; done` may do the trick.

Frenzie:
But that was some 8 months ago. :-/

It should've already shown up since https://github.com/koreader/koreader-base/pull/917 at the very latest (August) if not https://github.com/koreader/koreader-base/pull/896 (April).

poire-z:
Dunno, circunstances :) git does not store timestamps, so it uses the current time - depending on the CircleCI VM CPU, the order of things in git index or the order of how they end up being created, it's possible `hb-ot-shape-complex-indic-machine.hh` gets created before `hb-ot-shape-complex-indic-machine.rl` - and rl being newer than hh causing the need to use ragel to update the hh (?).

NiLuJe:
It did show up (occasionally) before ;).

Frenzie:
Guess I missed it ^_^

Still weird that it's now doing it a lot more in any case.

poire-z:
Mhhhh. I was ready to PR:
```diff
--- a/thirdparty/harfbuzz/CMakeLists.txt
+++ b/thirdparty/harfbuzz/CMakeLists.txt
@@ -15,6 +15,13 @@ assert_var_defined(FREETYPE_DIR)
 ep_get_source_dir(SOURCE_DIR)
 ep_get_binary_dir(BINARY_DIR)

+if($ENV{CI_BUILD_DIR})
+    # Avoid the need for ragel on CI hosts, in case git clone sets varying timestamps, and
+    # cause the .rl files to be newer than their autogenerated .hh counterparts (which are
+    # also in git): just replace it with /bin/true in "AM_MISSING_PROG([RAGEL], [ragel])".
+    set(PATCH_CMD sh -c "${ISED} \"s|^AM_MISSING_PROG..RAGEL., .ragel..|AM_MISSING_PROG([RAGEL], [/bin/true])|\"  configure.ac")
+endif()
+
 set(CFG_ENV_VAR "CC=\"${CC}\" CXX=\"${CXX}\" CFLAGS=\"${CFLAGS}\" CXXFLAGS=\"${CXXFLAGS}\" LDFLAGS=\"${LDFLAGS}\"")
 set(CFG_ENV_VAR "${CFG_ENV_VAR} FREETYPE_CFLAGS=\"-I${FREETYPE_DIR}/include/freetype2\"")
 set(CFG_ENV_VAR "${CFG_ENV_VAR} FREETYPE_LIBS=\"-L${FREETYPE_DIR} -lfreetype\"")
@@ -43,7 +50,7 @@ ExternalProject_Add(
     ${PROJECT_NAME}
     DOWNLOAD_COMMAND ${CMAKE_COMMAND} -P ${GIT_CLONE_SCRIPT_FILENAME}
     BUILD_IN_SOURCE 1
-    PATCH_COMMAND NOCONFIGURE=1 ./autogen.sh
+    PATCH_COMMAND COMMAND ${PATCH_CMD} COMMAND NOCONFIGURE=1 ./autogen.sh
     CONFIGURE_COMMAND ${CFG_CMD}
     BUILD_COMMAND ${KO_MAKE_RECURSIVE} -j${PARALLEL_JOBS}
     INSTALL_COMMAND ${KO_MAKE_RECURSIVE} -j${PARALLEL_JOBS} install
```

But testing on my machine that it compiles fine without this section executed, I got the ragel missing hell!
Strange thing is that the .hh are missing in the build dir, but they are still there in the git_checkout dir!
```
/koreader/base $ la thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/*machin*
-rw-r--r-- 1  10466 Jan  3 21:07 thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-machinery.hh
-rw-r--r-- 1  17888 Jan  3 21:07 thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-indic-machine.hh
-rw-r--r-- 1   3591 Jan  3 21:07 thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-indic-machine.rl
-rw-r--r-- 1  10458 Jan  3 21:07 thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-khmer-machine.hh
-rw-r--r-- 1   2999 Jan  3 21:07 thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-khmer-machine.rl
-rw-r--r-- 1  13050 Jan  3 21:07 thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-myanmar-machine.hh
-rw-r--r-- 1   3331 Jan  3 21:07 thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-myanmar-machine.rl
-rw-r--r-- 1  16718 Jan  3 21:07 thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-use-machine.hh
-rw-r--r-- 1   5732 Jan  3 21:07 thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-use-machine.rl
/koreader/base $ la thirdparty/harfbuzz/build/x86_64-linux-gnu-debug/harfbuzz-prefix/src/harfbuzz/src/*machin*
-rw-r--r-- 1  10466 Jan  3 21:07 thirdparty/harfbuzz/build/x86_64-linux-gnu-debug/harfbuzz-prefix/src/harfbuzz/src/hb-machinery.hh
-rw-r--r-- 1   3591 Jan  3 21:07 thirdparty/harfbuzz/build/x86_64-linux-gnu-debug/harfbuzz-prefix/src/harfbuzz/src/hb-ot-shape-complex-indic-machine.rl
-rw-r--r-- 1  10458 Jan  3 21:07 thirdparty/harfbuzz/build/x86_64-linux-gnu-debug/harfbuzz-prefix/src/harfbuzz/src/hb-ot-shape-complex-khmer-machine.hh
-rw-r--r-- 1   2999 Jan  3 21:07 thirdparty/harfbuzz/build/x86_64-linux-gnu-debug/harfbuzz-prefix/src/harfbuzz/src/hb-ot-shape-complex-khmer-machine.rl
-rw-r--r-- 1   3331 Jan  3 21:07 thirdparty/harfbuzz/build/x86_64-linux-gnu-debug/harfbuzz-prefix/src/harfbuzz/src/hb-ot-shape-complex-myanmar-machine.rl
-rw-r--r-- 1  16718 Jan  3 21:07 thirdparty/harfbuzz/build/x86_64-linux-gnu-debug/harfbuzz-prefix/src/harfbuzz/src/hb-ot-shape-complex-use-machine.hh
-rw-r--r-- 1   5732 Jan  3 21:07 thirdparty/harfbuzz/build/x86_64-linux-gnu-debug/harfbuzz-prefix/src/harfbuzz/src/hb-ot-shape-complex-use-machine.rl
```
`hb-ot-shape-complex-khmer-machine.hh is` there, but not `hb-ot-shape-complex-myanmar-machine.hh` nor `hb-ot-shape-complex-indic-machine.hh`.
Checking again: they are copied, but disappeared at some point.

```
/koreader/base $ stat thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-khmer-machine.rl thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-myanmar-machine.hh
  File: thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-khmer-machine.rl
  Size: 2999            Blocks: 8          IO Block: 4096   regular file
Access: (0644/-rw-r--r--)  
Access: 2020-01-03 21:07:06.501531090 +0100
Modify: 2020-01-03 21:07:06.501531090 +0100
Change: 2020-01-03 21:07:06.501531090 +0100
 Birth: -
  File: thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-myanmar-machine.hh
  Size: 13050           Blocks: 32         IO Block: 4096   regular file
Access: (0644/-rw-r--r--)  
Access: 2020-01-03 21:07:06.501531090 +0100
Modify: 2020-01-03 21:07:06.501531090 +0100    ======= same as .rl
Change: 2020-01-03 21:07:06.501531090 +0100
 Birth: -

/koreader/base $ stat thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-indic-machine.rl thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-indic-machine.hh
  File: thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-indic-machine.rl
  Size: 3591            Blocks: 8          IO Block: 4096   regular file
Access: (0644/-rw-r--r--) 
Access: 2020-01-03 21:07:07.285515921 +0100
Modify: 2020-01-03 21:07:07.285515921 +0100
Change: 2020-01-03 21:07:07.285515921 +0100
 Birth: -
  File: thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/hb-ot-shape-complex-indic-machine.hh
  Size: 17888           Blocks: 40         IO Block: 4096   regular file
Access: (0644/-rw-r--r--)  
Access: 2020-01-03 21:07:06.501531090 +0100
Modify: 2020-01-03 21:07:06.501531090 +0100    ======= OLDER than .rl
Change: 2020-01-03 21:07:06.501531090 +0100
```

That seems to confirm my theory :)
But hacking that just for CI is not a solution, as it may happen on my own (and so everybody) machine :)

Any idea?
- git clone option/trick to force same timestamps on original git clone?
- we can't touch *hh, as it could cause full recompile
- do the /bin/true substitution only when ragel is not found? (how to test for that in a CMakeLists.txt?

Frenzie:
It's not like I actively wish to avoid ragel in the Docker image or the build prerequisites, at most I'm trying to avoid spending a few minutes updating the Docker image(s) for what feels non-problem adjacent.

poire-z:
> - do the /bin/true substitution only when ragel is not found? (how to test for that in a CMakeLists.txt?

Forget that one, it will not help, as the .hh will have been removed (by CMake?) before it's called.

poire-z:
If you can spend these few minutes :) I already spent 2 hours :|

Frenzie:
A few minutes is understating it of course, given that it also involves a fair bit of follow-up effort to push the stuff to Docker Hub, pointing the various CIs to the new images, etc.

But at the basis, provided nothing goes wrong, it's just adding the word ragel [here](https://github.com/koreader/virdevenv/blob/2d19c789dc5ed6775b1e84696bf088c120dd7e2c/docker/ubuntu/baseimage/bootstrap.sh#L16-L17) and starting the image building process.

As an aside, I've never had this ragel program on any of my machines, ever.

NiLuJe:
I have ragel installed, so I wouldn't have noticed it on my end ;).

poire-z:
For those building the emulator who are blocked by this ragel hell, this should get you going:
```
touch base/thirdparty/harfbuzz/build/git_checkout/harfbuzz/src/*hh
rm -rf base/thirdparty/harfbuzz/build/x86_64-linux-gnu-debug/
```

poire-z:
Note that this ragel virus also affects nightly builds building - each day a few are failing.
Seems it started behaving that way 4 days ago, may be a Y2K'20 bug in git :)

Frenzie:
That's weird, GitLab is supposed to send me emails about failing builds.

Frenzie:
But indeed, the evidence quite convincingly points at this 2020 thing somehow causing it.

yparitcher:
I know i am a little late but Ragel should not be needed by harfbuzz see https://github.com/harfbuzz/harfbuzz/commit/abc82b493e301e5588d689baa1434b3ce6577fcd "Add Ragel-generated files to git"

i have never had Ragel and builds are working just fine

NiLuJe:
Yeah, but apparently there's a timestamp/make dependency quirk happening on the Docker bots, and it's just plain easier to simply have ragel and let it regen instead of trying to figure out what's going wonky ;).

Frenzie:
I remain quite puzzled by the fact that 2020 changed the frequency of this timestamp issue from a completely negligible, minuscule chance (i.e., the nightly builds on GitLab all succeeded for weeks if not months on end, barring the occasional failure to download some dependency) to a chance of at least 1 in 3 if not approaching 1 in 2.

Please feel free to figure out why this is happening, but since I wanted to update the custom-built TC anyway it was only a few minutes of additional effort really, and the couple extra megabytes in the image barely matter. It's about as benign a workaround as they come.

I tried to reproduce it locally in the same container some seven or eight times before I did in fact manage to reproduce the issue. I'd primarily suspect something about Docker/containerization itself (effectively unfixable), something about Git (2.7.4) or something about CMake (3.5.1). Playing around with that would be a fairly big timesink. I might still do it in the background but it's definitely something not worth prioritizing experiments with upgraded Git/CMake versions that may not even resolve anything, when just spending a few minutes adding ragel was sure to take care of the more pressing immediate issue that almost half the builds were consistently failing.

Frenzie:
PS But before I do any of that I should also try it at least a dozen times on my regular, uncontainerized environment.

yparitcher:
could it be ccache? i do not use it, but maybe it is messing with the timestamps?

Frenzie:
ccache only messes with the timestamps in ~/.ccache. It would be an extremely severe bug if it messed with timestamps of files it's only supposed to hash, and it's not supposed to hash ragel input files at all. Only regular C/C++ files.

I was, however, able to reproduce the problem outside of Docker too. This means there's no point to mucking about with updated Git or CMake, nor in blaming containers.

At first I thought the problem might be introduced when CMake copies the Git repo. CMake apparently resolves the original timestamps with a resolution of only one second.

### CMake's handiwork
```sh
frans@frenzie-desktop:~/src/kobo/koreader/base/thirdparty/harfbuzz/build/x86_64-linux-gnu-debug/harfbuzz-prefix/src/harfbuzz/src$ ls -l --time-style=full-iso
total 4452
-rwxr-xr-x 1 frans frans    742 2020-01-07 17:38:48.000000000 +0100 check-c-linkage-decls.sh
-rwxr-xr-x 1 frans frans    463 2020-01-07 17:38:48.000000000 +0100 check-externs.sh
-rwxr-xr-x 1 frans frans    686 2020-01-07 17:38:48.000000000 +0100 check-header-guards.sh
-rwxr-xr-x 1 frans frans   1093 2020-01-07 17:38:48.000000000 +0100 check-includes.sh
-rwxr-xr-x 1 frans frans    907 2020-01-07 17:38:48.000000000 +0100 check-libstdc++.sh
-rwxr-xr-x 1 frans frans    989 2020-01-07 17:38:48.000000000 +0100 check-static-inits.sh
-rwxr-xr-x 1 frans frans   1553 2020-01-07 17:38:48.000000000 +0100 check-symbols.sh
-rw-r--r-- 1 frans frans   1528 2020-01-07 17:38:48.000000000 +0100 dump-indic-data.cc
-rw-r--r-- 1 frans frans   1431 2020-01-07 17:38:48.000000000 +0100 dump-khmer-data.cc
-rw-r--r-- 1 frans frans   1540 2020-01-07 17:38:48.000000000 +0100 dump-myanmar-data.cc
-rw-r--r-- 1 frans frans   1343 2020-01-07 17:38:48.000000000 +0100 dump-use-data.cc
-rwxr-xr-x 1 frans frans   6824 2020-01-07 17:38:48.000000000 +0100 gen-arabic-table.py
-rwxr-xr-x 1 frans frans    707 2020-01-07 17:38:48.000000000 +0100 gen-def.py
-rwxr-xr-x 1 frans frans   1716 2020-01-07 17:38:48.000000000 +0100 gen-emoji-table.py
-rwxr-xr-x 1 frans frans   6730 2020-01-07 17:38:48.000000000 +0100 gen-indic-table.py
-rwxr-xr-x 1 frans frans   1427 2020-01-07 17:38:48.000000000 +0100 gen-os2-unicode-ranges.py
-rwxr-xr-x 1 frans frans  33302 2020-01-07 17:38:48.000000000 +0100 gen-tag-table.py
-rwxr-xr-x 1 frans frans   5418 2020-01-07 17:38:48.000000000 +0100 gen-ucd-table.py
-rwxr-xr-x 1 frans frans  15357 2020-01-07 17:38:48.000000000 +0100 gen-use-table.py
-rwxr-xr-x 1 frans frans   7765 2020-01-07 17:38:49.000000000 +0100 gen-vowel-constraints.py
[…]
-rw-r--r-- 1 frans frans  23397 2020-01-07 17:38:49.000000000 +0100 hb-aat-layout.cc
-rw-r--r-- 1 frans frans  24927 2020-01-07 17:38:48.000000000 +0100 hb-aat-layout-common.hh
[…]
-rw-r--r-- 1 frans frans  30828 2020-01-07 17:38:49.000000000 +0100 hb-algs.hh
-rw-r--r-- 1 frans frans  11860 2020-01-07 17:38:49.000000000 +0100 hb-array.hh
-rw-r--r-- 1 frans frans  10428 2020-01-07 17:38:48.000000000 +0100 hb-atomic.hh
-rw-r--r-- 1 frans frans   4211 2020-01-07 17:38:49.000000000 +0100 hb-bimap.hh
-rw-r--r-- 1 frans frans  16950 2020-01-07 17:38:49.000000000 +0100 hb-blob.cc
```

### The original Git clone
```
frans@frenzie-desktop:~/src/kobo/koreader/base/thirdparty/harfbuzz/build/git_checkout/harfbuzz/src$ ls -l --time-style=full-iso
total 3972
-rwxr-xr-x 1 frans frans    742 2020-01-07 17:38:48.889331936 +0100 check-c-linkage-decls.sh
-rwxr-xr-x 1 frans frans    463 2020-01-07 17:38:48.889331936 +0100 check-externs.sh
-rwxr-xr-x 1 frans frans    686 2020-01-07 17:38:48.889331936 +0100 check-header-guards.sh
-rwxr-xr-x 1 frans frans   1093 2020-01-07 17:38:48.889331936 +0100 check-includes.sh
-rwxr-xr-x 1 frans frans    907 2020-01-07 17:38:48.889331936 +0100 check-libstdc++.sh
-rwxr-xr-x 1 frans frans    989 2020-01-07 17:38:48.889331936 +0100 check-static-inits.sh
-rwxr-xr-x 1 frans frans   1553 2020-01-07 17:38:48.889331936 +0100 check-symbols.sh
-rw-r--r-- 1 frans frans   1528 2020-01-07 17:38:48.889331936 +0100 dump-indic-data.cc
-rw-r--r-- 1 frans frans   1431 2020-01-07 17:38:48.889331936 +0100 dump-khmer-data.cc
-rw-r--r-- 1 frans frans   1540 2020-01-07 17:38:48.889331936 +0100 dump-myanmar-data.cc
-rw-r--r-- 1 frans frans   1343 2020-01-07 17:38:48.889331936 +0100 dump-use-data.cc
-rwxr-xr-x 1 frans frans   6824 2020-01-07 17:38:48.889331936 +0100 gen-arabic-table.py
-rwxr-xr-x 1 frans frans    707 2020-01-07 17:38:48.889331936 +0100 gen-def.py
-rwxr-xr-x 1 frans frans   1716 2020-01-07 17:38:48.889331936 +0100 gen-emoji-table.py
-rwxr-xr-x 1 frans frans   6730 2020-01-07 17:38:48.893331899 +0100 gen-indic-table.py
-rwxr-xr-x 1 frans frans   1427 2020-01-07 17:38:48.893331899 +0100 gen-os2-unicode-ranges.py
-rwxr-xr-x 1 frans frans  33302 2020-01-07 17:38:48.893331899 +0100 gen-tag-table.py
-rwxr-xr-x 1 frans frans   5418 2020-01-07 17:38:48.893331899 +0100 gen-ucd-table.py
-rwxr-xr-x 1 frans frans  15357 2020-01-07 17:38:48.893331899 +0100 gen-use-table.py
-rwxr-xr-x 1 frans frans   7765 2020-01-07 17:38:49.037330561 +0100 gen-vowel-constraints.py
```

Clearly the problem is fundamentally caused by some interaction between Make and Git's checkout timestamps, but it's possible that CMake cements it into a problem due to its one second resolution.

But regardless if I'm on the right track or not, I just don't understand what 2020 has to do with anything.

yparitcher:
for reference https://www.gnu.org/software/autoconf/manual/autoconf-2.61/html_node/Timestamps-and-Make.html

>  11.15 Timestamp Resolution and Make
> 
> Traditionally, file timestamps had 1-second resolution, and make used those timestamps to determine whether one file was newer than the other. However, many modern file systems have timestamps with 1-nanosecond resolution. Some make implementations look at the entire timestamp; others ignore the fractional part, which can lead to incorrect results. Normally this is not a problem, but in some extreme cases you may need to use tricks like `sleep 1' to work around timestamp truncation bugs.
> 
> Commands like `cp -p' and `touch -r' typically do not copy file timestamps to their full resolutions (see Limitations of Usual Tools). Hence you should be wary of rules like this:
> 
>      dest: src
>              cp -p src dest
> 
> as dest often appears to be older than src after the timestamp is truncated, and this can cause make to do needless rework the next time it is invoked. To work around this problem, you can use a timestamp file, e.g.:
> 
>      dest-stamp: src
>              cp -p src dest
>              date >dest-stamp
> 

my guess is one in every N checkouts the seconds change is between the .hh and .rl file leading make to rebuild

Frenzie:
Well yes, that's what I said. ;-) But the fact remains that this obscure edge condition suddenly turned into a nearly 1 in 2 chance of failure in 2020, when previously it was a chance of 1 in hundreds or maybe even less.

PR #1032, opened 2020-01-20
Minor third-party bumps

NiLuJe:
* libjpeg-turbo 2.0.4
* LodePNG 20200120

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1032)

### DIFF
--- a/thirdparty/libjpeg-turbo/CMakeLists.txt
+++ b/thirdparty/libjpeg-turbo/CMakeLists.txt
@@ -46,7 +46,7 @@ endif()
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME
     https://github.com/libjpeg-turbo/libjpeg-turbo.git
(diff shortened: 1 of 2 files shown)


Frenzie:
@NiLuJe I can't rerun you workflow (bit annoying, that) so let's see if closing and reopening does it…

NiLuJe:
Once more, with feeling ;).

NiLuJe:
Whee! \o/

(What's left is the usual Clang/MD5 mystery, and https://github.com/koreader/virdevenv/pull/46#issuecomment-570902518 ;)).

Frenzie:
Like I wrote in https://github.com/koreader/virdevenv/issues/25#issuecomment-570903643 I don't think it looks very different from our regular base image, except that it has a bit less cruft in it. It'd be nice to know for sure though.

PR #1034, opened 2020-01-31
Re-enable HW dithering on Kindle

NiLuJe:
Without doing something stupid this time around, like requesting monochromatic output when we really, really shouldn't ;).

c.f., #819 & #820
As well as koreader/koreader#4602

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1034)

### DIFF
--- a/ffi-cdecl/include/einkfb.h
+++ b/ffi-cdecl/include/einkfb.h
@@ -1,3 +1,5 @@
+//  NOTE: Upstream kernels available here: https://www.amazon.com/gp/help/customer/display.html?nodeId=200203720
+
 #ifndef _EINKFB_H
 #define _EINKFB_H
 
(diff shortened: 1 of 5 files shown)


NiLuJe:
Why, yes, I did just realize that I had made a similar, but not quite the same, mistkae in [FBInk](https://github.com/NiLuJe/FBInk/commit/55fd356c582d2063bf2b1abbb2f1e7a8e7e2fce3)... :D.

NiLuJe:
Oh, and also don't wreck B&W dithering (applies to Kobo Mk. 7, too).

Sidenote: q1 dithering looks like ass, though. But it's free.
Legacy Y1 dithering looks better (except on Kindle, where it's a Sierra Lite matrix, which i find doens't work out all that well). But we're not doing that, anyway ^^.

NiLuJe:
Let's see if I got it right this time :D.

PR #1035, opened 2020-01-31
Update SQLite to 3.31.1

NiLuJe:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1035)

### DIFF
--- a/thirdparty/sqlite/CMakeLists.txt
+++ b/thirdparty/sqlite/CMakeLists.txt
@@ -37,7 +37,7 @@ set(SQLITE_VER "3310100")
 ExternalProject_Add(
     ${PROJECT_NAME}
     DOWNLOAD_DIR ${KO_DOWNLOAD_DIR}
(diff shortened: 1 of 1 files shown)


thirdparty/sqlite/CMakeLists.txt
```diff
@@ -33,12 +33,12 @@ set(CFG_CMD "${CFG_CMD} --disable-readline")
 set(CFG_CMD "${CFG_CMD} --disable-static-shell")
 
 include(ExternalProject)
-set(SQLITE_VER "3300100")
+set(SQLITE_VER "3310100")
```
Frenzie: That's quite right?

```
CMake Error at sqlite-stamp/download-sqlite.cmake:27 (message):
  error: downloading
  'https://www.sqlite.org/2019/sqlite-autoconf-3310100.tar.gz' failed
```

NiLuJe: Oh, crap, we're in 2020 ;).

PR #1036, opened 2020-02-03
Switch to zsync2

NiLuJe:
* Drop legacy zsync
* Switch to zsync2, and only build it where we need it (i.e., !Android eInk devices).
* That implies building & shipping cURL, so, do that, too.
* Minor buildystem tweaks & cleanups along the way

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1036)


NiLuJe:
Enjoy the insane amount of commit it took to get this to build with new CMake, then, worse, old CMake >_<".

Frenzie:
What's that about old CMake? The 3.5.x in Docker isn't set in stone or anything. I do think it's advantageous if potential contributors can just run (the equivalent of) `sudo apt install cmake` but with Ubuntu 16.04 aging, changing that to CMake 3.10.x should be fairly unproblematic by now.

NiLuJe:
Dunno really, just that it was building just fine locally with slightly
less fussing than the bots required ;/.

Makefile.defs
```diff
@@ -803,6 +801,13 @@ OPENSSH_DIR=$(CURDIR)/$(OPENSSH_BUILD_DIR)/openssh-prefix/src/openssh-build
 FBINK_BUILD_DIR=$(THIRDPARTY_DIR)/fbink/build/$(MACHINE)
 FBINK_DIR=$(CURDIR)/$(FBINK_BUILD_DIR)/fbink-prefix/src/fbink-build
 
+CURL_BUILD_DIR=$(THIRDPARTY_DIR)/curl/build/$(MACHINE)
+CURL_DIR=$(CURDIR)/$(CURL_BUILD_DIR)/curl-prefix/src/curl-build
+CURL_LIB=$(OUTPUT_DIR)/libs/libcurl.so.4
```
Frenzie: While my Mac OS VM kind of broke and I deleted it, you still ought to write something like this. ;-)
```
SQLITE_LIB_EXT=$(if $(WIN32),.dll,$(if $(DARWIN),.dylib,.so))
SQLITE_LIB=$(OUTPUT_DIR)/libs/libsqlite3$(SQLITE_LIB_EXT)
```

thirdparty/cmake_modules/CMakeCross.cmake
```diff
@@ -1,9 +1,12 @@
 # CMake Cross ToolChain config file. Adapted from Debian's dpkg-cross ;).
 # c.f., https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html#cross-compiling-toolchain
+# NOTE: Remember that CMake is a special snowflake that completely *ignores* CPPFLAGS,
+# because of course it does...
+# (https://gitlab.kitware.com/cmake/cmake/issues/12928)
```
Frenzie: I'm surprised that issue was only opened 8 years ago.

Makefile.defs
NiLuJe: I took a shortcut because it's only (cross) built for Linux targets ;).

thirdparty/cmake_modules/CMakeCross.cmake
NiLuJe: :D

NiLuJe: I'm surprise no-one decided this was grounds to throw it into a fire, while autotools actually shouts at you for using CPPFLAGS wrong.

Makefile.defs
Frenzie: cURL is the kind of thing that feels like it could be more generally useful. /shrug

Frenzie: Anyway, just a thought. No objections. ;-)

NiLuJe: True ;).  7712b66

PR #1037, opened 2020-02-04
Update LuaJIT

NiLuJe:
* 2020/02/02

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1037)


NiLuJe:
I'll leave the Android side of things to @pazos if nothing implodes first ^^.

NiLuJe:
Now includes a free zsync2 cleanup!

NiLuJe:
Okay, this should be ready, I'm done w/ zsync2.

We should now have the same latest bugfixes & performance improvement than we had before.

NiLuJe:
Actually tested on device, it still handles the 2019.12 -> 2020.01 (I didn't break it, yay!).

NiLuJe:
Also tested the LuaJIT update, nothing blew up either at first glance ;p.

PR #1038, opened 2020-02-06
Update zsync2

NiLuJe:
* Rebase against zlib 1.2.11
* Workaround the actual issue that made us switch to zsync2 in the first place (which just happened to be masked by zsync2's range requests optimization). (Spoiler alert: [ugly hack inside](https://github.com/NiLuJe/zsync2/commit/e98a34570a9159ac86c9be8da84362b3a916e2b8)).
* Also fixes a dodgy EoF check in the process.
* Update FBInk to allow it to more easily digest zsync2's output so we can print the output live.
* Update zsync2's logging to play nice w/ said FBInk hack :D.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1038)


NiLuJe:
@Frenzie: the patch is starting to get a bit insane, should I simply switch the repo to my fork?

Frenzie:
@NiLuJe Yeah, probably. You could also put it in koreader/zsync2 or something if that makes sense.

Is that workaround something that can be contributed upstream?

NiLuJe:
I'll do a recap of my changes for upstream eventually, but this bit is
prooobaaably not the sanest, so, we'll see ;).

NiLuJe:
@Frenzie: Yeah, the plan is to see if nothing breaks on our end, then recap the changes to upstream, and depending on if/what gets merged, move to a clean rebased tree in a repo relocated under the koreader wing ;).

NiLuJe:
Okay, for real done messing with zsync2 this time ;p.

poire-z:
Is it normal/unavoidable/ugly :) to have that libzsyn2.a in a lib/static/ instead of in libs/ like all the others libs (even if it's a .a and not a .so, but is it right to have a .a ?!)

<kbd>image</kbd>
(That's in the kobo nightly.)

NiLuJe:
@poire-z : [Way ahead of you ;p](https://github.com/koreader/koreader-base/pull/1044)

PR #1039, opened 2020-02-07
unbreak KT4 dithering (broken by #1034)

yparitcher:
@NiLuJe 
KT4 doesnt support this HW dithering.
#1034 causes the display to be faded (i almost thought i damaged my screen by mistake)

if you think there is anything else that can be tried i am happy to help/test

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1039)

### DIFF
--- a/ffi/framebuffer_mxcfb.lua
+++ b/ffi/framebuffer_mxcfb.lua
@@ -606,7 +606,9 @@ function framebuffer:init()
 
         -- NOTE: Devices on the Rex platform essentially use the same driver as the Zelda platform, they're just passing a slightly smaller mxcfb_update_data struct
         if isZelda or isRex then
-            self.device.canHWDither = yes
+            if not isNightModeChallenged then
+                self.device.canHWDither = yes
+            end
             if isZelda then
                 self.mech_refresh = refresh_zelda
             else


yparitcher:
after further investigation.
it only happens if there are images on the page.
if i then refresh the screen (diagonal swipe) continuously, it happens 9/10 times but not every time.

yparitcher:
screenshots look regular so i had to take a picture

HW dithering 9/10
HW

1/10 look ok 
SW

NiLuJe:
Ugh, looks like the KT4 really is the ugly duckling of the lineup :/.

Is this actually enough to disable HW dithering on your end? I seem to recall that the mxcfb code didn't actually really check that flag :D.

Sidebar: That 90% thing looks like an (unrelated) bug, it should definitely be 100%. Where can you reproduce it? I'm seeing a progress bar, is that the test ePub I posted on MR a few days ago? So, inside CRe, with diagonal swipes?

yparitcher:
On Fri, Feb 7, 2020, at 11:47 AM, NiLuJe wrote:
> Ugh, looks like the KT4 really is the ugly duckling of the lineup :/.

> Is this actually enough to disable HW dithering on your end? I seem to recall that the mxcfb code didn't actually really check that flag :D.

not sure, there might be some other solution, but turning it off does solve the problem
> Sidebar: That 90% thing looks like an (unrelated) bug, it should definitely be 100%. Where can you reproduce it? I'm seeing a progress bar, is that the test ePub I posted on MR a few days ago? So, inside CRe, with diagonal swipes?


everywhere,
 the picture is from the eink tourture png you posted in the remarkable pr. in image viewer
i dont have a good camera to show you right now.
for example if i load a FM page with cover thumbnails on the side of the list, the whole page (including the text) is not rendered clearly.
however before the thumbnail renders the page is fine, only when the thumbnail loads the whole page messes up.
you cant realy see it in the picture i took but the status bar is also fuzzy and not sharp.

my guess the 90/10 might have to do if the hw dithering actually updated the screen as opposed to leaving whatever the software rendered

NiLuJe:
It's done *during* the refresh, and we only request it if there's image content on screen (so the FM being undithered with no thumbnails is perfectly normal). That means it's applied to the full refresh region, so it affecting the progress bar makes sense, too ;).

On the other hand, that also means that refreshing a page w/ image content should *never* display it undithered ^^.

----

TL;DR: This LGTM, I'll tweak the actual mxcfb check later and see if I can reproduce the "dithering sometimes slips through the cracks" thing ^^.

yparitcher:
is it possible that the KT4 does dithering differently, or does it for sure not work?
better pic 
HW dithered:


HW dither disabled:

NiLuJe:
It might expect the flags to be setup differently, but that'd be weird, as the kernel is consistent with other devices. And there's not a lot of leeway on how the flags can be setup to begin with ^^.

On the other hand, we know that the kernel also exposes unsupported waveform modes on this specific device, and that it does a few other things in a very un-Kindle-like manner, so, I'd be inclined to say that the HW actually lacks support for some reason.

This may actually also apply to the full Zelda/Rex range, as I've never seen it being used by the stock system. I'm not quite sure why they'd willingly hobble the HW like that, though.

Ideally, someone with a PW4 or a KOA2 would be needed to confirm the behavior on those ;).

PR #1047, opened 2020-02-12
Oops, make sure we use the C bb on all the eInk devices...

NiLuJe:
Note that it's still blacklisted on PB because of their crappy unreliable fb info on some devices, though ;).

So this will be mainly of interest to Cervantes & reMarkable (ping @pazos / @avsej; @tcrs)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1047)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ $(OUTPUT_DIR)/data:
 
 libs: \
 	$(if $(or $(SDL),$(ANDROID)),,$(OUTPUT_DIR)/libs/libkoreader-input.so) \
-	$(if $(or $(SDL),$(ANDROID),$(KINDLE),$(KOBO),$(POCKETBOOK)),$(OUTPUT_DIR)/libs/libblitbuffer.so,) \
+	$(if $(or $(CERVANTES),$(KINDLE),$(KOBO),$(POCKETBOOK),$(REMARKABLE),$(SONY_PRSTUX)),$(OUTPUT_DIR)/libs/libblitbuffer.so,) \
 	$(if $(APPIMAGE),$(OUTPUT_DIR)/libs/libXss.so.1,) \
 	$(OUTPUT_DIR)/libs/libkoreader-lfs.so \
 	$(OUTPUT_DIR)/libs/libkoreader-djvu.so \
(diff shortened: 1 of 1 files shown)


NiLuJe:
I don't expect any trouble at all on reMarkable, because it's @ 8bpp, so it's essentially the same as on Kindle & Kobo.

But Cervantes is probably still @ 16bpp, and we didn't test that for very long on Kobo, because RGB565 is the worst.

pazos:
> So this will be mainly of interest to Cervantes ...

As far as I am concerned it works fine as is. I'm not too much interested in fiddling with my cervantes besides platform maintenance. 

If you think it is good to have the same bb on all platforms or it can provide *any* benefit for cervantes users I think it is safe to merge. On the worst case we can revert it soon. :+1:

NiLuJe:
@pazos: Ideally, Cervantes would be switching to 8bpp like Kobo/reMarkable. It'll probably need some testing to have fbdepth behave because of NTX board quirks, though (unlike on reMarkable, where it turned out to be straightforward).

Frenzie:
@NiLuJe I never got the impression that the C BB was really any faster than the LuaJIT one? (Except on malfunctioning Android of course.)

NiLuJe:
Try panning in an imagewidget ;).

NiLuJe:
The main gain is arguably in consistency: when the JIT is happy, everything's mostly fin (close/very close to the C perf), but when it isn't, it can be 2 to 10 times slower for no discernible reason.

That's mostly obvious when rotation is involved, or with RGB565, or with alpha-blending.

(Also, dithering is slower in the pure-Lua version, but that's probably mostly because everything's a double).

Frenzie:
> Try panning in an imagewidget ;).

It only seems to "pan" by swipes or something? I don't really use that one.

By "impression" I meant a touch more than just eyeballometric measurements (though not too much more), and I noticed very little meaningful difference. In your terms, probably "very close"/practically identical to "close" in some cases, but that last one not necessarily to the benefit of C… (The C one has been optimized better since.)

> That's mostly obvious when rotation is involved, or with RGB565, or with alpha-blending.

Right, alpha-blending is the only one I noticed and/or recall. Probably didn't look into 16-bit too much if at all either.

NiLuJe:
What I actively measured was blitting CRe's buffer to the framebuffer (i.e., page turns), as that's what I was mainly concerned with. The rest (alpha, image, dithering) is more of an observation than anything ;).

PR #1048, opened 2020-02-13
Switch to custom PB TC

NiLuJe:
c.f. https://github.com/koreader/koxtoolchain/pull/22

Got rid of all the GCC 4.8 patches.
Got rid of the dedicated toolchain target.
Stopped flagging PB as LEGACY (??!!), because that made no sense in the first place anyway.

Will probably need a rebase because I caught https://github.com/koreader/koreader-base/pull/1047 while working on this, so it's currently in here too ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1048)


NiLuJe:
NOTE: I'm not quite sure why the glib shared library is shipped on PB?

I *think*/*hope* that we're linking sdcv statically... (We might not have historically always been, since Kindle ships a glib).

Makefile
```diff
@@ -87,7 +87,7 @@ $(OUTPUT_DIR)/data:
 
 libs: \
 	$(if $(or $(SDL),$(ANDROID)),,$(OUTPUT_DIR)/libs/libkoreader-input.so) \
-	$(if $(or $(SDL),$(ANDROID),$(KINDLE),$(KOBO),$(POCKETBOOK)),$(OUTPUT_DIR)/libs/libblitbuffer.so,) \
+	$(if $(or $(CERVANTES),$(KINDLE),$(KOBO),$(POCKETBOOK),$(REMARKABLE),$(SONY_PRSTUX)),$(OUTPUT_DIR)/libs/libblitbuffer.so,) \
```
Frenzie: Did SDL and Android just disappear? :-P

NiLuJe: Oh, crap. :D. (I c/p pasted the zsync2 lines :D).

NiLuJe:
(Rebased without #1047)

Makefile
NiLuJe: Fixed in #1047, thanks ;).

PR #1049, opened 2020-02-16
xtext: flag tabs, optionally handle tabstops

poire-z:
Handle tabstops in makeLine() only (to get the segment that fits in the targeted width, with each tab width adjusted so the tab reaches the next tabstop - all that in string logical order).
(But not in shapeLine(), which will just report flags: it's the frontend code job to then handle tabstops just like it handles text justification and alignment.)
Buddy frontend PR: https://github.com/koreader/koreader/pull/5870
Will help closing https://github.com/koreader/koreader/issues/5837,

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1049)

### DIFF
--- a/xtext.cpp
+++ b/xtext.cpp
@@ -246,6 +246,7 @@ int Utf8ToUnicode(const char * src,  int srclen, uint32_t * dst, int dstlen, boo
 #define CHAR_IS_PARA_END                 0x0200
 #define CHAR_PARA_IS_RTL                 0x0400 /// to know the line with this char is part
                                                 /// of a paragraph with main dir RTL
+#define CHAR_IS_TAB                      0x1000 /// char is '\t'
 
 // Info, after measure(), about each m_text char
 typedef struct {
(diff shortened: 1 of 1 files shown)


poire-z:
@NiLuJe : feel free to bring this one (which is a no-op without the frontend buddy PR) with when you bump base in front for your 3 previous commits.

NiLuJe:
@poire-z: Waiting on https://github.com/koreader/virdevenv/pull/49 and/or having confirmation that the bots are up to date (@Frenzie ?) ;).

Frenzie:
@NiLuJe What do you mean exactly? The 0.1.x PB image won't quite build the changes in base properly, so it seems to me there's no point switching it to the new 0.2.x image until base is bumped. But unless you changed something in koxtoolchain I didn't notice (in which case I'd have to rebuild it to pick up the changes) the new image is good to go. Just a little `0.2.0` in the CircleCI config here and on GitLab.

NiLuJe:
@Frenzie: Oh, Catch-22 :D. I was waiting for the bump to the new images to bump base :D. (Remember, I don't do Docker, I only have a vague idea of how all of this interacts ^^).

Frenzie:
Practically speaking it's basically just a Linux system like yours or mine that's told to build KOReader once a day. The technical container details are quite elegantly abstracted away by Docker as an interface to the relevant Linux kernel features (cgroups).

On GitLab the pipeline automatically run if you change the config, so from that perspective it's the least amount of work to bump base first followed by updating GitLab. Otherwise you have to bother with canceling the pipeline. ;-) (Or not, but that'd be wasteful.)

PR #1052, opened 2020-02-21
[CI] Mac OS build

Frenzie:
GitHub Actions provides readily available Mac OS (and Windows) builds, so we should prevent the habitual Mac OS breakage from here on out.

Differences from README:
* ccache for obvious reasons (faster builds)
* md5sha1sum caused some conflict with coreutils

<hr>

Hopefully @NiLuJe will be able to help resolve the gettext/glib issue because I have no idea what's going on there.

GH Actions docs: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1052)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,34 @@
+name: test
+
+on: push
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [macos-latest]
+
+    steps:
+      - name: Check out Git repository
+        uses: actions/checkout@v2
+
+      - name: Cache for ccache
+        uses: actions/cache@v1
+        env:
+          cache-name: cache-ccache
+        with:
+          path: ~/.ccache # ccache cache files are stored in `~/.ccache` on Linux/macOS
+          key: ${{ runner.os }}-build-${{ env.cache-name }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+
+      - name: Homebrew install dependencies
+        # Compared to the README, adds ccache for faster compilation times and removes sh5sum to prevent some conflict with coreutils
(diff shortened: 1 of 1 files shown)


NiLuJe:
What glib/gettext issue? ^^

Frenzie:
glib apparently can't find the gettext that was just built right prior to it:

```
configure: error: 
*** You must have either have gettext support in your C library, or use the
*** GNU gettext library. (http://www.gnu.org/software/gettext/gettext.html)

make[3]: *** [glib-prefix/src/glib-stamp/glib-configure] Error 1
make[2]: *** [CMakeFiles/glib.dir/all] Error 2
make[1]: *** [all] Error 2
make: *** [/Users/runner/runners/2.165.2/work/koreader-base/koreader-base/thirdparty/glib/build/x86_64-apple-darwin19.3.0/glib-prefix/src/glib-build/lib/libglib-2.0.dylib] Error 2
```

Frenzie:
Full log: https://github.com/Frenzie/koreader-base/runs/460432876

NiLuJe:
Would ideally need glib's configure config.log ;).

https://github.com/koreader/koreader-base/pull/1000 should take care of the actual build.

I don't remember ever having to modify my PATH, but it's been a while (but since I use ZSH, I certainly wouldn't have recommended updating Bahs's profile ;p).

That said, that might be a hint: export PATH for real, don't rely on a profile that might not be used by the CI.

Frenzie:
Ftr, I simply copied what it said in https://github.com/koreader/koreader/blob/7b1ae4e486b9a920c0553cd654adc08668265116/doc/Building.md#macos ;-)

NiLuJe:
It's entirely possible that I do actually have /usr/local/* in front of my own PATH in my ZSH setup, though, let me check ;).

NiLuJe:
Yep ^^.

```
┌─(niluje@Felicity:s002)────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────(~/Dev/koreader)─┐
└─(3.51:50%:20:20:%)── echo $PATH                                                                                                                                            ──(Fri, Feb 21)─┘
/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
```

NiLuJe:
Which [makes sense](http://trac.ak-team.com/trac/browser/niluje/Configs/trunk/DotFiles/OSX/dot.zlogin) ;).

Frenzie:
@NiLuJe Alright, that was the ticket!

PR #1056, opened 2020-02-29
Tell SDL2 to handle HiDPI shenanigans for us

NiLuJe:
Makes the emulator behave properly on macOS Retina screens, at least.

Contributed by @magicmirror1365 in
https://github.com/koreader/koreader/issues/5909, thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1056)


ffi-cdecl/SDL2_0_decl.c
```diff
@@ -220,6 +220,7 @@ cdecl_const(SDL_WINDOWPOS_UNDEFINED)
 cdecl_const(SDL_WINDOW_FULLSCREEN)
 cdecl_const(SDL_WINDOW_FULLSCREEN_DESKTOP)
 cdecl_const(SDL_WINDOW_RESIZABLE)
+cdecl_const(SDL_WINDOW_ALLOW_HIGHDPI)
```
Frenzie: Btw, do you want to put it in alphabetically?

NiLuJe: I'd actually prefer it in increasing *value* order (i.e., the same order as the header), which I'm not sure is actually the case either :D.

Frenzie: That's an interesting point. RESIZABLE should go in between FULLSCREEN and FULLSCREEN_DESKTOP then. ;-)

magicmirror1365: Fwiw I think they should go in value order too

Frenzie:
Sidebar, GH Actions *is* running the build [here](https://github.com/NiLuJe/koreader-base/actions/runs/47329144) but for some reason it's not showing it in the PR.

NiLuJe:
Sidebar: this made me realize that LuaJIT bitops module accepts > 2 args for the bitops. I'm fairly sure I've left a mess of ugly nested bit.op() stuff in the blitbuffer code that could be cleaned up this way ;p.

Frenzie:
Haha, awesome. :+1:

PR #1053, opened 2020-02-23
Use awk instead of bc for version comparisons

NiLuJe:
The Docke rimages don't ship w/ bc.

As discussed on gitter ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1053)

PR #1060, opened 2020-03-07
Get rid of GNUism in the awk call

NiLuJe:
Apparently >= is hard?

(re #1053).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1060)

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -19,8 +19,9 @@ NDK?=$(TOOLCHAIN_DIR)/android-ndk-r15c
 NDKABI?=14
 
 # Detect LuaRocks 3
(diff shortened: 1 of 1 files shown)


NiLuJe:
(Also switch to non-recursively expanded Make variable,s because those are the worst).

NiLuJe:
I mean, it's in the docs and everything: https://invisible-island.net/mawk/manpage/mawk.html#h3-5_-Expressions-and-operators

:?

NiLuJe:
Okay, that makes slightly much more sense than not having a >= operator :D.

PR #1062, opened 2020-03-12
Make the PB build compatible with awfully old FW

NiLuJe:
i.e., glibc 2.5

Thankfully, that matches legacy Kindle, too, so we're kind of used to it ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1062)


NiLuJe:
(re: https://github.com/koreader/koreader/issues/5939)

NiLuJe:
We might have been confusing LEGACY/POCKETBOOK in some/most before, so it was probably inheriting the Kindle legacy workarounds ;).

PR #1063, opened 2020-03-12
bump crengine: handle block elements inside inline element

poire-z:
Includes https://github.com/koreader/crengine/pull/330
- FB2 footnotes: disable hardcoded handling as in-page 2
- CSS: also support gray colors spelled "grey"
- Text: move past floats when wide image or inlineBox does not fit
- Fix list item marker clipping check
- Ensure nodeDisplayStyleHash change when boxing prevented by cache
- Text: adds LFormattedText.is_reusable
- Handle block elements inside inline element
- Fix getBookmark() on multi-pages embedded block
- (Upstream) Support raw picture in RTF
- (Upstream) Fixed integer overflow in cache size calculations
- Add libCURL's CA bundle to the ignore list https://github.com/koreader/crengine/pull/332
- FB2 footnotes: fix setBoxingWishedButPreventedByCache() https://github.com/koreader/crengine/pull/333
- fb2.css: handle body[name="comments"] as footnotes

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1063)


Frenzie:
Perhaps we'd better wait a couple of days so we can release a 2020.03.2 with https://github.com/koreader/koreader-base/pull/1062.

poire-z:
Fine with me.
(Also, I'm a bit GitHub-blind at the moment, as not much of github works anymore with my prefered browser - preview, edit... and the only other browser I have on this PC is IE11 :) so waiting a bit for this discussion to resurrect: `https://github.com/MrAlex94/Waterfox/issues/1360` and may be have a solution other than switching my world to... current firefox or chrome... fawk u github!)

Frenzie:
Also see https://otter-browser.org/ and https://vivaldi.com/ (and possibly https://www.opera.com/)

poire-z:
Github works fully with all of them?

Frenzie:
I don't use Otter as much as I'd like but it seems okay. Vivaldi (and Opera) work for sure.

My main browser here is Firefox ESR (68) though.

poire-z:
Well, that waterfox issue has revived :)
Turns out github works again for me after I have enabled in about:config `dom.webcomponents.customelements.enabled` - so it's not a User-agent black/white listing issue.

PR #1064, opened 2020-03-13
Update turbo to v2.1.3

NiLuJe:
And stop shipping a useless CA bundle copy

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1064)


NiLuJe:
(Re https://github.com/koreader/crengine/issues/331)

NiLuJe:
Apparently :D.  Not that we're actually using it anyway ^^.

PR #1065, opened 2020-03-14
Rebase MD5 module against upstream

NiLuJe:
(Same upstream as luxl, which is why I fell into this rabbit hole).

My vague hope that this might fix the mysterious Clang bot issues have been crushed.
On the other hand, it doesn't seem to break anything, so, eh.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1065)

### DIFF
--- a/ffi/MD5.lua
+++ b/ffi/MD5.lua
@@ -1,5 +1,6 @@
 --[[--
 MD5 hash library.
+https://github.com/Wiladams/LAPHLibs/blob/master/laphlibs/md5.lua
 
 @module ffi.md5
 ]]
@@ -22,6 +23,7 @@ typedef struct MD5Context {
   unsigned char input[64];
 } MD5_CTX;
 ]]
+local MD5_CTX = ffi.typeof("MD5_CTX");
 
 local function byteReverse(buf, len)
     -- TODO: implement for big-endian architectures?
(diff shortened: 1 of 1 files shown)


ffi/MD5.lua
```diff
@@ -220,14 +228,16 @@ md5_proto.__index = md5_proto
...
+    local p = ffi.cast("const unsigned char *", luastr)
+    MD5Update(self.ctx, p, len)
 end
 
 --- Calculate md5 sum with the state of a md5 hashing instance.
 ---- @string luastr stream to process
 ---- @treturn string md5 sum in Lua string
 function md5_proto:sum(luastr)
-    local buf = ffi.new("char[33]")
+    local buf = ffi.new("unsigned char[33]")
```
Frenzie: Doubtful that it would affect anything, I actually quite explicitly played around with this very change because I wondered if putting in an `unsigned` might help.

Frenzie:
Hold up, this looks new? O_o

```
[ RUN      ] ./spec/base/unit/koptcontext_spec.lua @ 171: KOPTContext module should convert koptcontext to/from table
error: Unable to read ICC workflow
warning: Attempt to read Output Intent failed
./spec/base/unit/koptcontext_spec.lua:180: Expected objects to be the same.
Passed in:
(string) ''
Expected:
(string) 'eng'

stack traceback:
```

ffi/MD5.lua
NiLuJe: It's unsigned on arm anyway ;).

(And mostly always explicitly an uint8_t in upstream's examples).

NiLuJe:
Nope, that one has been happening sporadically since as far back as I can remember ;).

ffi/MD5.lua
Frenzie: Right, I'm just saying I did a few x86_64 experiments in Docker none of which made a difference on Clang-built LuaJIT. ;-) Regardless of cross-platform compatibility/best practices/etc., just to see if I could isolate something, anything at all.

Frenzie: It's still especially confusing to me that locally it's a toss up or possibly as low as 1 in 4 that the problem occurs, while in the CI it's assured to go wrong.

NiLuJe: Yeah, I think I sometimes saw only 2 of the tests failing instead of the three during last night's runs. :?

Frenzie: Oh, it does? Huh, I've always seen all three on CircleCI.

PR #1068, opened 2020-03-19
Update OpenSSL to 1.1.1e

NiLuJe:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1068)


NiLuJe:
Yeah, we only use it for sftp-server & scp.

I've been using the updated versions for a while, I'd just forgotten to bump them here ;).

PR #1072, opened 2020-03-29
Minor win32 fix for libjpeg-turbo

Frenzie:
See <https://github.com/koreader/koreader/issues/4512#issuecomment-605517909>.

Follow-up to <https://github.com/koreader/koreader-base/pull/1071>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1072)

### DIFF
--- a/Makefile.third
+++ b/Makefile.third
@@ -67,15 +67,15 @@ $(LIBUNIBREAK_LIB) $(LIBUNIBREAK_DIR): $(THIRDPARTY_DIR)/libunibreak/*.*
 	cp -fL $(LIBUNIBREAK_DIR)/lib/$(notdir $(LIBUNIBREAK_LIB)) $@
 	chmod 755 $@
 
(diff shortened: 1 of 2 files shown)


Frenzie:
Whoops, I overlooked that this was the first part of the patch command, luckily the android-arm build caught it. :-P

NiLuJe:
Ha, right, not enough context in the diff view to catch that, but I'm familiar with this predicament ;).

Frenzie:
Also the problem is likely not unique. I'll just revert the djvulibre fix/workaround in this PR while possibly pondering it some more. Win32 might be useful for users; not sure it has any generic cross-platform advantages though.
https://github.com/koreader/koreader-base/blob/60c8dc7f43910ccaf0457dd560ba37dceab45246/thirdparty/libzmq/CMakeLists.txt#L30

PR #1074, opened 2020-04-04
Better workaround for HB issues on armv6

NiLuJe:
Closes https://github.com/koreader/koreader/issues/6024
Better workaround than #1046 for https://github.com/koreader/koreader/issues/5780

Added PB to the list, on the off-chance it might help there...

----

Also, don't ship glib on Kindle:

High kablooey potential, because we pull liblipc, which depends on it and
libgthread, so we'd be mixing glib versions between the two...

Regression since #1048

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1074)


NiLuJe:
Not directly related to https://github.com/koreader/koreader/issues/6024, but found out about it while checking smaps for that one ;p.

PR #1078, opened 2020-04-14
thirdparty/gettext 0.20.2

Frenzie:
See <http://savannah.gnu.org/forum/forum.php?forum_id=9716>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1078)

### DIFF
--- a/thirdparty/gettext/CMakeLists.txt
+++ b/thirdparty/gettext/CMakeLists.txt
@@ -27,12 +27,12 @@ if($ENV{ANDROID})
     set(CFG_CMD "${CFG_CMD} && ${ISED} \"s|#define HAVE_PWD_H 1||\" gettext-tools/config.h")
 endif()
 
-set(GETTEXT_VER "0.20.1")
+set(GETTEXT_VER "0.20.2")
 include(ExternalProject)
 ExternalProject_Add(
     ${PROJECT_NAME}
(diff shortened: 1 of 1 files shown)


Frenzie:
Argh, apparently the mirror didn't have it yet and all I see is "rerun workflow" even though I only want to rerun Android?

Frenzie:
Android build fails with:
```
/home/ko/base/thirdparty/gettext/build/arm-linux-androideabi/gettext-prefix/src/gettext/libtextstyle/lib/glib/gstrfuncs.c: In function 'g_ascii_formatd':

/home/ko/base/thirdparty/gettext/build/arm-linux-androideabi/gettext-prefix/src/gettext/libtextstyle/lib/glib/gstrfuncs.c:591:30: error: 'struct lconv' has no member named 'decimal_point'

   decimal_point = locale_data->decimal_point;
```
According to <https://bugs.python.org/issue20305> that's easy to work around with a:
```
#ifdef __ANDROID__
    decimal_point = "."; 
```
I may or may not do that. The gettext version here is pretty irrelevant. Starting with API 21 supposedly it's in Android properly.

Frenzie:
I'll ignore this since it'd probably take at least 10-20 minutes for no benefit.

PR #1081, opened 2020-04-15
move round % to utils

mwoz123:
required for https://github.com/koreader/koreader/pull/6048

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1081)

### DIFF
--- a/ffi/util.lua
+++ b/ffi/util.lua
@@ -658,4 +658,8 @@ function util.template(str, ...)
     return result
 end
 
+function util.roundPercent(percent)
+    return math.floor(percent * 10000) / 10000
+end
+
 return util


mwoz123:
this is common part - I can duplicate it in my Change but prefered to extract common part to util.lua

Frenzie:
But why *base* utils?

mwoz123:
Rounding seems to be one of the basic operations that might be used at different layers of application. 
Or am I not enought familiar with the architecture of Koreader?

Frenzie:
It seems to me like something for the Math module or front util.

https://github.com/koreader/koreader/blob/ba4e175ee63d69086723e3366492eb1c79f505e5/frontend/optmath.lua#L11-L21

https://github.com/koreader/koreader/blob/ba4e175ee63d69086723e3366492eb1c79f505e5/frontend/util.lua#L717-L725

mwoz123:
ok

PR #1082, opened 2020-04-18
bump crengine: text typography by language

poire-z:
Includes https://github.com/koreader/crengine/pull/337 :
- Fix a few clang-tidy warnings
- Add support for 
- XML parsing: add more HTML5 named entities, optimize search
- Text: fix standalone BR not making an empty line
- Fix BR with "display: block" not making an empty line
- Fix hyphens from soft-hyphens not part of highlighted segments
- Use libunibreak for line breaking
- Adds TextLangMan for text typography by language

The added `cre.getTextLangStatus()` will allow us to get a feel of what languages have been seen and which hyph dicts loaded:

<kbd>image</kbd>

Noting a little thing that got me searching for a bug for quite some time:
The TextLangCfg tags we will see there are the ones that were actually seen by the code.
For a same book:
- on initial load or re-rendering: the full book content will be visited, so all the existing lang= attributes tags will appear here
- after exit and load from cache, this list will be smaller, and lang= attributes tags will appear here only when they are met: when you turn pages and meet new sections of the book that are to be rendered and drawn.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1082)


poire-z:
@Frenzie: are you fine if I bump this into frontend later this afternoon?
Is your tag v2020.04.1 enough for you to get binaries to add to the release?

poire-z:
And btw, didn't you need this month to do as `[i18n] Bump translations submodule for v2020.03 release` last month?

Frenzie:
You're right, I forgot that. It doesn't matter for the resulting builds; it's just a matter of syncing up translations with the code if you should want to check out that tag in the future.

Issue #1086, opened 2020-04-23
Dropbear URL broken


Frenzie:
See https://gitter.im/koreader/koreader?at=5ea16d63568e5258e47badb3
```
4905 CMake Error at dropbear-stamp/download-dropbear.cmake:27 (message):
4906   error: downloading
4907   'http://gentoo.osuosl.org/distfiles/dropbear-2018.76.tar.bz2' failed
4908     status_code: 22
4909     status_string: "HTTP response code said error"
4910     log:   Trying 140.211.166.134...
4911   Connected to gentoo.osuosl.org (140.211.166.134) port 80 (#0)
4912   GET /distfiles/dropbear-2018.76.tar.bz2 HTTP/1.1
4913   Host: gentoo.osuosl.org
4914   User-Agent: curl/7.47.0
4915   Accept: */*
4916   
4917   The requested URL returned error: 404 Not Found
4918   Closing connection 0
4919
```

Frenzie:
Possible source: http://deb.debian.org/debian/pool/main/d/dropbear/dropbear_2018.76.orig.tar.bz2

NiLuJe:
Ah, yeah, old dropbear was dropped from Portage.

Debian sounds like a good alternative, yeah ;).

PR #1087, opened 2020-04-23
thirdparty/dropbear: Change URL to Debian mirror

Frenzie:
Fixes <https://github.com/koreader/koreader-base/issues/1086>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1087)


NiLuJe:
See gitter for my answers about why it's built on Debuntu & co ;).

PR #1091, opened 2020-04-28
SDL: Enable Screen saver

### LINKED ISSUE #1090: SDL: Disables screensaver
SDL 2.0 defaults to "disable screensaver", so leaving the emulator running keeps the screensaver from starting (at least it does on a Mac).

The attached files call SDL_EnableScreensaver to fix this.

[SDL2_0.lua.txt](https://github.com/koreader/koreader-base/files/4541553/SDL2_0.lua.txt)
[SDL2_0_h.lua.txt](https://github.com/koreader/koreader-base/files/4541554/SDL2_0_h.lua.txt)

JasonInOttawa:
The default behaviour of SDL 2.0 is to disable the screen saver. This call turns it back on so the emulator does not keep the screen saver from starting automatically.

Closes #1090

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1091)


JasonInOttawa:
Done, I think

PR #1098, opened 2020-05-09
Update to FreeType 2.10.2

NiLuJe:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1098)


poire-z:
> Support of WOFF2 fonts

Would that mean we could easily support embedded .wof fonts in EPUBs ? (I don't think I ever witness them, or them being not handled :) but it's supposed to be supported by EPUB3).

NiLuJe:
We'd need to build Brotli for that to work.

AFAIU, WOFF2 is pretty much a webfont-only format, so I'm not convinced there'd be a real interest in doing so.

Unless the ePub standard does something stupid like recommending WOFF2 fonts over proper OTFs.

PR #1099, opened 2020-05-12
Workaround potential stalls in OpenSSL

NiLuJe:
This effectively disables https://github.com/openssl/openssl/pull/9595 (introduced in OpenSSL 1.1.1d), as I've been noticing numerous cases of stalls well after the early boot, most notably with scp transfers.

We do mainly target shitty old kernels & shitty old glibcs where this kicks in.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1099)


NiLuJe:
(Actually honor CPPFLAGS while we're in there).

NiLuJe:
That one has been bothering me for quite a while, and was incredibly sneaky to catch, because in quite a few instances, strace'ing the scp process was enough to unblock it and slow enough that the relevant select calls never made it to the trace...

PR #1102, opened 2020-05-23
Explicitly disable WOFF2 support in FT

NiLuJe:
Apparently, Android sometimes likes to try to link against an inexistant
brotli otherwise...

Fix https://github.com/koreader/koreader/issues/6178 (thanks to @zwim)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1102)

### DIFF
--- a/thirdparty/freetype2/CMakeLists.txt
+++ b/thirdparty/freetype2/CMakeLists.txt
@@ -15,7 +15,7 @@ ep_get_source_dir(SOURCE_DIR)
 ep_get_binary_dir(BINARY_DIR)
 
 set(CFG_ENV_VAR "CC=\"${CC}\" CXX=\"${CXX}\" CFLAGS=\"${CFLAGS}\" CXXFLAGS=\"${CXXFLAGS}\" LDFLAGS=\"${LDFLAGS}\"")
-set(CFG_OPTS "-q --prefix=${BINARY_DIR} --disable-static --enable-shared --with-zlib=no --with-bzip2=no --with-png=no --with-harfbuzz=no --host=\"${CHOST}\"")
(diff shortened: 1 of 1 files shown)


yparitcher:
btw this is not only android.
my emulator builds fine, but my kindle build fails with:
```
koreader/base/thirdparty/freetype2/build/arm-kindlepw2-linux-gnueabi/freetype2-prefix/src/freetype2/src/sfnt/sfwoff2.c:28:10: fatal error: brotli/decode.h: No such file or directory
 #include <brotli/decode.h>
```

This fixes it.

side note CircleCi is not letting me see build logs without logging in anymore, so if a test fails there is no way to see why

yparitcher:
If I delete my system brotli package it works fine, 
somehow freetype's cmake is picking up on the system brotli even when cross compiling

NiLuJe:
Yay CMake. >_<".

Frenzie:
@yparitcher 
> side note CircleCi is not letting me see build logs without logging in anymore, so if a test fails there is no way to see why

Hm, that's definitely annoying. I figured I'd open a topic about it on the CircleCI forum but good @#$@# they want me to set up two factor auth and who knows what to access the account I created three years ago. Grmbl.

NiLuJe:
Hmm, it's a pkg-config check, which makes sense, and it defaults to auto, which also makes sense.

Can't reproduce on my end because I have the PKG_CONFIG searchpaths set properly in my env.

I thought I'd taken care of that here, but it's commented out form some reason...

NiLuJe:
Namely: https://github.com/koreader/koreader-base/blame/bf263ec6544e946e188d91ad72097041cb63ac03/thirdparty/cmake_modules/CMakeCross.cmake#L76

PR #1106, opened 2020-05-30
Tweak RTC alarm disable

NiLuJe:
Do it like busybox & Nickel

The other approach would be a WKALM_RD + WKALM_SET.
I'm not *quite* sure what the previous behavior was actually doing
(No idea what an epoch set to -1 actually does in practice?)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1106)

### DIFF
--- a/ffi-cdecl/rtc_cdecl.c
+++ b/ffi-cdecl/rtc_cdecl.c
@@ -7,6 +7,8 @@
 cdecl_struct(rtc_time)
 cdecl_struct(rtc_wkalrm)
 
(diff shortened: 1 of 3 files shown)


NiLuJe:
Not that anything was ever actually calling `unsetWakeupAlarm`, so the `-1` epoch thingy never actually ran anywhere ;).

Frenzie:
> (No idea what an epoch set to -1 actually does in practice?)

Must've done something. :-P Might just be a shortcut to get something prior to the current time.

PR #1108, opened 2020-06-02
Update zsync2

NiLuJe:
Resync w/ upstream
Properly handle filesystems that don't support hard-links, like legacy
zsync used to.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1108)

### DIFF
--- a/thirdparty/zsync2/CMakeLists.txt
+++ b/thirdparty/zsync2/CMakeLists.txt
@@ -72,7 +72,7 @@ endif()
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME
     https://github.com/NiLuJe/zsync2.git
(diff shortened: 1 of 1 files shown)


NiLuJe:
This ought to fix https://www.mobileread.com/forums/showthread.php?t=330375

(That codepath is only triggered if a previous incomplete download is found. Logic seemed dodgy, so I basically replaced it with the original legacy one, which does make sense.
And restore the original comments to explain it, because, admittedly, not everybody remembers how and why link() fails on vfat ;)).

NiLuJe:
Actual changeset: https://github.com/NiLuJe/zsync2/compare/7565c081d581706c6a7859cf41eec976811180b6...e618d18f6a7cbf350cededa17ddfe8f76bdf0b5c

PR #1115, opened 2020-06-11
A few third party updates

NiLuJe:
* LuaJIT (should be fairly tame)
* LuaFileSystem (ditto)
* LodePNG (ditto)
* CRe (for the ePub3 metadata stuff)

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1115)


NiLuJe:
Huh. Tests recap is only showing the ones that hand't yet run to completion on my account before the PR.

(But, FWIW, all green except the usual).

PR #1123, opened 2020-06-24
bump crengine: add support for <ruby> elements

poire-z:
Includes https://github.com/koreader/crengine/pull/347 :
- Fix some non-recursive subtree walkers
- Text: skip formatting optimisations when inlineBoxes
- CSS parsing: parse selectors starting with [attrib]
- Use T="" for internal elements attributes
- CSS: re-order css_d_* (display) enum for easier checking
- Rendering methods: remove erm_table_cell
- Rendering methods: remove erm_table_caption
- Rendering methods: remove erm_list_item
- getRenderedWidths(): better estimate table width
- styleToTextFmtFlags(): adds is_block parameter
- Adds support for "display: ruby" and `<ruby>` elements
- Fix prev/next-VisibleWord-Start/End() with CJK text

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1123)


poire-z:
(Thanks. Possibly the first time I come back without any bug or annoyance noticed, and nothing to hurry fixing :)

PR #1128, opened 2020-06-26
remove W>H startup rotation

yparitcher:
It was broken (only the FB was rotated not the touch) and prevents using a W>H device

this keeps `isAlwaysPortrait()` for devices that have a bad starting orientation (can be set in the emulator with EMULATE_READER_FORCE_PORTRAIT)

to test try `kodev run -h=600 -w=800`. Fixes: koreader/koreader#5910 may need koreader/koreader#6309 to be complete

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1128)

### DIFF
--- a/ffi/framebuffer.lua
+++ b/ffi/framebuffer.lua
@@ -78,19 +78,15 @@ function fb:init()
     -- asking the framebuffer for orientation is error prone,
     -- so we do this simple heuristic (for now)
     self.screen_size = self:getSize()
-    if self.screen_size.w > self.screen_size.h then
-        self.native_rotation_mode = self.ORIENTATION_LANDSCAPE
+    if self.screen_size.w > self.screen_size.h and self.device:isAlwaysPortrait() then
         self.screen_size.w, self.screen_size.h = self.screen_size.h, self.screen_size.w
-        if self.device:isAlwaysPortrait() then
-            -- some framebuffers need to be rotated counter-clockwise (they start in landscape mode)
-            self.debug("enforcing portrait mode by doing an initial rotation")
-            self.bb:rotate(-90)
-            self.blitbuffer_rotation_mode = self.bb:getRotation()
-            self.native_rotation_mode = self.ORIENTATION_PORTRAIT
-        end
-    else
+        -- some framebuffers need to be rotated counter-clockwise (they start in landscape mode)
+        self.debug("enforcing portrait mode by doing an initial rotation")
+        self.bb:rotate(-90)
+        self.blitbuffer_rotation_mode = self.bb:getRotation()
         self.native_rotation_mode = self.ORIENTATION_PORTRAIT
     end
+    self.native_rotation_mode = self.ORIENTATION_PORTRAIT
     self.cur_rotation_mode = self.native_rotation_mode
 end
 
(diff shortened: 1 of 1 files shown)


NiLuJe:
Note that this is an expected quirk on Kobo devices (the touch being off by +1/-1 * 90°).

NiLuJe:
Now that I actually think about it, we probably don't even need `isAlwaysPortrait` anymore on Kobo, since fbdepth should rotate us to Portrait on startup (among other things, like getting the touch translation right, specifically to avoid the rotated blitbuffer performance penalty I mentioned in the other issue).

ffi/framebuffer.lua
```diff
@@ -78,19 +78,15 @@ function fb:init()
...
-            self.bb:rotate(-90)
-            self.blitbuffer_rotation_mode = self.bb:getRotation()
-            self.native_rotation_mode = self.ORIENTATION_PORTRAIT
-        end
-    else
+        -- some framebuffers need to be rotated counter-clockwise (they start in landscape mode)
+        self.debug("enforcing portrait mode by doing an initial rotation")
+        self.bb:rotate(-90)
+        self.blitbuffer_rotation_mode = self.bb:getRotation()
         self.native_rotation_mode = self.ORIENTATION_PORTRAIT
```
NiLuJe, line 87: That one line can go poof, since it's already set to this outside the branch.

NiLuJe: I'd also replace the debug message with a sterner warning to device maintainers that this should be fixed properly to avoid the rotated blitbuffer blitting penalty.

e.g.,

```
io.write("FB: Enforcing portrait mode by doing an initial BB rotation")                                                                                                                                                                                                                                                                                         
         io.flush()                                                                                                                                                                                                                                                                                                                                                      
         self.debug("FB: This prevents the use of blitting optimizations. This should instead be fixed on the device's side on startup.") 
```

(The first one is public, matching the framebuffer_linux logging, because it's good to know in bug reports. The other one is debug only for device maintainers).

NiLuJe:
Oh, and before I forget: I checked, and, no, we can't get rid of `isAlwaysPortrait` on Kobo, because it *could* theoretically still be tripped in some extreme corner-cases (most probably: old devices, 8bpp switch disabled).

yparitcher:
I also deleted `setScreenMode` as it is no longer used

yparitcher:
Can I commit this so i can update base for koreader/koreader#6309?

NiLuJe:
> Can I commit this so i can update base for [koreader/koreader#6309](https://github.com/koreader/koreader/pull/6309)?

Sorry, I'd missed this comment.

Do you want to bundle the `DLANDSCAPE_CLOCKWISE_ROTATION` removal before I merge, or do it in a separate PR?

yparitcher:
> > Can I commit this so i can update base for [koreader/koreader#6309](https://github.com/koreader/koreader/pull/6309)?
> 
> Sorry, I'd missed this comment.
> 
> Do you want to bundle the `DLANDSCAPE_CLOCKWISE_ROTATION` removal before I merge, or do it in a separate PR?

https://github.com/koreader/koreader/pull/6309#issuecomment-651507599

PR #1131, opened 2020-06-30
[CI] Use Clang 10

Frenzie:
See <https://github.com/koreader/virdevenv/pull/52>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1131)


Frenzie:
@NiLuJe Still happily failing at MD5. ;-)

PR #1132, opened 2020-07-04
Minor clarification on rotation constants and buffer rotation

NiLuJe:
Following the discussions in https://github.com/koreader/koreader/pull/6309 ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1132)


NiLuJe:
I forgot to merge that?

Yay, me.

PR #1144, opened 2020-07-17
workflow: use 10.13 for CI

pazos:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1144)


pazos:
eh, whatever

PR #1145, opened 2020-07-19
remarkable: fix button-listen when resuming from suspend

tcrs:
When the home button is used to wake the device from suspend the evdev
events have surprising time stamps. There is a pair of press & release
events, but the press has a timestamp from when the device went into
suspend and the release has a timestamp from after. This means that the
wake keypress will launch koreader as button-listen thinks the key was
held down for the whole time the device was suspended!

Instead, just record CLOCK_BOOTTIME timestamps for the events and use
those instead of the event timestamps. CLOCK_BOOTTIME is monotonic but
includes time spend in suspend.

Also keep track of whether button-listen saw a key press before each key
release. Probably not strictly required but seemed sensible to add.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1145)

### DIFF
--- a/button-listen.c
+++ b/button-listen.c
@@ -2,12 +2,13 @@
 #include <stdio.h>
 #include <time.h>
 #include <stdlib.h>
+#include <stdbool.h>
 
-static double timeval_diff(struct timeval const* start, struct timeval const* end)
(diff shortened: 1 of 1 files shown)


tcrs:
Fixes https://github.com/koreader/koreader/issues/6029

PR #1149, opened 2020-07-21
[fix, CI] Deprecate MD5 module in favor of sha2

Frenzie:
This will require a few minor changes in front. Fixes the tests in the CircleCI clang build.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1149)


Frenzie:
@NiLuJe Alternatively I could transparently offer the same API to front, but the usage is so limited it doesn't seem at all worthwhile.

Alternatively alternatively one could implant the apparently slightly better LuaJIT algorithm in the old MD5 module. Only for masochists. xD

Issue #1148, opened 2020-07-21
Is it doable/reasonable to add opencv?


Galunid:
I was thinking about writing a module that would allow users to double tap a panel in manga to zoom to a selected panel(basically cut out part of the image and display it).
**Base image:**
![image1](https://user-images.githubusercontent.com/10298730/88061322-a6e02980-cb67-11ea-9b89-e774dc75fc4d.png)
**Cut out panels:**
![image4](https://user-images.githubusercontent.com/10298730/88061334-a8a9ed00-cb67-11ea-8c66-eafa83c8effc.png)
![image3](https://user-images.githubusercontent.com/10298730/88061331-a8a9ed00-cb67-11ea-9482-bbb3e9bedbf4.png)
![image2](https://user-images.githubusercontent.com/10298730/88061327-a8115680-cb67-11ea-8165-b98fef40f1a7.png)

To get this result I used opencv library and a short python script https://pastebin.com/hKKGgzNw. My question is whether it's reasonable to include OpenCV, and if so, how would I go around it?

Frenzie:
OpenCV seems like it might be rather excessively large for our purposes, even setting aside some potential doubts about its usability on a poor little Kindle or Kobo.

But perhaps more important, Leptonica should mostly likely do the same thing faster and we already have it. In fact I implicitly mentioned it in https://github.com/koreader/koreader/pull/6408#issuecomment-661877691 although I didn't speak of some of the underlying technology. In fact the current zoom in page flipping mode already functions roughly the way you sketched. ;-)

Some examples:
https://github.com/koreader/koreader-base/blob/0e6406696f233d4f7033803bcd38e4b7559e7e40/ffi/koptcontext.lua#L232-L258

https://github.com/koreader/koreader-base/blob/0e6406696f233d4f7033803bcd38e4b7559e7e40/ffi/koptcontext.lua#L260-L335

Clearly the fact that it's hidden in page flipping mode means almost no one knows it exists. So there are multiple issues. 

* The zoom to box feature is great *as a starting point* and it's neat most of the time but it doesn't always work out. You want much freer zoom.
  Some details on things we might want in free zoom here: https://github.com/koreader/koreader/issues/5524
* Ideally all this would somehow be available in the main reader mode easily without having to trigger a mostly hidden special mode.
  I'm not sure page flipping mode still has much purpose with the greatly improved skim widget.

NiLuJe:
basically what @Frenzie said ;).

[...]

PR #1159, opened 2020-08-16
KOPTContext: add comic panel detection function/additional leptonica declarations

Galunid:
Currently a draft, this will be used by panel detection feature in .cbz/.pdf/.djvu. Not all definitions are newly added leptonica functions are used now, but they'll be used in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1159)


ffi/koptcontext.lua
```diff
@@ -257,6 +257,45 @@ function KOPTContext_mt.__index:findPageBlocks()
     end
 end
 
+function KOPTContext_mt.__index:getPanelFromPage(pos)
+    if self.src.data then
+        local pixs = k2pdfopt.bitmap2pix(self.src, 0, 0, self.src.width, self.src.height)
+        local pixg
+        if leptonica.pixGetDepth(pixs) == 32 then
+            pixg = leptonica.pixConvertRGBToGrayFast(pixs)
```
Frenzie, line 265: Does this mean the output will always be grayscale? If so, that may not be desirable.

Galunid: No, grayscale image is converted later to binary image, then we process the binary image to get coordinates of the panels. Then we cut those panels from color image

Galunid: Right now, it cuts from the gray image, but that was just for testing purposes and that part of the code will be removed later, since this method will probably return coordinates of the panels.

Galunid:
This should be ready for the final review

NiLuJe:
Nothing jumps out, didn't realize leptonica had all this fun stuff available ;). :+1:

poire-z:
At a quick read, I thought it would returns multiple panels and suggest you to rename it to getPanel**s**FromPage(). But no, you do return only one thingy.
So, my question is: what is a pannel ?
And please add your answer as a comment above that function, so future readers don't have to wonder.

And btw, if you're discovering and understanding things in the kopt related stuff (kopt parameters, what some leptonica function expect and return) that was hard to get into, and you felt comments describing what happens would have helped, and if you feel confortable writting a bit of prose, please write what you understood :) 
(Because all this stuff is CJK to me, and some comments would have helped me not immediately retreating when I had a look at that :)

Frenzie (approved):
Looks fine to me.

Galunid:
@poire-z https://github.com/koreader/koreader-base/issues/1148 answers what panel is (basically it's individual enclosed region on a page, the link provides example. I'll try to add more comments. For basic interaction between KOPTContext/Leptonica/Koptinterface take a look at https://github.com/koreader/koreader/issues/6498#issuecomment-674389353. I'll try to write some more in the evening/tomorrow, since I need a break ;-)

poire-z:
Oh, allright. I missed it has a `pos` argument, so it's for getting the panel at pos - and a page can indeed have multiple panels. A small comment like "Get panel that contains pos (a panel is a region surrounded by blank space" (or what it really is technically) will be enough to not have me confused later :) 

And so I guess later in some other context, you might add a getPanel**s**FromPage(page).

> I'll try to write some more in the evening/tomorrow, since I need a break ;-)
What are interested in the most?

Nothing special in mind. It's just that you're probably the first one having a look at how all that works for a long time - so, just don't lose the non-obvious information you're discovering. (That's what I've been doing since I discovered crengine - so I hope I have made it easier to get into for future hackers.)
But don't feel obliged. Just later on, as you meet some non-obvious stuff, write a note for later visitors :)

Frenzie (changes requested):
@Galunid I just noticed you didn't yet add the relevant lines to https://github.com/koreader/koreader-base/blob/8efd649950d296fc79d8ce5ef50060ef815f6f5a/ffi-cdecl/leptonica_cdecl.c

Galunid:
@Frenzie Good catch, thanks!

Frenzie:
@Galunid Finished now?

Galunid:
This one should be finished, although I didn't test the latest change yet. Would it fail to build/crash if I made some typos? I shouldn't, but just in case. I'll try to add some comments to explain how some other koptcontext functions work. I can do that either in this PR, or open a new one, let me know which one you'd prefer.

Frenzie:
A separate PR is fine by me; I'll give this one a test before merging first though.

Galunid:
@Frenzie I just tested it and it should be alright.

PR #1165, opened 2020-08-23
Build KoboUSBMS on Kobo

NiLuJe:
And sprinkle a bit of `CLOEXEC` magic in `framebuffer_linux` while we're at it.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1165)


NiLuJe:
Just checking if the bot builds it properly (libevdev's buildsystem requires Python...).

I'll also tag a proper KoboUSBMS release before merge.

PR #1168, opened 2020-08-25
Implement poll mode for pocketbook.

ezdiy:
This currently lives on the side, switched on by a global from koreader.
The C driver can be nuked some time later, for now it's in to avoid merge conflicts and keep koreader-base working with both frontends for a while.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1168)


ffi/input_pocketbook.lua
```diff
@@ -0,0 +1,147 @@
+local ffi = require("ffi")
+local C = ffi.C
+local inkview = ffi.load("inkview")
+local util = require("ffi/util")
+require("ffi/linux_input_h")
+
+-- luacheck: push
+-- luacheck: ignore
+local EVT_INIT = 21
```
NiLuJe, line 9: I wouldn't mind a reference as to where those magic numbers come from ;). (<inkview.h>?)

ffi/input_pocketbook.lua
```diff
@@ -0,0 +1,147 @@
...
+local EVT_POINTERMOVE = 31
+local EVT_ORIENTATION = 32
+local EVT_MTSYNC = 39
+local EVT_FOREGROUND = 151
+local EVT_BACKGROUND = 152
+
+local PANEL_DISABLED = 0
+-- luacheck: pop
+
+ffi.cdef [[
```
NiLuJe, line 30: Same general idea, if there's a public header somewhere, I'd prefer those handled via a .cdecl like the rest of our FFI stuff.

ffi/util.lua
```diff
@@ -546,6 +546,10 @@ function util.isAndroid()
     return isAndroid
 end
 
+function util.isPocketbook()
+    return lfs.attributes("/ebrmain")
```
NiLuJe, line 550: Might need something more specific ;).

Bookeen might still have traces of this somewhere, as do potentially old Kobos.

NiLuJe:
Other than the small nits, :+1; thanks!

ezdiy:
There are some multitouch issues still due to messed up ffi calls, so will force-push that with fixes for your notes.

ffi/util.lua
ezdiy: Hmm, what about `/ebrmain/pocketbook`? This one is a symlink that points to cramfs image is mounted, ie leftovers are unlikely to have valid one.

NiLuJe: Yup, that would do the trick ;).

pazos:
Could you add `hasExitOptions = no,` and `canRestart = no,` to generic pocketbook device?, this way it will supersed #6541 completely.

ffi/input_pocketbook.lua
ezdiy: I was tempted to use cdef generator at first, but ran into issues of inkview versions being utter and complete mess, plus  part of the API being undocumented, ie not in the inkview header at all.

Including it only for the sake of single enum *all the while* awkwardly avoiding the minefield of the version mess stuff of theirs isn't worth the effort imo. The bonus side of this is that since there's no hard version linkage to inkview anymore, the SDK stuff becomes a bit less fragile due to room for ad-hoc fixups in ffi.

NiLuJe: Okay, that's what I was afraid of, just wanted to make sure ;). :+1:

ezdiy: @NiLuJe After a bit of digging, turns out the the header at https://github.com/pocketbook-free/SDK_481/blob/5.12/arm-obreey-linux-gnueabi/sysroot/usr/local/include/inkview.h *seems* somewhat useable - at least the MT struct size seem  corrected now, and there's even function signatures for bit of the stuff that used to be undocumented - will have to test it first to see if it actually runs.

Provided it runs ok with no need to retouch stuff, it might be worth the trouble. The build process will have to commit to using only this recent versions of the sdk to get the header, which may spell ABI issues due to unrelated libraries in the SDK that koreader depends on being too new on too old readers. So it's up to you, generated cdef, at the cost of possibly having to deal with tardiness of the SDK, or ugly code of hardcoded cdefs, with a little more wiggle room build wise.

NiLuJe: The build-process is entirely irrelevant, the burden to run ffi-cdecl is left to the developer ;).

So, basically, just a .cdecl with a comment about it if there are API issues would be enough (especially as compared to what the TC pulls: https://github.com/koreader/koxtoolchain/blob/2440ef904d88a3d4f8bff98d97df9a61e4afd86f/gen-tc.sh#L171-L172).

(There's already quite a few things that wouldn't quite work if they were simply being auto-generated at compile-time ;)).

NiLuJe: i.e., basically, it's just a question of traceability more than anything else, in case someone else stumbled onto the cdef block ;).

ffi/util.lua
ezdiy: Done.

ffi/input_pocketbook.lua
ezdiy: Welp, that was more obtuse than anticipated, but brought the entirety of inkview as ffi-cdecl in the end. Might come handy when frontend or plugin wants to bring in something from it.

ezdiy:
> There are some multitouch issues still due to messed up ffi calls, so will force-push that with fixes for your notes.

Hopefuly fixed for real now.

@pazos Done nevertheless (in the right repo :).

ffi/input_pocketbook.lua
NiLuJe: Yeah, it can be a bit.... awkward to grok ;).

ezdiy:
Currently blocked on https://github.com/koreader/koreader/issues/5114#issuecomment-681602222 - will proceed to port the change here if it works.

ezdiy:
If #5114 resolves, this should be good to go reliability-wise.

pazos:
@ezdiy: is this waiting for something?. It works fine on my end :)

ezdiy:
@pazos Summoning o' masters @Frenzie and @NiLuJe

Ye this seems to work fine, 2 people with old fw confirmed the poll emulation works ok.

Frenzie:
Nothing jumps out of me at a glance, looks fine to me. I might quibble that normally we only add cdecl stuff we actually use but I imagine at least some of it might be useful to already have in there, and I understand there was manual correction involved?

Frenzie (approved):

NiLuJe (approved):
@Frenzie: Yeah, the housekeeping re: the FFI stuff was handled early on ;).

So, nothing much to add, I'll obviously trust you guys on the PB shenanigans ;).

PR #1188, opened 2020-09-17
Generate proper PocketBook mxcfb constants

NiLuJe:
As in, from the actual kernel sources, instead of pulling magic constants out of a hat...

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1188)


Frenzie (approved):
:+1:

ezdiy:
ACK as-it-is now as there's high confidence it fixes #6000 (and keeps the wait working) and works ok at least on newer PBs. There are some more bold changes (like spoofing temp to overdrive DU waveform). In case that introduces issues would be nice to know from people running snapshot going forward.

ezdiy (commented):
> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1188)* status: 1 of 4 files reviewed, all discussions resolved

PR #1189, opened 2020-09-17
KOPTContext - add comments explaining Leptonica's usage

Galunid:
Work in progress, the idea is to document what I learned so far, so that others interested in this stuff don't have to rediscover America again ;-). For now I'm adding a cheatsheet, with Leptonica's most common functions we use and short explanation how it's used/what it's used for. Later I'll try to add documentation to a few functions.
@poire-z Is something like this acceptable? I'm open to suggestions.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1189)

### DIFF
--- a/ffi/koptcontext.lua
+++ b/ffi/koptcontext.lua
@@ -1,3 +1,53 @@
+--[[--
+Leptonica cheatsheet:
+-- Data structures:
+PIX - basic data structure - stores image
+PIXA - array of PIX
+BOX - stores rectangle (x, y, w, h)
+BOXA - array of BOX
+SEL - selector for morphological operations
+SELA - array of SEL
+SELA - array of SEL
+NUMA - array of numbers
+
+-- Functions:
+boxaGetCount(boxa) - get number of elements in boxa array
+boxaGetBox(boxa, index, flag) - flag can be C.L_COPY (creates new copy) or C.L_CLONE (returns ref-counted handle)
+pixThresholdToBinary(pix, threshold) - returns new black-white PIX, if the source pixel is < threshold,
+                                       resulting pixel is 1 (black), otherwise 0 (white)
+pixGetRegionsBinary(input_pix,
+                    output_halftonemask,
+                    output_textline_mask,
+                    output_textblock_mask,
+                    debug) - returns 0 on success and 1 on failure, output_* are PIX
+pixDestroy(pix) - free memory from PIX
+boxDestroy(box) - free memory from BOX
+selDestroy(sel) - free memory from SEL
+pixaDestroy(pixa) - free memory from PIXA
+boxaDestroy(boxa) - free memory from BOXA
+selaDestroy(sela) - free memory from SELA
+
+numaGetIValue(nai, counter_w, counter_l) - returns int value from NUMA
+numaGetFValue(nai, counter_w, counter_l) - returns float vlaue from NUMA
+
+pixGetWidth(pix)
+pixGetHeight(pix)
+pixSplitIntoBoxa
+boxAdjustSides
+boxaCombineOverlaps
+boxCreate
+boxaClipToBox
+boxOverlapRegion
+boxaAddBox
+pixDrawBoxaRandom
+pixWritePng
+pixConvertTo32
+pixMultiplyByColor
+pixWriteMemPng
+numaCreateFromFArray
+@module ffi.koptcontext
+]]
+
 local ffi = require("ffi")
 local C = ffi.C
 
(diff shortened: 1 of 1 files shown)


poire-z:
Yep! That kind of doc is great!
(I would usually be happy to see that around where it's used, but I guess it's used all over the place, so this cheatsheet :) 

I know now about PIX, BOX and SEL, but I don't know (yet) which is usually the input and which are the output, etc...
A small intro about what k2pdfopt and leptonica are and do (why we need both, how they interoperate...) would be nice too. (Because really, I have no idea :) I guess what k2pdfopt can do for having used the Windows too, but that's it...)

ffi/koptcontext.lua
```diff
@@ -15,6 +15,34 @@ else
     k2pdfopt = ffi.load("libs/libk2pdfopt.so.2")
 end
 
+--[[
```
Frenzie, line 18: If you add a couple more dashes and move it to the top then ldoc could pick it up? 
```suggestion
--[[--
```

ffi/koptcontext.lua
```diff
@@ -15,6 +15,34 @@ else
...
+                    output_textline_mask,
+                    output_textblock_mask,
+                    debug) - returns 0 on success and 1 on failure, output_* are PIX
+pixDestroy(pix) - free memory from PIX
+boxDestroy(box) - free memory from BOX
+selDestroy(sel) - free memory from SEL
+pixaDestroy(pixa) - free memory from PIXA
+boxaDestroy(boxa) - free memory from BOXA
+selaDestroy(sela) - free memory from SELA
+]]
```
Frenzie, line 43:
```suggestion

@module ffi.koptcontext
]]
```
See a couple of the other files around here for examples. You can generate the docs to see if they look as expected by running `ldoc .` in the /doc folder in front.
https://github.com/koreader/koreader-base/blob/3fdf18c280d821492abb51befb202fe96f08d1f5/ffi/MD5.lua#L1-L8

Frenzie (commented):
See http://koreader.rocks/doc/modules/gettext.html for one of the better docs I wrote. That's in https://github.com/koreader/koreader/blob/f1918cdfbb41a36665b010ef1f4a0a6fd4d6774d/frontend/gettext.lua

See http://koreader.rocks/doc/modules/ffi.mupdf.html for what's more of a stub.

Just something to consider. :-)

Galunid:
image
@Frenzie Any advice on how to format this? I can't annotate those functions, since they are part of leptonica and not `koptcontext`

Frenzie:
Looks pretty good for starters? ^_^

pazos (approved):

Galunid:
I'd like to hold it for a while, I'll try to work on this soon. It wraps lines strangely and it just looks bad.

ffi/koptcontext.lua
```diff
@@ -1,3 +1,57 @@
...
+## Data structures:
+`PIX - basic data structure - stores image`
+
+`PIXA - array of PIX`<br/>
+`BOX - stores rectangle (x, y, w, h)`<br/>
+`BOXA - array of BOX`<br/>
+`SEL - selector for morphological operations`<br/>
+`SELA - array of SEL`<br/>
+`SELA - array of SEL`<br/>
+`NUMA - array of numbers`<br/>
```
Frenzie, line 12: I'd say if you want it to look like code indent it by four spaces, and if you want it to be an HTML list… well, use a Markdown list. :-)

```
* `BOX` - stores things
* `BOXA` - array of BOX
```

Or something like that.

Galunid:
Oh wow, I just realized this is still not merged 😅. I'll work on this on Monday hopefully.

Galunid:
Note to self: `make doc` is the command.

This looks way better, I'll probably have to move comments above functions, since we are limited to ~80 characters, otherwise ldoc adds vertical scroll.
image

Galunid:
This is the other version, I'm not sure which is preferred?

Frenzie:
Maybe both? By the side for the short ones like all the destroys, and above for the ones that'd have to be split across so many lines?

Galunid:
Here's the hybrid
image

Frenzie:
I think that's the best of both 👍

ffi/koptcontext.lua
```diff
@@ -1,3 +1,77 @@
...
+
+    -- creates png from pix, useful for debugging
+    pixWritePng(path, pix, gamma)
+
+    -- multiplies box part of pixs by color and writes it to pixd
+    pixMultiplyByColor(pixd, pixs, box, color)
+
+    -- converts float_array to NUMA, flag is either C.L_INSERT or C.L_COPY
+    numaCreateFromFArray(float_array, size, flag)
+    @module ffi.koptcontext
```
Frenzie, line 72: I'm a bit surprised it even works indented (since to my eye that means it's part of the code block).
```suggestion

@module ffi.koptcontext
```

Frenzie:
> On an unrelated note, I think it's a good idea to block directly pushing to main repos. After this is merged `koptcontext-comments` branch needs to be deleted ;)

The master branch is protected from direct pushing to prevent mistakes. Besides that I don't think we're particularly concerned, or at least I don't think it's every really been an issue. :-)

PR #1191, opened 2020-09-19
Lua blitbuffer support for CFA displays (Kaleido)

ezdiy:
Very experimental as of now, needs feedback from https://github.com/koreader/koreader/issues/6479

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1191)

### DIFF
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -581,7 +581,6 @@ local SHIFT_ROTATED = 2
 local MASK_TYPE = 0x70
 local SHIFT_TYPE = 4
 local MASK_CFA = 0x80
-local SHIFT_CFA = 7
 
 local TYPE_BB4 = 0
 local TYPE_BB8 = 1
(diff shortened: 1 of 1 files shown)


ffi/blitbuffer.lua
```diff
@@ -596,8 +596,10 @@ local MASK_INVERSE = 0x02
 local SHIFT_INVERSE = 1
 local MASK_ROTATED = 0x0C
 local SHIFT_ROTATED = 2
-local MASK_TYPE = 0xF0
+local MASK_TYPE = 0x70
```
NiLuJe, line 599: This'll need a matching change in blitbuffer.c ;).

NiLuJe:
I'm not *quite* sure I grok everything happening in there, but, in any case, the comments are very welcome ^^.

(And, yeah, ultimately this should probably be implemented as a meta method once we have confirmation that it works).

ezdiy:
> I'm not quite sure I grok everything happening in there

You can force your kobo to render CFA too just by giving it:

```
    canUseCBB = no,
    hasColorScreen = yes,
    hasCFAScreen = yes,
```

And then guesstimating under looking glass it indeed looks like the staircase patterns posted in koreader/koreader#6479. In the end the issues will be mostly rotation related (again!) because when the pattern can go BGR or RGB separately for each X or Y + starting offset when we're going backwards (display is even number pixels, but CFA is multiplle of 3...).

ffi/blitbuffer.lua
ezdiy: Yep. Provided I'll get ad-hoc rotation going on correctly (ie no SW rotations ever), this is one of the places where C blitter can benefit a lot (the operation can be parallelized on neon for batches of 12 pixels).

pazos:
Superseeded by #1451

poire-z:
> Superseeded 

Until someday some other manufacturer like Kobo comes up with a device that uses one of these color eink panels :)

NiLuJe:
Would probably be a non-issue on any other manufacturer, because in the updated sunxi drivers used on Kobo, enabling CFA support just takes toggling a flag in the ioctl struct, all the magic happen in kernel-land ;).

PR #1192, opened 2020-09-19
Build & ship FBInk on PB

NiLuJe:
Should allow using fancy OTA feedback and the crash screen on PB.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1192)


Frenzie (approved):
Cool!

PR #1193, opened 2020-09-19
reMarkable: don't exclusive grab input for remarkable build

raisjnn:
as per https://github.com/Eeems/oxide/issues/69, we have a couple changes to koreader for remarkable build that will make koreader play nicer with launchers. this diff turns off the exclusive input grab because we don't need that for remarkable

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1193)

### DIFF
--- a/input/input.c
+++ b/input/input.c
@@ -71,6 +71,8 @@ static int openInputDevice(lua_State *L) {
     int childpid;
     int fd = findFreeFdSlot();
     if (fd == -1) return luaL_error(L, "no free slot for new input device <%s>", inputdevice);
+    char *ko_dont_grab_input = getenv("KO_DONT_GRAB_INPUT");
+
 
 #ifdef POCKETBOOK
     int inkview_events = luaL_checkint(L, 2);
@@ -104,7 +106,10 @@ static int openInputDevice(lua_State *L) {
     } else {
         inputfds[fd] = open(inputdevice, O_RDONLY | O_NONBLOCK, 0);
         if (inputfds[fd] != -1) {
-            ioctl(inputfds[fd], EVIOCGRAB, 1);
+            if (ko_dont_grab_input == NULL) {
+              ioctl(inputfds[fd], EVIOCGRAB, 1);
+            }
+
             /* prevent background command started from exec call from grabbing
              * input fd. for example: wpa_supplicant. */
             fcntl(inputfds[fd], F_SETFD, FD_CLOEXEC);
--- a/input/input.c
+++ b/input/input.c
@@ -107,7 +107,7 @@ static int openInputDevice(lua_State *L) {
         inputfds[fd] = open(inputdevice, O_RDONLY | O_NONBLOCK, 0);
         if (inputfds[fd] != -1) {
             if (ko_dont_grab_input == NULL) {
-              ioctl(inputfds[fd], EVIOCGRAB, 1);
+                ioctl(inputfds[fd], EVIOCGRAB, 1);
             }
 
             /* prevent background command started from exec call from grabbing


raisjnn:
i've also considered using an environment variable here, instead of the #ifdef, like: `char *grab_input = getenv("KO_GRAB_INPUT")`, so anyone can turn on and off this behavior instead of just hardcoding it to remarkable. thoughts?

Frenzie:
I'm fine with this approach until some other platform comes along.

NiLuJe:
Yeah, that could probably come in handy in the future ;).

Frenzie:
But don't let me dissuade you, also what @NiLuJe said. ;-)

Frenzie:
Actually, do we even need to grab this on Kobo, say?

NiLuJe:
I'm on mobile right now so it's a bit clunky to pull my musings on the subject from yesterday's oxide thread, but, yeah, it's quite likely useless on Kobo (at least as far as KFMon & NM are concerned, KSM might be another kettle of fish).

The only corner case I can think of is where MiniClock is concerned, because we're essentially silencing it, which is desirable here given that it's not useful on KOReader ;).

TL;DR: Yeah, env var would be nice, we'll deal with it where it makes sense ;).

NiLuJe:
The gist of yesterday's post was that I think that it's essentially a remnant of the Kindle centric days, where it was necessary to play nice with the stock system.

Frenzie:
Right, found it. :-)

https://github.com/Eeems/oxide/issues/69#issuecomment-695112177

raisjnn:
I'll push the env variable version after testing locally, then. I prefer env variable too. I will name variable `KO_DONT_GRAB_INPUT` (instead of `KO_GRAB_INPUT`)

raisjnn:
also: i went through the history of the line that has the exclusive grab and it was around from the very first commit to koreader base (the initial import): https://github.com/koreader/koreader-base/commit/0c24e3c7e0b8fd4a8e6d82b8df6f2f0004fb4d77#diff-8a39c51d28caad2ea5ee8ccc87680cbbR36

Frenzie:
Anything prior to 2013 or so would be tailored to Kindle, yup.

raisjnn:
looks like everything built ok on CI. i've tested locally with and without the env variable set, works as expected for me

NiLuJe (approved):
:+1:, Thanks!

input/input.c
```diff
@@ -104,7 +106,10 @@ static int openInputDevice(lua_State *L) {
     } else {
         inputfds[fd] = open(inputdevice, O_RDONLY | O_NONBLOCK, 0);
         if (inputfds[fd] != -1) {
-            ioctl(inputfds[fd], EVIOCGRAB, 1);
+            if (ko_dont_grab_input == NULL) {
+              ioctl(inputfds[fd], EVIOCGRAB, 1);
```
Frenzie, line 110:
```suggestion
                ioctl(inputfds[fd], EVIOCGRAB, 1);
```

Frenzie: (You didn't allow me to push, so please fix the spacing. ^_^)

raisjnn: i don't understand the "allow you to push" comment - are you saying to add you to the repo as write?

Frenzie: No, not at all. I thought you had disabled edits from maintainers, but it may be a GitHub issue with organizations.

https://github.community/t/how-can-we-enable-allow-edits-from-maintainers-by-default/2847

See https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork

PR #1194, opened 2020-09-19
linuxfb: Refactor and remove kludges, add support for HW rotation.

ezdiy:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1194)

### DIFF
--- a/blitbuffer.c
+++ b/blitbuffer.c
@@ -456,7 +456,7 @@ void BB_blit_to_BB8(BlitBuffer *src, BlitBuffer *dst,
                 // (i.e., setPixel, no rota, no invert).
                 // The cbb codepath ensures setPixel & no invert, so we only check for rotation.
                 if (sbb_rotation == 0 && dbb_rotation == 0) {
-                    if (offs_x == 0 && dest_x == 0 && w == src->phys_w && w == dst->phys_w) {
+                    if (offs_x == 0 && dest_x == 0 && w == src->w && w == dst->w && src->pitch == dst->pitch) {
                         // Single step for contiguous scanlines (on both sides)
                         //fprintf(stdout, "%s: full copy blit from BB8 to BB8\n", __FUNCTION__);
                         // BB8 is 1 byte per pixel
@@ -1066,7 +1066,7 @@ void BB_blit_to_BB32(BlitBuffer *src, BlitBuffer *dst,
                 // (i.e., setPixel, no rota, no invert).
                 // The cbb codepath ensures setPixel & no invert, so we only check for rotation.
                 if (sbb_rotation == 0 && dbb_rotation == 0) {
-                    if (offs_x == 0 && dest_x == 0 && w == src->phys_w && w == dst->phys_w) {
+                    if (offs_x == 0 && dest_x == 0 && w == src->w && w == dst->w && src->pitch == dst->pitch) {
                         // Single step for contiguous scanlines (on both sides)
                         //fprintf(stdout, "%s: full copy blit from BBRGB32 to BBRGB32\n", __FUNCTION__);
                         // BBRGB32 is 4 bytes per pixel
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -1023,8 +1023,8 @@ function BB8_mt.__index:blitTo8(dest, dest_x, dest_y, offs_x, offs_y, width, hei
         return self:blitDefault(dest, dest_x, dest_y, offs_x, offs_y, width, height, setter, set_param)
     end
 
-    -- NOTE: We need to compare against the fb's *actual* scanline width, off-screen/padding regions included (i.e., xres_virtual instead of xres)
-    if offs_x == 0 and dest_x == 0 and width == self.phys_w and width == dest.phys_w then
+    -- Perform block copies only in the simple case of the blit width being equal to width of both buffers of same pitch.
+    if offs_x == 0 and dest_x == 0 and width == self.w and width == dest.w and dest.pitch == self.pitch then
         -- Single step for contiguous scanlines (on both sides)
         --print("BB8 to BB8 full copy")
         -- BB8 is 1 byte per pixel
@@ -1050,8 +1050,8 @@ function BBRGB32_mt.__index:blitToRGB32(dest, dest_x, dest_y, offs_x, offs_y, wi
         return self:blitDefault(dest, dest_x, dest_y, offs_x, offs_y, width, height, setter, set_param)
     end
 
-    -- NOTE: We need to compare against the fb's *actual* scanline width, off-screen/padding regions included (i.e., xres_virtual instead of xres)
-    if offs_x == 0 and dest_x == 0 and width == self.phys_w and width == dest.phys_w then
+    -- Perform block copies only in the simple case of the blit width being equal to width of both buffers of same pitch.
+    if offs_x == 0 and dest_x == 0 and width == self.w and width == dest.w and dest.pitch == self.pitch then
         -- Single step for contiguous scanlines (on both sides)
         --print("BBRGB32 to BBRGB32 full copy")
         -- BBRGB32 is 4 bytes per pixel
--- a/ffi/framebuffer.lua
+++ b/ffi/framebuffer.lua
@@ -22,10 +22,21 @@ local fb = {
     night_mode = false,
     hw_dithering = false, -- will be setup via setupDithering @ startup by reader.lua
     sw_dithering = false, -- will be setup via setupDithering @ startup by reader.lua
+    is_always_portrait = false, -- true = SW flip landscape into portrait (historically a bit of misnomer)
+    forced_rotation = nil, --[[{
+        -- canonically the order below - but frontend may specify their own mapping
+        fb.ORIENTATION_PORTRAIT,
+        fb.ORIENTATION_LANDSCAPE,
+        fb.ORIENTATION_PORTRAIT_ROTATED,
+        fb.ORIENTATION_LANDSCAPE_ROTATED,
+        restore = false, -- true if hw rot mode is to be restored after painting, see framebuffer_linux:afterPaint()
+        default = nil, -- if not nil, use this as default rotation value (this one is canonical, not HW)
+        every_paint = false, -- true if HW mode should be enforced for every paint batch
+    },]]
 }
 
 --[[
-Codes for rotation modes:
+Canonical values for rotation modes:
 
 0 for no rotation (i.e., Portrait),
 1 for landscape with bottom on the right side of screen, etc.
@@ -57,6 +68,42 @@ fb.ORIENTATION_LANDSCAPE = 1
 fb.ORIENTATION_PORTRAIT_ROTATED = 2
 fb.ORIENTATION_LANDSCAPE_ROTATED = 3
 
+--[[
+Rotation modes are not guaranteed to be canonical however, as HW driver may interpret the value as CW or CCW, as well
+as starting from elsewhere than portrait so in practice the values can be pretty much anything.
+For such an occasion the frontend can provide custom mappings, or simply override the HW setter/getter/map function.
+]]
+
+-- Invert HW value to canonical one, needed to interpret the one given by OS.
+-- Some devices may wish to override this if the mapping is not as simple as 1:1.
+function fb:getCanonicalRotationMode(hw)
+    for k,v in ipairs(self.forced_rotation) do
+        if v == hw then
+            return k-1
+        end
+    end
+end
+
+-- To be overridden. Returns the current FB rotation exactly as reported by hw
+-- Pipe the result into getCanonicalRotationMode() if you want canonical (ie the koreader uses) result.
+function fb:getHWRotation()
+    return 0
+end
+
+-- To be overridden. Set HW rotation of the FB. Note that the argument is canonical value, ie this
+-- call should do translation from forced_rotation[] table - or any other necessary dance to convert
+-- canonical fb.ORIENTATION_ mode to HW reality.
+function fb:setHWRotation(canon)
+end
+
+-- Rotation modes may cause desync of rotation for touch translation, so frontend may need to override this
+-- to provide a fixup.
+function fb:getTouchRotation()
+    -- The touch matrix stays the same for things like linuxfb so just output the rotation we're doing either hw or sw.
+    -- But this may not be necessarily so with other backends.
+    return self.cur_rotation_mode
+end
+
 local EMULATE_READER_DPI = tonumber(os.getenv("EMULATE_READER_DPI"))
 
 function fb:extend(o)
@@ -80,7 +127,7 @@ function fb:init()
     -- asking the framebuffer for orientation is error prone,
     -- so we do this simple heuristic (for now)
     self.screen_size = self:getSize()
-    if self.screen_size.w > self.screen_size.h and self.device:isAlwaysPortrait() then
+    if self.screen_size.w > self.screen_size.h and self.is_always_portrait then
         self.screen_size.w, self.screen_size.h = self.screen_size.h, self.screen_size.w
         -- some framebuffers need to be rotated counter-clockwise (they start in landscape mode)
         io.write("FB: Enforcing portrait mode by doing an initial BB rotation")
@@ -88,9 +135,19 @@ function fb:init()
         self.debug("FB: This prevents the use of blitting optimizations. This should instead be fixed on the device's side on startup.")
         self.bb:rotate(-90)
         self.blitbuffer_rotation_mode = self.bb:getRotation()
+        assert(not self.forced_rotation, "If forced HW rotation is used, isAlwaysPortrait should not be set.")
     end
-    self.native_rotation_mode = self.ORIENTATION_PORTRAIT
+    self.native_rotation_mode = self.forced_rotation and self.forced_rotation.default or self.ORIENTATION_PORTRAIT
     self.cur_rotation_mode = self.native_rotation_mode
+end
+
+-- This method must be called just before drawing a sequence of blits into a framebuffer.
+-- It's ok to spam it for each paint, drivers should ensure to become nop for subsequent calls until final afterPaint().
+function fb:beforePaint()
+end
+
+-- This method must be called once we're done drawing all batched updates into a framebuffer, and *after* all the necessary fb:refresh* calls.
+function fb:afterPaint()
 end
 
 -- the ...Imp methods may be overridden to implement refresh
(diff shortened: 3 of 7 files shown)


ffi/framebuffer.lua
```diff
@@ -50,6 +50,10 @@ i.e., this matches <linux/fb.h> FB_ROTATE_* constants ;).
 
 This corresponds to the user-facing *physical device* rotation (in 90° CW steps).
 
+Of course this is not guaranteed, as drivers may interpret the value as CW or CCW, as well
+as starting from elsewhere than a portrait. For such an occasion the frontend can provide
```
Frenzie, line 54:
```suggestion
as starting from elsewhere than portrait. For such an occasion the frontend can provide
```

ffi/framebuffer.lua
```diff
@@ -149,6 +154,14 @@ end
...
 end
 
+-- overriden by HW rotation methods
+function fb:forceRotation()
+end
+
+function fb:restoreRotation()
+end
+
+
```
Frenzie, line 164:
```suggestion
```

Frenzie (commented):
Nothing jumps out.

Frenzie (approved):
Pinging @NiLuJe

ezdiy:
There will be probably more to it than just this due to cre acting when we shuffle the FB underneath its nose.

ffi/framebuffer_linux.lua
```diff
@@ -245,11 +269,47 @@ function framebuffer:init()
...
+        -- Go the standard route
+        return self.parent.setRotationMode(self, mode)
+    end
+    self.debug("setRotationMode:", mode, "old:",self.cur_rotation_mode)
+    if mode ~= self.cur_rotation_mode then
+        -- Do this dance only if rotation actually changes
+        self.cur_rotation_mode = mode
+        self:forceRotation()
+        self:reinit()
+        self:restoreRotation()
```
NiLuJe, line 287: Needing a restore there is extremely PB-specific, that's basically the only thing mildy bothering me with this implementation ;).

NiLuJe: And of course I'd missed the `needsRestoredRotation` device cap, so, never mind me ;).

NiLuJe:
Note that I'm volunteering, but shit is even weirder on some NTX boards, so a 1:1 mapping wouldn't necessarily always cut it there.

Just throwing it out there, because we know that on some NTX boards, some non-standard rotations are crashy @ 8bpp anyway, so it wouldn't necessary be a worthwhile pursuit ^^.

NiLuJe:
But it's (the rota table map) quite possibly much more user-friendly for Joe Random to come up with for his own device than the tricksy way we try to auto-compute those either in FBInk or Plato, so, eh ^^.

ffi/framebuffer_linux.lua
ezdiy: There's still device specifity though: We do forceRotation on every batch of paint. This is not necessarily needed for devies that can set and "hold" the rotate once it set once in here. But then again, better safe than sorry, and the call isn't all that expensive compared to bunch of paints, ergo, no bothering with special cap just to signal this.

poire-z: Mhhh, I spent a few minutes trying to understand what this is all about ... and didn't :)
Could you add a few comments about what all this new is about, why you forceRotation to soon after restoreRotation please...?

ffi/framebuffer_linux.lua
```diff
@@ -219,6 +241,8 @@ function framebuffer:init()
                            0)
     assert(tonumber(ffi.cast("intptr_t", self.data)) ~= C.MAP_FAILED,
            "can not mmap() framebuffer")
+
+    -- New invariant: self.bb MUST NOT be cached by anyone if we reinit, as the old one will point to junk now.
```
poire-z, line 245: I think you use "invariant" a few times before. English not being my first language, can you explicite what you mean when you use it, like in here?

ezdiy: It denotes location of contract that is invariable. Ie when an invariant is introduced, it implies there's a contract too - or warns that this end might be at odds with invariant of pre-existing contract somewhere else. It's useful to mark invariant because then you just grep source for 'contract|invariant' and notice if some are at odds with each other.

poire-z: OK :) So it doesn't mean anything more than what I would have written at your place:
`-- Note: don't ever cache self.bb, as we may replace it at any time` :)
It's just using some kind of rare keyword instead of my "Nore", to allow grep'ing or indexing.

Frenzie: Invariant just means not changing. This is awkward at best imho.

Frenzie: Btw, we use the `@warning` annotation for that.

ezdiy: I won't pretend this isn't pretentious as all hell (don't make me use something really awkward like "congruence"!), but indeed this is about caveats that specifically deal with "spooky effect at a distance" one looks specifically for. `note` is indeed far too general.

ffi/framebuffer_linux.lua
poire-z: Thanks for the added comments, that's clearer and I get it. May be just add a comment mentionning "i.e. on Pocketbooks, where xxx/yyy may draw over koreader" so we don't have to wonder who would do that on Kobo or Kindle :)

ffi/framebuffer_linux.lua
Frenzie: Like I said, our standard is:
```
-- @warning This will blow up.
```

There's no need to tag a note.

ezdiy:
Aight, this should be good to go (and see what else explodes elsewhere due to invasive changes of linuxfb logic)

Frenzie (approved):
lgtm

NiLuJe:
The pitch things makes me mildly curious if *replacing* the phys_w check by a pitch one wouldn't make more sense...

(Since pitch is actually what's used to calculate the pointer offset anyway...).

ezdiy:
@NiLuJe Yep, ye it should be enough. Meanwhile I'm doing more extensive rewrite of all this stuff as it turns out one needs to account for touch and refresh rotations too.

NiLuJe:
I'd be grateful if you could do a test with a few debug prints that output the actual values, because I'm still mildly confused as to how phys_w could match but not line_length ^^.

ezdiy:
This is now almost a complete rewrite of `framebuffer_linux.lua`, keeping it as separate patch for a while. Primarily it's to get rid of fragile device quirk handling (there's technically no need for any of that) and to provide some more reasonable hooks for rotation modes. Snooping it `self.device` is also restricted so as to make shimming of linuxfb in tests somewhat viable even without koreader frontend.

NiLuJe:
I'm afraid you're being a tad optimistic on the "hopefully it's all been fixed on Pb's side now" front, but I guess time will tell, and I'm all for making the common code cleaner anyway ;).

(We've had *a lot* of issues about that in the past, hence the heuristics ^^).

blitbuffer.c
```diff
@@ -456,7 +456,7 @@ void BB_blit_to_BB8(BlitBuffer *src, BlitBuffer *dst,
                 // (i.e., setPixel, no rota, no invert).
                 // The cbb codepath ensures setPixel & no invert, so we only check for rotation.
                 if (sbb_rotation == 0 && dbb_rotation == 0) {
-                    if (offs_x == 0 && dest_x == 0 && w == src->phys_w && w == dst->phys_w) {
+                    if (offs_x == 0 && dest_x == 0 && w >= src->w && w >= dst->w && src->pitch == dst->pitch) {
```
NiLuJe, line 459: In that one, w *has* to match the stride in pixels (of both src & dst). It's a full-screen blit of w * h, so if w doesn't match the stride, it'll either copy slightly too much or slightly not enough data.

Which, granted, might be nearly unnoticeable in the latter and possibly harmless in the former, but, still ^^.

blitbuffer.c
```diff
@@ -1066,7 +1066,7 @@ void BB_blit_to_BB32(BlitBuffer *src, BlitBuffer *dst,
...
```
NiLuJe, line 1069: Ditto.

blitbuffer.c
NiLuJe: Okay, >= means that it could only copy too *much* data, granted.

That may be harmless on Y8, but it won't be on RGB32, where we usually do *NOT* have any scratch space past the screen.

ffi/framebuffer_linux.lua
```diff
@@ -13,228 +13,124 @@ local framebuffer = {
...
-    io.flush()
+    -- Apply frontend kludges (color lux, very old eink...)
+    self:fbinfoOverride(finfo, vinfo)
+
+    assert(finfo.type == C.FB_TYPE_PACKED_PIXELS, "video type not supported")
+    assert(vinfo.xres > 0 and vinfo.yres > 0, "invalid framebuffer resolution")
+
+    -- We can forgo all messy fb detection logic by simply assuming only f.line_length and v.yres are valid,
+    -- because if they aren't, the world would be on fire no matter what sooner or later.
+    self.fb_size = finfo.line_length * vinfo.yres
```
NiLuJe, line 88: Probably a good idea ;).

I don't think I've ever seen line_length being broken anywhere.

blitbuffer.c
ezdiy: w is always <= pitch. If there's no scratch space, then w==pitch. Additional w<pitch check could be done, but I'm simply assuming the given rect dimensions are always within pitch from the caller (is that a valid assumption though?)

ffi/framebuffer_linux.lua
```diff
@@ -13,228 +13,124 @@ local framebuffer = {
...
+
+    local bpp = vinfo.bits_per_pixel
+    local stride_pixels = finfo.line_length * 8
+    assert(stride_pixels % bpp == 0, "line_length doesn't end at pixel boundary")
+    stride_pixels = stride_pixels / bpp
```
NiLuJe, line 93: Doing it this way neatly sidesteps the usual mishaps @ 4bpp when doing stride = line_length / (bpp/8), I like it :).

ffi/framebuffer_linux.lua
```diff
@@ -13,228 +13,124 @@ local framebuffer = {
...
+    assert(vinfo.xres > 0 and vinfo.yres > 0, "invalid framebuffer resolution")
 
     -- Make sure we never try to map a larger memory region than the fb reports
-    assert(self.fb_size <= finfo.smem_len, "computed fb memory region too large")
+    -- @warning Feel free to remove this check if it burns. There are chinese things out there that even happily report smem as 0x1000 and such.
+    assert(self.fb_size <= finfo.smem_len or finfo.smem <= 0x1000, "computed fb memory region too large")
 
     self.data = C.mmap(nil,
-                           self.fb_size,
+                           bit.band(self.fb_size+4095,-4096),
```
NiLuJe, line 121: I wouldn't mind a little comment about the alignment trickery here for future readers ;).

NiLuJe:
Looks good to me, besides the few comments above ;).

Mainly concerned about what the reasoning was being the width thing for the full screen blits ;).

blitbuffer.c
NiLuJe: I vaguely recall PDFs *maybe* doing some super-wonky things like sometimes passing the document's width, which may be very very large indeed, but I *may* be wrong.

NiLuJe: I'm vaguely wondering if the simplest solution is not simply to drop the w checks entirely, and simply use pitch*h as the copy size?

NiLuJe: Wait, no, that's dumb, we do need to ensure that we actually requested a blit of at least a scanline's worth of pixels for that to work.

NiLuJe: Well, technically, a scanline's worth of pixels *exactly*, which was my original point ^^.

NiLuJe: c.f., https://github.com/koreader/koreader-base/commit/6e7cd40b9425926ef22d8bcb6527396b910516f7#commitcomment-42580213; I think checkBounds takes care of w <= pitch for us ;).

ezdiy: The idea here is "be as permissive as possible without breaking stuff" but that indeed probably isn't wise, as even the old code managed to get it wrong. Perhaps just forgo this altogether except the trivial safe case (x=0,w=dst->w,pitch==pitch) as the *practical* speed up of block copy compared to copy-per-scanline are most likely academic (sub 0.1%) when we're copying near full scanline length anyway..

ffi/framebuffer_linux.lua
ezdiy: Well, I vaguely remember first gen kindle being evil and reporting line_length in pixels at 4bpp (affectively doubling the byte count than it really is). You don't happen to remember some old fbink issue relating to that so as to add quirk for it?

ezdiy:
I'd like to see kobo owners give it a spin, especially if someone happens to have older ones. I'll do the same for KT (gen 4 4bpp) and KPW2 (gen 10) in the meanwhile. That should hopefuly cover the most common devices out there a kludge is possibly missing for.

blitbuffer.c
NiLuJe: Agreed ;).

And, yeah, I did test that at the time, and the difference between a screen copy vs. a height * scanlines copy are indeed minimal ;).

But I like it nonetheless ;p.

ffi/framebuffer_linux.lua
NiLuJe: Hmm, that doesn't ring a bell, at least the stuff I have here on the final FW versions does report line_length as 300 on those 600x800 screens.

NiLuJe: Then again, if by "first gen Kindle" you mean an actual Kindle 1 (the one with the roller), I've never had one myself, so I wouldn't really know ^^.

The earliest thing I can test is a K2i ;).

ffi/blitbuffer.lua
```diff
@@ -1024,8 +1024,8 @@ function BB8_mt.__index:blitTo8(dest, dest_x, dest_y, offs_x, offs_y, width, hei
         return self:blitDefault(dest, dest_x, dest_y, offs_x, offs_y, width, height, setter, set_param)
     end
 
-    -- NOTE: We need to compare against the fb's *actual* scanline width, off-screen/padding regions included (i.e., xres_virtual instead of xres)
-    if offs_x == 0 and dest_x == 0 and width == self.phys_w and width == dest.phys_w then
+    -- NOTE: We may copy past (visible) width of both buffers, for as long strides match this is fine.
```
NiLuJe, line 1027: Those comments are no longer accurate with the stricter checks ;).

ffi/blitbuffer.lua
```diff
@@ -1051,8 +1051,8 @@ function BBRGB32_mt.__index:blitToRGB32(dest, dest_x, dest_y, offs_x, offs_y, wi
...
```
NiLuJe, line 1054: Ditto

NiLuJe:
Nothing blew up on a Kobo H20 (Mk.5) ;).

ffi/framebuffer_linux.lua
NiLuJe: Yep, it's san on a K2i:

```
Fixed framebuffer info:
id: "eink_fb"
start of fb mem: 0xc8889000
length of fb mem: 483328 bytes
type: packed pixels
interleave for interleaved planes: 0
visual: pseudo color readonly
hw panning x step: 0
hw panning y step: 0
hw panning y wrap: 0
line length: 300 bytes (600 pixels)
start of mmio: 0
length of mmio: 0
accel chip/card: 0

Variable framebuffer info:
visible resolution: 600x800
virtual resolution: 600x800
offset from virtual to visible resolution: 0x0
bits per pixel: 4
grayscale: true
red:   bitfield offset: 0, bitfield length: 0, MSB is right: no
green: bitfield offset: 0, bitfield length: 0, MSB is right: no
blue:  bitfield offset: 0, bitfield length: 0, MSB is right: no
alpha: bitfield offset: 0, bitfield length: 0, MSB is right: no
pixel format: standard
activate: 2
height: 121 mm
width: 91 mm
obsolete accel_flags: 0
pixel clock: 0 ps
time from sync to picture (left_margin): 0 pixclocks
time from picture to sync (right_margin): 0 pixclocks
time from sync to picture (upper_margin): 0 pixclocks
lower_margin: 0 pixclocks
length of horizontal sync: 0 pixclocks
length of vertical sync: 0 pixclocks
sync: undefined!
vmode: non interlaced
rotate: 0

Color map info:
start: 0, len: 16
  N      #R      #G      #B
00-00   FFFF    FFFF    FFFF
01-01   EEEE    EEEE    EEEE
02-02   DDDD    DDDD    DDDD
03-03   CCCC    CCCC    CCCC
04-04   BBBB    BBBB    BBBB
05-05   AAAA    AAAA    AAAA
06-06   9999    9999    9999
07-07   8888    8888    8888
08-08   7777    7777    7777
09-09   6666    6666    6666
0A-0A   5555    5555    5555
0B-0B   4444    4444    4444
0C-0C   3333    3333    3333
0D-0D   2222    2222    2222
0E-0E   1111    1111    1111
0F-0F   0000    0000    0000
```

NiLuJe:
Nothing imploded on a K2i either ;).

NiLuJe:
I'll round up with a K4 later tonight, it's currently flat.

NiLuJe:
And, yeah, we can probably get rid of phys_h indeed ;).

ezdiy:
@NiLuJe Yea will do in some chore PR later on. For now it would probably introduce plethora of merge conflicts wrt #1195 and possibly others that are kept standalone from this.

NiLuJe:
No issues either on a K4 :+1: 

```
Fixed framebuffer info:
id: "eink_fb"
start of fb mem: 0xd0b8e000
length of fb mem: 966656 bytes
type: packed pixels
interleave for interleaved planes: 0
visual: pseudo color readonly
hw panning x step: 0
hw panning y step: 0
hw panning y wrap: 0
line length: 600 bytes (600 pixels)
start of mmio: 0
length of mmio: 0
accel chip/card: 0

Variable framebuffer info:
visible resolution: 600x800
virtual resolution: 600x800
offset from virtual to visible resolution: 0x0
bits per pixel: 8
grayscale: true
red:   bitfield offset: 0, bitfield length: 0, MSB is right: no
green: bitfield offset: 0, bitfield length: 0, MSB is right: no
blue:  bitfield offset: 0, bitfield length: 0, MSB is right: no
alpha: bitfield offset: 0, bitfield length: 0, MSB is right: no
pixel format: standard
activate: 2
height: 121 mm
width: 91 mm
obsolete accel_flags: 0
pixel clock: 0 ps
time from sync to picture (left_margin): 0 pixclocks
time from picture to sync (right_margin): 0 pixclocks
time from sync to picture (upper_margin): 0 pixclocks
lower_margin: 0 pixclocks
length of horizontal sync: 0 pixclocks
length of vertical sync: 0 pixclocks
sync: undefined!
vmode: non interlaced
rotate: 0

Color map info:
start: 0, len: 256
  N      #R      #G      #B
00-10   FFFF    FFFF    FFFF
11-21   EEEE    EEEE    EEEE
22-32   DDDD    DDDD    DDDD
33-43   CCCC    CCCC    CCCC
44-54   BBBB    BBBB    BBBB
55-65   AAAA    AAAA    AAAA
66-76   9999    9999    9999
77-87   8888    8888    8888
88-98   7777    7777    7777
99-A9   6666    6666    6666
AA-BA   5555    5555    5555
BB-CB   4444    4444    4444
CC-DC   3333    3333    3333
DD-ED   2222    2222    2222
EE-FE   1111    1111    1111
FF-FF   0000    0000    0000
```

NiLuJe:
This is good to go as far as I'm concerned, just give the word and we'll merge ;).

ezdiy:
rebased to master, should be gtg so that frontend can be piled on

NiLuJe:
Finally updated the Forma, (which, incidentally, is probably one of the rare devices where we can use the block copy codepath, because of a glorious xres == xres_virtual == line_length trinity ^^), and everything's fine on that front ;).

```
Fixed framebuffer info:
id: "mxc_epdc_fb"
start of fb mem: 0x9d000000
length of fb mem: 11796480 bytes
type: packed pixels
interleave for interleaved planes: 0
visual: pseudo color readonly
hw panning x step: 1
hw panning y step: 1
hw y wrap: 0
line length: 1440 bytes (1440 pixels)
start of mmio: 0
length of mmio: 0
accel chip/card: 0

Variable framebuffer info:
visible resolution: 1440x1920
virtual resolution: 1440x3840
offset from virtual to visible resolution: 0x0
bits per pixel: 8
grayscale: true
red:   bitfield offset: 0, bitfield length: 8, MSB is right: no
green: bitfield offset: 0, bitfield length: 8, MSB is right: no
blue:  bitfield offset: 0, bitfield length: 8, MSB is right: no
alpha: bitfield offset: 0, bitfield length: 0, MSB is right: no
pixel format: standard
activate: 128
height: 4294967295 mm
width: 4294967295 mm
obsolete accel_flags: 0
pixel clock: 160000000 ps
time from sync to picture (left margin): 36 pixclocks
time from picture to sync (right margin): 248 pixclocks
time from sync to picture (upper margin): 4 pixclocks
time from picture to sync (lower margin): 8 pixclocks
length of horizontal sync: 52 pixclocks
length of vertical sync: 1 pixclocks
sync: undefined! (0)
vmode: non interlaced
rotate: 3
```

PR #1195, opened 2020-09-21
blitbuffer refactor: Make CBB modular

ezdiy:
* Detecting stuff we can't handle, and fall back to Lua
* It's also possible to toggle CBB on the go without koreader restarts.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1195)

### DIFF
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -1918,13 +1918,13 @@ if not os.getenv("KO_NO_CBB") then
 end
 
 -- Set the actual enable/disable CBB flag. Returns the flag of whether it is (actually) enabled.
-function BB:enableCBB(ison)
+function BB:enableCBB(enabled)
     local old = use_cblitbuffer
-    use_cblitbuffer = ison and self.has_cblitbuffer
+    use_cblitbuffer = enabled and self.has_cblitbuffer
     if old ~= use_cblitbuffer then
         -- NOTE: This works-around a number of corner-cases which may end up with LuaJIT's optimizer blacklisting this very codepath,
         --       which'd obviously *murder* performance (to the effect of a soft-lock, essentially).
-        --       c.f., #4137, #4752, #4782
+        --       c.f., koreader/koreader#4137, koreader/koreader#4752, koreader/koreader#4782
         local val = use_cblitbuffer and 15 or 45
         jit.opt.start("loopunroll="..tostring(val))
         jit.flush()
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -651,11 +651,11 @@ end
 -- Determine if a pair of buffers can use CBB in relation to each other, or whether CBB is used at all.
 -- Used to skip unsupported modes such as unrelated inverses.
 -- TODO: Possibly some RGB24/32 stuff too?
-function BB_mt.__index:canCBB2(other)
+function BB_mt.__index:canUseCbbTogether(other)
     return use_cblitbuffer and self:getInverse() == other:getInverse()
 end
 
-function BB_mt.__index:canCBB()
+function BB_mt.__index:canUseCbb()
     return use_cblitbuffer and self:getInverse() == 0
 end
 
@@ -1082,7 +1082,7 @@ function BB_mt.__index:blitFrom(source, dest_x, dest_y, offs_x, offs_y, width, h
     if width <= 0 or height <= 0 then return end
 
     if not setter then setter = self.setPixel end
-    if self:canCBB2(source) and setter == self.setPixel then
+    if self:canUseCbbTogether(source) and setter == self.setPixel then
         cblitbuffer.BB_blit_to(ffi.cast("struct BlitBuffer *", source),
             ffi.cast("struct BlitBuffer *", self),
             dest_x, dest_y, offs_x, offs_y, width, height)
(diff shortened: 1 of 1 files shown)


ffi/blitbuffer.lua
```diff
@@ -658,6 +647,18 @@ end
 function BB_mt.__index:getType()
     return rshift(band(MASK_TYPE, self.config), SHIFT_TYPE)
 end
+
+-- Determine if a pair of buffers can use CBB in relation to each other, or whether CBB is used at all.
+-- Used to skip unsupported modes such as unrelated inverses.
+-- TODO: Possibly some RGB24/32 stuff too?
+function BB_mt.__index:canCBB2(other)
```
poire-z, line 654: You named it canCBB**2** because it is a 2nd function and you did not find a better name?
Or because it's supposed to read/sound like "can CBB *to* (other)" ? :)
If the latter, I'd rather see it spelled canCBBto(other) or canCBBwith(other), rather than us starting writting code like SMSs :)

Frenzie: Or a cozy `canCBBTogether`. ^_^

ezdiy: can rename it to canCbbTo if you're getting this impression.

ffi/blitbuffer.lua
```diff
@@ -1912,4 +1911,28 @@ BB.TYPE_BBRGB16 = TYPE_BBRGB16
 BB.TYPE_BBRGB24 = TYPE_BBRGB24
 BB.TYPE_BBRGB32 = TYPE_BBRGB32
 
+BB.has_cblitbuffer = false
+if not os.getenv("KO_NO_CBB") then
+    -- Load C blit buffer, we'll decide whether to use it later on
+    BB.has_cblitbuffer, cblitbuffer = pcall(ffi.load, 'blitbuffer')
```
Frenzie, line 1917: Nit
```suggestion
    BB.has_cblitbuffer, cblitbuffer = pcall(ffi.load, "blitbuffer")
```

ffi/blitbuffer.lua
ezdiy: Hmm, or better yet, collapse the two by making the argument optional. Not really sure why I kept it split in two in the first place.

Frenzie: I think the naming/splitting is quite reasonable. At a glance I'd think `canCBB` was a property of a single one and the fact that an optional argument points to an "other" is far from self-explanatory, while it is in `canCBBTo`.

Ease of maintenance/skimming/reading > being clever to me. ;-)

poire-z: Reads indeed better after the "Cosmetics." commit, thanks.

ffi/blitbuffer.lua
```diff
@@ -1912,4 +1911,28 @@ BB.TYPE_BBRGB16 = TYPE_BBRGB16
...
 BB.TYPE_BBRGB32 = TYPE_BBRGB32
 
+BB.has_cblitbuffer = false
+if not os.getenv("KO_NO_CBB") then
+    -- Load C blit buffer, we'll decide whether to use it later on
+    BB.has_cblitbuffer, cblitbuffer = pcall(ffi.load, "blitbuffer")
+end
+
+-- Set the actual enable/disable CBB flag. Returns the flag of whether it is (actually) enabled.
+function BB:enableCBB(ison)
```
Frenzie, line 1921: For legibility, our variable equivalent of isOn (function) is is_on. ^_^
```suggestion
function BB:enableCBB(is_on)
```

Frenzie: But actually is_on seems a bit of a weird name here.

Frenzie: I think we generally use:
```suggestion
function BB:enableCBB(enabled)
```
(feel free to suggest alternatives of course, but in that case we should probably update some others too)

ffi/blitbuffer.lua
```diff
@@ -1912,4 +1911,28 @@ BB.TYPE_BBRGB16 = TYPE_BBRGB16
...
+    local old = use_cblitbuffer
+    use_cblitbuffer = ison and self.has_cblitbuffer
+    if old ~= use_cblitbuffer then
+        -- NOTE: This works-around a number of corner-cases which may end up with LuaJIT's optimizer blacklisting this very codepath,
+        --       which'd obviously *murder* performance (to the effect of a soft-lock, essentially).
+        --       c.f., #4137, #4752, #4782
```
Frenzie, line 1927: While we're unlikely to hit 4000 anytime soon (if ever), I think it's nevertheless clearer/safer to specify.

```suggestion
        --       c.f., koreader/koreader#4137, koreader/koreader#4752, koreader/koreader#4782
```

```suggestion
        --       c.f., koreader#4137, koreader#4752, koreader#4782
```

ffi/blitbuffer.lua
```diff
@@ -658,6 +647,18 @@ end
...
 end
+
+-- Determine if a pair of buffers can use CBB in relation to each other, or whether CBB is used at all.
+-- Used to skip unsupported modes such as unrelated inverses.
+-- TODO: Possibly some RGB24/32 stuff too?
+function BB_mt.__index:canUseCbbTogether(other)
+    return use_cblitbuffer and self:getInverse() == other:getInverse()
+end
+
+function BB_mt.__index:canUseCbb()
```
Frenzie, line 658: Not 100% convinced of `Cbb` but that aside. I'll leave any further potential nitting to @pazos and/or @NiLuJe.

Frenzie (approved):
lgtm!

poire-z (approved):
No more nitpick.

ffi/blitbuffer.lua
```diff
@@ -651,11 +651,11 @@ end
 -- Determine if a pair of buffers can use CBB in relation to each other, or whether CBB is used at all.
 -- Used to skip unsupported modes such as unrelated inverses.
 -- TODO: Possibly some RGB24/32 stuff too?
```
NiLuJe, line 653: What did you have in mind with that specific comment?

ffi/blitbuffer.lua
NiLuJe: I utterly failed to find anything better at the time, so, I'm fine with it, FWIW ;).

Frenzie: Talking about canUseCBB vs canUseCbb. I prefer the former but I feel I've met my nitpick quota, plus those were more substantial nits. :-P

NiLuJe: Makes the C stands out more this way, which isn't necessarily a bad thing, I think ;).

NiLuJe (approved):
Thanks!

PR #1200, opened 2020-09-25
SDL + Android: Clean up inverse blit modes

ezdiy:
Get rid of the triple/quad inversion mess.

Tested only on SDL, but the mechanism is now same on Android:
Inversion is done during writes to the shadow BB that is configured
to invert by noHWInvert from the base fb class, same way it works for
linuxfb eink that lacks inversion mode.

During the final update we then copy the shadow buffer 1:1 to the
backing texture.

Should fix

https://github.com/koreader/koreader/issues/6708
https://github.com/koreader/koreader/issues/6703

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1200)

### DIFF
--- a/ffi/framebuffer.lua
+++ b/ffi/framebuffer.lua
@@ -403,7 +403,7 @@ function fb:toggleNightMode()
         self.bb:invert()
         if self.viewport then
             -- invert and blank out the full framebuffer when we are working on a viewport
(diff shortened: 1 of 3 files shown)


Frenzie (approved):
Works for me.

Screenshot_2020-09-25_06-08-24

PR #1201, opened 2020-09-25
blitbuffer: Some more refactoring + SDL FB debug tools     

ezdiy:
More of an RFC/WIP for now, to agree on stuff etc.
TODO: This turned out to be fairly invasive in a lot of places, so need to do some daily use testing to see if things still hold together.

* phys_w/pitch renamed to pixel_stride/stride, phys_h removed
* Get rid of getPhys*() uses. stride is hardwired to rotation,
   so returning swapped coordinates from there doesn't work
* make proper distinctions in uses between stride/pixel_stride where such
   wasn't the case
* add some more getenv() options to SDL to emulate arbitrary bb formats
  and b&w device
* make BB.new() infer optional stride/pixel_stride parameters

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1201)

### DIFF
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -59,64 +59,57 @@ typedef struct ColorRGB32 {
 
 typedef struct BlitBuffer {
     int w;
-    int phys_w;
+    int pixel_stride;   // nb of pixels from the start of a line to the start of next line
     int h;
-    int phys_h;
-    int pitch;
+    int stride;         // nb of bytes from the start of a line to the start of next line
     uint8_t *data;
     uint8_t config;
 } BlitBuffer;
 typedef struct BlitBuffer4 {
     int w;
-    int phys_w;
+    int pixel_stride;
     int h;
-    int phys_h;
-    int pitch;
+    int stride;
     uint8_t *data;
     uint8_t config;
 } BlitBuffer4;
 typedef struct BlitBuffer8 {
     int w;
-    int phys_w;
+    int pixel_stride;
     int h;
-    int phys_h;
-    int pitch;
+    int stride;
     Color8 *data;
     uint8_t config;
 } BlitBuffer8;
 typedef struct BlitBuffer8A {
     int w;
-    int phys_w;
+    int pixel_stride;
     int h;
-    int phys_h;
-    int pitch;
+    int stride;
     Color8A *data;
     uint8_t config;
 } BlitBuffer8A;
 typedef struct BlitBufferRGB16 {
     int w;
-    int phys_w;
+    int pixel_stride;
     int h;
-    int phys_h;
-    int pitch;
+    int stride;
     ColorRGB16 *data;
     uint8_t config;
 } BlitBufferRGB16;
 typedef struct BlitBufferRGB24 {
     int w;
-    int phys_w;
+    int pixel_stride;
     int h;
-    int phys_h;
-    int pitch;
+    int stride;
     ColorRGB24 *data;
     uint8_t config;
 } BlitBufferRGB24;
 typedef struct BlitBufferRGB32 {
     int w;
-    int phys_w;
+    int pixel_stride;
     int h;
-    int phys_h;
-    int pitch;
+    int stride;
     ColorRGB32 *data;
     uint8_t config;
 } BlitBufferRGB32;
@@ -721,11 +714,11 @@ end
 -- getPixelP (pointer) routines, working on physical coordinates
 function BB_mt.__index:getPixelP(x, y)
     --self:checkCoordinates(x, y)
-    return ffi.cast(self.data, ffi.cast(uint8pt, self.data) + self.pitch*y) + x
+    return ffi.cast(self.data, ffi.cast(uint8pt, self.data) + self.stride*y) + x
 end
 function BB4_mt.__index:getPixelP(x, y)
     --self:checkCoordinates(x, y)
-    local p = self.data + self.pitch*y + rshift(x, 1)
+    local p = self.data + self.stride*y + rshift(x, 1)
     if band(x, 1) == 0 then
         return ffi.cast(P_Color4U, p)
     else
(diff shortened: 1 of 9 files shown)


NiLuJe (approved):
Nothing jumps out, LGTM ;).

`pixel_pitch` is fine by me.

ffi/framebuffer_SDL2_0.lua
```diff
@@ -15,6 +15,10 @@ function framebuffer:init()
     else
         self.bb = BB.new(600, 800)
     end
+    if os.getenv("EMULATE_BW_SCREEN") then
+        self.device.hasColorScreen = function() return false end
+        self.isColorEnabled = function() return false end
+    end
```
Frenzie, line 21: I'm not quite sure if this should be here in base in this form?

Frenzie (commented):
Seems fine by me too.

ffi/blitbuffer.lua
```diff
@@ -1818,27 +1792,21 @@ ffi.metatype("ColorRGB16", ColorRGB16_mt)
...
-    elseif buffertype == TYPE_BB8 then bb = BlitBuffer8(width, phys_width, height, phys_height, pitch, nil, 0)
-    elseif buffertype == TYPE_BB8A then bb = BlitBuffer8A(width, phys_width, height, phys_height, pitch, nil, 0)
-    elseif buffertype == TYPE_BBRGB16 then bb = BlitBufferRGB16(width, phys_width, height, phys_height, pitch, nil, 0)
-    elseif buffertype == TYPE_BBRGB24 then bb = BlitBufferRGB24(width, phys_width, height, phys_height, pitch, nil, 0)
-    elseif buffertype == TYPE_BBRGB32 then bb = BlitBufferRGB32(width, phys_width, height, phys_height, pitch, nil, 0)
+    local pixel_size = assert(BB.TYPE_TO_BPP[tonumber(buffertype)], "unknown buffer type " .. tostring(buffertype))
+
+    -- If no pitch is given, make up one from buffer width
+    pitch = pitch or rshift(width * pixel_size + 7, 3)
+    pixel_pitch = pixel_pitch or (pitch / pixel_size)
```
poire-z, line 1802: It's been a long time since I looked at this module, and I'm not competent enough to give any comment.
The word "pitch" was already there, but I'm not familiar with the term and its meaning here - and that rshift formula looks a bit cryptic to me :)
Can you add a comment somewhere to define what pitch/pixel_pitch actually is (the other stuff, bpp, something_w, I can guess, pitch I can't :/)

NiLuJe: I'm not necessarily super fond of the (previous) usage of pitch here, but the alternative is something like stride, which isn't any better? :D.

We *could* arguably come up with something ourselves line `(scan)line_length` (in pixels) / `(scan)line_size` (in bytes) or something?

---

>> 3 is / 8

ffi/framebuffer_SDL2_0.lua
ezdiy: The idea here is that env specifically instructs fb to do all the emulation stuff directly, often at costs coercing frontends into things they have no idea about. So its either this or polluting frontend with debug brutalism (test spec are fine with this too, as opposed to *reading* funky flags we'd prefer to not shim).

ffi/blitbuffer.lua
NiLuJe: Barring something custom, `stride` is possibly more commonly recognizable in that context than `pitch`?

In which case `pixel_stride` would work for me for its pixel counterpart.

ezdiy: Pitch is kind of a misnomer indeed. I'm tempted to wholesale replace pitch for stride everywhere (not just comments).

> >> 3 is / 8

Hmm, this code looks outright wrong, probably should be

```
pixel_pitch = pixel_pitch or (pitch * 8 / pixel_size)
```

ezdiy: also probably rename pixel_size to bits_per_pixel to make things more obvious

poire-z: > Barring something custom, stride is possibly more commonly recognizable in that context than pitch?

Not for me :/ you haven't yet told me clearly what it is :)
```
stride (n): a long step in walking.
stride (v): to pass over or across in one long step: 
```
So, I guess it is the length of a bb _horizontal_ (?) line in... bytes ? So, it's width * bpp/8 ? Right ?

So, you're talking about scan_line_size / scan_line_nbpixels ?
Please just explain somewhere what is this thing, whatever  you decide to name it :) So I (and future readers) don't have to guess and stay unsure of my (their) guess.

ffi/framebuffer_SDL2_0.lua
Frenzie: Alright, it's a niche development option anyway. :-)

ffi/blitbuffer.lua
NiLuJe: The physical analogy is because it's the amount of bytes you have to walk over to step across a scanline.

Which, in plaintext, translates to the length of a scanline (i.e., visible + padding) in bytes ;).

ezdiy: @poire-z Stride is specifically for bitmap blitting. It's either amount of bytes or pixels until next row (assuming no rotation), usually both are tracked so as to avoid issues with bpp shifts. Pitch is conceptually related to stride, but it can measured in any unit - for instance, font libraries often use *points*. Pitch is typically used for *on screen* transforms (eg path tracing of font raster with subpixel antialiasing), whereas stride is exclusively about computing next row, pixel or byte.

poire-z: Thanks !
So, I guess a better comment on these 2 lines would be enough (for me to get it):
```
52    int pixel_pitch;   // nb of pixels from the start of a line to the start of next line
53    int h;
54    int pitch;         // nb of bytes from the start of a line to the start of next line
```

poire-z (approved):

ezdiy:
Oops, somehow this glitched posting the commit message twice (no more junk beyond that luckily). Moral of the story: Don't use github UI right after you squash/rebase under its nose, it will do just this.

Frenzie:
That's, um, interesting. To make sure I understand correctly, we should definitely reload the web GUI page after force pushing something? ;-)

NiLuJe:
Yeah, I've had that happen once in the past ;).

ezdiy:
@Frenzie Other way around. I squashed in cli and force pushed that. The UI has been opened with the commits before squashing. The UI  notices the force-push, but only halfway somehow - when you click 'squash and merge' at that point, just after the push from cli, and *then* it will slap another commit on top of the force-pushed one only with the commit message

Frenzie:
@ezdiy I'm not sure what you think I meant, but that's what I meant. ;-) (Although I would've expected a "wrong" commit message or something, not a stray extra commit.)

PR #1206, opened 2020-10-04
Slightly less crappy screenshot performance

NiLuJe:
By using our existing blitting methods if possible, and outputting in a pixel-format as close to the input one as possible.

(LodePNG itself appears to have been smart enough to squash grayscale-as-RGBA32 to grayscale all on its own in the encoding stage, so, no space savings on the encoded files).

re: https://github.com/koreader/koreader/issues/6736

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1206)


NiLuJe:
I didn't duplicate the code in BBTYPE specific functions because this is not performance ciritical, so the mt and some minimal branching will have to do ;).

NiLuJe:
Also squashed the crap fallback to RGB24 instead of RGBA32. We drop the alpha anyway, and we don't gain anything by pixels being pointer-aligned here, since we're doing byte by byte load/stores anyway.

NiLuJe:
Thankfully, no-one should be using said crap fallback anyway, so BB8 & BB32 fbs should enjoy significantly less crappy performance.

Frenzie (approved):
Fine by me, saves a pass?

NiLuJe:
Also benefits from the blitting fast-paths if applicable (i.e., type matches, and source is not rotated), and uses the C blitter if enabled.

NiLuJe:
Actually unbroke it in !BB8 (wrong stride :s), and made it zero-copy if possible (which, granted, will be rare. basically the same cases as the block copy, e.g., and unrotated Forma).

NiLuJe:
And testing the zero-copy thingy on a Forma revealed that it was hilariously slower, because CPUs are weird, I guess?

poire-z:
(Might add the proper ffiutil.strcoll stuff in this PR ? :) Just mentionning it because you're familiar with ffi and I don't want to have to do it, so ... :)

NiLuJe:
Sure, I can do that ;).

PR #1208, opened 2020-10-05
bump crengine: more than 64 styles, table layout fixes

poire-z:
Includes https://github.com/koreader/crengine/pull/385 :
- TextLang: German: loosen quotation marks handling
- Fonts: add LVFontManager::getFontFileNameAndFaceIndex()
- fb2.css: avoid break before footnote links
- FlowState: fix possible wrong baseline with nested inline-table
- Tables: fix possible layout issues with rowspan
- Table cells: minor refactoring of baseline handling
- getRenderedWidths(): better estimate table width
- lvstyles: allow for more than 64 styles

cre.cpp: add getFontFaceFilenameAndFaceIndex().
ffi/freetype.lua: add optional 'faceindex" parameter to FT.newFace() so we can request a font other than the first one from font collections (.ttc).
Needed to show font names in the menu in their own font, see https://github.com/koreader/koreader/issues/6739#issuecomment-703131167 .

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1208)

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -5,6 +5,7 @@ std = "luajit"
 self = false
 
 read_globals = {
+    "DALPHA_SORT_CASE_INSENSITIVE",
     "DLANDSCAPE_CLOCKWISE_ROTATION",
     "lfs",
 }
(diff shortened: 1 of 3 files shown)


Frenzie (approved):
Big hooray on the styles! (The rest looks fine too. ^_^)

PR #1210, opened 2020-10-06
Drop unused variable

NiLuJe:
The rest of them were already doing that ;).

I don't *rightly* recall *why* we're killing debug hooks in there in the first place (performance?), but, oh, well.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1210)

### DIFF
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -973,7 +973,7 @@ end
 
 function BB_mt.__index:blitDefault(dest, dest_x, dest_y, offs_x, offs_y, width, height, setter, set_param)
     -- slow default variant:
-    local hook, mask, _ = debug.gethook()
+    local hook, hmask, _ = debug.gethook()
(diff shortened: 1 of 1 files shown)


ffi/blitbuffer.lua
```diff
@@ -973,7 +973,7 @@ end
 
 function BB_mt.__index:blitDefault(dest, dest_x, dest_y, offs_x, offs_y, width, height, setter, set_param)
     -- slow default variant:
-    local hook, mask, count = debug.gethook()
+    local hook, hmask, _ = debug.gethook()
```
Frenzie, line 976: Bit late to comment I suppose, but to drop it you could also just ignore that third one fully.

NiLuJe: True, I do generally like the _ placeholder though (mostly for documentation purposes ;)).

PR #1211, opened 2020-10-08
Add harfbuzz FFI

ezdiy:
Full binding, so as to avoid type dep annoyances in the future.
Primarily for mupdf font queries, but perhaps also more generally,
cf https://github.com/koreader/koreader/issues/6763

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1211)


Frenzie (approved):
Fine by me, we probably don't really need the whole shebang from the get-go though?

ezdiy:
@Frenzie Almost-automated fire and forget thing via bunch of :%s macros.

@NiLuJe was wondering the same after I spammed inkview the same way. Indeed there's no need for full cdecl, but figuring *what* to cherry pick *while* you're prototyping splits loyalties, as you need to regenerate the header repeated during the course of prototyping stuff and deal with the annoying cascades of type dependencies there. Albeit I a admit it's a little bit bloaty, perf wise (5kb vs 50k lua file) the impact is not measurable.

NiLuJe:
It's fine by me if there's no perf impact ;).

I tend to like hand-tailored cdecls because it serves as a bit of documentation in the sense of: hey, here are the things I'll probably be using, but it's more of a personal choice than anything, really ;).

Frenzie:
That's why I like it too, I think it makes it easier to grok what's going on. Besides that it's irrelevant.

PR #1212, opened 2020-10-10
Minor lua-ljsqlite3 tweaks

NiLuJe:
* Merges a few patches from forks
* Adds a traceback to `err()`, because it's nigh unusable without it.
* Get rid of the xsys dep and use (hopefully) slightly less arcane versions of split & trim.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1212)


NiLuJe:
(I've been testing https://github.com/koreader/koreader/pull/6778 w/ this, FWIW).

thirdparty/lua-ljsqlite3/init.lua
```diff
@@ -342,10 +376,9 @@ end
 
 -- Connection exec, __call, rowexec --------------------------------------------
 function conn_mt:exec(commands, get) T_open(self)
-  local cmd1 = split(commands, ";")
   local res, n
-  for i=1,#cmd1 do
-    local cmd = trim(cmd1[i])
+  for command in commands:gsplit(";", true) do
```
poire-z, line 380: No issue with such SQL ?:
`select * from book where title like '%;%'; select * from book where title = ';';`
(that might probably not happen with out code, but well :)

poire-z:
(Mentionning this, just in case you think it should be in this module, now that we're hacking it :):
https://github.com/koreader/koreader/blob/39faa8b860325c3c18372a4b24e50e1a1b407a8b/plugins/coverbrowser.koplugin/xutil.lua#L35-L46
)

thirdparty/lua-ljsqlite3/init.lua
NiLuJe: I expect that this will implode in fun and interesting ways, yeah, and that stuff like that might have been what the first fork was trying to fix, let me check ;).

NiLuJe: (Sidebar: xsys was also blowing up on that)

NiLuJe:
```
split_sql
1       SELECT * FROM book WHERE title LIKE '%;%';
2
SELECT * FROM book WHERE title = ';';
3

xsys
1       SELECT * FROM book WHERE title LIKE '%
2       %'
3
SELECT * FROM book WHERE title = '
4       '
5

Cosmin
1       SELECT * FROM book WHERE title LIKE '%
2       %'
3
SELECT * FROM book WHERE title = '
4       '
5
```

NiLuJe: Except it might be more robust in theory, but in practice it breaks on our code, because it's getting confused by comments ;).

NiLuJe: So, better be safe than sorry, and stay with a similar "dumb" split ^^.

poire-z:
Looks fine (haven't looked the splitter options).

(Probably not our war to find the right splitter, but just in case, I have this I might have made 10 years ago in python:
```python
    # Commands parsing (do not split if ; in quoted string)
    # commands = commands.split(";")
    cmds = commands
    theLength = len(cmds)
    inQuote = False
    start = 0
    commands = []
    for i in range(theLength):
        nextch = cmds[i]
        if nextch == ';' and not inQuote:
            commands.append(cmds[start:i])
            start = i+1
        if nextch == "\'":
            inQuote = not inQuote
    commands.append(cmds[start:])
```
Might need to catch `--` and zap until end of line to skip comments.
)

poire-z:
No `common/lua-ljsqlite3/init.lua` in this morning Kobo nightly:
```
10/15/20-09:06:18 WARN  Error when loading plugins/coverbrowser.koplugin/main.lua plugins/coverbrowser.koplugin/bookinfomanager.lua:8: module 'lua-ljsqlite3/init' not found:
    no field package.preload['lua-ljsqlite3/init']
    no file 'plugins/coverbrowser.koplugin/lua-ljsqlite3/init.lua'
    no file 'common/lua-ljsqlite3/init.lua'
10/15/20-09:06:19 WARN  Error when loading plugins/statistics.koplugin/main.lua plugins/statistics.koplugin/main.lua:15: module 'lua-ljsqlite3/init' not found:
    no field package.preload['lua-ljsqlite3/init']
    no file 'plugins/statistics.koplugin/lua-ljsqlite3/init.lua'
    no file 'common/lua-ljsqlite3/init.lua'
```

ezdiy:
https://github.com/koreader/koreader-base/pull/1212/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L44 killed a dependency chain that did the init.lua copy. It should be named instead of xsys now.

ezdiy:
@NiLuJe 

10/16/20-00:52:45 WARN  Error when loading plugins/coverbrowser.koplugin/main.lua common/lua-ljsqlite3/init.lua:107: ';' expected near '='

Not really sure what goes on there with that cdef, but I don't remember seeing this in CI.

NiLuJe:
It's always been broken on the CI ;).

PR #1213, opened 2020-10-11
Get rid of debug hooks shenanigans

NiLuJe:
It's currently horribly broken on CircleCI under Busted, for... reasons.
And, surprise surprise, it was originally implemented as... a CI hack.
(in 897906d91a3cbba0eb2aa27583a6510c99ac720d).
Plus, I don't think we actually ever set debug hooks in the first place
anyway?

I fully expect the testsuite to implode on some random new stuff instead,
but, oh, well.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1213)

### DIFF
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -23,7 +23,6 @@ local uint32pt = ffi.typeof("uint32_t*")
 local uint16pt = ffi.typeof("uint16_t*")
 local uint8pt = ffi.typeof("uint8_t*")
 local posix = require("ffi/posix_h") -- luacheck: ignore 211
-local debug = debug
 
 -- the following definitions are redundant.
 -- they need to be since only this way we can set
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
Probably okay, but you could also just say something like `debug = {sethook = function() end}` so a human can easily swap it in and out.

NiLuJe:
AFAIU, nothing is ever actually setting a hook, so I'm not sure it serves
any purpose now, even in the best of times ;).

NiLuJe:
Also, C BB now ;p.

Frenzie:
> AFAIU, nothing is ever actually setting a hook

Depends a bit, if they're in Really Useful™ positions I tend to prefer keeping such things around for the future over cleaning it up, so to speak. ymmv I suppose

PR #1214, opened 2020-10-11
Bump FreeType to 2.10.3

NiLuJe:
c.f., https://git.savannah.gnu.org/cgit/freetype/freetype2.git/tree/docs/CHANGES

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1214)

### DIFF
--- a/thirdparty/freetype2/CMakeLists.txt
+++ b/thirdparty/freetype2/CMakeLists.txt
@@ -29,7 +29,7 @@ endif()
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME
     https://gitlab.com/koreader/freetype2.git
-    VER-2-10-2
+    VER-2-10-3
     ${SOURCE_DIR}
 )
 
--- a/ffi-cdecl/freetype2_decl.c
+++ b/ffi-cdecl/freetype2_decl.c
@@ -2,9 +2,9 @@
 
 #include "ffi-cdecl.h"
 
-#include FT_FREETYPE_H
-#include FT_SYNTHESIS_H
-#include FT_OUTLINE_H
+#include <freetype/freetype.h>
+#include <freetype/ftsynth.h>
+#include <freetype/ftoutln.h>
 
 cdecl_type(FT_Error)
 cdecl_type(FT_Pos)
--- a/ffi/freetype_h.lua
+++ b/ffi/freetype_h.lua
@@ -222,15 +222,15 @@ struct FT_CharMapRec_ {
   short unsigned int platform_id;
   short unsigned int encoding_id;
 };
-FT_Error FT_Init_FreeType(FT_Library *);
-FT_Error FT_New_Face(FT_Library, const char *, long int, FT_Face *);
-FT_Error FT_Set_Pixel_Sizes(FT_Face, unsigned int, unsigned int);
-FT_Error FT_Done_Face(FT_Face);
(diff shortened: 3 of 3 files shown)


NiLuJe:
(Shouldn't cause any trouble, I'm pretty much always running FT master on my desktop).

Frenzie (approved):

yparitcher:
After this when rebuilding Koreader I have to manually delete `libs/libfreetype.so.6` & rebuild (kodev/make don't take care of this themselves) or it crashes with:
```
ffi.load: libs/libmupdf.so
ffi.load (warning): /mnt/us/koreader/libs/libfreetype.so.6: undefined symbol: ft_smooth_lcd_renderer_class
./luajit: ./setupkoenv.lua:27: Not able to load dynamic library: libs/libmupdf.so
stack traceback:
	[C]: in function 'error'
	./setupkoenv.lua:27: in function 'load'
	./ffi/mupdf.lua:25: in main chunk
	[C]: in function 'require'
	frontend/document/canvascontext.lua:14: in main chunk
	[C]: in function 'require'
	./reader.lua:139: in main chunk
	[C]: at 0x0001414d
```

NiLuJe:
That... doesn't surprise me all that much, I rarely do incremental builds for a reason: I don't trust 'em not to mess up the depchain after any significant change ;).

yparitcher:
They almost always work on my end, and ./kodev clean deletes deps downloads so not an option with slow internet.

NiLuJe:
I don't rightly recall where the git checkout ends up, so this may be a moot point, but I tend to do those manually, too ;p (e.g., rm -rf base/thirdparty/*/build or something, plus the actual target dir in build/something IIRC).

Granted, unmetered FTTH means I've never really thought all that much about the download aspect of things.

NiLuJe:
I'm not even sure the buildsystem groks that a version update requires a clean rebuild of at least that subproject?

(And I do know that there are... weird interactions where FT & HB are involved when touching 'em on an incremental build).

yparitcher:
> I'm not even sure the buildsystem groks that a version update requires a clean rebuild of at least that subproject?
> 
> (And I do know that there are... weird interactions where FT & HB are involved when touching 'em on an incremental build).

usually it does, this time it did download and rebuild FT but i guess it didn't replace the library or something.

Frenzie:
*If* it doesn't, and that's a pretty big if except for some minor trouble with libk2pdfopt/leptonica/tesseract I never bothered to fix, it's a simple dependency issue in our Makefile. I probably haven't done a full clean build in years on my main Git checkout.

> I don't rightly recall where the git checkout ends up, so this may be a moot point, but I tend to do those manually, too ;p (e.g., rm -rf base/thirdparty/*/build

Checkout is separate from build, CMake copies it into build so if you delete build it only needs to recompile, not redownload. But it also does that if the CMakeLists.txt changes or one of the patch files, or one of the dependencies. If it didn't it means FT is a dependency for something that isn't specified properly in Makefile.third.

thirdparty/freetype2/CMakeLists.txt
```diff
@@ -29,7 +29,7 @@ endif()
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME
     https://gitlab.com/koreader/freetype2.git
-    VER-2-10-2
+    VER-2-10-3
```
Frenzie, line 32: Okay, changed cmakelists = this will be recompiled as will everything that depends on FT.

PR #1215, opened 2020-10-12
freetype/harfbuzz: APIs to query font metadata

ezdiy:
This is necessary for proper language support in mupdf with arbitrary
fonts in html/pdf - mupdf expects a real font server for its callbacks,
we're told abstract typeface, script and language that must be
accounted for whenever offering close-kin substitutes of a typeface.

There are marginal uses for this elsewhere, eg. localized font
names in UI - https://github.com/koreader/koreader/issues/6763

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1215)


ezdiy:
This is one of the few things that can be chipped away from (already humongous) mupdf as there are no cross dependencies + some marginal use outside of it, so better make splinter PRs.

ffi-cdecl/freetype2_decl.c
```diff
@@ -2,10 +2,23 @@
 
 #include "ffi-cdecl.h"
 
-#include <freetype/freetype.h>
-#include <freetype/ftsynth.h>
-#include <freetype/ftoutln.h>
-
+#include FT_FREETYPE_H
+#include FT_SYNTHESIS_H
+#include FT_OUTLINE_H
```
poire-z, line 7: These #`include FT__xxx` were replaced with `#include <freetype/xxx.h>` in #1214 .
Is that an explicite revert, or you did not see it/forgot it while rebasing?

ffi/harfbuzz_coverage.lua
```diff
@@ -0,0 +1,359 @@
...
+--   * `delta` denotes start of next range since the end of the previous one
+--   * `count` is number of glyphs in range
+-- For instance set of ranges:
+-- 0041..005A, 0061..007A, 00AA, 00BA ...
+-- Becomes:
+-- 65,26, 8,26, 49,1 17,1 ...
+return {
+    -- Table of ranges for entire scripts, ordered by UCDN_SCRIPT_* macros, starting with UCDN_SCRIPT_COMMON
+    -- Source data is https://www.unicode.org/Public/5.2.0/ucd/Scripts.txt
+    -- UCDN is internally used to map character classes (and detect scripts) both by harfbuzz and mupdf
```
poire-z, line 12: Thanks for the comments.
No risk these macros/indices could get out of sync between HB and MuPDF if a new script is introduced between others?

ffi/harfbuzz.lua
```diff
@@ -1,3 +1,118 @@
...
+
 require("ffi/harfbuzz_h")
-return ffi.load("libs/libharfbuzz." .. (ffi.os == "OSX" and "0.dylib" or "so.0"))
+
+local hb_face_t = {}
+hb_face_t.__index = hb_face_t
+ffi.metatype("hb_face_t", hb_face_t)
+
+-- Dump contents of OT name fields
+function hb_face_t:getNames(maxlen)
```
poire-z, line 13: I'm a bit unexperienced with all these metatype and ffi tricks.
But what's this `hb_face_t` ? Where does it end up - as it's not attached to the HB object we get when `require(harfbuzz)` ?
Or is that just attaching additional methods to the libharfbuzz internal hb_face_t type - that other code may use ? (so, these additional methods would be accessible only when used via ffi?)

ffi/harfbuzz.lua
```diff
@@ -1,3 +1,118 @@
...
+    -- nglyph   coverage %
+    0,          100,    -- Simple alphabets of 0-100 glyphs. Be strict, all glyphs are typically in use.
+    100,        99,     -- 100-250 glyphs. abundant diacritics, allow for 1% missing, typically some archaisms
+    250,        98,     -- 250-1000 glyphs. even more diacritics (eg cyrillic dialects), allow for 2% missing
+    1000,       85,     -- 1000 and more = CJK, allow for 15% missing
+}
+
+-- Get script and language coverage
+function hb_face_t:getCoverage()
+    local set, tmp = hb.hb_set_create(), hb.hb_set_create()
```
poire-z, line 47: (Not trying to understand all that part :)

ffi/freetype.lua
```diff
@@ -125,10 +125,46 @@ function FTFace_mt.__index:getHeightAndAscender()
 end
 
 function FTFace_mt.__index:done()
+    -- Note that past this point, accessing self is illegal, FT2 frees the pointer
     assert(ft2.FT_Done_Face(self) == 0, "freetype error when freeing face")
 end
 
-FTFace_mt.__gc = FTFace_mt.__index.done;
```
poire-z, line 131: No more need for this `__gc` ? We should not forget to explicitely call `done()` when done ? :)

ffi/freetype.lua
```diff
@@ -125,10 +125,46 @@ function FTFace_mt.__index:getHeightAndAscender()
...
+        finfo.serif = true
+    elseif lname:match("sans") then
+        finfo.serif = false
+    elseif bit.band(tonumber(self.face_flags), ft2.FT_FACE_FLAG_SFNT) ~= 0 then
+        local os2 = ft2.FT_Get_Sfnt_Table(self, ft2.FT_SFNT_OS2)
+        if os2 ~= nil then
+            local kls = tonumber(ffi.cast("TT_OS2*", os2).sFamilyClass)
+            if kls ~= 0 then -- 0 is usually bogus
+                -- class 8 = sans, otherwise serif
+                finfo.serif = bit.band(kls, 0x0800) == 0
```
poire-z, line 158: Mhhh, in crengine, I killed the loosy support for serif/sans as it looked we couldn't get that info :)
I'll pretend I didn't see that :) (and it would be painful as we'd need the user to choose a main font for serif and another for sans - and I don't really care)

poire-z (commented):
Looks alright.

ffi-cdecl/freetype2_decl.c
NiLuJe: It was officially deprecated in 2.10.3, which is why it was coincidental to #1214 ;).

(It's a remnant of... DOS (!) era short filename FS limitations).

TL;DR: Yes, please use sane "modern" includes ^^.

ffi/freetype.lua
```diff
@@ -125,10 +125,46 @@ function FTFace_mt.__index:getHeightAndAscender()
...
+            end
+        end
+        if finfo.serif == nil then
+            --TODO: Maybe it's worth to consult panose tables too?
```
NiLuJe, line 162: They're also often depressingly wrong ^^.

It's been a recurrent source of issues for Access (the !RMSDK rendered used on Kobos).

To the extent that I usually run my fonts through this: http://trac.ak-team.com/trac/browser/niluje/Configs/trunk/Kindle/Kobo_Hacks/Patches/fix-panose.py

ffi/freetype.lua
ezdiy: It never worked properly (and be glad, as it would lead to a double-free). __gc is used only for types that do ffi.new() - which we don't (we ffi new its pointer box, different type).

ffi/freetype.lua
ezdiy: its not really all that accurate (panose tells a bit more for *some* fonts), but it's still 50/50, with heavy preference for serif, so eh.

ffi/harfbuzz.lua
ezdiy: LuaJIT attaches it implicitly to any `hb_face_t*` typed C pointer we spawn through `hb_face_create_*` calls in various places. This is not as unparadigmatic as it may seem - lua does same sort of thing if you attach metatable to string, number or a function - in which case you too get mt per *type*, as opposed to *value* as would happen to UD/table.

ffi-cdecl/freetype2_decl.c
ezdiy: oops, indeed accidental rebase resolution

ffi/harfbuzz_coverage.lua
ezdiy: In mu/hb, the order is hardcoded *without* regard for the source data (that is compiled into UC DB contents later). So I'm essentially doing the same here (in gen.lua). The symbolic indices aren't all that interesting to us - mupdf tells us this indice directly as int, for HB its a bit awkward because they translate the UCD canonical order to their own `hb_script_t`, maintaining a lookup table between the two which sucks. Luckily, these indices are probably never gonna be used with HB directly.

PR #1217, opened 2020-10-14
Pocketbook: support for /dev/input

ezdiy:
We can bypass inkview event loop entirely. Needs root, though.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1217)


NiLuJe:
Ooh, shiny! :]

ffi/input_pocketbook.lua
```diff
@@ -26,23 +32,32 @@ local function genEmuEvent(t,c,v)
...
+    return ts
+end
+
 local function translateEvent(t, par1, par2)
     if eventq == nil then
         return 0
     end
-    local sec, usec = util.gettime()
-    ts = { sec = sec, usec = usec }
+    getTimestamp()
```
NiLuJe, line 50: Stupid question: isn't there a scope issue here?

(I genuinely have no idea how local functions behave in Lua: is `getTimestamp`'s scope tighter than `translateEvent`'s, or does this behave like a C macro, so that `translateEvent` now has access to `getTimestamp`'s `sec` & `usec` variables?

ezdiy: Nope, `ts` is simply a "global" (as in, in the outermost scope) `gts()` just updates it with current value, as we're soon to be generating events with same timestamp in bulk (so they all share same ts). Later on, genEmuEvent picks it up from the same outer scope. I suppose this could be renamed to updateTimestamp() to make it more clear.

NiLuJe: Oh, duh, I missed the scoping the `ts`, was too focused on `sec` & `usec`, thanks ;)

PR #1218, opened 2020-10-15
Pocketbook: Add support for waveform presets.

ezdiy:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1218)


ezdiy:
This is hopefuly more general solution to problems like https://www.mobileread.com/forums/showpost.php?p=4046967&postcount=320 as we have now "slow, but always work" baseline shipped by default, that can be bumped from UI up to a level that's useable by the user.

NiLuJe:
Time will tell, but, yeah, this kind of approach probably makes sense given the vast variety of PB devices, which would make hard-coded, but more fine-grained tweaks like on Kindle/Kobo a somewhat hairy proposition ;).

NiLuJe (approved):

ffi-cdecl/include/mxcfb-pb.h
```diff
@@ -107,20 +107,22 @@ struct mxcfb_rect {
...
 #define EPDC_WFTYPE_A2IN			6
 #define EPDC_WFTYPE_A2OUT			7
-#define EPDC_WFTYPE_DU4				8
-#define EPDC_WFTYPE_AA				9
-#define EPDC_WFTYPE_AAD				10
-#define EPDC_WFTYPE_GC16HQ			15
+#define EPDC_WFTYPE_DU4				8 // not B288
+#define EPDC_WFTYPE_AA				9 // not B288
+#define EPDC_WFTYPE_AAD				10 // not B288
+#define EPDC_WFTYPE_GS16			14 // B288 only. More aggressive than GC16, less than GL16.
```
NiLuJe, line 116: Just for information, is that one constant buried somewhere in the B288 kernel?

ezdiy: yep, https://github.com/pocketbook/kernel-b288/blob/master/drivers/video/sunxi/disp2/disp/eink.h

PR #1228, opened 2020-11-03
Bump crengine for non-linear flow support

Jellby:
To be merged after https://github.com/koreader/crengine/pull/391


poire-z:
You'll need to update the crengine bump/hash _once_ https://github.com/koreader/crengine/pull/391 is merged.

poire-z:
This looks allright.
But as there's quite a bit of activity in koreader-base, and so that it does not block other things and is not carried in frontend with a base bump by another PR, we'll merge it once we reviewed your frontend part - so go ahead with the frontend PR even if it fails CI because base does not yet bring this.

poire-z:
I won't be able to rename the 2nd commit if we keep them two. So, could you make it more explicite, like the title of this PR or `bump crengine: support for EPUB non-linear flows`.
And if you're fine with the frontend PR (I am), we can merge this one so you can update the other one.

Jellby:
Yes, I'm fine with the frontend, I'll update it once this is merged.

PR #1205, opened 2020-10-03
blitbuffer: Always enable JIT workaround

ezdiy:
This is necessary even with CBB on, for Lua fallbacks.

(maybe) fixes https://github.com/koreader/koreader/issues/6736

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1205)

PR #1233, opened 2020-11-21
Add a function to write BMP-files

zwim:
This function works on the emulator and on Tolino.

I have tested it with the cover-image plugin. On the Tolino even a BMP file is shown as supend image if it is renamed to suspend_others.jpg

We would need a function to write BMP files, for making cover-image usable on Pocketbooks.
https://github.com/koreader/koreader/pull/6813#issuecomment-725569462

I have put the function in `pic.lua` as I am not quite sure, how to do implement this in `blitbuffer.lua`.
When I put it just as `function BB:writeBMP(filename)` I get the error `'struct BlitBufferRGB24' has no member named 'writeBMP'`. 
~~How can I implement it in `blitbuffer.lua`? ~~

#1205 has to be reverted -> see #1235 

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1233)


Frenzie:
Search for writePNG and you'll see. ;-)

ffi/pic.lua
```diff
@@ -372,4 +372,84 @@ function Pic.openDocument(filename)
     end
 end
 
+--[[
+function for writing BMP to file
+]]
+
+local function write_uint32(of, data )
+    of:write( string.char(data % 256) )
```
NiLuJe, line 380: An `& 255` (well, `bit.and(data, 255)` in Lua ;p)  will be faster (and identical, because unsigned maths and 256 is a power of two).

ffi/pic.lua
```diff
@@ -372,4 +372,84 @@ function Pic.openDocument(filename)
...
+
+    for i = 1,24 do
+      of:write(string.char(0))
+    end
+
+    local pos = 0
+    -- start with lowes row
+    for y = h-1, 0, -1 do
+        pos = y * w * output_channels
+        for x = 0,w-1 do
```
NiLuJe, line 438: nit: space after the comma

ffi/pic.lua
```diff
@@ -372,4 +372,84 @@ function Pic.openDocument(filename)
...
+     -- biHeight (4 Bytes)
+    write_uint32(of, h )
+    -- biPlanes ( 2 Bytes)
+    of:write(string.char(1))
+    of:write(string.char(0))
+    -- biBitCount (2 Bytes)
+    of:write(string.char(output_channels * 8))
+    of:write(string.char(0))
+
+    for i = 1,24 do
```
NiLuJe, line 430: ditto

ffi/pic.lua
```diff
@@ -372,4 +372,84 @@ function Pic.openDocument(filename)
...
+
+    local filesize = output_channels * w * h + 54
+
+    -- bfType (2 Bytes)
+    of:write("B")
+    of:write("M")
+    -- bfSize (4 Bytes)
+    write_uint32(of, filesize)
+    -- bfReserved (4 Bytes )
+    write_uint32(of, 0 )
```
NiLuJe, line 413: nit: space before closing paren

ffi/pic.lua
```diff
@@ -372,4 +372,84 @@ function Pic.openDocument(filename)
...
+    -- biBitCount (2 Bytes)
+    of:write(string.char(output_channels * 8))
+    of:write(string.char(0))
+
+    for i = 1,24 do
+      of:write(string.char(0))
+    end
+
+    local pos = 0
+    -- start with lowes row
```
NiLuJe, line 435: lowest? (a mention that BMP is weird and bottom-to-top would make that clearer).

ffi/pic.lua
```diff
@@ -372,4 +372,84 @@ function Pic.openDocument(filename)
...
+    for y = h-1, 0, -1 do
+        pos = y * w * output_channels
+        for x = 0,w-1 do
+            of:write(string.char(data[pos+2]))
+            of:write(string.char(data[pos+1]))
+            of:write(string.char(data[pos]))
+            pos = pos + 3
+        end
+        -- fill up a row to a multiply of 4 bytes
```
NiLuJe, line 444: a multiple

ffi/pic.lua
```diff
@@ -372,4 +372,84 @@ function Pic.openDocument(filename)
...
+        if pos % 4 ~= 0 then
```
NiLuJe, line 445: The same AND trick applies here.

Namely, MOD n == AND n-1 where n is a power of two (and unsigned).

NiLuJe:
Hint: it involves the BB_mt meta table ;p.

zwim:
> Hint: it involves the BB_mt meta table ;p.

Thank you, I had it there, but after the metatable-set :-(

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,81 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+        pos = y * w * output_channels
+        for x = 0, w-1 do
+            of:write(string.char(data[pos+2]))
+            of:write(string.char(data[pos+1]))
+            of:write(string.char(data[pos]))
+            pos = pos + 3
+        end
+        -- fill up a row to a multiple of 4 bytes
+        if band(pos, 3) ~= 0 then
+            for i = 0, (4 - pos % 4) * output_channels do
```
NiLuJe, line 1895: Still a MOD left ;p.

zwim: OK

zwim:
How can I correct the CircleCI warning on `ffi/blitbuffer.lua:1883:11: value assigned to variable 'pos' is unused`

NiLuJe:
Move the declaration inside the loop, on the first actual store (the y * w * stride) one ;).

zwim:
> Move the declaration inside the loop, on the first actual store (the y * w * stride one) ;).
> 
> (i.e., tighter scoping, you don't actually need pos outside the loop, and you don't need its result across iterations, since it's recomputed first thing each iter).

I see.

zwim:
The code is ready from my side:

I have done some benchmarks found surprising results (always for the same cover image).

On the Emulator (small screen: bmp-filesize=1166454  png-filesize=469032)
* Strogin the same png takes 0,24s.
* Time is between 0,027s and 0,041s.

On Tolino (big screen: bmp-filesize=8294454  png-filesize=469032)
* Storing the same png takes 1,4s.
* Most of the time the bmp-creation takes  ~1,7s
* sometimes ~1,4s and 
* seldom 97s! 

If I restart the Tolino or reinstall the same apk again, the creation is fast again.
Killing KOReader does not help.

Am I right, we can not blame JIT for this?
Is this an Android feature?
How can I track down the problem?

zwim:
Update: When the creation takes long, then I can see `proc/meminfo` that something is eating up the memory slowly.
On start of the creation there are ~31MB free, at the end ~15MB.
I have tried `of:setvbuf("full", 8192)` but with no success.

NiLuJe:
You can check by disabling the JIT for this function (`jit.off(func)` right below it). It s gold be consistently slow this way ;p.

zwim:
You are right its consistently slow with `jit.off`, but way faster than the 100s.
It takes 9.9s

NiLuJe:
That's the device where the super-slow screenshots were originally reproduced, too, right?

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,79 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+    local stride = w * output_channels
+
+    local of, err = io.open(filename, "wb")
+    if err ~= nil then
+        return err
+    end
+
+    local bbdump = BB.new(w, h, TYPE_BBRGB24, nil)
+    bbdump:blitFrom(self)
+    local data = ffi.cast("unsigned char *", bbdump.data)
```
NiLuJe, line 1852: You *should* be able to use self.data directly if self:getType() is BBRGB24 or BBRGB32

(In which case, keep track of this optional intermediate buffer, because of the final explicit free).

NiLuJe: (If you want to go the extra mile, you can make type-speciifc functions like writePNG, as there's possibly a Grayscale 8bpp BMP pixel-format, but it may or may not require building a palette).

NiLuJe:
I would also possibly try working on an in-memory buffer, and only at the end dump the' whole thing to a file, as the whole mess of io + `string.char` is possibly not helping...

ffi/blitbuffer.lua
zwim: I see that this would save a `blitFrom(self)`, but would that help with the timing?

I have a `logger.dbg` after `local data = ffi.cast`. The message appears immediately, so the bbdump is not the cause.

zwim:
> That's the device where the super-slow screenshots were originally reproduced, too, right?

Yes. 

It seems a problem with the jit optimizer. If hava a `jit.opt.start(1)` before the `writeBMP` I have a constant timing of 3.4s.
This is about half so fast as the fastest measures time.

zwim:
> I would also possibly try working on an in-memory buffer, 

Good point. I will try that tomorrow.

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,79 @@ function BB_mt.__index:writePNGFromBGR(filename)
     C.free(cdata)
 end
 
+
+local function write_uint32(of, data)
```
pazos, line 1830: This function is only called from writeBMP. You can put it inside.

ffi/blitbuffer.lua
NiLuJe: Probably not, because in these cases, the type matches, so the blit will be fast (sub 10ms).

But it might help alleviate pressure on the JIT nonetheless ;).

NiLuJe:
> > That's the device where the super-slow screenshots were originally reproduced, too, right?
> 
> Yes.
> 
> It seems a problem with the jit optimizer. If hava a `jit.opt.start(1)` before the `writeBMP` I have a constant timing of 3.4s.
> This is about half so fast as the fastest measured time.

If it's the same deal than what the current workarounds in place attempt to help with, yeah, probably the loop optimizer, which isn't very happy with our BB pixel loops.

It's already enabled at -O2, so you *should* be able to reproduce the inconsistencies at `jit.opt.start(2)`, too.

If you feel like a few rounds of running around like a headless chicken, feel free to try to tweak the existing JIT opt knobs in there, the giant comment above it has all the history and the few things we tried.

zwim:
`jit.opt.start(2)` yields to the same slow performance.

I have played with `loopunroll`:
When I put `jit.opt.start("loopunroll=x")` with 5<=x<=20 just before `writeBMP`, there is no performance problem; with x>25 the problem arises.
When I delete `jit.opt.start("loopunroll=45")` on line 145 in blitbuffer.lua, the problem is gone. I found no negative performance impacts in KOReader (of course, I have not tested everything).

In my understanding the history was the following:
I created an issue https://github.com/koreader/koreader/issues/6736.
On Oct. 4 @ezdiy proposed a solution for the slow screenshot (the loopunroll), which worked at that time.
On Oct. 5 @NiLuJe improved blitbuffer.lua for "less crappy performance". This was a massive change in blitbuffer.lua which counteracted the improvement of the loopunroll. But nobody noticed that, because nobody used the folded loops on a slow device.

What should I do now? Remove line 145 or add `jit.opt.start("loopunroll=15")` before writeBMP?

NiLuJe:
I'd be all for getting rid of the loopunroll hack, but I'm slightly wary of that, because the screenshot thing was simply the more visible aspect of the issue: *everything* in BB was slower.

Does simply reverting #1205 yield the same results?

NiLuJe:
Also, re-reading #1206: if you try my earlier "don't go through a temp bb" idea, do benchmark it, because it really was dramatically slower for PNGs ;).

zwim:
> Also, re-reading #1206: if you try my earlier "don't go through a temp bb" idea, do benchmark it, because it really was dramatically slower for PNGs ;).

I have it in, since you suggested it. It is an improvement anyway. (But it does not solve the performance problem.)

NiLuJe:
It'll need to be conditional on rota being 0, though, I'd forgotten about that bit ;).

(The stride issue shouldn't be a problem here, since you're doing it scanline by scanline manually because of the BMP alignment quirks).

zwim:
> Does simply reverting #1205 yield the same results?

With reverted #1205 I get stable timings. 
Timings for reverted #1205 and `deleted unroll=45`:
`writePNG` takes 2.0s (this is the same as with HEAD)
`writeBMP` takes 1.5s to 1.8s varying.

On a screenshot I don't feel a difference.
What do you mean, shall I revert #1205?

I will add the copy on rota and push what I have, soon. (the `logger.err` will be removed later)

zwim:
Do we really need to check `getRotation` of the cover image bb? Isn't it always 0?

When testing #6ce10f3, I noticed this does not work on Tolino. The generated image is OK, but the FW displays it rotated to portrait and stretched to fit the screen. It looks ugly. (But performance is OK.)

On the emulator, the generated image is OK.

Maybe we have to add another switch to ignore the orientation of cover-image.

NiLuJe:
I'd rather #1205 be reverted in a dedicated PR to make it easier to track (or re-revert ;p) ;).

NiLuJe:
> Do we really need to check `getRotation` of the cover image bb? Isn't it always 0?

That's a good point, I was stuck on writePNG where the source BB is the screen, not an image ;p.

zwim:
The FW thing on Tolino (that any image is shown in portait mode) also affects the created `png`. So I would like to handle this on a different PR, as this would mean to also change `writePNG`.
Is this OK for you?

zwim:
> I'd rather #1205 be reverted in a dedicated PR to make it easier to track (or re-revert ;p) ;).

I will revert the revert on my PR. 
Could you do a PR with the reverted #1205 on koreader-base, or should I do it?

NiLuJe:
Feel free to do it :) (via a `git revert 6705c0281996e4ab96472c89fbb50b471a2d5bd1` and a cross-link to this issue in the log message).

zwim:
I would be ready for review now. Is there something to do on the failing (macos-latest) test?

NiLuJe:
Nope, XCode is the worst :).

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,99 @@ function BB_mt.__index:writePNGFromBGR(filename)
     C.free(cdata)
 end
 
+--jit.opt.start(1)
+--jit.opt.start("loopunroll=10")
```
NiLuJe, line 1830: Get rid of those?

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,99 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+                of:write(string.char(0))
+            end
+        end
+    end
+    of:close()
+    if bbdump ~= nil then
+        bbdump:free()
+    end
+end
+--jit.off(BB_mt.__index.writeBMP)
```
NiLuJe, line 1911: Ditto ;).

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,99 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+    local stride = w * output_channels
+
+    local of, err = io.open(filename, "wb")
+    if err ~= nil then
+        return err
+    end
+
+    local bbdump
+    local data
+    if self:getType() == TYPE_BBRGB24 then
```
NiLuJe, line 1853: You should be able to use this on BBRGB32, too, you'll just have to keep track of two different `pos` later on: the input one (using the source bb's stride), and the output one (using the output bb's stride).

NiLuJe:
Any news on the "woorking on a buffer and not a file" front?

ffi/blitbuffer.lua
zwim: Yes.

ffi/blitbuffer.lua
zwim: Yes.

ffi/blitbuffer.lua
zwim: OK, I can do it. But how should I test it. Although I know how to do it, I don't feel comfortable just to program it without a test.
Is there a switch on the emulator for using RGB32?

zwim:
> Any news on the "working on a buffer and not a file" front?

Not now, as the performance of this solution is as good as the PNG one, I don't quite see the need.

My Tolino has at worst I have seen 12MB memfree. The image buffer would need 8MB

ffi/blitbuffer.lua
NiLuJe: Both Android and the emulator should be running on an RGB32 screen buffer, IIRC, so running that from the screen BB like the Screenshotter does instead of an image file should do the trick?

NiLuJe: But, again, I'm thinking in terms of Screen buffers: this may be a terrible idea on actual RGBA images, since you wouldn't be doing alpha-blending.

TL;DR: Eh, probably don't bother ;p.

NiLuJe:
> > Any news on the "working on a buffer and not a file" front?
> 
> Not now, as the performance of this solution is as good as the PNG one, I don't quite see the need.
> 
> My Tolino has at worst I have seen 12MB memfree. The image buffer would need 8.1MB.
> I don't think it is clever to waste that memory, maybe luajit is upset again :-)

This is Linux, actually "free" memory is always terribly low ;p.

I still think there's gigantic overhead here, because this should barely be any slower than a rotated blitFrom (which sits in at around 100ms).

pazos:
> This is Linux, actually "free" memory is always terribly low ;p.

There's a still relevant link: https://www.linuxatemyram.com/ :smile:

zwim:
> > > Any news on the "working on a buffer and not a file" front?
> > 
> > 
> > Not now, as the performance of this solution is as good as the PNG one, I don't quite see the need.
> > My Tolino has at worst I have seen 12MB memfree. The image buffer would need 8.1MB.
> > I don't think it is clever to waste that memory, maybe luajit is upset again :-)
> 
> This is Linux, actually "free" memory is always terribly low ;p.
> 
> I still think there's gigantic overhead here, because this should barely be any slower than a rotated blitFrom (which sits in at around 100ms on a Forma) (outside of file I/O).

Should not be too hard to implement, I'll do it tomorrow, then we know more. ;-)

NiLuJe:
You'll want to go with FFI an C data types, to avoid Lua's insane "everything's a double" mantra ;).

Hell, if it weren't for the alignment quirk, you probably *could* just make the inner loop a rotated blitFrom (or near enough ;) ^^.

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,99 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+    else
+        bbdump = BB.new(w, h, TYPE_BBRGB24, nil)
+        bbdump:blitFrom(self)
+        data = ffi.cast("unsigned char *", bbdump.data)
+    end
+
+    local filesize = stride * h + 54
+    -- update filesize, if stride is not a multiple of 4
+    if band(stride, 3) ~= 0 then
+        filesize = filesize + (4 - band(stride,3))*h
```
NiLuJe, line 1864: You could also remember the amount of padding bytes computed here (i.e., `4 - (stride & 3)`) and then use that later when you need to fill-in the padding bytes at the end of each row, instead of re-computing it based on `pos`.

(This becomes a moot point if you switch to a `calloc` buffer, since it'll be zero-initialized, so, it'll come pre-padded already ;p).

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,99 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+    for i = 1, 24 do
+        of:write(string.char(0))
+    end
+
+    local pos
+    -- start with bottom line, because BMP stores from bottom to top
+    for y = h-1, 0, -1 do
+        pos = y * stride
+        for x = 0, w-1 do
+            of:write(string.char(data[pos+2]), string.char(data[pos+1]), string.char(data[pos]))
```
NiLuJe, line 1896: Okay, I'd forgotten that this nastiness was BGR, which makes scanline-by-scanline copies a pipedream, so, forget my blitFrom comparisons, it's going to be slower ;).

ffi/blitbuffer.lua
zwim: This would be an optimization, but maybe it is already done by CSE.

zwim:
In the commit `Buffer not File` i have done a first implementation of that idea.

Good result: The generated files are identical and on the emulator the buffer impl. is faster

Bad result: On Tolino the generation is ridiculously slow (89s).
From the debuggin output: "xxxxxxxxxxxx target_pos=54" to the next "xxxxxxxxxxxx begin dump" it takes 87s.
I speculated it was something to do with the f**k**g jit again. Applying a `jit.opt.start(1)` before `writeBMP` the perframance increase to 0.39s which IS impressive.

Any ideas what to do with jit?

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,216 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+        data = math.floor(data / 256)
+        target_ptr[target_pos] = band(data, 255)
+        target_pos = target_pos + 1
+        data = math.floor(data / 256)
+        target_ptr[target_pos] = band(data, 255)
+        target_pos = target_pos + 1
+        data = math.floor(data / 256)
+        target_ptr[target_pos] = data
+        return target_pos + 1
+    end
```
NiLuJe, line 1842: Can be rewritten in a much more C-like fashion:

```
*ptr++ = (data & 0xFF);
*ptr++ = ((data >> 8) & 0xFF);
*ptr++ = ((data >> 16) & 0xFF);
*ptr++ = ((data >> 24) & 0xFF);
```

Since pointer derefs use array notation in Lua/FFI, this should turn into something like

```
target_ptr[target_pos] = band(data, 0xFF)
target_ptr[target_pos + 1] = band(rshift(data, 8), 0xFF)
target_ptr[target_pos + 2] = band(rshift(data, 16), 0xFF)
target_ptr[target_pos + 3] = band(rshift(data, 24), 0xFF)
return target_pos + 4
```

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,216 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+    -- start with bottom line, because BMP stores from bottom to top
+    for y = h-1, 0, -1 do
+        pos = y * stride
+        for x = 0, w-1 do
+            target_ptr[target_pos] = data[pos+2]
+            target_pos = target_pos + 1
+            target_ptr[target_pos] = data[pos+1]
+            target_pos = target_pos + 1
+            target_ptr[target_pos] = data[pos+0]
+            target_pos = target_pos + 1
```
NiLuJe, line 1921: I'd also rewrite this, mainly to make the RGB <-> BGR swap exceedingly obvious:

```
            target_ptr[target_pos] = data[pos + 2]
            target_ptr[target_pos + 1] = data[pos + 1]
            target_ptr[target_pos + 2] = data[pos]
            target_pos = target_pos + 3
```

NiLuJe: Or even `target_pos = target_pos + output_channels`, to match its friend `pos` ;).

ffi/blitbuffer.lua
zwim: Much better.

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,216 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+    target_pos = write_uint32(target_ptr, target_pos, 40)
+    -- biWidth (4 Bytes)
+    target_pos = write_uint32(target_ptr, target_pos, w)
+     -- biHeight (4 Bytes)
+    target_pos = write_uint32(target_ptr, target_pos, h)
+    -- biPlanes ( 2 Bytes)
+    target_ptr[target_pos] = 1
+    target_pos = target_pos + 1
+    target_ptr[target_pos] = 0
+    target_pos = target_pos + 1
```
NiLuJe, line 1891: You could either whip up a `write_uint16` for those if you want to make it clearer, or just write the 1 manually, and just increase the pos by 2

(Remember, calloc, everything's zero ;)).

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,216 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+    -- biBitCount (2 Bytes)
+    target_ptr[target_pos] = output_channels * 8
+    target_pos = target_pos + 1
+    target_ptr[target_pos] = 0
+    target_pos = target_pos + 1
+
+    for i = 1, 24 do
+        target_ptr[target_pos] = 0
+        target_pos = target_pos + 1
+    end
```
NiLuJe, line 1901: Same idea here, I'd keep it commented with a comment that those few remaining fields belong to the header, but we don't care about 'em, so leave them set to 0.

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,216 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+
+    local filesize = stride * h + 54
+    local padding = 0
+    -- update filesize, if stride is not a multiple of 4
+    if band(stride, 3) ~= 0 then
+        padding = 4 - band(stride, 3)
+        filesize = filesize + padding * h
+    end
+
+    local target_ptr = ffi.cast("unsigned char*", C.calloc(filesize, 1))
```
NiLuJe, line 1866: I would *possibly* store the original result of calloc in a separate variable, use *that* one for the `free` call, and cast it to `uint8pt` separately.

This shouldn't matter much in practice, but you never know ;).

ffi/blitbuffer.lua
NiLuJe: Without forgetting the a `target_pos =  target_pos + 24` ;p.

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,111 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+            target_pos = target_pos + output_channels
+            pos = pos + output_channels
+        end
+        -- no need to fill up a row to a multiple of 4 bytes, as calloc initializes with 0
+        target_pos = target_pos + padding
+    end
+
+    if target_pos ~= filesize then
+        require("logger").err("Cover image: internal error:  filesize=" .. filesize .. " target_pos=" .. target_pos)
+    end
```
NiLuJe, line 1914: You should probably make that a real `assert`, I think importing frontend stuff is frowned upon in base ;).

zwim: Of course, I forgot that.

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,109 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+        target_ptr[target_pos + 2] = band(rshift(data, 16), 0xFF)
+        target_ptr[target_pos + 3] = band(rshift(data, 24), 0xFF)
+        return target_pos + 4
+    end
+
+    local output_channels = 3
+    local w, h = self:getWidth(), self:getHeight()
+    local stride = w * output_channels
+
+    -- we do this before all buffer alocations
```
NiLuJe, line 1842: nit: allocations

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,109 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+
+    local filesize = stride * h + 54
+    local padding = 0
+    -- update filesize, if stride is not a multiple of 4
+    if band(stride, 3) ~= 0 then
+        padding = 4 - band(stride, 3)
+        filesize = filesize + padding * h
+    end
+
+    local target_buff = C.calloc(filesize, 1) -- initiate array with zero
```
NiLuJe, line 1866: nit: initialize

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,109 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+    local target_ptr = ffi.cast("unsigned char*", target_buff)
```
NiLuJe, line 1867: nit: You can use `uint8pt` instead of `"unsigned char*"`. (It's from a cached `ffi.typeof` call).

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,109 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+    -- bfReserved (4 Bytes)
+    target_pos = write_uint32(target_ptr, target_pos, 0)
+    -- bfOffBits (4 Byte)
+    target_pos = write_uint32(target_ptr, target_pos, 54)
+
+    -- biSize (4 Byte)
+    target_pos = write_uint32(target_ptr, target_pos, 40)
+    -- biWidth (4 Bytes)
+    target_pos = write_uint32(target_ptr, target_pos, w)
+     -- biHeight (4 Bytes)
```
NiLuJe, line 1886: nit: stray space before the comment ^^.

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,109 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+    target_pos = write_uint32(target_ptr, target_pos, h)
+    -- biPlanes ( 2 Bytes)
```
NiLuJe, line 1888: nit: stray space before the opening paren

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,109 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+    -- the next 24 bytes are zero (calloc already does that)
+    target_pos = target_pos + 24
+
+    -- start with bottom line, because BMP stores from bottom to top
+    for y = h-1, 0, -1 do
+        local pos = y * stride
+        for x = 0, w-1 do
+            target_ptr[target_pos] = source_ptr[pos+2]
+            target_ptr[target_pos + 1] = source_ptr[pos+1]
+            target_ptr[target_pos + 2] = source_ptr[pos]
```
NiLuJe, line 1904: nit: Unify the `p+n` vs. `p + n` syntax. I don't have a preference, and there's no clear consensus in the rest of the module's code, so, whichever you prefer ;).

ffi/blitbuffer.lua
```diff
@@ -1863,8 +1863,8 @@ function BB_mt.__index:writeBMP(filename)
         filesize = filesize + padding * h
     end
 
-    local target_buff = C.calloc(filesize, 1) -- initiate array with zero
-    local target_ptr = ffi.cast("unsigned char*", target_buff)
+    local target_buff = C.calloc(filesize, 1) -- initialize array with zero
+    local target_ptr = ffi.cast("uint8pt", target_buff)
```
NiLuJe, line 1867: Unquoted, it's a variable ;).

zwim: Oh!

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,109 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+    -- we do this before all buffer allocations
+    local of, err = io.open(filename, "wb")
+    if err ~= nil then
+        return err
+    end
+
+    local bbdump
+    local source_ptr
+    if self:getType() == TYPE_BBRGB24 then
+        source_ptr = ffi.cast("unsigned char *", self.data)
```
NiLuJe, line 1851: nit: `uint8pt`

ffi/blitbuffer.lua
```diff
@@ -1826,6 +1826,109 @@ function BB_mt.__index:writePNGFromBGR(filename)
...
+    else
+        bbdump = BB.new(w, h, TYPE_BBRGB24, nil)
+        bbdump:blitFrom(self)
+        source_ptr = ffi.cast("unsigned char *", bbdump.data)
```
NiLuJe, line 1855: ditto

NiLuJe (approved):
LGTM \o/

PR #1236, opened 2020-11-23
Minor third-party updates

NiLuJe:
* FBInk v1.23.0 (basically what we were running before, but on a tagged release ;p)
* KoboUSBMS v1.0.0 (I expect that the version number alone will trigger new and interesting failures!).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1236)


Frenzie (approved):
1.0? w00t!

PR #1238, opened 2020-11-24
BB: Ensure consistent types are being used in writePNG

NiLuJe:
Momentary brainfart lead to me using a char* instead of an unsigned
char*
Which doesn't change much in the grand scheme of things, but, oh, well.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1238)


NiLuJe:
(This should be a purely cosmetic tweak).

Frenzie (approved):
BGR huh :-P

NiLuJe:
AKA the bane of my existence after RGB565 ;p.

Issue #1237, opened 2020-11-23
Android: remove versioning from soname


pazos:
So all libraries are called `whatever.so` instead of `whatever.so.x.y`. 

This is a prerequisite for https://github.com/koreader/android-luajit-launcher/pull/175 and would improve https://github.com/koreader/koreader/pull/6881 by moving libraries extraction from first run to install time.

I have no idea how to do that, pinging @NiLuJe

NiLuJe:
Most of 'em already do that by default, otherwise it's messy, and, for autotools stuff, it involves reverting the libtool hack that we need for I don't even remember what that disabled Android-specific quirks, one of which was that soname trickery, IIRC.

As for Cmake stuff, jpeg-turbo is broken when built in Android mode, which also did the same thing to the soname, besides breaking it ;p.

NiLuJe:
Doing it post-build involves modifying the ELF header, which patchelf can probably do, but then patchelf is known to randomly break some binaries, so, eh ^^.

Frenzie:
A prerequisite? Seems unlikely. ;-)

For many of them you could simply remove my patch, but I put that in there
to work around specific Android issues (among other things it's a simple
hack to avoid clashes with system libs like zlib iirc 'cause bionic kinda
sucks). Cf. https://github.com/koreader/koreader-base/pull/521

pazos:
Uuuuuh, In principle adding `-Wl,-soname whatever.so` in ldflags is what the NDK does automatically. I see that we do that for some libs: https://github.com/koreader/koreader-base/search?q=-Wl%2C-soname

As I don't have any idea about autotools (or generally about build tools) I have no idea if this is feasible.

Anyways, the list of libs that need a change in soname are:

![sos](https://user-images.githubusercontent.com/975883/100015987-0b8ee780-2dd9-11eb-83fa-f4687922eb33.png)

pazos:
> For many of them you could simply remove my patch, but I put that in there
to work around specific Android issues (among other things it's a simple
hack to avoid clashes with system libs like zlib iirc 'cause bionic kinda
sucks). Cf. #521

Oh, that's interesting. Indeed they can clash with system libraries, but I have no way to check that. Did it happen at some point?

Frenzie:
Yes, see some of the linked issues. That libjpeg.so.8 you mentioned is
guaranteed to break.

You can of course change the name to androidsuckslibjpeg.so instead.

NiLuJe:
[...]

PR #1241, opened 2020-11-28
Add functionality to write jpg-files

zwim:
This PR allows to store jpg-files. 

Benchmarks on some selected books on the
Emulator time in seconds: 
      png: 0.19, 0.055
      bmp: 0.0017, 0.0019
      jpg: 0.0025, 0.0022 (does not depend on quality)
Tolino:
     png: 2.19, 6.00, 4.80
     bmp: 0.21, 0.18, 0.19 **Sometimes I get the problems with "loopunroll" again, I will look into it.**
     jpg: 0.30, 0.38, 0.37 (does not depend on quality)

Filesize on Tolino (big screen) and a detailed image:
    jpg: 8294kB, 2852kb, 304kB (quality 75)
Filesize on Tolino (big screen) and a simple image:
    jpg: 8294kB, 628kb, 140kB (quality 75)

Questions:
I have put it in blitbuffer, as the other writefile functions are also in there (but then I have to load the shared libs also in blitbuffer, dunno if this is a problem), or would it be better to put it in pic and write just a wrapper function in `blitbuffer:writeFile`? The drawback of the latter would be to "load" pic.lua in blitbuffer.lua.

IMHO the best would be to add a `jpeg.lua` file to deal with all jpg things and "lacy load" this in `pic.lua` and `blitbuffer.lua`.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1241)


ffi/blitbuffer.lua
```diff
@@ -24,6 +24,17 @@ local uint16pt = ffi.typeof("uint16_t*")
 local uint8pt = ffi.typeof("uint8_t*")
 local posix = require("ffi/posix_h") -- luacheck: ignore 211
 
+local _ = require("ffi/turbojpeg_h")
+local turbojpeg
+if ffi.os == "Windows" then
+    turbojpeg = ffi.load("libs/libturbojpeg.dll")
```
pazos, line 30: The libraries are named the same. 

```lua
local turbojpeg = ffi.load("turbojpeg")
```
should work.

zwim: I have copied this from `pic.lua`, so should it be changed there too?

pazos: > I have copied this from pic.lua, so should it be changed there too?

If it works, yeah. But please keep it in a different PR.

At the end `libs` is part of cpath and we shouldn't hardcode the full so name if there's no reason to. It makes some things more difficult. For instance: https://github.com/koreader/koreader-base/issues/1237

Sadly it is used on a bunch of files: https://github.com/search?q=org%3Akoreader+ffi.load%28%22libs&type=code. Also replacement is not as easier as it seems, because we break expectations on some libraries (for example we built some of them as .so in osx, where the OS expect dylib)

But for new cases it is indeed better (if it works :p)

Frenzie: > If it works, yeah. But please keep it in a different PR.

^ That. I imagine pic.lua predates the cpath changes or some such.

NiLuJe:
Yep, definitely move the encode function to a dedicated jpeg.lua, like png.lua ;).

ffi/jpeg.lua
```diff
@@ -0,0 +1,99 @@
+--[[--
+Module for JPEG decoding/encoding.
+
+Currently, this is a LuaJIT FFI wrapper for turbojpeg lib.
```
Frenzie, line 4: "Currently" would just make me worry it's about to change. ;-)
```suggestion
This is a LuaJIT FFI wrapper for turbojpeg.
```

zwim: You are right, razor sharp.

ffi/blitbuffer.lua
```diff
@@ -1914,12 +1914,38 @@ function BB_mt.__index:writeBMP(filename)
...
+end
+
+function BB_mt.__index:writeToFile(filename, format, quality)
+    format = format or "jpg" -- set default format
+    format = format:lower()
+    if format == "png" then
+        self:writePNG(filename)
+    elseif format == "bmp" then
         self:writeBMP(filename)
     else -- default all other extensions to png
```
NiLuJe, line 1947: Code says jpg ;).

zwim: Yepp, comment is wrong.

ffi/jpeg.lua
```diff
@@ -0,0 +1,99 @@
...
+        image_bb = BB.new(width[0], height[0], BB.TYPE_BBRGB24)
+        components = 3
+        format = turbojpeg.TJPF_RGB
+    else
+        image_bb = BB.new(width[0], height[0], BB.TYPE_BB8)
+        components = 1
+        format = turbojpeg.TJPF_GRAY
+    end
+
+    if turbojpeg.tjDecompress2(handle, ffi.cast("unsigned char*", data), #data,
```
NiLuJe, line 57: This one takes a `const unsigned char*`. Speaking of, I'd possibly switch to `tjDecompressHeader3`, which *also* takes a const (as it should), to avoid the puzzling madness.

zwim: I have to look into what I should do with the colorspace tomorrow.

NiLuJe: Assuming you meant the subsampling constants, there's a `TJSAMP_GRAY` for the `TJPF_GRAY` pixel format ;).

ffi/jpeg.lua
```diff
@@ -0,0 +1,93 @@
...
+This is a LuaJIT FFI wrapper for turbojpeg.
+
+@module ffi.jpeg
+]]
+
+local BB = require("ffi/blitbuffer")
+local ffi = require("ffi")
+
+local _ = require("ffi/turbojpeg_h")
+local turbojpeg = ffi.load("turbojpeg")
```
NiLuJe, line 13: Well, apparently, the hoop-jumping was necessary, c.f., the testsuite:

```
[----------] Running tests from ./spec/base/unit/pic_spec.lua
./ffi/jpeg.lua:13: libturbojpeg.so: cannot open shared object file: No such file or directory
```

(ping @Frenzie/@poire-z).

ffi/jpeg.lua
```diff
@@ -0,0 +1,93 @@
...
+    turbojpeg.tjDestroy(handle)
+    return image_bb, width[0], height[0], components
+end
+
+function Jpeg.encodeToFile(filename, source_ptr, w, h, quality, color_type, subsample)
+    quality = quality or 75
+    color_type = color_type or turbojpeg.TJPF_RGB
+    subsample = subsample or turbojpeg.TJSAMP_420
+
+
```
NiLuJe, line 66: not: stray LF

ffi/jpeg.lua
Frenzie: It'd work in combination with front where cpath is set and ffi.load is replaced with a wrapper. But indeed this is base.

ffi/jpeg.lua
```diff
@@ -0,0 +1,93 @@
...
+    local jpeg_size = ffi.new("unsigned long int [1]")
+
+    local jpeg_image = ffi.new("unsigned char* [1]")
+    jpeg_image[0] = ffi.new("unsigned char*")
```
NiLuJe, line 70: That looks fishy.

First, no need for the second line, jpeg_image is a `unsigned char**`, as expected, which already makes `jpeg_image[0]` (i.e., *jpeg_image`) a `unsigned char*`.

Secondly, looking at the API doc, you have three choices:

* Pre-allocate it to an arbitrary size via `tjAlloc`, in which case `jpeg_size` should be set to the buffer size.
* Set *jpeg_image to NULL (e.g., `jpeg_image[0] = nil`), and let TJ realloc as-needed.
* Pre-allocate to worst-case size via `tjBufSize`, in which case `jpeg_size` should be set to the buffer size.

You appear to be going with option two, except that new looks fishy to me ^^.

ffi/jpeg.lua
```diff
@@ -0,0 +1,93 @@
...
+    if turbojpeg.tjCompress2(handle, source_ptr, w, 0, h, color_type,
+        jpeg_image, jpeg_size, subsample, quality, 0) == 0 then
+
+        local of, err = io.open(filename, "wb")
+        if err ~= nil then
+            return err
+        else
+            of:write(ffi.string(jpeg_image[0], jpeg_size[0]))
+            of:close()
+        end
```
NiLuJe, line 84: I'd be tempted to avoid the Lua string interning roundtrip and just do this in C (e.g. open -> write -> close; we should have all  of it in ffi/posix).

NiLuJe: Thinking about it, the same applies to the BMP codepath ;p.

(Speaking of, TJ has a tjSaveFile function that... does just that for BMPs ^^).

ffi/jpeg.lua
zwim: > First, no need for the second line, jpeg_image is an `unsigned char**`, as expected, which already makes `jpeg_image[0]` (i.e., `*jpeg_image`) an `unsigned char*`.

You are right, the second line is a leftover.

> You appear to be going with option two, except that new looks fishy to me ^^.

Yes, I let turbojpeg do the work.

ffi/jpeg.lua
NiLuJe: Aaah, that makes sense, thanks!

ffi/jpeg.lua
NiLuJe: (Remember that ffi.string, besides the Lua string interning, allocates a new GC-tracked buffer to store that. e.g., it's slow).

ffi/jpeg.lua
zwim: What does that mean? Can I leave `local turbojpeg = ffi.load("turbojpeg")`?

NiLuJe: No you can't, as it at the very least breaks the testsuite ;).

zwim:
I have a problem now: when I use tjSaveImage the image format is not bmp but ppm.
gwenview shows: `org.kde.kdegraphics.gwenview.lib: Image format is actually "ppm" not "bmp"`.

ffi/jpeg.lua
```diff
@@ -90,4 +87,8 @@ function Jpeg.encodeToFile(filename, source_ptr, w, h, quality, color_type, subs
     end
 end
 
-return Jpeg
\ No newline at end of file
+function Jpeg.writeBMP(filename, source_ptr, w, h)
+    turbojpeg.tjSaveImage(ffi.cast("const char *", filename), source_ptr, w, 0, h, turbojpeg.TJPF_BGR, 2)
```
NiLuJe, line 91: You *might* be able to pass filename directly and let LuaJIT sort it out (I *think* that works).

NiLuJe: Not *quite* sure about the pixelformat either, from the docs, I'm not sure it matters if it's not gray or cmyk, but, in any case, that's the format of the *input* buffer: and ours is RGB, not BGR ;).

ffi/turbojpeg_h.lua
```diff
@@ -49,4 +49,8 @@ int tjCompress2(tjhandle, const unsigned char *, int, int, int, int, unsigned ch
 int tjCompress2(tjhandle handle, const unsigned char *srcBuf, int width, int pitch, int height,
                 int pixelFormat, unsigned char **jpegBuf, unsigned long *jpegSize, int jpegSubsamp,
                 int jpegQual, int flags);
+
+int tjSaveImage(const char *filename, unsigned char *buffer, int width, int pitch, int height,
```
NiLuJe, line 53: Needs matching updates in `ffi-cdecl/turbojpeg_decl.c` ;).

ffi/jpeg.lua
```diff
@@ -63,24 +65,19 @@ function Jpeg.encodeToFile(filename, source_ptr, w, h, quality, color_type, subs
...
     if turbojpeg.tjCompress2(handle, source_ptr, w, 0, h, color_type,
         jpeg_image, jpeg_size, subsample, quality, 0) == 0 then
 
-        local of, err = io.open(filename, "wb")
-        if err ~= nil then
-            return err
-        else
-            of:write(ffi.string(jpeg_image[0], jpeg_size[0]))
-            of:close()
+        local fhandle = C.open(ffi.cast("char *", filename), C.O_RDWR)
```
NiLuJe, line 77: We might be missing a few flags to make this saner, `O_CREAT`, `O_APPEND`, `O_TRUNC`.

NiLuJe: i.e.,

```diff
diff --git a/ffi-cdecl/posix_decl.c b/ffi-cdecl/posix_decl.c
index ea3ee18d..0c3cb307 100644
--- a/ffi-cdecl/posix_decl.c
+++ b/ffi-cdecl/posix_decl.c
@@ -27,6 +27,9 @@ cdecl_func(fork)
 cdecl_func(dup)
 cdecl_func(dup2)
 
+cdecl_const(O_APPEND)
+cdecl_const(O_CREAT)
+cdecl_const(O_TRUNC)
 cdecl_const(O_RDWR)
 cdecl_const(O_RDONLY)
 cdecl_const(O_NONBLOCK)
diff --git a/ffi/posix_h.lua b/ffi/posix_h.lua
index 999ce719..897c871a 100644
--- a/ffi/posix_h.lua
+++ b/ffi/posix_h.lua
@@ -16,6 +16,9 @@ int pipe(int *) __attribute__((nothrow, leaf));
 int fork(void) __attribute__((nothrow));
 int dup(int) __attribute__((nothrow, leaf));
 int dup2(int, int) __attribute__((nothrow, leaf));
+static const int O_APPEND = 1024;
+static const int O_CREAT = 64;
+static const int O_TRUNC = 512;
 static const int O_RDWR = 2;
 static const int O_RDONLY = 0;
 static const int O_NONBLOCK = 2048;
```

And use `O_RDWR | O_CREAT | O_TRUNC` here.

ffi/jpeg.lua
zwim: > You _might_ be able to pass filename directly and let LuaJIT sort it out (I _think_ that works).

Yes, that works.

ffi/jpeg.lua
NiLuJe: (Updated diff above).

ffi/jpeg.lua
zwim: > and ours is RGB, not BGR ;).

Right, this is a leftover. The fileformat stays netbmp.

zwim:
> I have a problem now: when I use tjSaveImage the image format is not bmp but ppm.
> gwenview shows: `org.kde.kdegraphics.gwenview.lib: Image format is actually "ppm" not "bmp"`.

I have got it. If the filename ends with `.bmp` tjSaveImage creates a bmp. But if the filename does not end with bmp it creates a netpbm/ppm file.
Also if the file ends with `.` a netpbm file is created.

ffi/jpeg.lua
zwim: Don't we want to use "+" instead of "|" in this case? The code would be more readable.

zwim:
JPG, PNG and BMP file creation works now. PR has to be applied together witch #1243

ffi/jpeg.lua
NiLuJe: The idiomatic C syntax is an OR, which, granted, mostly looks like an ADD in this context, but I'd rather not deviate too much from that ;).

ffi/jpeg.lua
```diff
@@ -0,0 +1,110 @@
...
+    local jpeg_size = ffi.new("unsigned long int [1]")
+    local jpeg_image = ffi.new("unsigned char* [1]")
+
+    local handle = turbojpeg.tjInitCompress()
+    assert(handle, "no TurboJPEG API compressor handle")
+
+    if turbojpeg.tjCompress2(handle, source_ptr, w, 0, h, color_type,
+        jpeg_image, jpeg_size, subsample, quality, 0) == 0 then
+
+        local fhandle = C.open(filename, C.O_WRONLY + C.O_CREAT + C.O_TRUNC, C.I_IRUSR + C.I_IWUSR)
```
NiLuJe, line 85: Is there a specific issue with the default umask that requires passing the mode?

zwim: Yes, if no modes are given the file permissions are zero.

NiLuJe: That's... weird, but, okay ;).

ffi/jpeg.lua
```diff
@@ -0,0 +1,110 @@
...
+    turbojpeg.tjDestroy(handle)
+    if jpeg_image[0] ~= nil then
+        turbojpeg.tjFree(jpeg_image[0])
+    end
+end
+
+function Jpeg.writeBMP(filename, source_ptr, w, h)
+    -- if file extension is not ".bmp" tjSaveImage uses netpbm format!
+    if filename:sub(-#".bmp") == ".bmp" then
+        turbojpeg.tjSaveImage(filename .. ".bmp", source_ptr, w, 0, h, turbojpeg.TJPF_RGB, 0)
```
NiLuJe, line 101: I'd pass bb.stride to pitch, it'll ensure this works when used with a Screen bb as input ;).

ffi/jpeg.lua
```diff
@@ -0,0 +1,110 @@
...
+    color_type = color_type or turbojpeg.TJPF_RGB
+    subsample = subsample or turbojpeg.TJSAMP_420
+
+    local jpeg_size = ffi.new("unsigned long int [1]")
+    local jpeg_image = ffi.new("unsigned char* [1]")
+
+    local handle = turbojpeg.tjInitCompress()
+    assert(handle, "no TurboJPEG API compressor handle")
+
+    if turbojpeg.tjCompress2(handle, source_ptr, w, 0, h, color_type,
```
NiLuJe, line 82: Ditto about stride.

ffi/jpeg.lua
```diff
@@ -0,0 +1,110 @@
...
+    local jpeg_image = ffi.new("unsigned char* [1]")
+
+    local handle = turbojpeg.tjInitCompress()
+    assert(handle, "no TurboJPEG API compressor handle")
+
+    if turbojpeg.tjCompress2(handle, source_ptr, w, stride, h, color_type,
+        jpeg_image, jpeg_size, subsample, quality, 0) == 0 then
+
+        local fhandle = C.open(filename, bit.bor(C.O_WRONLY, C.O_CREAT, C.O_TRUNC), bit.bor(C.I_IRUSR, C.I_IWUSR))
+        if fhandle > 0 then
```
NiLuJe, line 86: `!= -1`. WHile it's highly unlikely we'd ever get fd `0`, it's entirely plausible as a successful return value from open ;).

NiLuJe (approved):
LGTM, thanks ;).

zwim:
Thank you, for your support and help.

ffi/jpeg.lua
```diff
@@ -0,0 +1,110 @@
...
+    turbojpeg = ffi.load("libs/libturbojpeg.dylib")
+else
+    turbojpeg = ffi.load("libs/libturbojpeg.so")
+end
+
+local Jpeg = {}
+
+function Jpeg.openDocument(filename, color)
+    local fh = io.open(filename, "rb")
+    assert(fh, "couldn't open JPG file")
```
Frenzie, line 29: These asserts are guarded with a pcall in front?

zwim:
Hopefully, this is the last commit. #1245 is necessary for more file modes.
When `jpg` file is _not present,_ `writeJPG` creates one. The mode argument has to be casted to `int` because of the ellipsis in `C.open(const char*, int, ...) `. If this is not done (or wrong mode bits are set), then the newly created file has zero permissions.

When the exists, the mode bits  are not important, as the old file permissions are kept.

zwim:
@Frenzie: Now there is a `pcal` protection in `writeToFile`.

`openDocument` should be protected.
`openDocumentFromMem` should only called from `openDocument` (at least I didn't find any other occurrences in the code. Could you double check this?)

Frenzie (approved):
I think that should be fine.

PR #1242, opened 2020-11-29
what happens with homebrew?

pazos:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1242)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,10 +11,8 @@ jobs:
         os: [macos-latest]
 
     steps:
-      - name: setup-xcode
-        uses: maxim-lobanov/setup-xcode@v1
-        with:
-          xcode-version: 11.3.1 # Use Xcode 11.3.1, because 11.4, 11.5, and 11.6 have a segfaulting clang.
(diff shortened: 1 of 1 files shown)


NiLuJe:
That's a good question. It's actually utterly borked on my laptop, and I can't be arsed to fix it there, so I'm a tad out of the loop on that front ^^.

pazos:
@Frenzie: it seems that last xcode (12.0) builds fine **today**. I wonder what will happen tomorrow 😄

Frenzie:
@pazos What do they keep doing over there? O_o I had previously naively thought that most breakage was due to us/me not testing on macOS, but ever since I set up this GitHub action it's broken itself… 3 or 4 times by now without any change on our side. :-/

pazos:
No idea.

The brew error got fixed automagically so the change in brew deps order is no longer relevant
The issue with leptonica build and the C compiler cannot generate executables got fixed automagically ¿?

The only issue left which breaks the build in our end happens with "otooling" libkoreader-nnsvg.so. Once that's fixed everything else should work :p

PR #1110, opened 2020-06-04
Added htmlparser Luarock for newsdownloader

lich-tex:
My setup does not allow me to check how does it integrate with the complete build (Raspberry Pi). So, would be thankful if someone could verify if it works and then is used by [the patched newsdownloader](https://github.com/koreader/koreader/pull/6228).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1110)

PR #1244, opened 2020-12-01
LuaRocks 3: use lua5.1 instead of the system default

yparitcher:
this fixes on a default archlinux install.

Can someone test with Luarocks 2 to make sure it does not break?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1244)


NiLuJe:
The buildbots run on LR 2, IIRC, so that should answer that question ;).

yparitcher:
in the first commit (which failed) I don't think i changed anything in LR2 but i can't access CircleCi, so i have no way of knowing why :(

yparitcher:
They failed because they were cancelled ?

poire-z:
Yes, first commit fails are because cancelled.
Second commit just has (for now) the macos test failing, but that's usual. All other arch builds are still running.

yparitcher:
my second commit will only affect a rebuild so the CI won't help can someone with LR2 let me know the htmlparser rockspec name.
is the rockspec in `base/build/x86_64-pc-linux-gnu-debug/rocks/lib/luarocks/rocks-5.1/htmlparser` or in the subfolder `scm-0`?

in LR3 it is `base/build/x86_64-pc-linux-gnu-debug/rocks/lib/luarocks/rocks-5.1/htmlparser/scm-0/htmlparser-scm-0.rockspec`

yparitcher:
Confirmed LR2 uses the same path from looking in the github release (i think these are still built with LR2).
these were probably CP errors in #1110 & #1113

Frenzie:
What I just wrote [here](https://github.com/koreader/koreader/pull/6935#commitcomment-44725284) might also be relevant.

> I didn't mean I can't toss Bash 4.4+ [LuaRocks 3] on the Docker image or alternatively perhaps the time has arrived to consider upgrading to 18.04. It's fully under our/my control. ;-) I didn't realize readarray was a recent invention.

That being said, it'd be better to build our own LR as part of the build process in base in that case (actually cf. what the Docker image does [here](https://github.com/koreader/virdevenv/blob/aec9d7e55e3f4f2980e2739005970ff3134d36e6/docker/ubuntu/baseimage/install_luarocks.sh)), since I'd definitely prefer if users of current Debian/Ubuntu LTS releases could just start hacking out of the box.

yparitcher:
LR2 works currently, this pr will allow a default LR3 to work also, they don't conflict so there should be no need to build LR to allow default distros to work.

Frenzie:
Excellent. :+1:

NiLuJe:
Works for me on Gentoo w/ LuaRocks 3.4.0 (defaulting to LuaJIT 2.1) ;).

PR #1250, opened 2020-12-04
Adds libkoreader-nnsvg.so, Lua/C module wrapping NanoSVG

poire-z:
Will allow rendering SVG images in frontend. See https://github.com/koreader/koreader/pull/6937 https://github.com/koreader/koreader/pull/6937#issuecomment-738347035

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1250)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -194,7 +194,7 @@ $(OUTPUT_DIR)/libs/libkoreader-nnsvg.so: nnsvg.c \
 			$(if $(USE_LUAJIT_LIB),$(LUAJIT_LIB),) \
 			$(NANOSVG_HEADERS)
 	$(CC) -I$(NANOSVG_INCLUDE_DIR) \
-	$(CFLAGS) $(DYNLIB_CFLAGS) -Wall -o $@ nnsvg.c
+	$(CFLAGS) $(DYNLIB_CFLAGS) -fvisibility=hidden -Wall -o $@ nnsvg.c
 ifdef DARWIN
 	install_name_tool -change \
 		`otool -L "$@" | grep "libluajit" | awk '{print $$1}'` \
--- a/Makefile
+++ b/Makefile
@@ -190,11 +190,11 @@ ifdef DARWIN
 		$@
 endif
 
-$(OUTPUT_DIR)/libs/libkoreader-nnsvg.so: nnsvg.cpp \
+$(OUTPUT_DIR)/libs/libkoreader-nnsvg.so: nnsvg.c \
 			$(if $(USE_LUAJIT_LIB),$(LUAJIT_LIB),) \
 			$(NANOSVG_HEADERS)
-	$(CXX) -I$(NANOSVG_INCLUDE_DIR) \
-	$(DYNLIB_CXXFLAGS) -static-libstdc++ -Wall -o $@ nnsvg.cpp
+	$(CC) -I$(NANOSVG_INCLUDE_DIR) \
+	$(CFLAGS) $(DYNLIB_CFLAGS) -Wall -o $@ nnsvg.c
 ifdef DARWIN
 	install_name_tool -change \
 		`otool -L "$@" | grep "libluajit" | awk '{print $$1}'` \
--- a/Makefile
+++ b/Makefile
@@ -102,6 +102,7 @@ libs: \
 	$(OUTPUT_DIR)/libs/libkoreader-djvu.so \
 	$(OUTPUT_DIR)/libs/libkoreader-cre.so \
 	$(OUTPUT_DIR)/libs/libkoreader-xtext.so \
+	$(OUTPUT_DIR)/libs/libkoreader-nnsvg.so \
 	$(OUTPUT_DIR)/libs/libwrap-mupdf.so
 
 $(OUTPUT_DIR)/libs/libinkview-compat.so: input/inkview-compat.c
@@ -186,6 +187,18 @@ ifdef DARWIN
 	install_name_tool -change \
 		`otool -L "$@" | grep "$(notdir $(FRIBIDI_LIB)) " | awk '{print $$1}'` \
 		libs/$(notdir $(FRIBIDI_LIB)) \
+		$@
+endif
+
+$(OUTPUT_DIR)/libs/libkoreader-nnsvg.so: nnsvg.cpp \
+			$(if $(USE_LUAJIT_LIB),$(LUAJIT_LIB),) \
+			$(NANOSVG_HEADERS)
+	$(CXX) -I$(NANOSVG_INCLUDE_DIR) \
+	$(DYNLIB_CXXFLAGS) -static-libstdc++ -Wall -o $@ nnsvg.cpp
+ifdef DARWIN
+	install_name_tool -change \
+		`otool -L "$@" | grep "libluajit" | awk '{print $$1}'` \
+		libs/$(notdir $(LUAJIT_LIB)) \
 		$@
 endif
 
--- a/nnsvg.c
+++ b/nnsvg.c
@@ -106,6 +106,12 @@ static int nnSVGImage_drawTo(lua_State *L) {
 	luaL_typerror(L, -1, "BlitBuffer");
     }
     BlitBuffer * bb = (BlitBuffer*) lua_topointer(L, 2);
+    /* TODO: check bb type is TYPE_BBRGB32
+    if ( GET_BB_TYPE(bb) != TYPE_BBRGB32 ) {
+        lua_pushstring(L, "BlitBuffer BBRGB32 expected.");
+        lua_error(L);
+    }
+    */
 
     // nanosvg's rasterizer won't automatically scale the image
     // to adjust it to the target buffer w/h.
@@ -124,7 +130,7 @@ static int nnSVGImage_drawTo(lua_State *L) {
         lua_pushstring(L, "Could not init rasterizer.");
         lua_error(L);
     }
-    nsvgRasterize(rast, image, offset_x, offset_y, scale_factor, bb->data, bb->w, bb->h, 4*bb->w);
+    nsvgRasterize(rast, image, offset_x, offset_y, scale_factor, bb->data, bb->w, bb->h, bb->stride);
     nsvgDeleteRasterizer(rast);
     return 0;
 }
--- a/nnsvg.c
+++ b/nnsvg.c
@@ -1,16 +1,13 @@
 // nsvg.cpp
 // Lua interface to wrap nanosvg.
 
-extern "C"
-{
 #include <lua.h>
 #include <lauxlib.h>
 #include <lualib.h>
-#include "blitbuffer.h"
-#include "nnsvg.h"
-}
 
-#include <math.h>
+#include "nnsvg.h"
+#include "blitbuffer.h"
+
 #define NANOSVG_ALL_COLOR_KEYWORDS
 #define NANOSVG_IMPLEMENTATION
 #define NANOSVGRAST_IMPLEMENTATION
@@ -35,7 +32,7 @@ static int nnsvg_new(lua_State *L) {
     input = luaL_checklstring(L, 1, &len);
 
     // Second arg specified what the string is: by default, a filepath
-    bool is_svg_data = false;
+    int is_svg_data = 0;
     if ( lua_isboolean(L, 2) ) {
         is_svg_data = lua_toboolean(L, 2);
     }
@@ -80,7 +77,7 @@ NSVGimage * check_NSVGimage(lua_State * L, int n) {
 // the NSVGimage early) and as :__gc() (called by the async Lua GC).
 static int nnSVGImage_free(lua_State *L) {
     NSVGimage ** udata = (NSVGimage **)luaL_checkudata(L, 1, NNSVG_METATABLE_NAME);
-    printf("nnSVGImage_free %p\n", *udata);
+    // printf("nnSVGImage_free %p\n", *udata);
     if ( *udata )
         nsvgDelete(*udata);
     *udata = NULL;
--- a/nnsvg.c
+++ b/nnsvg.c
@@ -5,6 +5,8 @@
 #include <lauxlib.h>
 #include <lualib.h>
 
+#include <stdbool.h>
+
 #include "nnsvg.h"
 #include "blitbuffer.h"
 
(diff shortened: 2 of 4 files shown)


NiLuJe:
Why C++? Unlike xtext, this time I don't think there's any actual C++ involved here ;p.

poire-z:
Because ... it's ++, it's better ?! Why use something -- ?! :)
See, C-- has no `bool` ! loser!
```
nnsvg.c:36:5: error: unknown type name 'bool'; did you mean '_Bool'?
     bool is_svg_data = false;
     ^~~~
     _Bool
nnsvg.c:36:24: error: 'false' undeclared (first use in this function); did you mean 'fabsl'?
     bool is_svg_data = false;

```

pazos:
@poire-z: you can `#include <stdbool.h>` in a c99 fashion

NiLuJe:
`#include <stdbool.h>` ;p.

pazos:
I won by 14 seconds! :1st_place_medal:

NiLuJe:
I'm not familiar at all with the symbol visibility requirements for Lua module, but, assuming the only visible module ought to be the one in the header, if you just prefix it with `__attribute__((visibility("default")))` (e.g., __attribute__((visibility("default"))) int luaopen_nnsvg(lua_State *L);), and build the thing with `-fvisibility=hidden`, this will ensure we don't leak NanoSVG's own symbols in there ;).

poire-z:
> you can #include <stdbool.h> in a c99 fashion

I just had one `bool` and one `false` to remove - so I can do it the 1980s way :)

Is it normal that on arm (nor on the emulator), I get:
arm-linux-gnueabihf-gcc **-static-libstdc++** (that ++ thing added) ?

NiLuJe:
You definitely ought to remove that, too, once you switch to C ;p

(It's hardcoded in the Makefile, right now, IIRC).

NiLuJe:
Speaking of Makefile, it's also missing a `$(CFLAGS)` (they're not included in the DYNLIB_* stuff).

NiLuJe:
> Is it normal that on arm (nor on the emulator), I get:
> arm-linux-gnueabihf-gcc **-static-libstdc++** (that ++ thing added) ?

Oh, we hardcode it everywhere, even outside of C++, because everything is terrible, don't mind it ;).

nnsvg.c
```diff
@@ -35,7 +32,7 @@ static int nnsvg_new(lua_State *L) {
     input = luaL_checklstring(L, 1, &len);
 
     // Second arg specified what the string is: by default, a filepath
-    bool is_svg_data = false;
+    int is_svg_data = 0;
```
NiLuJe, line 35: I'd really rather we not perpetuate a history of using terrible data types, so, C99 bool, please ;p.

NiLuJe: (We require C99/C++11, so, not a compat issue).

poire-z:
> if you just prefix it with __attribute__((visibility("default"))) (e.g., __attribute__((visibility("default"))) int luaopen_nnsvg(lua_State *L);), and build the thing with -fvisibility=hidden, this will ensure we don't leak NanoSVG's own symbols in there ;).

Looks like it changes something:
`   219: 000000000000ce02   133 FUNC    GLOBAL DEFAULT   12 nsvgParse`
becomes:
`   196: 000000000000bd81   133 FUNC    LOCAL  DEFAULT   12 nsvgParse`

Among the FUNC GLOBAL, we get only these 2 kinds of stuff:
```
    51: 000000000000c5ce   132 FUNC    GLOBAL DEFAULT   12 luaopen_nnsvg
   210: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND free@@GLIBC_2.2.5
   212: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND strncpy@@GLIBC_2.2.5
   216: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND qsort@@GLIBC_2.2.5
   217: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND fread@@GLIBC_2.2.5
   218: 000000000000c5ce   132 FUNC    GLOBAL DEFAULT   12 luaopen_nnsvg
   219: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND fclose@@GLIBC_2.2.5
```
So, ok, adding that.

NiLuJe:
Yep, that's the idea ;).

(I'll do a pass on our other Lua modules later ;)).

nnsvg.c
```diff
@@ -0,0 +1,169 @@
...
+    const float scale_factor = (float)luaL_optnumber(L, 3, 1.0);
+    const int offset_x = (int)luaL_optint(L, 4, 0);
+    const int offset_y = (int)luaL_optint(L, 5, 0);
+
+    NSVGrasterizer *rast = nsvgCreateRasterizer();
+    if (rast == NULL) {
+        lua_pushstring(L, "Could not init rasterizer.");
+        lua_error(L);
+    }
+    nsvgRasterize(rast, image, offset_x, offset_y, scale_factor, bb->data, bb->w, bb->h, 4*bb->w);
```
NiLuJe, line 127: The final argument ought to be `bb->stride`, with an earlier check to make sure sure that the dest BB type is BBRGB32 ;).

poire-z: Is there a proper way to check for BBRGB32 ?
`#define TYPE_BBRGB32 5` is in blitbuffer.c, not in blitbuffer.h.
Should I just check for `5`, or move the #define in blitbuffer.h (and which, only the TYPE_BB*?)

poire-z: Oh, but there's no type in the structs:) So, what is the proper way to be sure it's BBRGB32?

poire-z:
```c
#define MASK_TYPE 0xF0
#define SHIFT_TYPE 4
#define TYPE_BB4 0
#define TYPE_BB8 1
#define TYPE_BB8A 2
#define TYPE_BBRGB16 3
#define TYPE_BBRGB24 4
#define TYPE_BBRGB32 5
#define GET_BB_TYPE(bb) (((MASK_TYPE & bb->config) >> SHIFT_TYPE))
```

NiLuJe: You're including `"blitbuffer.h"` already, right?

You should be able to do `if (GET_BB_TYPE(bb) == TYPE_BBRGB32)` ;).

NiLuJe: Oh, yeah, right, it's in the C file, not the header.

Well, that's a mistake :D.

NiLuJe: So, just use stride for now, I'll take care of the rest later ;).

poire-z:
Feel free to merge if it's ok and you want to add the final touches - and/or bring it with if you bump some stuff.

NiLuJe:
Stuff is not *quite* entirely ready on my end, so, yeah, this is good to go as far as I'm concerned ;).

NiLuJe (approved):

poire-z:
This fails to load on Android :( 
`dlopen failed: cannot locate symbol "pow" referenced by "/data/.../libs/likoreader-nnsvg.so"`

We met that kind of error in https://github.com/koreader/koreader/issues/3493 and https://github.com/koreader/koreader/issues/4469#issuecomment-453056533

Makefile
```diff
@@ -189,6 +190,18 @@ ifdef DARWIN
 		$@
 endif
 
+$(OUTPUT_DIR)/libs/libkoreader-nnsvg.so: nnsvg.c \
+			$(if $(USE_LUAJIT_LIB),$(LUAJIT_LIB),) \
+			$(NANOSVG_HEADERS)
+	$(CC) -I$(NANOSVG_INCLUDE_DIR) \
+	$(CFLAGS) $(DYNLIB_CFLAGS) -fvisibility=hidden -Wall -o $@ nnsvg.c
```
Frenzie, line 197: @poire-z Wrt https://github.com/koreader/koreader-base/pull/1250#issuecomment-739471075 what if you stick a `-lm` in there?

poire-z: I don't have (not want :/) any android build tools. So, if you do, and you get the same crash as on phones, what if ? :)

poire-z:
Any thoughts on https://github.com/koreader/koreader-base/pull/1252#issuecomment-739413550 ?

Makefile
Frenzie: I spent the past 15-20 minutes building for Android as I haven't since… oh, October or so when I didn't have any free time at all due to moving, so now I'll be able to do some quick tests. 

I only have to comment out these lines to test?
```lua
-- Uncomment to use SVG icons from one of these directories
-- ICONS_ALT_SVG_DIR = "resources/icons/src/"
-- ICONS_ALT_SVG_DIR = "resources/icons/svg/"
```

poire-z: Yes, the /src/ one where we already have a few SVGs (it will look ugly, some will be unparsable, black...).

For testing https://github.com/koreader/koreader/pull/6937, you can just download the zip from https://github.com/johnbeard/koreader/tree/icons , and extract the svg from its resources/icons/src/ into that directory (I put them into "resources/icons/svg/", so it's clean to update).

Frenzie:
`USE_LUAJIT_LIB` is for Android (and Mac OS as well as Windows)

You need it as a dependency to link against, but as written that strikes me as meaningless at best, unless you stick it in LDFLAGS or some such.

Makefile
Frenzie: Yup, -lm gets around that error. But it looks like we do indeed need to link luajit too:

```
12-06 11:41:46.724  4000  4030 E KOReader: 	dlopen failed: cannot locate symbol "luaL_checklstring" referenced by "/data/data/org.koreader.launcher/files/libs/libkoreader-nnsvg.so"...
```

Frenzie: I.e., works like this, but the -lluajit should be behind a `$(if $(USE_LUAJIT_LIB),$(LUAJIT_LIB),)` type check.
```diff
diff --git a/Makefile b/Makefile
index b49b784e..751c8f71 100644
--- a/Makefile
+++ b/Makefile
@@ -194,7 +194,7 @@ $(OUTPUT_DIR)/libs/libkoreader-nnsvg.so: nnsvg.c \
                        $(if $(USE_LUAJIT_LIB),$(LUAJIT_LIB),) \
                        $(NANOSVG_HEADERS)
        $(CC) -I$(NANOSVG_INCLUDE_DIR) \
-       $(CFLAGS) $(DYNLIB_CFLAGS) -fvisibility=hidden -Wall -o $@ nnsvg.c
+       $(CFLAGS) $(DYNLIB_CFLAGS) -lm -lluajit -fvisibility=hidden -Wall -o $@ nnsvg.c
 ifdef DARWIN
        install_name_tool -change \
                `otool -L "$@" | grep "libluajit" | awk '{print $$1}'` \
```

Frenzie: Screenshot_2020-12-06_11-48-17

NiLuJe:
That's the reasoning for `$^` in the other recipes: they feed the recipe dependencies, which include the canoncal paths to the .so/.dylib/.dll depedencies to the compiler, which translates to it linking against it.

(This is similar to what kpvcrlib does to build CRe, if you're more familiar with that).

It's... not pretty, and not how you're really supposed to do that (which is simply `-L$DESTDIR)/lib -llib`, but the buildsystem uses this trick extensively.

NiLuJe:
(And, yeah, -lm is a dependency of nanosvg, I should have remembered that, my bad).

poire-z:
So, how should this be done ? With this: ?
```diff
--- a/Makefile
+++ b/Makefile
@@ -194,7 +194,7 @@ $(OUTPUT_DIR)/libs/libkoreader-nnsvg.so: nnsvg.c \
                        $(if $(USE_LUAJIT_LIB),$(LUAJIT_LIB),) \
                        $(NANOSVG_HEADERS)
        $(CC) -I$(NANOSVG_INCLUDE_DIR) \
-       $(CFLAGS) $(DYNLIB_CFLAGS) -fvisibility=hidden -Wall -o $@ nnsvg.c
+       $(CFLAGS) $(DYNLIB_CFLAGS) -lm $(if $(USE_LUAJIT_LIB),-lluajit,) -fvisibility=hidden -Wall -o $@ nnsvg.c
 ifdef DARWIN
        install_name_tool -change \
                `otool -L "$@" | grep "libluajit" | awk '{print $$1}'` \
```

Simulating using it with `$(if $(USE_LUAJIT_LIB),-lluajit,-lluajit)` causes the following. Is this normal when !USE_LUAJIT_LIB ?
```
gcc -I/koreader/base/thirdparty/nanosvg/build//include \
-Og -g -pipe -fno-omit-frame-pointer   -march=native -fPIC -L/koreader/base/build/x86_64-linux-gnu-debug/libs -Og -g -pipe -fno-omit-frame-pointer   -march=native -fPIC -I/koreader/base/thirdparty/luajit/build/x86_64-linux-gnu-debug/luajit-prefix/src/luajit/src -shared -Wl,-E -Wl,-rpath,'$ORIGIN' -lm -lluajit -fvisibility=hidden -Wall -o build/x86_64-linux-gnu-debug/libs/libkoreader-nnsvg.so nnsvg.c
/usr/bin/ld: cannot find -lluajit
collect2: error: ld returned 1 exit status
make[1]: *** [Makefile:196: build/x86_64-linux-gnu-debug/libs/libkoreader-nnsvg.so] Error 1
```

(I remember I removed a `#include <math.h>` when picking stuff from lvimg.cpp, because it compiled fine without :) If I had let it in, would it have pick -lm by itself?)

NiLuJe:
> Simulating using it with `$(if $(USE_LUAJIT_LIB),-lluajit,-lluajit)` causes the following. Is this normal when !USE_LUAJIT_LIB ?

Yes, there's no dynamic library built *at all* for LuaJIT on platforms where `!USE_LUAJIT_LIB` ;).

> (I remember I removed a `#include <math.h>` when picking stuff from lvimg.cpp, because it compiled fine without :) If I had let it in, would it have pick -lm by itself?)

No, it wouldn't have ;).

(That was nonetheless the right thing to do, you're not using libm, nanosvg is, and nanosvg/nanosvgraster.h should be the ones taking care of that).

NiLuJe:
> So, how should this be done ? With this: ?

That looks good, yeah.

(Incidentally, when I told you DYNLIB_CFLAGS didn't include CFLAGS, I was wrong, so the CFLAGS are actually redundant here, my bad ;)).

NiLuJe:
> > So, how should this be done ? With this: ?
> 
> That looks good, yeah.

Actually, maybe not, as I don't know *where* the LuaJIT lib ends up, and if it's even in the searchpath, so, you can simply duplicate the earlier `$(if $(USE_LUAJIT_LIB),$(LUAJIT_LIB),)` ;).

PR #1257, opened 2020-12-11
Another great bundle of messy stuff

NiLuJe:
Companion to https://github.com/koreader/crengine/pull/399

Much like over there, I intend to somewhat cleanup the history before merge.

----

* CBB:
  * Reformat in a C99 fashion (tighter scope, restricted pointers).
     (This makes the diff a horrible mess to read, so, don't, and hope there aren't any more typos than those I already caught ;D).
  * Made some helper macros available in the header (e.g., GET_BB_*).
  * Very few logic changes, but a few which'll probably get back to me later :D.
  * Update the BlitBuffer structs to use unsigned types, and made the fact that stride is a byte count explicit by using a size_t.
  * Add support for straight-alpha blitFrom (useful for NanoSVG, which spits out straight alpha, unlike MµPDF).
  * Don't clobber the alpha byte in paintRect & fill (c.f., the tail end of https://github.com/koreader/koreader/pull/6937)

* LBB:
  * Minor updates for feature-parity w/ CBB.
  * Fair warning about the stride is now size_t thing: a Lua number isn't precise enough to fit the full range of an uint64_t, so, on platforms where size_t is 64bit, it's *not* automatically coerced to a Lua number. The frontend companion PR takes care of the very few cases w<here this matters.
  * Fix a few instances where the invert flag was mishandled.

* Add ZSTD and an FFI binding.
  * ZSTD is now used by CRe to compress its cache
  * And it's also used by BookInfo to compress the thumbnails
  * BookInfo's now unused xutil has been moved to ffi/zlib

* Rejig our understanding and usage of the `__gc` metamethod: it doesn't run on plain tables in Lua 5.1, only on userdata. In LuaJIT, it's also usable on cdata metatypes, and cdata finalizers can also be setup explicitly.
Convert stuff to the sanest solution in each case, and, in cases where it's not possible, use the hacky workaround of introducing an empty `newproxy` userdata in the mix, via ffi/__gc.
This finally solves a number of puzzling code comments that were wondering why the hell our __gc methods were never called (hint: because it's not supported, or, in a few cases it's not supported and they were misnamed anyway ^^).

* Tweak lj-sqlite3 to stop coercing BLOBs as Lua strings. Instead return a raw pointer and a byte count, like the SQLite C API. This allows avoiding a few useless copies in BookInfo, our sole user of BLOB. It also just plain makes sense, because if I wanted strings, I'd be using a TEXT column ;).

* Use proper symbol visibility discipline for our Lua/C modules, much like was done in nnsvg. This avoids leaking a ton of useless symbols, and makes the dynamic loader's job that much easier.

* Update the stbiw version shipped by NanoSVG to the latest one (only used by a CRe helper function).

* Fun Valgrind experiments in cre.cpp if DEBUG_CRENGINE is set ;p.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1257)


poire-z:
Does that mean KOReader doesn't leak memory anymore ?! Well done ! :)

NiLuJe:
In practiced, it showed that most of our manual cleanup did the job mostly correctly ;).

(Hence the extra-shitty double-free guards in KOptContext, for instance ^^).

As far a BB is concerned, I haven't really caught much that we missed, maybe a few tiny BBs during startup.

NiLuJe:
I'm still utterly puzzled where roughly 8MB of memory go each time a CRe document is opened, but the PDF stuff seem pretty solid, which clearly visible chunks of memory being reclaimed at runtime in a single document.

DjVu appears to be a fairly bad student a mess, and will eat a metric shitton of RAM each time you open a *new* page.

poire-z:
> I'm still utterly puzzled by where roughly 8MB of memory go each time a CRe document is opened 

I haven't been seeing that (looking at the footer Memory info).
I can see it increase sometimes (on load, as you browse pages), but I sometimes also see decreases.
Switching now between 2 small Wikipedia EPUBs, I see 29M > 30 > 29 > 30 > 29 > 30.

What I see is it raises a log when you do a dict lookup, show history,... and then stay there.
(Some previous comments of mine about it at https://github.com/koreader/koreader/issues/5998#issuecomment-602180998)

NiLuJe:
> > I'm still utterly puzzled by where roughly 8MB of memory go each time a CRe document is opened
> 
> I haven't been seeing that (looking at the footer Memory info).
> I can see it increase sometimes (on load, as you browse pages), but I sometimes also see decreases.
> Switching now between 2 small Wikipedia EPUBs, I see 29M > 30 > 29 > 30 > 29 > 30.
> 
> What I see is it raises a log when you do a dict lookup, show history,... and then stay there.
> (Some previous comments of mine about it at [koreader/koreader#5998 (comment)](https://github.com/koreader/koreader/issues/5998#issuecomment-602180998))

Haven't tried a CRe->CRe swap, so this may not actually all be at CRe's feet, but doing a CRe -> FM -> CRe nets me a ~4 to 5MB increase each time on Kobo.

NiLuJe:
(A CRe -> CRe swap doesn't make a lick of difference over here, both with or without my mess applied ^^).

poire-z:
> A CRe -> CRe swap doesn't make a lick of difference

A difference vs what ? :) You mean you get the same memory increase as CRe>FM>CRe ? Or that you get the same memory usage (no increase) ?

Frenzie (commented):
note to self, read through blitbuffer and zstd.lua later

thirdparty/kpvcrlib/CMakeLists.txt
```diff
@@ -20,15 +20,17 @@ set(ANTIWORD_INCLUDE_DIR ${CR_3RDPARTY_DIR}/antiword)
 set(CHM_INCLUDE_DIRS ${CR_3RDPARTY_DIR}/chmlib/src)
 set(CREGINE_INCLUDE_DIR ${CRE_DIR}/include)
 
+assert_var_defined(JCONFIG_INCLUDE_DIR)
```
poire-z, line 23: Any reason for you to move this at top ? Looks a bit related to JPEGLIB - or may be not, dunno :)

poire-z (approved):
Looks fine, no real comment.

xtext.cpp
```diff
@@ -360,7 +360,7 @@ class XText {
         if (m_bidi_ctypes) { free(m_bidi_ctypes); m_bidi_ctypes = NULL; }
         if (m_bidi_btypes) { free(m_bidi_btypes); m_bidi_btypes = NULL; }
         if (m_bidi_levels) { free(m_bidi_levels); m_bidi_levels = NULL; }
-        if (m_lang)        { delete m_lang;       m_lang = NULL; }
+        if (m_lang)        { delete[] m_lang;     m_lang = NULL; }
```
poire-z, line 363: Does this change anything in any way in what the compiler would do ? Or it's just for sake of correctness?

pazos: It deletes the entire array of characters instead of the first one. Until now the memory of the rest of characters was leaked on each deallocation

poire-z: I get the idea.
But it's one single thing I created/allocated with `m_lang = new char[strlen(lang)+1];` - so I find it strange the compiler would just delete the first char (a fragment of that thing, while it knows how much was allocated). It's my idea the compiler would do the right thing, but I don't know him personally :)

pazos: >  I don't know him personally :)

Me neither :smile: 

But C++ is a humongous beast with operator overloading and other stuff and this specific behaviour is forbidden. See https://isocpp.org/wiki/faq/freestore-mgmt#delete-array-built-ins

thirdparty/kpvcrlib/CMakeLists.txt
NiLuJe: It was to keep the same order as the order in which they're defined/used below ;).

xtext.cpp
NiLuJe: It's *apparently* mostly harmless on Linux, but, it is admittedly an implementation detail, and Valgrind was certainly unhappy about it ;).

NiLuJe:
> > A CRe -> CRe swap doesn't make a lick of difference
> 
> A difference vs what ? :) You mean you get the same memory increase as CRe>FM>CRe ? Or that you get the same memory usage (no increase) ?

Yep, same increase.

Checking statm manually, the increase is actually in tow steps: on close, and then on open.

c.f., 

emu:

```
echo $(( $(cut -f2 -d' ' /proc/$(pidof luajit | cut -f2 -d' ')/statm) * 4096 / 1024 / 1024. ))
```

Kobo:

```
echo $(( $(cut -f2 -d' ' /proc/$(pidof reader.lua | cut -f2 -d' ')/statm) * 4096 / 1024 / 1024. ))
```

NiLuJe:
I also think I have a vague idea of why djvu is stuffing itself like crazy, I'll see if I can tame it a bit.

NiLuJe:
> Checking statm manually, the increase is actually in two steps: on close, and then on open.

Slightly more subtle than that, which means possibly not all-that-CRe-related ;).

* Restart KOReader: 19.921875
* Open folder with single book: 23.6328125
* Tap book: 38.68359375
* Open top menu (on search tab): 43.12890625 <- Oh, hi, 4MB I was talking about ;p.
* Back to FM: 40.7109375
* Tap book: 44.76171875
* Menu: 47.63671875
* Back to FM: 45.5546875
* Tap book: 49.28125
* Menu: 52.49609375
* Back to FM: 49.51171875
* Tap book: 53.36328125
* Menu: 56.609375
* FM: 53.6171875

etc. ;).

Soooo, probably not so much CRe's fault ;p.

zwim:
Horray. I admire zstd. I wIll test this bunch of commit tomorrow. +5 from my side :))

NiLuJe:
@poire-z: FWIW, now that I finally have a somewhat readable Valgrind output (hi, https://github.com/koreader/koreader/pull/6980), most, if not all, of the "still reachable" stuff is from FT/HB.

It's a bit hard to dig into that, because AFAICT, LuaJIT doesn't unload packages on exit (and we don't really have teardown facilities in place anyway), so it's hard to judge which of those pointers are actually expected to stay alive (i.e., cached stuff).

NiLuJe (commented):
> Reviewed 30 of 30 files at r1.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1257)* status: all files reviewed, 2 unresolved discussions (waiting on @poire-z)

NiLuJe:
History cleaned up (use a merge commit to merge); detailed history archived in https://github.com/NiLuJe/koreader-base/tree/cbb-c99_archive

poire-z:
316a72d feels like it doesn't deserve its own commit - and could be squased in the main commit(s) that touched these parts.

> use a merge commit to merge

Isn't "Rebase and merge" cleaner ? (no strange "Merge pull request..." additional commit that seems to redo what the previous do).
Or there's something I don't get about that "Create a merge commit" - what is it ? :)

NiLuJe:
> [316a72d](https://github.com/koreader/koreader-base/commit/316a72da759e4c8945c228627da80d0239f977a8) feels like it doesn't deserve its own commit - and could be squased in the main commit(s) that touched these parts.
> 

Agreed, but it was simply too hellish to re-order in terms of merge conflicts ;).

> Or there's something I don't get about that "Create a merge commit" - what is it ? :)

It makes it clear that a series of commits belonged to a single PR (besides the tiny, tiny mention about that in GitHub's UI).

poire-z:
> It makes it clear that a series of commits belonged to a single PR 
> I'm not generally a fan of it, but when the series is as "messy" as these, it kinda helps a bit?

A tiny little wee bit, ok :) 

There's another little thing I don't like much, it's the commit timestamps. I'd rather have them all the date of the merge, that a mix of misordered date from the last 2 weeks (which may make following the changes a bit confusing).
On my side, when there's work spanning multiple days, or when re-ordering/squashing/cleaning, I just do before pushing:
`git rebase --force-rebase --exec "sleep 2" --exec "git commit --amend --date=now" master`

NiLuJe:
`git rebase --ignore-date` ;o).

poire-z:
I think I needed the `sleep 2`, or sometimes, _Github_ would not show them in _my_ original order (they may be in the correct order once merged, dunno).

NiLuJe:
Okay, Busted apparently has a really hard time dealing with cdata finalizers (even when running Busted master, which supposedly does it better), so, ensure the tests properly cleanup behind themselves.

Also, finally fix the KOptContext language thingy so that it stops randomly failing because that's not how you copy a C string ;p.

NiLuJe (commented):
> Reviewed 11 of 11 files at r2.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1257)* status: all files reviewed, 2 unresolved discussions (waiting on @poire-z)

NiLuJe:
(Latest history archived in https://github.com/NiLuJe/koreader-base/tree/cbb-c99-front_archive_r2 (nice typo!) :D).

PR #1258, opened 2020-12-12
framebuffer: Make sure scaleBySize honors a DPI override

NiLuJe:
Even if said override happens to match the native DPI, because the scaleBySize heuristic is weird and results would differ.

c.f., https://github.com/koreader/koreader/pull/6977#issuecomment-743276911

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1258)

### DIFF
--- a/ffi/framebuffer.lua
+++ b/ffi/framebuffer.lua
@@ -337,10 +337,19 @@ function fb:getDPI()
 end
 
 function fb:setDPI(dpi)
-    -- Nothing except having a DPI override ("screen_dpi" reader setting) set calls Device:setScreenDPI -> fb:setDPI
+    -- Nothing except having a real DPI override ("screen_dpi" reader setting non nil) set calls Device:setScreenDPI -> fb:setDPI
     self.dpi = dpi
     -- Remember that we're using an override for scaleBySize
     self.dpi_override = true
+end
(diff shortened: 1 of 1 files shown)


Frenzie (approved):

poire-z:
Ref https://github.com/koreader/koreader/pull/6977#issuecomment-743300572

So, we continue getting a different result whether auto 300 dpi vs. user-set 300 dpi ?
You're less puzzled than yesterday ? Or just don't want to introduce a change ?

NiLuJe:
Getting consistent results would involve just *always* going the scale_size + scale_dpi way (with scale_dpi actually being scaleByDpi), which would change a lot of things, yeah.

At least this way it does what it probably intended to do without the weird corner-case of "except when override matches native".

NiLuJe:
And this way, we can actually *see* what would happen if we went that way, by setting an override that matches our device's native DPI ;).

NiLuJe:
It's definitely not a bad idea, and would probably not change much on legacy devices, because scale_size is 1 there, and scaleByDpi would be almost exactly 1, too (167 / 160).

poire-z:
OK, so _not_ less puzzled :)

PR #1260, opened 2020-12-19
Fix ZSTD rpath

NiLuJe:
We don't really need it, but it was broken by the switch to env from `MOREFLAGS` ;).

Also fix CRe's dependencies on macOS to avoid it depending on brew's zstd.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1260)


Frenzie (approved):
lgtm

PR #1268, opened 2020-12-23
Android: Never ever touch the Lua blitter

NiLuJe:
As a side-effect of #1200, software nightmode was no-longer handled in the C blitter, it instead went through to the Lua blitter.

This exacerbated the latent issues with LuaJIT and mcode_alloc on that platform (remember, the C blitter was initially implemented *specifically* to avoid stressing LuaJIT with the blitter code on Android because of that fact).

So, without going back to the admittedly messy way this was handled before, tweak a few things to make Android a special snowflake again, so that nightmode on Android no longer goes through the Lua blitter.

In fact, port `paintRoundedRect` to the C blitter to make doubly sure of that (and get rid of an ugly hack that was necessary to handle inversion quirks in setPixel otherwise ;p).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1268)

### DIFF
--- a/blitbuffer.c
+++ b/blitbuffer.c
@@ -221,7 +221,7 @@ static inline void BBRGB32_SET_PIXEL_CLAMPED(BlitBuffer * restrict bb, int rotat
 }
 
 static inline unsigned int BB_GET_WIDTH(BlitBuffer * restrict bb) {
-    if (GET_BB_ROTATION(bb) & 1 == 0) {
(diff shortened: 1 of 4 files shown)


NiLuJe:
Fun fact: remember the days when the one sure-fire way we could reproduce the same kind of weird JIT hiccups on Kobo back in the day, @poire-z?

That was with AlphaContainer & MovableContainer... And what do they have to do with this? They're the main caller of paintRoundedRect... :o)

PR #1270, opened 2020-12-24
thirdparty/libzmq 4.3.1

Frenzie:
It worked fine in my basic tests, but I haven't tested Android yet.

4.3.2+ has a build error that'd require a little more energy to investigate.

Don't merge before <https://github.com/koreader/koreader-base/pull/1266>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1270)


NiLuJe:
I seemed to remember that we actually require the hilariously old API?

Also, it kinds of go in a set with all the ømq stuff (c.f., ffi-cdecl/zeromq_cdecl.c).

It's been a while since I checked, but the last time I did, I went: ahahahahahah, nope.

Frenzie:
Oh yeah, you're right. I just did an isolated check to see if the library worked. /whistles

PR #1273, opened 2020-12-27
Mac launcher: do not pass directory as argument

yparitcher:
depends on koreader/koreader#7053

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1273)


pazos (approved):
Cool!, now with 50% less boilerplate 😄

PR #1277, opened 2020-12-28
SDL: Handle text input

johnbeard:
Turn on text input for SDL, which means it decodes key events
into text events according to the keyboard layout and delivers
it as Unicode.

This can then be wrapped in an SDL event and passed upwards.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1277)

### DIFF
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -121,6 +121,11 @@ function S.open(w, h, x, y)
     S.renderer = SDL.SDL_CreateRenderer(S.screen, -1, 0)
     S.texture = S.createTexture()
 
+    -- Start delivering Unicode text as well as keypresses - this will
+    -- handle things like Shift-numbers and letters according to the layout
+    -- and will deliver SDL_TEXTINPUT with that text.
+    SDL.SDL_StartTextInput()
+
     openGameController()
 end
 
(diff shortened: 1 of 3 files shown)


ffi/SDL2_0.lua
```diff
@@ -121,6 +121,11 @@ function S.open(w, h, x, y)
     S.renderer = SDL.SDL_CreateRenderer(S.screen, -1, 0)
     S.texture = S.createTexture()
 
+    -- Start delivering Unicode text as well as keypresses - this will
+    -- handle things like Shift-numbers and letters according to the layout
+    -- and will deliver SDL_TEXTINPUT with that text.
+    SDL.SDL_StartTextInput()
```
Frenzie, line 127: Given that "On some platforms using this function activates the screen keyboard.", this shouldn't be done here but only where it's used.

ffi/SDL2_0_h.lua
```diff
@@ -883,4 +883,5 @@ struct SDL_version {
   Uint8 patch;
 };
 void SDL_GetVersion(struct SDL_version *) __attribute__((visibility("default")));
+void SDL_StartTextInput(void);
```
Frenzie, line 886: Please also add it to ccdecl (and is just at the end the best place? :-) ).

Frenzie (approved):
@pazos Do you agree with the general architecture as I proposed it, as implemented by this PR and koreader/koreader#7037? Because part of the idea there is that (theoretically) Android could hook into the same mechanism too.

pazos:
> @pazos Do you agree with the general architecture as I proposed it, as implemented by this PR and [koreader/koreader#7037](https://github.com/koreader/koreader/pull/7037)? Because part of the idea there is that (theoretically) Android could hook into the same mechanism too.

Sure. It will do the trick

PR #1289, opened 2021-01-20
Bundle lj_wpaclient on rM

NiLuJe:
Re https://github.com/koreader/koreader/issues/7175

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1289)


Frenzie:
Merging since we don't seem to have a reMarkable set up over here to double check if everything builds.

PR #1291, opened 2021-01-22
SDL: Handle and forward SDL_DROPTEXT events

johnbeard:
This allows the client to handle things like dragged URLs and other
text.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1291)

### DIFF
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -340,6 +340,9 @@ function S.waitForEvent(usecs)
         elseif event.type == SDL.SDL_DROPFILE then
             local dropped_file_path = ffi.string(event.drop.file)
             genEmuEvent(EV_SDL, SDL.SDL_DROPFILE, dropped_file_path)
+        elseif event.type == SDL.SDL_DROPTEXT then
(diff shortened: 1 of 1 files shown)


Frenzie (commented):
I think you forgot to commit the ffi and cdecl stuff but otherwise sure, why not? ^_^

johnbeard:
What cdecl stuff is needed for this? `DROPFILE` doesn't seem to have any?

Frenzie:
I think you meant to say it's already in there, don't worry.

https://github.com/koreader/koreader-base/blob/8614f997e487c440a9bb57252c20401f2aec54f3/ffi/SDL2_0_h.lua#L306-L307

https://github.com/koreader/koreader-base/blob/b8ac18ef9f31a76c9a1e8e9de3741908dcb19077/ffi-cdecl/SDL2_0_decl.c#L83-L84

Frenzie (approved):

johnbeard:
Ah right, I thought there was something _extra_ I was supposed to be doing :laughing:

PR #1293, opened 2021-01-23
Android: Fix nighmode on devices w/ RGB565 windows w/ the C blitter

NiLuJe:
Regardless of whether we actually could potentially coerce those buffers to RGB32, the sanity check makes sense ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1293)


NiLuJe:
xref: https://github.com/koreader/koreader/discussions/7187

NiLuJe:
In practice, this might end up being slower than the C BB, and the invert is not accurate.

Yay, RGB565!

Frenzie (approved):
/shrug ;-)

PR #1296, opened 2021-01-27
Update sdcv

NiLuJe:
Drop patch that was merged upstream

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1296)


Frenzie (approved):
Was going to do that later. :-)

PR #1306, opened 2021-02-15
Really fix gettext on Android

NiLuJe:
Wheee!

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1306)


Frenzie:
Just a note that https://savannah.gnu.org/news/?id=10513 (gettext 0.22.1 release notes) says "Building on Android is now supported. "

PR #1310, opened 2021-02-17
Fix sdcv build on macOS

NiLuJe:
Maybe?

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1310)


NiLuJe:
Let's see how the CI likes this, because the mails are starting to drive me nuts...

NiLuJe:
The underlying issue being we build glib & gettext against our own GNU libiconv, but scdv was somehow picking up macOS's own libiconv, which uses different symbol names...

NiLuJe:
Eh, it's the other way around: gettext is built against the OS's libiconv :s.

NiLuJe:
Good news: I could reproduce it on my 10.14 system.

Bad news: I hope this takes, because gettext's configure script takes fucking forever to run.

NiLuJe:
Note to self: update the deps list/workflow in front for bison.

PR #1312, opened 2021-02-21
MXCFB: Match the Kobo behavior for "fast" on every device where I know it'll behave sanely.

NiLuJe:
We've effectively switched to DU for "fast" everywhere, and we've
limited our usage of "fast" to plain highlights.
Relying on FORCE_MONOCHROME for these use-cases is unnecessary, and has
known side-effects on various platforms, ranging from minor refresh
glitches to more severe and pernicious issues.

(That leaves PB, which I can't test, and where DU and especially A2
are already special snowflakes to begin with, even on NXP SoCs).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1312)

### DIFF
--- a/ffi/framebuffer_mxcfb.lua
+++ b/ffi/framebuffer_mxcfb.lua
@@ -646,7 +646,7 @@ function framebuffer:init()
         self.mech_wait_update_complete = kindle_pearl_mxc_wait_for_update_complete
         self.mech_wait_update_submission = kindle_mxc_wait_for_update_submission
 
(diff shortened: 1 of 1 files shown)


NiLuJe:
In short: stop requesting `FORCE_MONOCHROME` for `fast`. We don't need it, and it's hella buggy on EPDCv2.

yparitcher:
regression:
https://github.com/koreader/koreader/issues/7326#issuecomment-783801531
on PW4 (rex)

PR #1314, opened 2021-02-23
MXCFB:  Unbreak Zelda/Rex

NiLuJe:
~~* One that actually *uses* REAGLD only on a select few devices, those where REAGL appears extra finicky (right now, Kindle Zelda & Rex).~~
~~* Re-enable FORCE_MONOCHROME on Zelda & Rex. (Anyway, Rex can be lower res, so it somewhat weighs in the balance, too).~~
* Fix a terrible typo in #1312 that used the wrong A2 constant on Zelda/Rex, causing the Reader to get squished to 2bit by FORCE_MONOCHROME. Ouch.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1314)

### DIFF
--- a/ffi/framebuffer.lua
+++ b/ffi/framebuffer.lua
@@ -163,6 +163,10 @@ function fb:refreshPartialImp(x, y, w, h, d)
     -- default is fallback
     return self:refreshFullImp(x, y, w, h, d)
 end
+function fb:refreshReagldImp(x, y, w, h, d)
+    -- default is fallback
+    return self:refreshPartialImp(x, y, w, h, d)
+end
 function fb:refreshFlashPartialImp(x, y, w, h, d)
     -- default is fallback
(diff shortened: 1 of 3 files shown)


yparitcher (approved):
Fixes Both issues

NiLuJe:
Thanks, that means I can scrap the whole REAGLD idea :D.

We've also definitely confirmed that I should stop writing these things at 4AM ^^.

NiLuJe:
And this much simpler version should still fix all the things ;).

yparitcher:
You still left some REAGLD stuff (ZELDA_GLD16)

NiLuJe:
Yeah, can't hurt leaving the actual flag selection logic in, given how prone I am to forget that the Zelda/Rex constants are specific ^^.

PR #1317, opened 2021-02-26
Fix visibility of custom non-wrapped MuPDF symbols.

NiLuJe:
Missed those :/.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1317)


Frenzie (approved):
(probably won't help the reporter though ;-) )

PR #1318, opened 2021-03-03
Add reMarkable2 generateFakeEvents

snelg:
When the reMarkable2 is suspended and the user presses the Power button, the device wakes up but does NOT send the power button "pressed" or "released" events to `/dev/input`.
This PR is an alternate solution to https://github.com/koreader/koreader/pull/7345

If someone has a cleaner way to watch DBus events on the reMarkable2, let me know.
This code simply launches the command
```
dbus-monitor --system member='PrepareForSleep'
```
and then watches for a line that starts with
```
[three spaces]boolean false
```

That feels a bit klunky and brittle, but it's working on my machine, so [*shrug*]

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1318)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,13 +109,13 @@ libs: \
 $(OUTPUT_DIR)/libs/libinkview-compat.so: input/inkview-compat.c
 	$(CC) $(DYNLIB_CFLAGS) -linkview -o $@ $<
 
-$(OUTPUT_DIR)/libs/libkoreader-input.so: input/*.c input/*.h $(if $(KINDLE),$(POPEN_NOSHELL_LIB),)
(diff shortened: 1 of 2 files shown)


NiLuJe:
I'd say look into using the actual C DBus API, but libdbus is a nightmarish hellscape to use ^^.

(I mean, when the official docs points you to *another* C implementation, you know that you *really* don't want to keep looking at it any further ^^).

Said recommended binding is GDBus, which may or may not be a part of glib in the ancient glib version we build to avoid meson.

TL;DR: Yeah, I wouldn't worry overmuch about using the CLI tools, despite them being *also* somewhat clunky to use.

NiLuJe (approved):
:+1:

NiLuJe:
The upside is that the glib stuff doesn't actually depend on libdbus, so we *should* be building it already, if someone down the line cares enough to do it this way.

In the same vein, the Kindle variant would also be ported to openlipc, FWIW ;).

PR #1321, opened 2021-03-06
Update to KoboUSBMS 1.2.0

NiLuJe:
* User-friendly check if the SD card is busy, instead of a hard failure during umount, like we already did for onboard.
* Fancier clock updates ;p.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1321)


Frenzie (approved):
mkay

PR #1327, opened 2021-03-14
Allow working with PDF annotations

Toromtomtom:
This exposes mupdf functions for searching PDF annotations and changing their contents.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1327)


NiLuJe (approved):
Thoroughly unfamiliar with the APIs, but this looks sound ;).

PR #1331, opened 2021-03-15
Add a unit test for editing annotations

Toromtomtom:
This adds a unit test for the annotation editing introduced in https://github.com/koreader/koreader-base/pull/1327.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1331)


spec/unit/mupdf_spec.lua
```diff
@@ -137,6 +139,29 @@ describe("mupdf module", function()
...
+            assert.is_not_nil(page)
+            local annot = page:getMarkupAnnotation(ffi.new("float[8]", {
+                 70,  930,
+                510,  930,
+                510,  970,
+                 70,  970 }),
+                1)
+            page:updateMarkupAnnotation(annot, "annotation contents")
+            page:close()
+            doc:writeDocument(simple_pdf_out_annotated)
```
NiLuJe, line 156: This is probably missing a `doc:close()` (As does the existing test above, apparently :/).

spec/unit/mupdf_spec.lua
```diff
@@ -137,6 +139,29 @@ describe("mupdf module", function()
...
+            local out_f = io.open(simple_pdf_out_annotated, "r")
+            out_f:read("*a")
+            out_f:close()
```
NiLuJe, line 159: Also applies to the existing test above: I have no idea what this is attempting to test :D.

spec/unit/mupdf_spec.lua
Toromtomtom: Thanks, I added the `doc:close()` to both tests.

spec/unit/mupdf_spec.lua
Toromtomtom: Thanks, removed it in both tests.

NiLuJe (approved):
Thanks!

Frenzie (approved):

PR #1332, opened 2021-03-16
The great input handling spring cleanup

NiLuJe:
Because we usually want a clock that never goes backward ;).

As far as performance goes, AFAICT, on devices we care about, either *nothing* is implemented via vdso (so it isn't particularly any slower than `gettimeofday`) [e.g., on Mk. 5], or they *both* are, so it isn't any slower either [e.g., Mk. 7].

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1332)


NiLuJe:
More vdso comments: https://github.com/koreader/koreader/pull/6798#issuecomment-711445144

Frenzie (commented):
Backward? That a Forma thing? :-P

NiLuJe:
Nope, just a sysclock running ever so fast, and getting corrected on resume by the RTC, which is not entirely uncommon ;).

NiLuJe:
(Draft because it's not much more tightly coupled w/ https://github.com/koreader/koreader/pull/7415; and I haven't updated the other waitForEvent implementations ;p).

NiLuJe:
You know you've been staring at strace's output for too long when you recognize an INT32_MAX split into a timeval struct ;p

NiLuJe:
Whee, accurate timestamps on Android :}

```
03-23 22:35:50.669  7454  7493 I KOReader:  GestureDetector:probeClockSource: Touch event timestamps appear to use CLOCK_MONOTONIC
```

NiLuJe:
* Updated <linux/input.h> FFI stuff in order to be able to use it in front instead of re-declaring everything there.
* Fixed lua-rapidjson build w/ CMake >= 3.20
* Added some errno consts to the POSIX ffi module
* Added clock_gettime & friends to the POSIX ffi module
* Added setpriority & friends to the POSIX ffi module
* Added sched_setscheduler & friends to the POSIX ffi module
* Stopped storing useless unused references to require'd FFI modules
* Switched SDL2 input events to synthetic CLOCK_MONOTONIC (native ones are not helpful, they're relative to the app's startup).
* Simplified SDL2 input polling to actually rely on SDL2 instead of implementing a weird timeout handling ourselves.
* Preserve input event timestamps on Android, they're in a sane time base (MONOTONIC).
* Simplified and documented input polling on Android. We don't actually use the weird delayed-callback mechanism from android_native_app_glue,
  so, make that clear by relying on ALooper's native API instead of a weird mix of the two.
* Made ffiUtil.runInSubProcess tweak the subprocess scheduling and priority to match the use-case: non-interactive background tasks.
  This ensures that the kernel will schedule it accordingly without requiring weird yield hackery in front.
* Revamped input handling on most platforms, and especially on embedded pure-Linux eInk devices:
  * The C/Lua input module has been modernized and simplified.
  * Of specific note, the API of waitForEvent has been updated to get rid of protected mode calls in front.
    This allows us to get rid of the cost of handling throws in an interactivity-sensitive and hot codepath,
    and makes error-checking much easier, because parsing dynamically generated strings is not my idea of fun.
    Timeout values are also preserved in a timeval-like format, preventing truncation.
  * A timerfd backend has been implemented for the "timed callbacks" API required by GestureDetector to handle holds & double-tap detection.
    This allows us, among other things, to honor the input event's timestamps without having to resort to time-base conversions or hackery
    on devices where input events use the REALTIME clock source.

NiLuJe (commented):
> Reviewed 25 of 31 files at r1, 6 of 6 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1332)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

NiLuJe (commented):
> Reviewed 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1332)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

NiLuJe (commented):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1332)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

input/.clang-format
```diff
@@ -0,0 +1,151 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Mozilla
```
poire-z, line 3: What's this file? (erroneous addition, or does it make sense to have it in the repo?)

poire-z (commented):
Not really familiar with any of all that to comment much, so trusting you :)

input/.clang-format
NiLuJe: It's the clang-format config being used for the code formatting, so, yes ;).

poire-z: Code formatting by what ?

NiLuJe: Well, Clang's `clang-format` ;).

NiLuJe:
Welp, here goes nothing ;).

NiLuJe:
As a final nail on the coffin of the vdso discussions: on ARM, clock_gettime (and, well, gettimeofday) got a vdso implementation in Linux 4.1.

Which means we technically could have vdso support on kobo Mk. 7... But, of course, kobo's hilariously outdated userland strikes again: handling of this vdso codepath on arm was only added in glibc 2.21 (at the earliest).

As they're currently running glibc 2.19, that explains why I can still see those in `strace` :(.

NiLuJe:
And it requires hardware that is optional on Cortex A8 & A9 anyway.

Linux:
https://lwn.net/Articles/612835/
https://patchwork.kernel.org/project/linux-arm-kernel/patch/1426886003-30123-4-git-send-email-nathan_lynch@mentor.com/

glibc:
http://patchwork.sourceware.org/project/glibc/patch/1433442735-28785-1-git-send-email-nathan_lynch@codesourcery.com/

PR #1338, opened 2021-03-26
FBInk updates

NiLuJe:
There was a rather nasty bug in the OT codepath that could possibly cause KoboUSBMS to do Bad Things in some circumstances.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1338)


Frenzie (approved):
Is that the OpenSSL recent vulnerability?

NiLuJe:
Possibly? There's a CVE in the ChangeLog: https://github.com/openssl/openssl/blob/master/CHANGES.md#openssl-111

PR #1339, opened 2021-03-26
MXCFB: Move cap checks to front

NiLuJe:
This makes adding support for new devices less unwieldy and error-prone.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1339)


Frenzie (approved):
Sure.

PR #1343, opened 2021-04-01
SDL/Android: Actually honor polling timeouts

NiLuJe:
Hopefully can't happen on Android, but very common on SDL because we don't do anything with mouse pointer events.

Only noticed tonight because my gamepad happens to be spamming gyro events tonight (which we don't do anything with, too) ;p.

Basically, waitForEvent is designed to poll *once* and return *one* event. But the SDL & Android version need to generate evdev input events from Abdroid/SDL input events, which mean they nerd to handle 1 => many scenarios. This is handled via a FIFO queue, which means the polling is stuck in a forever loop.
That means we were never going back to Lua to recompute timeouts, which was bad: that meant scheduled task would only actually run on the next input event instead of the scheduled time.

PB is a special snowflake, and, while it's also a forever loop, it's one that happens to re-compute its timeout (an uses custom timeouts anyway).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1343)

### DIFF
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -257,6 +257,7 @@ function S.waitForEvent(sec, usec)
     local event = ffi.new("union SDL_Event")
     -- TimeVal's :tomsecs if we were passed one to begin with, otherwise, -1 => block
     local timeout = sec and math.floor(sec * 1000000 + usec + 0.5) / 1000 or -1
+    print("SDL:", timeout)
     while true do
         -- check for queued events
         if #inputQueue > 0 then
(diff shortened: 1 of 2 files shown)


NiLuJe:
That obviously led to an existential question: *why* do we handle events one by one?

Turns out, it doesn't make sense, because Linux evdev inputs can be read in batches, too.

So, err, I may have done just that ;p.

Frenzie:
> Turns out, it doesn't make sense, because Linux evdev inputs can be read in batches, too.

I never thought to check!

NiLuJe:
Turns out is as simple as that:

https://github.com/NiLuJe/koreader-base/blob/592105b99ef0fa8442a0ef01c49398ebfce88365/input/input.c#L323-L327

(That's the magic of O_NONBLOCK, it'll return EGAIN and we go on our merry way once the batch is drained ;).

NiLuJe:
I mean, we could technically read *once*, too, but that implies some list/queue handling in C, at which point, better just use libevdev to do that for us ;).

PR #1344, opened 2021-04-01
Input: Handle events in batches

NiLuJe:
The legacy design of consuming events one by one made the SDL & Android backends somewhat awkward, because they needed to generate many evdev events from possibly a single native event.

Moreover, even on plain Linux, you're mostly guaranteed that events will be readable in a batch, with boundaries at each `SYN_REPORT`.
Actually draining the queue ASAP pushes the possibility of a queue overflow (and as such, a `SYN_DROPPED`, which we don't actually handle) even further away from the realm of possibility, which is good, because, again, we're not handling this case anyway ;).

Except for PB being a special snowflake with its custom loop and timeout handling that I'm not touching, this makes the Android/SDL implementations code flow much closer to the C implementation's.

Speaking of, since I had to play with the C Lua API to handle the table wrapping, I documented the stack handling in the C backend, because that stuff is mind bending when you're not used to it.
And I switched to `rawset`/`rawseti` since we're dealing with a plain table with no metamethods.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1344)

### DIFF
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -257,156 +257,155 @@ function S.waitForEvent(sec, usec)
     local event = ffi.new("union SDL_Event")
     -- TimeVal's :tomsecs if we were passed one to begin with, otherwise, -1 => block
     local timeout = sec and math.floor(sec * 1000000 + usec + 0.5) / 1000 or -1
-    while true do
-        -- check for queued events
-        if #inputQueue > 0 then
-            -- return oldest FIFO element
-            return true, table.remove(inputQueue, 1)
-        end
 
-        -- otherwise, wait for event
-        local got_event = SDL.SDL_WaitEventTimeout(event, timeout)
-        if got_event == 0 then
-            -- ETIME
-            return false, C.ETIME
-        end
+    -- Reset the queue
+    inputQueue = {}
 
-        -- if we got an event, examine it here and generate events for koreader
-        if ffi.os == "OSX" and (event.type == SDL.SDL_FINGERMOTION or
-            event.type == SDL.SDL_FINGERDOWN or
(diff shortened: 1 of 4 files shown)


NiLuJe:
(Actually tested on Kobo/SDL/Android).

NiLuJe:
(Nixing whitespace changes makes the diff much more readable, because the SDL/Android diff basically just involves killing the outer while loop).

poire-z (approved):
Looks OK, at least for the Lua C API stuff, which is the only thing I can get :)

("Mind blowing" in a good sense, or a ridiculous one ? :) I found it quite rudimentary and strange, but once you get into it, it feels like it might be very clever. I wonder how much this simple stack idea makes it performant, vs. some other kind of API.)

NiLuJe:
No, just annoying to keep track of, which makes it hard to read ;).

Frenzie (approved):
input.c is the only place with actual changes, right? (besides removing that queue stuff) Otherwise I'll have to read it in more detail. ;-)

NiLuJe:
Yep! GH's ignore whitespace option makes the rest of the diff clearer ;).

Frenzie:
I don't see that, maybe a mobile issue

NiLuJe:
Ah, probably ;).

For reference, because even on desktop, you kinda have to know it's there:

Screenshot_20210402_155712

Frenzie:
Curious that GH detects the whitespace and can hide it, but doesn't show it as changed whitespace (darker green) in the diff.

NiLuJe:
True ;).

[git-delta](https://github.com/dandavison/delta/) sometimes manages to do that better, FWIW.

(And pretty much rules to begin with anyway ^^).

PR #1347, opened 2021-04-03
CRe: Disable CLOEXEC on LEGACY targets

NiLuJe:
For https://github.com/koreader/crengine/pull/433

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1347)


NiLuJe:
Possibly needed on older Android too, trying to find the exact point at which it started being supported...

PR #1349, opened 2021-04-04
CLOEXEC related tweaks

NiLuJe:
* CRe: Disable CLOEXEC on Kindle Legacy
* Use CLOEXEC in every one of our (ffi.C) open() calls.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1349)


NiLuJe:
Supersedes https://github.com/koreader/koreader-base/pull/1347 because I forgot that it was on master ;p.

For https://github.com/koreader/crengine/pull/433

PR #1351, opened 2021-04-09
CRe & nanosvg bump

NiLuJe:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1351)


Frenzie:
Sidenote, but there's a new rapidjson thingy too.

NiLuJe:
> Sidenote, but there's a new rapidjson thingy too.

Will need some code changes, as the JSON `null` is now a lightuserdata instead of a function, and I think we check that somewhere ;).

Frenzie:
I see; I hadn't looked at it. Just saw the tag(s). :-)

PR #1358, opened 2021-04-16
Android: Swith to pollOnce

NiLuJe:
pollAll is a wrapper around pollOnce useful when multiple fds are registered with callbacks, to ensure they *all* fire. Since we don't register callbacks at all, do away with this wrapper.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1358)

### DIFF
--- a/ffi/input_android.lua
+++ b/ffi/input_android.lua
@@ -68,6 +68,7 @@ local function genTouchUpEvent(event, id)
 end
 
 local function genTouchMoveEvent(event, id, index)
+    -- NOTE: May return a float for events w/ subpixel precision.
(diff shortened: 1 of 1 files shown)


NiLuJe:
As for the float coordinates, I get them for move events on my phone, and I *think* we're okay with that (because Lua, everything's already a double anyway).

NiLuJe:
(Also, wireless adb is fun :D.)

pazos:
https://github.com/android/ndk/discussions/2020 :)))

NiLuJe:
:sunglasses: :D

PR #1360, opened 2021-04-24
POSIX: Fix portability issues with clock_gettime

NiLuJe:
The constants are hilariously not portable. So, fix 'em on macOS & most common BSD flavors.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1360)


Frenzie (approved):
Ouch.

PR #1374, opened 2021-05-23
bump crengine: support for user hyphenation dictionary

zwim:
Apply this together with https://github.com/koreader/crengine/pull/439.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1374)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -3495,6 +3495,20 @@ static int getImageDataFromPosition(lua_State *L) {
     return 0;
 }
 
+static int setUserHyphenationDict(lua_State *L) {
+    CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+    const char *filename = luaL_checkstring(L, 2);
+    UserHyphenDict::init(filename);
+    return 0;
+}
+
+static int getHyphenation(lua_State *L) {
+    CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+    const char *word = luaL_checkstring(L, 2);
+    lString32 hyphenation = UserHyphenDict::getHyphenation(word);
+    lua_pushstring(L, UnicodeToLocal(hyphenation).c_str());
+    return 1;
+}
 
 static const struct luaL_Reg cre_func[] = {
     {"initCache", initCache},
@@ -3627,6 +3641,8 @@ static const struct luaL_Reg credocument_meth[] = {
     {"saveDefaults", saveDefaults},
     {"close", closeDocument},
     {"__gc", closeDocument},
+    {"setUserHyphenationDict", setUserHyphenationDict},
+    {"getHyphenation", getHyphenation},
     {NULL, NULL}
 };
 
--- a/cre.cpp
+++ b/cre.cpp
@@ -3495,14 +3495,6 @@ static int getImageDataFromPosition(lua_State *L) {
     return 0;
 }
 
-static int setUserHyphenationDict(lua_State *L) {
-    CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
-    const char *filename = luaL_checkstring(L, 2);
-    int no_sloppy_load = luaL_checkint(L, 3);
-    UserHyphenDict::init(filename, no_sloppy_load !=0);
-    return 0;
-}
-
 static int getHyphenationForWord(lua_State *L) {
     CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
     const char *word = luaL_checkstring(L, 2);
@@ -3511,7 +3503,7 @@ static int getHyphenationForWord(lua_State *L) {
     return 1;
 }
 
-static int getLower(lua_State *L) {
+static int getLowercasedWord(lua_State *L) {
     CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
     const char *word = luaL_checkstring(L, 2);
     lString32 word_lower = UserHyphenDict::getLower(word);
@@ -3659,9 +3651,8 @@ static const struct luaL_Reg credocument_meth[] = {
     {"saveDefaults", saveDefaults},
     {"close", closeDocument},
     {"__gc", closeDocument},
-    {"setUserHyphenationDict", setUserHyphenationDict},
     {"getHyphenationForWord", getHyphenationForWord},
-    {"getLower", getLower},
+    {"getLowercasedWord", getLowercasedWord},
     {"formatHyphenationSuggestion", formatHyphenationSuggestion},
     {NULL, NULL}
 };
--- a/cre.cpp
+++ b/cre.cpp
@@ -3500,7 +3500,7 @@ static int setUserHyphenationDict(lua_State *L) {
     CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
     const char *filename = luaL_checkstring(L, 2);
     bool reload = lua_toboolean(L, 3);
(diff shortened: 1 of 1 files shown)


cre.cpp
```diff
@@ -3495,6 +3495,29 @@ static int getImageDataFromPosition(lua_State *L) {
...
+
+static int getHyphenation(lua_State *L) {
+    CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+    const char *word = luaL_checkstring(L, 2);
+    lString32 hyphenation = UserHyphenDict::getHyphenation(word);
+    lua_pushstring(L, UnicodeToLocal(hyphenation).c_str());
+    return 1;
+}
+
+static int formatHyphenation(lua_State *L) {
```
poire-z, line 3513: Can we have clearer names ?
getUserHyphenationForWord ? getUserHyphenationFormForWord ?
getFormattedUserHyphenationForWord ?

Not sure what are the differences between all these :) Is formatHyphenation() to just not have to add utf8proc ffi stuff ? (which is fine as long as the function name hides this fact :)

cre.cpp
```diff
@@ -3495,6 +3495,38 @@ static int getImageDataFromPosition(lua_State *L) {
...
+    const char *word = luaL_checkstring(L, 2);
+    lString32 hyphenation = UserHyphenDict::getHyphenation(word);
+    lua_pushstring(L, UnicodeToLocal(hyphenation).c_str());
+    return 1;
+}
+
+static int getLower(lua_State *L) {
+    CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+    const char *word = luaL_checkstring(L, 2);
+    lString32 word_lower = UserHyphenDict::getLower(word);
```
poire-z, line 3517: That's what I aimed at in a previous comment:
> Is formatHyphenation() to just not have to add utf8proc ffi stuff ? (which is fine as long as the function name hides this fact :)

It feels (here and in hyphman.h) strange to export such a low level lowercasing API function.
I'm OK with not wasting time with making a ffi/utf8proc.lua - and it could make sense to be sure we get the exact same lowercasing result as crengine.
So, please juste give it a less low-level name: `getLowercasedWord()` here and `UserHyphenDict::getLowercasedWord(word)` - which makes it feel it's all hyphenation higher level business.

cre.cpp
```diff
@@ -3495,14 +3495,6 @@ static int getImageDataFromPosition(lua_State *L) {
     return 0;
 }
 
-static int setUserHyphenationDict(lua_State *L) {
-    CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
-    const char *filename = luaL_checkstring(L, 2);
-    int no_sloppy_load = luaL_checkint(L, 3);
-    UserHyphenDict::init(filename, no_sloppy_load !=0);
-    return 0;
-}
```
poire-z: Oh, you were really not using this function?
I somehow still prefer this one to the trick of hacking crengine.textlang.hyphenation.user.dict and appending -reload.
The support for setStringProperty("crengine.textlang.hyphenation.user.dict") in crengine is good to allow it to be used by other frontends (who would probably not implement editing) - but we're allowed to access lower level objects like UserHyphenDict in cre.cpp if we need to do a bit more that what we can do with setStringProperty.

zwim: Yes I don't need that function. 
The cool thing on the string.prop is, that I can apply a rerender if the dict changes. (And we need the rerender, because some heights of paragraphs might change.

poire-z: Yes, I understand the need for rerender.
What I mean is there is nothing currently like this for string properties in crengine:
https://github.com/koreader/crengine/blob/362fe344fa01dec188f33606f0257006594ea316/crengine/src/lvdocview.cpp#L6682-L6689
These properties get a single argument, and it's somehow stored/saved in CoolReader settings db.

If you need a 2nd argument (reload=true/false), it feels ugly to hack this string property to acccept dynamic strings :)
It feels more proper to do it like you did in that bit of code you removed - so I would keep it and use it in credocument.lua instead of the string property. (But keep the string property handling in crengine lvdocvien.cpp, without the "-reload" hack, as a normal CoolReader setting, without the ability to force a reload.)

zwim: What about one function to load the dict and one property just to rerender?

poire-z: Not sure what you mean by "rerender" (reloading the dict ? because re-rendering the book is never explicitely asked, it's done on page drawing by crengine when the hashes have changed).

But what's the issue with the above removed function? It seems just ideal as it was, no ?

zwim: Now, when I call the setproperty the dict is  loaded the hash changes and I call explicitly RENDER_REQUEST

poire-z: Ah, it's not auto re-rendered when you just set a new updated dict, as there was yet no need to check hash as not re-drawing? Just curious: would it be re-rendered when you go to the next page (then, their must be a rendering hashes check).
Anyway, you could do in the above function what's done for the string property:
```c
if ( UserHyphenDict::init(filename, no_sloppy_load !=0) ) {
    doc->text_view->requestRender(); /* this is public, and what the macro REQUEST_RENDER does */
}
```

zwim: Strange, it seems, that the rerendering works even without a `requestRender` now. Maybe that was necessary before I implemented the hash? Don't know anymore.

cre.cpp
```diff
@@ -3627,6 +3659,10 @@ static const struct luaL_Reg credocument_meth[] = {
     {"saveDefaults", saveDefaults},
     {"close", closeDocument},
     {"__gc", closeDocument},
+    {"setUserHyphenationDict", setUserHyphenationDict},
+    {"getHyphenationForWord", getHyphenationForWord},
+    {"getLowercasedWord", getLowercasedWord},
+    {"formatHyphenationSuggestion", formatHyphenationSuggestion},
```
poire-z, line 3665: Please move them just after hasNonLinearFlows (the last real added feature), from readDefaults to __gc it's mostly low-level internal stuff.

zwim:
Why does circleci not find `UserHyphenDict`? Isn't it in `hyphman.h`?

NiLuJe:
Not yet, since the CRe bit hasn't been merged & bumped ;).

poire-z (approved):

poire-z:
Your crengine hash is wrong.
You have:
```diff
--- a/thirdparty/kpvcrlib/crengine
+++ b/thirdparty/kpvcrlib/crengine
@@ -1 +1 @@
-Subproject commit 4946dfa6acdb560dc766e1f273ac2d8226c2c548     old one, good
+Subproject commit 96832ebdf36bec3a002674d8dc88b35862e664da     new one, wrong
```

while it should be the top of https://github.com/koreader/crengine/commits/master: fef32929a00406ad2fe73e1ace4ad67c641112f5

(May be it was the hash of your zwim/master and not the new one I created when merging in crengine?)

poire-z:
Uh, CircleCI works for you in koreader-base, but not in koreader ?!

Frenzie:
It's probably the annoying issue for which there seems to be no real solution. The frontend needs secrets to push updates (docs, translations, coverage) and those aren't made available to any random person. 

However, CircleCI has decided that means they shouldn't run at all. It's really silly.

poire-z:
But what @hius07 has done (or not done) to have CircleCI work for his PRs ?

Frenzie:
You can add them fake.

Frenzie:
On my phone atm but it'll be called something like koreader-vars. You can make an environment with that name and some nonsense (or empty?) values in CircleCI and then it'll be happy to run it.

PR #1382, opened 2021-07-10
Convert getLowercasedWord in lowercaseString

zwim:
The companion changes for #7956.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1382)


cre.cpp
```diff
@@ -3511,29 +3518,19 @@ static int getImageDataFromPosition(lua_State *L) {
 }
 
 static int setUserHyphenationDict(lua_State *L) {
-    CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
```
poire-z, line 3514: If these really don't require a document, and if you remove this, they should be part of `cre_func` like the other `cre.*Hyph*` stuff (that means, a function and not a method of the document) - and you should call them as `cre.setUserHyphenationDict(...)`. (As they would be used in a reader*.lua module loaded when a cre document is loaded, you don't need any `require("libkoreader-cre")` as `cre` has already been loaded and made global.

zwim:
How can I include  "libutf8proc.so"  instead of  "libutf8proc.so.2"?

ffi/utf8proc.lua
```diff
@@ -0,0 +1,63 @@
...
+    libutf8proc = ffi.load("libs/libutf8proc.so.2")
+end
+
+local Utf8Proc = {}
+
+function Utf8Proc.lowercase(str)
+    local str8_len = str:len()
+
+    local str8 = ffi.new("uint8_t[" .. str8_len .."]")
+    local str32 = ffi.new("int32_t[" .. str8_len .. "]")
```
NiLuJe, line 29: LuaJIT's FFI module has a dedicated syntax for that: ... `ffi.new("type[?]", len)`.

ffi/utf8proc.lua
```diff
@@ -0,0 +1,63 @@
...
+
+    -- copy string contents to uint8_t array
+    for i = 1,str8_len do
+        str8[i-1] = str:byte(i)
+    end
```
NiLuJe, line 32: :?

(Lua strings are immutable, you can use em as a `const char*`).

If you *do* need it mutable, there's `ffi.string` if you nee dit in Lua, or just `ffi.memcpy` in C?

NiLuJe:
> How can I include "libutf8proc.so" instead of "libutf8proc.so.2"?

You don't, the dynamic loader expects the SOVER to be there.

NiLuJe:
Also: doesn't `utf8proc_NFKC_Casefold` basically just do what you need without any weird faffing about?

poire-z:
Is `utf8proc_NFKC_Casefold()` really equivalent to the bunch of `utf8proc_tolower()` we only do in crengine ?
(Just asking, don't want to learn about NFKC&co.)
If different, does it matter in the way it is used in ReaderUserHyphenation?

zwim:
I have tested the user hyphenation with some lower/upper case combinations and did not find any difference (Umlauts at the beginning or in the middle of the word).
So I would say the answe is: They are equivalent.

poire-z (approved):
I'm fine with the cre.cpp change.
I let @NiLuJe say about the ffi stuff.

NiLuJe:
@poire-z: I haven't checked what CRe does, exactly, but, yeah, it's probably a possibly noticeable waste of time ^^.

On the other hand, `utf8proc_NFKC_Casefold` allocates memory, so, this needs to be taken care of.

ffi/utf8proc.lua
```diff
@@ -0,0 +1,27 @@
...
+elseif ffi.os == "OSX" then
+    libutf8proc = ffi.load("libs/libutf8proc.dylib")
+else
+    libutf8proc = ffi.load("libs/libutf8proc.so.2")
+end
+
+local Utf8Proc = {}
+
+function Utf8Proc.lowercase(str)
+    return ffi.string(libutf8proc.utf8proc_NFKC_Casefold(str))
```
NiLuJe, line 24: Re: memory

```suggestion
local folded_strz = libutf8proc.utf8proc_NFKC_Casefold(str)
local folded_str = ffi.string(folded_strz)
C.free(folded_strz)
return folded_str
```

poire-z:
> I haven't checked what CRe does, exactly, but, yeah, it's probably a possibly noticeable waste of time ^^.
> (NFKC + CaseFold is specifically designed for case-insensitive searches, AFAIU).

cre's HTML parser / DOM builder doesn't do any de/composition/canonisation of it's input, so we get Unicode/UTF32 as in the source, ie ü might be and stay one unicode char (`Ü`), or two unicode chars (`U`+`¨`).
For its lowercasing, it calls utf8proc_tolower(ch) on each of these chars.

`utf8proc_NFKC_Casefold()` might transform both `Ü` and `U`+`¨` into `ü`.
It's probably the right thing for comparing filenames in frontend when you do that on both inputs.

With cre, we might get a word with text selection that contains U+¨ - if lowercasing it via crengine lower, we'll get u+¨. Create a user hyphenation word for this, and it will match the selected word.
If using on one side `utf8proc_NFKC_Casefold()`, we may end up with adding in the hyph dict something with ü, which might then not match the original word having u+¨.

It might already be an issue with user hyph dict and having to enter a word in the various forms (althought I guess in a single book, you'll have always one form or the other) - but at least the one you're just editing will match in the current book.
I fear that by using `utf8proc_NFKC_Casefold()`, we might get no match/effect just after editing.
(And that's as much as I understand these composition/canonalisation :) 

But I guess most often, in books, you have the single char Ü - so if that's what NFKC would transform the rare U+¨ into, I guess we're fine - until someone complain :)

zwim:
> If using on one side utf8proc_NFKC_Casefold(), we may end up with adding in the hyph dict something with ü, which might then not match the original word having u+¨.

Wouldn't we also have a problem with our hyphenation patterns in the case the book uses u+¨ ?

poire-z:
> Wouldn't we also have a problem with our hyphenation patterns in the case the book uses u+¨ ?

Yes, I think we would, and we do:
image

so, let's say we expect NFC/NKFC from the source (and let's not fix crengine, it would just make it slower :)
https://fr.wikipedia.org/wiki/Normalisation_Unicode

So, go ahead, we'll see.

poire-z:
@NiLuJe likes ffi/utf8proc_h.lua to be generated from a ffi-cdecl/utf8proc_decl.c.
Dunno if there's a tool to create utf8proc_decl.c - so I made one manually :/
The ffi/utf8proc_h.lua can be generated then with https://github.com/koreader/ffi-cdecl (I might have an older version) with:
```
export CPPFLAGS="-I/space/kobo/koreader/base/thirdparty/utf8proc/build/git_checkout/utf8proc"
/path/to/ffi-cdecl/ffi-cdecl gcc koreader/base/ffi-cdecl/utf8proc_decl.c koreader/base/ffi/utf8proc_h.lua
```
Here are my/these 2 files: [ffi-cdecl_utfproc.zip](https://github.com/koreader/koreader-base/files/6796645/ffi-cdecl_utfproc.zip)

zwim:
@poire-z : Thank you for the `ffi-cdecl` information

poire-z:
Is this all done & fine ?
@NiLuJe : no more comments about the ffi stuff ?

zwim:
> Is this all done & fine ?
> @NiLuJe : no more comments about the ffi stuff ?

Done is all-ways relative :)  But for the near future I am done with this.

NiLuJe:
The typedefs are a bit wonky, but it's mostly harmless, I'll fix it next I'm around there ;).

poire-z:
Thanks, merged, ready to be bumped in the related frontend PR.

PR #1384, opened 2021-07-10
Remove zsync dependencies

Galunid:
It builds and doesn't crash, so it looks I didn't miss anything, but please double check

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1384)


Frenzie:
Definitely not zsync2! ;-)

NiLuJe:
IIRC, a few things actually *do* use some/all øMQ bits: Calibre Wireless & KOSync come to mind; Evernote probably does, but we don't really care ;).

Frenzie:
Yeah, I'll have do check in more detail. But last night I took a quick peek on my phone at the changelog and zsync2 jumped out.

Galunid:
I guess I'll close it, since it turned out to be more complicated than I initially thought. From what I found out, Calibre plugin needs only `zmq` and `czmq`, it builds and works without `FileMQ/Zyre`

poire-z:
Do you not still want to remove FileMQ/Zyre ? Or is that still complicated ?

Galunid:
Nevermind, I just realized how simple it was to check which library the stuff in `ffi/zeromq_h.lua` belong to and now I feel stupid, I guess at least I learned something new 😂

poire-z:
Is this ready and ok ?
Can we merge/bump with the related frontend PR, or is it safest to leave it for after the coming up v2021.07 release ?

Galunid:
@poire-z Pretty sure they're good to go, I tested them a bit (especially calibre plugin, since it mostly uses the remaining dependencies) and it worked alright. I'd rather wait for 2021.07 though, on the off chance something broke.

poire-z:
OK, so let's wait for after v2021.07.

Galunid:
In theory, should be gtg, @NiLuJe @Frenzie any comments?

NiLuJe:
Nothing jumps to mind ;).

PR #1396, opened 2021-08-07
BMP in 8bit grayscale

zwim:
Allows to save a BMP in 8 bit grayscale.
See https://github.com/koreader/koreader/issues/8046

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1396)

### DIFF
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -2051,12 +2051,12 @@ function BB_mt.__index:getBufferData()
     return bbdump, source_ptr, w, stride, h
 end
 
-function BB_mt.__index:writeBMP(filename)
+function BB_mt.__index:writeBMP(filename, grayscale)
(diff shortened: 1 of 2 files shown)


NiLuJe:
If you want to skip the floating point maths (which mpay be a moot point here, because Lua numbers are doubles), a decent approximation of an RGB -> Y8 conversion would be `(uint8_t) (((r * 77U) + (g * 150U) + (29U * b)) >> 8U)` (from stb).

zwim:
I would like to stay with double maths here. The integer variant does not speed things up here (in fact it might even be slower due to the normalization/division at the end).

zwim:
ping @NiLuJe  
Is this PR good from your side? Can it be merged?

NiLuJe (approved):

PR #1403, opened 2021-08-24
Update to OpenSSL 1.1.1l

NiLuJe:
Will check after dinner if the patches need to be refreshed ;p.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1403)


Frenzie (approved):
Builds passing = refresh not needed? :-)

NiLuJe:
Looks like it ;).

PR #1405, opened 2021-08-29
Update FBInk & co

NiLuJe:
Not much changes, besides properly tagged releases ;).

(And, well, `fbdepth` having been rewritten ;p).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1405)


Frenzie (approved):
That macOS build is starting to annoy me. >_< I thought it would keep us from accidentally breaking the macOS build; I didn't figure *it* would keep breaking us.

PR #1407, opened 2021-09-02
MµPDF: Don't accumulate rounding errors when computing page dimensions

NiLuJe:
The `getSize` methods of a page -- much like the various BBox methods -- all return an internal representation of rectangles, with floating point precision, for a fixed render DPI of 72.

Trying to round that to integer this early in the pipeline risks cascading large rounding errors after scaling back to the actual screen size & scale factor.

Noticed with a random CBZ where this returned weirdly tiny rectangles (still don't know why), but, most importantly, ones that featured meaningful decimals.

This lead to computing a scale factor that was slightly off, which meant we actually scaled it to a slightly unexpected size (here, bigger), while the images were actually perfectly screen sized.

So, extra scaling pass: bad!
Extra-scaling pass that breaks page mode + fit-to because it no longer actually fits on a single screen: *very* bad!

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1407)

### DIFF
--- a/ffi/mupdf.lua
+++ b/ffi/mupdf.lua
@@ -212,7 +212,6 @@ function document_mt.__index:isDocumentReflowable()
 end
 
 function document_mt.__index:layoutDocument(width, height, em)
-    print("layoutDocument", width, height, em)
     -- Reset the cache.
     self.number_of_pages = nil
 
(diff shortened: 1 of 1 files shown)


NiLuJe:
Floats all the way down ;o).

ffi/mupdf.lua
```diff
@@ -370,9 +370,20 @@ function page_mt.__index:getSize(draw_context)
 
     M.fz_bound_page(context(), self.page, bounds)
     M.fz_transform_rect(bounds, ctm)
-    M.fz_round_rect(bbox, bounds)
 
+    -- NOTE: fz_bound_page returns an internal representation computed @ 72dpi...
```
poire-z, line 374: It was called fz_round_rect, not fz_bound_page , if you're talking about the thing you commented out.
(And that's as much as I'm able to review this :)

NiLuJe: Yep, that's correct ;).

`fz_round_rect` is just the geometry helper that rounds a rectangle with floating point coordinates to a rectangle with integer coordinates. the comment is indeed about where the rectangle with floating point coordinates comes from (`fz_bound_page`) and why we don't run `fz_round_rect` on it right now (because we do it in from post-scaling via `Geom:transformByScale`)

Frenzie (approved):
If you're looking at MuPDF *hint, hint* :-P

When I tried https://github.com/koreader/koreader-base/pull/1203 earlier this year it seemed to work great, so I'm not really sure what still needs fixing.

NiLuJe:
There's still a few mysteries left in there, and I don't have time to properly look at it :/.

(Plus, it needs to be rebased so as not to scrap the few actual fixes that happened last year).

Frenzie:
> (Plus, it needs to be carefully rebased so as not to scrap the few actual fixes that happened over the last year).

Well yes, obviously it needs some work now. ^_^

PR #1408, opened 2021-09-09
bump crengine: CSS: parse/skip at-rules

poire-z:
Includes https://github.com/koreader/crengine/pull/454 : 
- XML parsing: strip leading \n in PRE and TEXTAREA
- Text: allow wrap on any space in "white-space: pre"
- initTableRendMethods(): avoid expensive processing
- Tables: fix handling of whitespace between nodes
- ldomDocument: store screen size
- CSS: more robust skipping of invalid content
- CSS: parse/skip at-rules, support `@media`, `@supports`
- CSS: handle media condition with `@import` and <link/>
- CSS: fix parsing of "font-family:inherit!important"
- LDOMNameIdMap::deserialize(): fix when >1024 items

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1408)

### DIFF
--- a/ffi/mupdf.lua
+++ b/ffi/mupdf.lua
@@ -362,7 +362,7 @@ calculate page size after applying DrawContext
 --]]
 function page_mt.__index:getSize(draw_context)
     local bounds = ffi.new("fz_rect[1]")
-    local bbox = ffi.new("fz_irect[1]")
+    -- local bbox = ffi.new("fz_irect[1]")
     local ctm = ffi.new("fz_matrix")
 
     M.fz_scale(ctm, draw_context.zoom, draw_context.zoom)


ffi/mupdf.lua
```diff
@@ -362,7 +362,7 @@ calculate page size after applying DrawContext
 --]]
 function page_mt.__index:getSize(draw_context)
     local bounds = ffi.new("fz_rect[1]")
-    local bbox = ffi.new("fz_irect[1]")
+    -- local bbox = ffi.new("fz_irect[1]")
```
Frenzie, line 365: Did @NiLuJe overlook this?

poire-z: Yes (but you too did not look at CI output with #1404 :)

Frenzie: It's because that stupid Mac build is training me to ignore failed checks. :-(

(And why did it not fail now? Ugh.)

poire-z: It did fail for my first commit - that's why I added the 2nd to fix the failure.

NiLuJe: Oh, right, I must have did my final test run while still running the "with debug print" version that still needed it to print it, my bad ;).

PR #1414, opened 2021-09-26
Add strlen for utf8 encoded strings

zwim:
This PR implements a function to count unicode codepoints in an utf8 encoded string

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1414)


poire-z:
What do you/we need it for ? (= what are you planning to add/fix ? :)

Strange that we don't have such a thing in frontend/util.lua or base/ffi/util.lua. The closest I would have used is `util.splitToChars()` and counting the nb of item in the resulting table.

poire-z:
Btw, we have a few more or less crappy functions such as `util.splitToChars()`, dunno if we'd better use things from libraries instead.
Although these crappy functions can be hacked to make them work with wilder content than these libraries - like the WTF-8 that we can get from Wikipedia JSON.

zwim:
> What do you/we need it for ? (= what are you planning to add/fix ? :)

See PR https://github.com/koreader/koreader/pull/8271
 
> Strange that we don't have such a thing in frontend/util.lua or base/ffi/util.lua. The closest I would have used is `util.splitToChars()` and counting the nb of item in the resulting table.

It would be an option to use `util.splitToChars()`, but we already have the utf8proc library.
I think translations are an ideal use-case for using utf8proc library.

ffi/utf8proc.lua
```diff
@@ -30,4 +30,16 @@ function Utf8Proc.lowercase(str)
     return folded_str
 end
 
+function Utf8Proc.strlen(str)
+    local str_buf = ffi.cast("const utf8proc_uint8_t *", str)
+    local buffer = ffi.new("utf8proc_int32_t[1]")
+    local retval = 0
+    local pos = 0
+    while str_buf[pos] ~= 0 do
+        pos = pos + libutf8proc.utf8proc_iterate(str_buf + pos, -1 , buffer)
```
NiLuJe, line 39: nit: stray space before the comma after `-1` ;).

NiLuJe:
> > What do you/we need it for ? (= what are you planning to add/fix ? :)
> 
> See PR [koreader/koreader#8271](https://github.com/koreader/koreader/pull/8271)
> 
> > Strange that we don't have such a thing in frontend/util.lua or base/ffi/util.lua. The closest I would have used is `util.splitToChars()` and counting the nb of item in the resulting table.
> 
> It would be an option to use `util.splitToChars()`, but we already have the utf8proc library.
> I think translations are an ideal use-case for using utf8proc library.

Yeah, +1 on going native/FFI ;).

ffi/utf8proc.lua
```diff
@@ -30,4 +30,16 @@ function Utf8Proc.lowercase(str)
...
 end
 
+function Utf8Proc.strlen(str)
+    local str_buf = ffi.cast("const utf8proc_uint8_t *", str)
+    local buffer = ffi.new("utf8proc_int32_t[1]")
+    local retval = 0
+    local pos = 0
+    while str_buf[pos] ~= 0 do
+        pos = pos + libutf8proc.utf8proc_iterate(str_buf + pos, -1, buffer)
+        retval = retval + 1
```
poire-z, line 40: May be make this easier to read by switching lines 39 and 40:
```
while char at pos is not NULL:
   count = count + 1
   advance pos
```
Also, name retval "count" or something, we know it's what will be returned with `return count` - `return retval` is redundant :)

poire-z (approved):

Frenzie (approved):
I'll leave final review to @NiLuJe but seems fine to me.

poire-z:
@NiLuJe : any last comment before we merge this?

ffi/utf8proc.lua
```diff
@@ -30,4 +30,16 @@ function Utf8Proc.lowercase(str)
...
+    local str_buf = ffi.cast("const utf8proc_uint8_t *", str)
+    local buffer = ffi.new("utf8proc_int32_t[1]")
+    local count = 0
+    local pos = 0
+    while str_buf[pos] ~= 0 do
+        count = count + 1
+        pos = pos + libutf8proc.utf8proc_iterate(str_buf + pos, -1, buffer)
+    end
+    return count
+end
```
NiLuJe, line 43: Doesn't handle errors sanely (`utf8proc_iterate` returns a negative value on error, so moving the pointer backward may lead to "interesting" results ;)).

This should do the trick (untested):

```suggestion
function Utf8Proc.strlen(str)
    local p = ffi.cast("const utf8proc_uint8_t *", str)
    local codepoint = ffi.new("utf8proc_int32_t[1]")
    local count = 0
    local pos = 0
    local bytes = 0
    while bytes >= 0 do
        bytes = libutf8proc.utf8proc_iterate(p + pos, -1, codepoint)
        if bytes > 0 then
            count = count + 1
            pos = pos + bytes
        end
    end
    return count
end
```

NiLuJe (changes requested):

ffi/utf8proc.lua
NiLuJe: It's unclear whether it can ever return `0`, in which case, either reworking the loop to avoid the funky check by preloading an initial call or breaking on 0 inside the loop might be necessary.

NiLuJe: TL;DR:

```suggestion
function Utf8Proc.strlen(str)
    local p = ffi.cast("const utf8proc_uint8_t *", str)
    local codepoint = ffi.new("utf8proc_int32_t[1]")

    local bytes = libutf8proc.utf8proc_iterate(p, -1, codepoint)
    if bytes <= 0 then
        -- No valid codepoints at all?
        return 0
    end
    
    local count = 1
    local pos = bytes
    while bytes > 0 do
        bytes = libutf8proc.utf8proc_iterate(p + pos, -1, codepoint)
        if bytes > 0 then
            count = count + 1
            pos = pos + bytes
        end
    end
    return count
end
```

poire-z:
If you're updating the code, you may as well rename the function from `:strlen()` to `:countCodepoints()`.

Or just `:count()` if we can expect a Utf8Proc module to provide functions that work _with_ Unicode codepoints _on_ UTF-8 strings.

NiLuJe:
Regardless of what I'm about to say, this is useful to have as-is ;).

But, a few notes (especially relating to how this is used in front):

* We have no guarantee that the input is composed, so, a count of *codepoints* does not guarantee a count of *glyphs*. This is mostly invisible in LGC, but may be an issue with more complex scripts.
* Moreover, a count of *glyphs* does not amount to a count of *columns* (and this may be an actual visible issue with the front code), as double-wide glyphs are definitely a thing (especially in CJK).

A completely fool-proof approach might want to combine this, after ensuring the input is normalized to a form where each codepoint is composed as a single glyph, and where we actually count the width of individual codepoints with `utf8proc_charwidth` ;).

poire-z:
I can't guess what's supposed to be displayed by https://github.com/koreader/koreader/pull/8271.
But I guess we're counting the nb of codepoints/glyphs so to add some amount of spaces so next content in each lines are aligned.
But will this work with the fact we don't use a monospace font ? Or are you using a monospace font ?

> A completely fool-proof approach might want to combine this, after ensuring the input is normalized to a form where each codepoint is composed as a single glyph, and where we actually count the width of individual codepoints 

And you're not even sure of that, it's all dependant on the font. So, probably a job for base/xtext.cpp.
But even then, you're not sure padding with spaces (`string.rep(" ", nb_spaces)`) will align them correctly, you'll probably have a few pixels shifts.
So, in the end, all this is probably better handled with our UI widgets, OverlapGroups, CenterContainer & co, with a TextWidget for each bit of text.

zwim:
For the desired application in koreader/koreader#8271 a monospace font is used.

poire-z:
Can you post a screenshot of the final desired result (the full screen) ?

zwim:
I will try it like you mentioned in https://github.com/koreader/koreader/pull/8271#issuecomment-932804732

But anyway this is the screenshot: (mind Bürgerlich)
grafik

ffi/utf8proc.lua
```diff
@@ -30,14 +30,19 @@ function Utf8Proc.lowercase(str)
...
-    local buffer = ffi.new("utf8proc_int32_t[1]")
+function Utf8Proc.count(str)
+    local str_p = ffi.cast("const utf8proc_uint8_t *", str)
+    local codepoint = ffi.new("utf8proc_int32_t[1]")
     local count = 0
     local pos = 0
-    while str_buf[pos] ~= 0 do
-        count = count + 1
-        pos = pos + libutf8proc.utf8proc_iterate(str_buf + pos, -1, buffer)
+    while str_p[pos] ~= 0 do
```
NiLuJe, line 38: I still don't like that as a loop condition ;p. (Lua strings can contain NULs).

zwim: Yes, but if I use
`while bytes > 0 do` I get random results for count. (Maybe depending on what is in the memory)
```
xxx str_len	18	uft8proc.count	35	text	Wahre Mitternacht:
xxx str_len	13	uft8proc.count	16	text	Astronomisch:
xxx str_len	9	uft8proc.count	36	text	Nautisch:
xxx str_len	11	uft8proc.count	35	text	Bürgerlich:
xxx str_len	14	uft8proc.count	36	text	Sonnenaufgang:
xxx str_len	14	uft8proc.count	16	text	Wahrer Mittag:
xxx str_len	16	uft8proc.count	44	text	Sonnenuntergang:
xxx str_len	11	uft8proc.count	35	text	Bürgerlich:
xxx str_len	9	uft8proc.count	36	text	Nautisch:
xxx str_len	13	uft8proc.count	16	text	Astronomisch:
xxx str_len	18	uft8proc.count	35	text	Wahre Mitternacht:
```

Correct results are calculated if I change the function to:
```
function Utf8Proc.count(str)
    local str_p = ffi.cast("const utf8proc_uint8_t *", str)
    local codepoint = ffi.new("utf8proc_int32_t[1]")
    local count = 0
    local pos = 0
    local bytes = 1
    local str_len = str:len()
    while pos < str_len and bytes > 0 do
        bytes = libutf8proc.utf8proc_iterate(str_p + pos, -1, codepoint)
        if bytes > 0 then
            count = count + 1
            pos = pos + bytes
        else
            return count, false
        end
    end
    return count
end
```
But this would mean to iterate to the string twice (str:len() and our loop). 

Would it be any good to use `while codepoint[0] ~= 0 and bytes > 0 do` or is this just the same as `str_p[pos] ~=0`?

NiLuJe (approved):

ffi/utf8proc.lua
```diff
@@ -35,16 +35,18 @@ function Utf8Proc.count(str)
     local codepoint = ffi.new("utf8proc_int32_t[1]")
     local count = 0
     local pos = 0
-    while str_p[pos] ~= 0 do
-        local bytes = libutf8proc.utf8proc_iterate(str_p + pos, -1, codepoint)
+    local bytes = 1
```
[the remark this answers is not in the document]
zwim: We don't need the `bytes > 0` check in the `while` condition, because we `return` in the else branch.

zwim:
Does anyone know what's the problem with CircleCI? 
```
FAILED: lua-Spore-prefix/src/lua-Spore-stamp/lua-Spore-download 
cd /home/ko/project/thirdparty/lua-Spore/build/x86_64-pc-linux-gnu/lua-Spore-prefix/src && /usr/bin/cmake -P /home/ko/project/thirdparty/lua-Spore/build/x86_64-pc-linux-gnu/git_checkout/tmp/lua-Spore-gitclone-tags/0.3.3.cmake && /usr/bin/cmake -E touch /home/ko/project/thirdparty/lua-Spore/build/x86_64-pc-linux-gnu/lua-Spore-prefix/src/lua-Spore-stamp/lua-Spore-download
Cloning into 'lua-Spore'...
fatal: unable to access 'https://framagit.org/fperrad/lua-Spore/': server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
Cloning into 'lua-Spore'...
fatal: unable to access 'https://framagit.org/fperrad/lua-Spore/': server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
Cloning into 'lua-Spore'...
fatal: unable to access 'https://framagit.org/fperrad/lua-Spore/': server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
-- Had to git clone more than once:
            3 times.
CMake Error at /home/ko/project/thirdparty/lua-Spore/build/x86_64-pc-linux-gnu/git_checkout/tmp/lua-Spore-gitclone-tags/0.3.3.cmake:66 (message):
  Failed to clone repository: 'https://framagit.org/fperrad/lua-Spore'


Makefile.third:678: recipe for target 'build/x86_64-pc-linux-gnu/rocks/lib/luarocks/rocks/lua-spore/0.3.1-1/lua-spore-0.3.1-1.rockspec' failed
make: *** [build/x86_64-pc-linux-gnu/rocks/lib/luarocks/rocks/lua-spore/0.3.1-1/lua-spore-0.3.1-1.rockspec] Error 1

Exited with code exit status 2


```

NiLuJe:
https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/

Frenzie:
Nevertheless I'm not sure I quite understand the question, since the error message is (somewhat deceptively perhaps) self-explanatory? ^_^

poire-z:
Is this PR obsolete/uneeded ?

zwim:
It is obsolete, as we currently dont need it. (but it should work)

So I let deccide you, how to proceed: merge ord drop.

Frenzie:
What do we use instead?

zwim:
In the beginnings of the fromdusktilldown plugin I used this for getting the length of a string. But after

https://github.com/koreader/koreader/blob/02ddd41b385977be9601322927c73e75d0f84f73/plugins/autowarmth.koplugin/main.lua#L778-L800

it is not necessary to know the strlen any more.

In any case I use a monospace font.

poire-z:
Let's have it even if not used, just to not waste the amount of brain power that was given above :)

PR #1415, opened 2021-10-02
thirdparty: fix build on openSUSE

cyphar:
On some distributions (mainly openSUSE), the host has a site-config
package which explicitly sets ${prefix}/lib64 as the libdir unless
explicitly modified. Because the build system relies on the libdir being
${prefix}/lib, it needs to be specified explicitly.

On systems without a site-package setup, this change is a no-op.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1415)


Frenzie (approved):
That's the full extent of it? I'm slightly surprised there wouldn't be more but I guess that's all the autoconf stuff. :-)

NiLuJe (approved):

yparitcher:
As you are a first time contributor, i manually approved the workflow so the CI runs.

cyphar:
> That's the full extent of it? I'm slightly surprised there wouldn't be more but I guess that's all the autoconf stuff. :-)

Well, I've got the emulator up and running without issues with this patch applied. I'm not sure if there's anything else I should test (maybe trying to build an APK or something), but building and running the emulator definitely works.

Frenzie:
> I'm not sure if there's anything else I should test (maybe trying to build an APK or something),

Yes, though probably more Kobo than Android.

PR #1419, opened 2021-10-09
thirdparty: sdcv: update to better support multi-word lookups

cyphar:
Current sdcv doesn't support multi-word lookups in the most ideal way.
In particular, as soon as any word specified in the command-line is
found to not have any dictionary results, it immediately bails. This
makes sdcv quite hard to use for Japanese deinflection and results in
less efficient batching when you have to look up many word candiates.

The patch applied is fairly trivial and was already submitted upstream
as a PR, so this patch can be dropped once it's merged upstream.

Ref: https://github.com/Dushistov/sdcv/pull/77
Needed by koreader/koreader#8312.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1419)


cyphar:
Maintainer said they will review it in the next week, so we can wait for a week and see if it's merged by then.

cyphar:
Merged upstream.

PR #1420, opened 2021-10-10
utf8proc: add utf8proc_NFC normalisation wrapper

cyphar:
Useful since Lua doesn't do this natively, and KOReader plugins might
need to be able to operate on utf8 strings using Lua's built in string
matching which won't match un-normalised strings.

Useful for koreader/koreader#8312.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1420)

### DIFF
--- a/ffi/utf8proc.lua
+++ b/ffi/utf8proc.lua
@@ -30,4 +30,11 @@ function Utf8Proc.lowercase(str)
     return folded_str
 end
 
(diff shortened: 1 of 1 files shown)


cyphar:
Converted to draft since I might need to add FFI wrappers for a few more utf8proc functions (`utf8proc_category` and some usage of `utf8proc_iterate`).

PR #1427, opened 2021-11-07
bump crengine: CSS fixes and optimisations

poire-z:
Includes https://github.com/koreader/crengine/pull/458 :
- lString32Collection::erase(): fix bounds checking
- ldomDocumentFragmentWriter: fix OnTagBody called twice
- CSS: optimize E.class rule check
- CSS: optimize E#id rule check
- EPUB: forward class attribute from the html element

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1427)


poire-z:
Stuff must have changed on the CircleCI checks (because we did something ? Or just changes on CircleCI sides ? and it now integrates a bit more into GitHub site).
I have here:
image

Previously, I think the ninja-debug one was always failing - but the other Kobo/Kindle.... checks were running nevertheless.

Anything one wants to check/re-run ? Or can I just merge this ?

Frenzie:
> Previously, I think the ninja-debug one was always failing - but the other Kobo/Kindle.... checks were running nevertheless.


You mean in this PR? It definitely wasn't always failing, and would've always blocked. :-)

Frenzie:
> Anything one wants to check/re-run ? Or can I just merge this ?

No, that's our main one. The clang thing isn't really used besides on MacOS and in the future on Android.

poire-z:
Oh, right. Didn't even look at the nija-debug build output :/
I think it may have been on TravisCI that there was one of the build items that was always failing (a more generic gcc or gcc-debut or clang-debug, not one of the device build).

Frenzie:
Are you on top of the latest commits in master here? HB hasn't changed in a month. O.o

(Sidenote: https://github.com/harfbuzz/harfbuzz/releases/tag/3.1.0 )

poire-z:
Guess I am on top of master. Didn't get the occasion to look at the output and see something related to HB, as you must have re-run it since, and now it is a success :/ so, well, good I guess :)

Frenzie:
Yes, I used rerun from failure. Very strange.

poire-z:
> (Sidenote: https://github.com/harfbuzz/harfbuzz/releases/tag/3.1.0 )

Good I did not bite on your hint :) We should wait for 3.1.1.

```
https://github.com/harfbuzz/harfbuzz/issues/3283
https://github.com/harfbuzz/harfbuzz/issues/3287
```

Frenzie:
Alright. :-)

PR #1428, opened 2021-11-08
Update button-listen for firmware update

### LINKED ISSUE #1426: button-listen doesn't seem to work on reMarkable 1
On the reMarkable 1 middle button launching of koreader seems to have stopped working. It appears to be linked to button-listen not working. Manually running "systemctl start koreader" works, but manually running button-listen and long-pressing buttons doesn't work.

(Tested on koreader v2021.10, and xochitl 2.10.0.324)

There is more detail at, though it seems like it might be the wrong project: https://github.com/koreader/koreader/issues/8242

klapaukh:
Suggestion to fix #1426 without breaking pre-firmware update compatibility.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1428)

### DIFF
--- a/button-listen.c
+++ b/button-listen.c
@@ -1,47 +1,47 @@
-#include <linux/input.h>
-#include <stdio.h>
-#include <time.h>
-#include <stdlib.h>
-#include <stdbool.h>
-#include <poll.h>
-#include <string.h>
 #include <errno.h>
+#include <linux/input.h>
+#include <poll.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
 
 static double timespec_diff(struct timespec const* start, struct timespec const* end)
 {
-    double s = difftime(end->tv_sec, (double)start->tv_sec);
-    double ns = (double)end->tv_nsec - (double)start->tv_nsec;
+    double s  = difftime(end->tv_sec, (double) start->tv_sec);
+    double ns = (double) end->tv_nsec - (double) start->tv_nsec;
     return s + (ns / 1e9);
 }
 
-static bool open_file(FILE **evf, struct pollfd *evf_poll, int file_idx, int expected_files, const char *filename)
+static bool open_file(FILE** evf, struct pollfd* evf_poll, int file_idx, int expected_files, const char* filename)
 {
     if (file_idx < 0 || file_idx >= expected_files) {
         fprintf(stderr, "Attempted to add file index %d, but only %d files are supported", file_idx, expected_files);
         return false;
     }
-    FILE *file = NULL;
-    int fd = -1;
-    file = fopen(filename, "re");
+    FILE* file = NULL;
+    int   fd   = -1;
+    file       = fopen(filename, "re");
     if (!file) {
         fprintf(stderr, "Could not open %s\n", filename);
         return false;
     }
-    fd = fileno(file);
-    evf[file_idx] = file;
+    fd                    = fileno(file);
+    evf[file_idx]         = file;
     evf_poll[file_idx].fd = fd;
     return true;
 }
 
 int main(int argc, char** argv)
 {
-    int num_input_files = 2;
-    FILE *evf[num_input_files];
+    int           num_input_files = 2;
+    FILE*         evf[num_input_files];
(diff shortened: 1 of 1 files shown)


klapaukh:
I am not sure why the build fails in the tests. The error appears to be in a file that I haven't changed at all.

Frenzie:
Could you please merge or rebase on top of master? See https://github.com/koreader/koreader-base/pull/1434

klapaukh:
@Frenzie  of course. Done.

button-listen.c
```diff
@@ -11,37 +14,77 @@ static double timespec_diff(struct timespec const* start, struct timespec const*
     return s + (ns / 1e9);
 }
 
+static bool open_file(FILE **evf, const char *filename)
+{
+    FILE *file = NULL;
+    int fd = -1;
+    file = fopen(filename, "rb");
```
NiLuJe, line 21: `"re"`, binary mode is meaningless on POSIX, and we usually prefer CLOEXEC by default (and there's no compatibility issue with that here).

button-listen.c
```diff
@@ -11,37 +14,77 @@ static double timespec_diff(struct timespec const* start, struct timespec const*
...
+    if(!file) {
```
NiLuJe, line 22: Consistent styling would be nice ;).

(Feel free to steal the input's clang-format, which I *think* I may have committed?).

button-listen.c
```diff
@@ -11,37 +14,77 @@ static double timespec_diff(struct timespec const* start, struct timespec const*
...
+        fprintf(stderr, "Could not open %s\n", filename);
+        return false;
+    }
+    fd = fileno(file);
+    if (fd < 0 || fd >= FD_SETSIZE) {
+        fprintf(stderr, "File descriptor for %s too large to handle (%d > %d)", filename, fd, FD_SETSIZE);
```
NiLuJe, line 28: Not: Not so much as large as too high for select to handle ;).

button-listen.c
```diff
@@ -11,37 +14,77 @@ static double timespec_diff(struct timespec const* start, struct timespec const*
...
+    }
+}
+
 int main(int argc, char** argv)
 {
-    FILE* evf = fopen("/dev/input/event2", "rb");
-    if(!evf) {
-        fprintf(stderr, "Could not open /dev/input/event2\n");
+    FILE *evf[FD_SETSIZE];
+    for (int i =0; i < FD_SETSIZE; i++){
```
NiLuJe, line 46: nit: style

button-listen.c
```diff
@@ -11,37 +14,77 @@ static double timespec_diff(struct timespec const* start, struct timespec const*
...
     struct input_event ev;
     struct timespec press_time = {0};
     bool pressed = false;
     while(1) {
-        size_t sz = fread(&ev, sizeof(ev), 1, evf);
-        if(sz == 0) {
+        fd_set evf_set;
+        FD_ZERO(&evf_set);
+        prep_evf_set(evf, &evf_set);
+        if(select(FD_SETSIZE, &evf_set, NULL, NULL, NULL) < 0){
```
NiLuJe, line 60: I'd prefer an actual computed `nfds` instead of just throwing `FD_SETSIZE` at `select`, but that's probably minor in the grand scheme of things ;).

(That, or just ditching select's clunky API and moving to poll ;p).

button-listen.c
klapaukh: Ah sorry. I had preserved that from the original.

button-listen.c
klapaukh: Good point. I'd not used `poll` before but that is **way** nicer.

button-listen.c
klapaukh: I didn't see a style config sorry. I've tried to tidy it up a bit.

klapaukh:
I think that should have addressed everything.

button-listen.c
NiLuJe: `input/.clang-format` ;). May or may not match the existing style, though, as it postdates this code, IIRC.

klapaukh: Cool. I've now run clang-format over it too.

NiLuJe (approved):

PR #1430, opened 2021-11-13
crengine: build with long file support

TFSThiagoBR98:
This allows to open files from rclone mounts in crengine

Closes koreader/koreader#7236

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1430)

### DIFF
--- a/thirdparty/kpvcrlib/CMakeLists.txt
+++ b/thirdparty/kpvcrlib/CMakeLists.txt
@@ -56,8 +56,19 @@ else()
     add_definitions(-DLINUX=1 -D_LINUX=1)
 endif()
 
+include(CheckTypeSize)
+
 if(DEFINED ENV{LEGACY})
     add_definitions(-DDISABLE_CLOEXEC)
+else()
+    # Test and Enable LFS Support in glibc, based on zlib
+    set(CMAKE_REQUIRED_DEFINITIONS -D_LARGEFILE64_SOURCE=1)
+    check_type_size(off64_t OFF64_T)
+    if(HAVE_OFF64_T)
+        add_definitions(-D_LARGEFILE64_SOURCE=1)
+        add_definitions(-DHAVE_STAT64=1)
+    endif()
+    set(CMAKE_REQUIRED_DEFINITIONS)
 endif()
 
 add_definitions(


Frenzie (commented):
Looks great! However, please stick it in https://github.com/koreader/koreader-base/blob/master/thirdparty/kpvcrlib/CMakeLists.txt instead? :-)

poire-z:
Any idea _why_ it magically helps with "rclone mounts" (whatever that is :) ?
And on which device/build do you need this ? Linux x64 desktop only ?

I'm a bit unconfortable with this - but I'm no C or low level expert, so dunno the implications, pinging @NiLuJe @Frenzie .

A few thoughts, from what I can grasp:

`-D_LARGEFILE64_SOURCE=1 -D_FILE_OFFSET_BITS=64` are not used by the crengine code - but they are used by its various shipped dependancies (sqlite, unrar, libpng) that KOReader doesn't use.
But we build some of these (sqlite, libpng) outside tree as dynamic libraries, and without any of these options. Building crengine with these may have #include headers for these libraries include different code/prototypes than what the dynamic libraries were build with. It feels this may cause issues.

`-DLVLONG_FILE_SUPPORT=1` will enable different bit of codes in lvstream.cpp. We only have ever tested it with it disabled (upstream CoolReader may have tested it with it enabled, but well, we may not use the same parts/features), so changing this will need testing.

`-DLVLONG_FILE_SUPPORT=1` will change:
```c
#if LVLONG_FILE_SUPPORT == 1
typedef   lUInt64   lvsize_t;    ///< file size type
typedef   lInt64    lvoffset_t;  ///< file offset type
typedef   lUInt64   lvpos_t;     ///< file position type
#else
typedef   lUInt32   lvsize_t;    ///< file size type
typedef   lInt32    lvoffset_t;  ///< file offset type
typedef   lUInt32   lvpos_t;     ///< file position type
#endif
```
lvsize_t is not only used in lvstream.cpp (that deals with file I/Os), but in various other bits.
I don't know how this can be bothering/slower on our ARM32 devices.
And if we wanted to enable it only on x64 linux builds, I would be unconfortable not having the same code running between our x64 emulator and our arm32 devices, it might complexify debugging.

cre.cpp
```diff
@@ -436,7 +436,7 @@ static int setCallback(lua_State *L) {
 
 static int initCache(lua_State *L) {
     const char *cache_path = luaL_checkstring(L, 1);
-    int cache_size = luaL_checkint(L, 2);
+    unsigned long long int cache_size = luaL_checklong(L, 2);
```
NiLuJe, line 439: Overkill, it's actually a macro that just does `(long) luaL_checkinteger`, and `lua_Integer`is `ptrdiff_t`, so this is never going to be 64 bits wide on 32 bits arches ;).

TL;DR: `long` ;).

NiLuJe:
I'd also echo @poire-z's worry, because this doesn't actually enable large file support outside of the C runtime, so an actual detailed use-case of why it failed before would be helpful, as w egenerally don't care about large files, because FAT32 ;).

NiLuJe:
Also, yeah, only building *some* of the dependencies with this enabled would be a hard veto, because here be dragons.

(And, again, we can't enable it on at least LEGACY platforms anyway).

TFSThiagoBR98:
> And on which device/build do you need this ? Linux x64 desktop only ?

Kindle PW4

> Any idea why it magically helps with "rclone mounts" (whatever that is :) ?

When I open an epub file i got this error "Cannot get file size for", after some debug, i get this [error](https://www.gnu.org/software/coreutils/faq/coreutils-faq.html#Value-too-large-for-defined-data-type), from fstat that indicates the crengine is not compiled with long files support (i don't know why need it to open an 18MB file)

NiLuJe:
An fstat on *what*, though? ;o).

(i.e., strace it, please).

TFSThiagoBR98:
> An fstat on _what_, though? ;o).
> 
> (i.e., strace it, please).

Here: https://github.com/koreader/crengine/blob/5facf9083d890c5f49b4d404b8c62366fe90df4d/crengine/src/lvstream.cpp#L1342

NiLuJe:
I don't see any reason why that would fail on a < 4GB file, though :?.

TFSThiagoBR98:
Yes, but for some reason it fails inside rclone mounts

NiLuJe:
That might actually be a fuse/rclone TC issue ;).

NiLuJe:
If you can, try building rclone with our own TC, it might magically behave (or horribly implode ;p).

TFSThiagoBR98:
@NiLuJe Recompiled with Kindle TC, it works, but i get same error when open this file with Cre

TFSThiagoBR98:
Also luafilesystem enable it by default [here](https://github.com/keplerproject/luafilesystem/blob/7c6e1b013caec0602ca4796df3b1d7253a2dd258/src/lfs.c#L26)

cre.cpp
TFSThiagoBR98: replaced with lvsize_t

TFSThiagoBR98:
@NiLuJe  I removed the DLVLONG_FILE_SUPPORT=1, and it also works

poire-z:
Can't we just try to do something in the if branch 
https://github.com/koreader/crengine/blob/5facf9083d890c5f49b4d404b8c62366fe90df4d/crengine/src/lvstream.cpp#L1342-L1345 when errno=EOVERFLOW (which I assume is what you get when you mentionned https://www.gnu.org/software/coreutils/faq/coreutils-faq.html#Value-too-large-for-defined-data-type).
Like hack a home made stat64 struct to receive the stuff (if that's even possible), and try again with it :)

```
       EOVERFLOW
              pathname or fd refers to a file whose  size,  inode  number,  or
              number  of  blocks  cannot  be represented in, respectively, the
              types off_t, ino_t, or blkcnt_t.  This error can occur when, for
              example,  an  application  compiled on a 32-bit platform without
              -D_FILE_OFFSET_BITS=64 calls stat() on a file whose size exceeds
              (1<<31)-1 bytes.
```

NiLuJe:
Hmm, possibly rclone and/or fuse prefers very large inode numbers for some reason :?.

We *might* be able to selectively use `fstat64` here, but I may be wrong ;o).

TFSThiagoBR98:
We can test if glibc support LFS and add the macros if possible

NiLuJe:
No we can't, because the legacy TC targets an earlier glibc version than it
is built against, so compile time headers are unreliable.

Frenzie:
> I'm a bit unconfortable with this - but I'm no C or low level expert, so dunno the implications, pinging @NiLuJe @Frenzie .

Neither am I I'm afraid. I don't understand what this could possibly change either (give or take very large files), and it's definitely hazardous. :-/

TFSThiagoBR98:
Ok, I've only enabled the stat64 to be avaliable. I'm now working on a patch to crengine to fallback to stat64 if the data is too big

NiLuJe:
I was going to suggest that whatever you might want to do would have to be done as a conditional patch based on !LEGACY, but you got the hang of it before I even made that clear ;o).

TFSThiagoBR98:
PR for cre: https://github.com/koreader/crengine/pull/459

thirdparty/kpvcrlib/CMakeLists.txt
```diff
@@ -58,6 +58,11 @@ endif()
 
 if(DEFINED ENV{LEGACY})
     add_definitions(-DDISABLE_CLOEXEC)
+else()
+    # Enable LFS Support in glibc
+    add_definitions(
+        -D_LARGEFILE64_SOURCE=1
+    )
```
poire-z, line 65: So, no `DISABLE_CLOEXEC` on non-legacy ? :)

I would also explicitely name this -D `HAS_STAT64` here and in cre if we don't need it for more things on the cre side - and to avoid any conflict with other bits of code (or even a `-D_CRE_CAN_USE_STAT64` to avoid more conflicts if HAS_STAT64 is too plain)

NiLuJe: Yeah, we *want* (and know that it's supported) CLOEXEC on other platforms ;).

TFSThiagoBR98:
Add an check method to see if is possible to enable LFS

NiLuJe (approved):
:lgtm:

thirdparty/kpvcrlib/CMakeLists.txt
```diff
@@ -56,8 +56,19 @@ else()
     add_definitions(-DLINUX=1 -D_LINUX=1)
 endif()
 
+include(CheckTypeSize)
+
 if(DEFINED ENV{LEGACY})
     add_definitions(-DDISABLE_CLOEXEC)
+else()
+    # Test and Enable LFS Support in glibc, based on zlib
```
poire-z, line 64: Still no -DDISABLE_CLOEXEC if !LEGACY

poire-z: Oups, I misread the original.

NiLuJe (approved):

poire-z:
I guess @NiLuJe will soon bump it into frontend with the other 3 updates waiting.

poire-z:
^ I might as well do these bumps now :)

Issue #1437, opened 2021-12-06
Clobbering of slots in MT protocol handling


mapinguari:
Hi all,

If a device has multiple input fds each of which is conforming to the MT B protocol you end up with clobbering of MT slots.

For example, if I have two touch devices as above.

fd 1 produces:
`ABS_MT_SLOT 0
ABS_MT_TRACKING_ID 45
ABS_MT_POSITION_X 0
ABS_MT_POSITION_Y 0
SYN_REPORT
`
then some time later produces

`ABS_MT_POSITION_X 100
ABS_MT_POSITION_Y 100
SYN_REPORT`

and fd2 produces:
`ABS_MT_SLOT 0
ABS_MT_TRACKING_ID 6
ABS_MT_POSITION_X 1000
ABS_MT_POSITION_Y 1000
SYN_REPORT`

From two distinct touch events, say, one from a finger and one from a pen tool then depending on the order of the reads from the fds you can end up with any permutation of touches at (0,0), (100,100) and (1000,1000).

The input functions need to make sure they are keeping track of which fd was the last to be read from, what the last slot it was producing events for was and if it is a different fd to ensure an ABS_MT_SLOT event for the current fd's last slot is inserted into the table of events being generated. It probably also needs to maintain a mapping between an individual fd's slot number and a "globally unique" slot number so you don't end up with clobbering of slots between different input devices.

I don't think this is a driver issue, each individual driver is doing the correct thing and respecting the MT protocol, it is just when you combine multiple input sources together into a combined stream you end up with unfortunate interlacing of the event streams.

I am not sure how much appetite you have to look at this tho. 

Any feedback apprecited!

NiLuJe:
It's a simplification, as this should be an exceedingly rare occurrence in practice.

(Also, we also clobber slots on some Kobo protocols later, anyway).

mapinguari:
Fair enough.

I will do by best to handle it further downstream. 

Want me to close this?

NiLuJe:
> Want me to close this?

Not necessarily, base's issue tracker isn't crowded, so leaving this open isn't really problematic.

As far as I'm concerned, this would be fairly low on the priority list, even if I actually had time to work on this (spoiler: I don't have *any* time ;p).

NiLuJe:
[...]

PR #1439, opened 2021-12-09
Clobbering of MT slots

mapinguari:
As per discussion on #1437.

I have made changes to return an fd on device open and to pass the fd that generated the event as part of the lua table returned.

Having made these changes I think it probably is unnecessary to pass the fd with EVERY event as ultimately the [main loop in waitForInput](https://github.com/koreader/koreader-base/blob/1432b9298e8a2ab00aaa2cc52ca14fa90665c70f/input/input.c#L360-L424) only returns a table of event tables for one device at a time, so could have waitForInput return a triple instead. hypothetically.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1439)

### DIFF
--- a/input/input.c
+++ b/input/input.c
@@ -150,7 +150,7 @@ static int openInputDevice(lua_State* L)
 
     // We're done w/ inputdevice, pop it
     lua_settop(L, 0);
-		lua_pushnumber(L,inputfds[fd_idx]);
+    lua_pushnumber(L, inputfds[fd_idx]);
 
     // Compute select's nfds argument.
     // That's not the actual number of fds in the set, like poll(),
@@ -204,8 +204,8 @@ static int closeInputDevices(lua_State* L __attribute__((unused)))
 static int fakeTapInput(lua_State* L)
 {
     const char* restrict inputdevice = luaL_checkstring(L, 1);
-    int                  x           = luaL_checkint(L, 2);
-    int                  y           = luaL_checkint(L, 3);
+    int x                            = luaL_checkint(L, 2);
+    int y                            = luaL_checkint(L, 3);
 
     int inputfd = open(inputdevice, O_WRONLY | O_NONBLOCK);
     if (inputfd == -1) {
@@ -288,8 +288,8 @@ static inline void set_event_table(lua_State* L, const struct input_event* input
     lua_rawset(L, -3);                 // ev.value = input.value
 
     lua_pushstring(L, "fd");
-    lua_pushinteger(L, fd);						 //	int 
-    lua_rawset(L, -3);                 // ev.fd = fd
+    lua_pushinteger(L, fd);  //	int
+    lua_rawset(L, -3);       // ev.fd = fd
 
     lua_pushstring(L, "time");
     // NOTE: This is TimeVal-like, but it doesn't feature its metatable!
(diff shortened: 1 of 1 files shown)


input/input.c
```diff
@@ -312,7 +317,7 @@ static inline size_t drain_input_queue(lua_State* L, struct input_event* input_q
 
     // Iterate over every input event in the queue buffer
     for (const struct input_event* event = input_queue; event < input_queue + ev_count; event++) {
-        set_event_table(L, event);  // Pushed a new ev table all filled up at the top of the stack (that's -1)
+        set_event_table(L, event,fd);  // Pushed a new ev table all filled up at the top of the stack (that's -1)
```
Frenzie, line 320: nit, I think clang-format will take care of it by itself?

input/input.c
```diff
@@ -406,7 +411,7 @@ static int waitForInput(lua_State* L)
 
                 if ((size_t) len == queue_available_size) {
                     // If we're out of buffer space in the queue, drain it *now*
-                    j = drain_input_queue(L, input_queue, ev_count, j);
+                    j = drain_input_queue(L, input_queue, ev_count,inputfds[i], j);
```
Frenzie, line 414: same nit

input/input.c
```diff
@@ -150,6 +150,7 @@ static int openInputDevice(lua_State* L)
 
     // We're done w/ inputdevice, pop it
     lua_settop(L, 0);
+		lua_pushnumber(L,inputfds[fd_idx]);
```
Frenzie, line 153: Double nit

Frenzie (approved):
lgtm besides the formatting nits (spaces for indentation, space after comma), but as I said, clang-format. ^_^

mapinguari:
ugh sorry. Ok, those should be updated

NiLuJe:
I'm gonna go with "Not good enough" for now, sorry. (This is just a quick veto to merging this as-is, until I have more than a shower's worth of time to think about it).

In decreasing order of importance:

* This breaks initial sttate handling (I allkuded to this in the other thread). The only way I could think of to avoid that would be to:
* Pass the same info along during the `openInputDevice` code, so that front's init has it on *init*, so it can tweak the initial state appropriately.
* This should probably be using a small-ish seed value in addition to the fd number, because that may be low enough to be lower than the maximum touch points of the input device, which could still leave room for conflicts.
* I like the idea of passing the info along in `w<aitForInput`... as long is it doesn't make the frontend code that much more horrible ;).
* I'd like to see the matching frontend changes before merging this ;).

NiLuJe (changes requested):
I did it wrong, see above ;p.

(This is just to generate a big red cross to avoid drive-by merges).

mapinguari:
Ok, points taken. 

I will close this for the moment on my end.

Frenzie:
> Pass the same info along during the openInputDevice code, so that front's init has it on init, so it can tweak the initial state appropriately.

I thought that was implied by this PR btw.

mapinguari:
I think my lack of clarity didn't aid this PR.

I was thinking that the fd would just be passed between the C and the lua so that frontend can deal with it as it sees appropriate. Not that the `ABS_MT_SLOT` values would be automatically converted to `ABS_MT_SLOT` + fd (this would obviously cause major differences with handling gestures and the inited state [not that I fully understand what they would be at this point :)]).

This confusion aside however I totally take that the lack of changes on the lua side makes this a bit futile at this point and adds a not insignificant amount to the size of the tables being sent back to lua. So I am happy for this to sit quietly until I have made some changes on the front that either make this PR worthwhile, or change the PR to reduce overhead and satisfy you guys and myself that it isn't going to cause some pernicious issues in the gesture detecting.

PR #1441, opened 2021-12-14
cre.cpp: fix isXPointerInCurrentPage()

poire-z:
Fix bookmark dogear not shown when revisiting some page when hanging punctuation is enabled and the first char of the page is hanging in the left margin. See https://github.com/koreader/koreader/pull/6456#issuecomment-993461335

Also add `cre.smoothScaleBlitBuffer()` to allow benchmarking BB scaling methods on devices.
See https://github.com/koreader/koreader/pull/3905#issuecomment-989789750

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1441)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -3484,7 +3484,11 @@ static int isXPointerInCurrentPage(lua_State *L) {
 	doc->dom_doc->createXPointer(lString32(xpointer)).getRect(xp_rect);
 	//CRLog::trace("page range: %d,%d - %d,%d", page_rect.left, page_rect.top, page_rect.right, page_rect.bottom);
 	//CRLog::trace("xp range: %d,%d - %d,%d", xp_rect.left, xp_rect.top, xp_rect.right, xp_rect.bottom);
-	lua_pushboolean(L, page_rect.isRectInside(xp_rect));
+	// lua_pushboolean(L, page_rect.isRectInside(xp_rect));
+	// Just check that the xpointer rect intersects with the page rect on the y-axis
+	// (as on the x-axis it may not, with hanging punctuation or negative margins)
+	bool res = (xp_rect.bottom > page_rect.top) && (xp_rect.top < page_rect.bottom);
+	lua_pushboolean(L, res);
 	return 1;
 }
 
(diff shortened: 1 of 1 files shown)


NiLuJe:
Possibly needs doc/a matching destructor, because it's not necessarily clear that the returned pointer is allocated and needs to be freed by the caller.

(That, or make it return an actual KO BlitBuffer flagged as allocated and stuff).

(I'll take a "don't use this" comment if this is just for testing purposes).

poire-z:
I can add:
```c
/* This was added for testing and benchmarking purpose: do not use it.
 * Note that the returned pointer 'data' is allocated and needs to freed */
static int smoothScaleBlitBuffer(lua_State *L) {
```

Thought about having it returned a BB, but I don't know if that's possible with just blitbuffer.h included and having just this available:
https://github.com/koreader/koreader-base/blob/2e53592573718a59399df1739a92b39cf8d92e2d/blitbuffer.h#L124-L131
blitbuffer.lua .new() return this, or rather this possibly wrapped by metatype and Lua stuff that I'm not familiar with :)
Obviously, such stuff, if needed, shouldn't be done in the plain cre.cpp, but possibly by some other ffi wrapper, that I'd rather not add, or make cre.cpp dependant on some other stuff.

So, well, unless you see an obvious cheap way to return a classic usable Lua BB, I'll go with the comment :)

cre.cpp
```diff
@@ -3545,6 +3549,19 @@ static int getImageDataFromPosition(lua_State *L) {
     return 0;
 }
 
+static int smoothScaleBlitBuffer(lua_State *L) {
```
Frenzie, line 3552: Besides a comment you could also consider giving it a less safe name? (Just a thought; I'm fine with just the comment.)

NiLuJe:
Comment's fine with me, yeah ;).

(I had a vague memory *something* might have been doing BBs-in-C-Lua, but, no, I was probably thinking of the C blitter, and that's FFI).

cre.cpp
poire-z: (Well, with a less safe name, if we end up one day using it :) I'd have to make a PR here to rename it. With this safe pretty name, we can just use it and pretend I forgot to mark that comment as obsolete :)

PR #1375, opened 2021-05-24
[CI] Update to Ubuntu 18.04 Docker images

Frenzie:
See https://github.com/koreader/virdevenv/pull/57 and https://github.com/koreader/virdevenv/pull/62

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1375)

PR #1442, opened 2021-12-27
[CI] Github Action for lint, build and tests

yparitcher:
There is a job for linters, a matrix for the emulator builds, and a matrix for cross compile

Fix macos build based on koreader/koreader#8525

Android Builds have not be implemented yet. (I still have to figure out what the toolchain requirements are vs what GHA provides)

~~This still has some work to be done to deduplicate the dependency install scripts with front and virdevenv~~
Requires: koreader/virdevenv#71

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1442)

### DIFF
--- a/.ci/test_script.sh
+++ b/.ci/test_script.sh
@@ -10,7 +10,6 @@ if [ "$TARGET" = "android" ]; then
         exit 1
     fi
 elif [ "$EMULATE_READER" = "1" ]; then
-    cp build/*/luajit "${HOME}/.luarocks/bin"
     # install tesseract trained language data for testing OCR functionality
     travis_retry wget https://src.fedoraproject.org/repo/pkgs/tesseract/tesseract-ocr-3.02.eng.tar.gz/3562250fe6f4e76229a329166b8ae853/tesseract-ocr-3.02.eng.tar.gz
     tar zxf tesseract-ocr-3.02.eng.tar.gz
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,20 +1,14 @@
-name: test
+name: macos
 
 on: [push, pull_request]
 
 jobs:
   build:
-    runs-on: ${{ matrix.os }}
-
-    strategy:
-      matrix:
-        os: [macos-latest]
+    runs-on: macos-11
 
     steps:
-      - name: setup-xcode
-        uses: maxim-lobanov/setup-xcode@v1
-        with:
-          xcode-version: 11.3.1 # Use Xcode 11.3.1, because 11.4, 11.5, and 11.6 have a segfaulting clang.
+      - name: XCode version
+        run: xcode-select -p
 
       - name: Check out Git repository
         uses: actions/checkout@v2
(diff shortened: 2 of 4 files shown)


poire-z (commented):
(I unfortunately can't comment on CI/build stuff, this all looks like LTR Hebrew to me :)

yparitcher:
Ready for merge

Frenzie:
NB I stupidly just merged this because it all looks fine and I didn't have time to look at it or think about it all week but this is essentially reverting to the pre-Docker days, which is imo not particularly desirable and also unnecessary in base because there are no mysterious segfaults in Docker.

Frenzie:
That is to say, it's nice as an experiment but I'm not fond of this approach because it triples my workload again.

Frenzie:
PS But also thanks a lot for looking into this stuff! It's just not here that anything is problematic. ^_^

yparitcher:
> PS But also thanks a lot for looking into this stuff! It's just not here that anything is problematic. ^_^

Phew :) I was getting a little worried you were totally against GHA. I had assumed that the CI in base was also giving issues.

Btw, if there is any interest now that all the pieces are in place (toolchains etc) it should not be so hard to make a gha to build a nightly on each commit to master (or every day) and/or to automatically build a release when a new git tag is made.

Frenzie:
> I was getting a little worried you were totally against GHA. 

I'm certainly not in favor of it. The only reason we want GHA is because CircleCI has developed annoying authorization-related issues when it should just run on every PR. Its horribly convoluted yet straitjacketed "actions" are atrociously complex compared to Travis CI, CirlceCI and GitLab CI. Simple things are made very, very hard. It's theoretically available if you parse some JSON API thing buried fifty million documentation pages away, when in Travis CI you could set up the same thing in like 5 minutes.

The CircleCI thing only an issue on the repo which requires authorization in the first place (i.e., if you want to remove the docs & translations from the CI then GHA is absolutely not necessary and I'd miss it like a toothache; not using secrets would *fix CircleCI*).

> I had assumed that the CI in base was also giving issues.

No, the abandoned PR in front would've never happened if these hadn't gone through as expected. See https://github.com/koreader/koreader-base/pull/1375

That was base first to work out the kinks, then GitLab CI nightlies, then front. Except front suddenly suffered from mysterious segfaults that didn't happen with the Ubuntu 16.04 images.

> Btw, if there is any interest now that all the pieces are in place (toolchains etc) it should not be so hard to make a gha to build a nightly on each commit to master (or every day) and/or to automatically build a release when a new git tag is made.

Each commit seems rather excessive. Building a new release every day and/or on a git tag could be interesting.

yparitcher:
> GHA's horribly convoluted yet straitjacketed "actions" are atrociously complex compared to Travis CI, CircleCI and GitLab CI, among others. Simple things are made very, very hard. It's theoretically available if you parse some JSON API thing buried fifty million documentation pages away, when in Travis CI you could set up the same thing in like 5 minutes.

:) very true.

I was annoyed by CircleCi refusal to allow be to se what went wrong withoug giving them access to my github account. However since the CirclCI check intergration was added we can now see all that info without logging in. 

> I.e., if you want to remove the docs & translations from the CI then GHA is absolutely not necessary and I'd miss it like a toothache; not using secrets would _fix CircleCI_. Doing that was always my plan B.

I might be able to make a separate gha that runs on master that triggers docs and translations, and the rest can be left to CircleCI

Frenzie:
I have no particular attachment to CircleCI to be clear. I like the setup you proposed in front very well since unfortunately we'd have to do with the same thing in CircleCI anyway.

Frenzie:
I suspect we'd also have to do the same thing in GitLab CI, since it can't be a coincidence that CircleCI and GHA both show the same weird segfaults problem that I can't reproduce with my local Docker.

Frenzie:
But anyway, the primary issue with CircleCI is the authorization thing. The segfaults are easily avoided, even though I don't really want to avoid Docker. :-)

PR #1443, opened 2022-01-05
For the upcoming terminal emulator

zwim:
This PR is for the upcoming KOTerm (a terminal emulator), which will be a full-blown vt52 terminal enriched with vt102-ansi ['cause some busybox applets won't understand the basic basics of at terminal (vt52 lives for ever!) and mksh (on Android is a bit bitchy)]  ;)

Now testing the KOTerm terminal emulator 
* in emulator (an emulated emulator so to say): looks very good
* on android a bit wonky: will need help later
* on kobo: looks very good

Draft by now, because the frontend is not quite finished yet. Stay tuned ... 

To the internals: I have put everything --- for the sake of simpicity --- in ffi*/posix*. Actually some defines are in `stdlib.h` and `unistd.h`, but I hope this won't hurt.
 
Nothing to do for now until the frontend will be ready .........

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1443)


ffi/posix_h.lua
```diff
@@ -53,6 +53,10 @@ ssize_t mq_receive(int, char *, size_t, unsigned int *);
 int mq_close(int) __attribute__((nothrow, leaf));
 int close(int);
 int fcntl(int, int, ...);
+int grantpt(int fd);
+int unlockpt(int fd);
+char *ptsname(int fd);
+int setsid(void);
```
Frenzie, line 59: That's a very different order. ;-)

Frenzie:
That sounds like a very imposing endeavor. I wish you luck!

zwim:
> That sounds like a very imposing endeavor. I wish you luck!

Thank's. Everyone needs luck now, especially in this special times.

Back to our case, I knew what to expect. I almost got this thing going (for a week by now). What works by now is in emulator (bash, busybox ash, mksh), kobo (busybox ash), andorid (mksh).
In all of them "normal" commands like pwd, date, ls, ls -al ...work.
In all of them: nano, vi, top, more and less (less is a bit bitchy on Android) work.

So some work has to be done the next days.

ffi/posix_h.lua
zwim: What are the/your concerns about the order? I have just added what I needed.
We can sort them anyway. What  is a better proposal?

Frenzie: I don't know what the order is in the original. I would normally propose to follow that unless it seems grossly illogical for our purposes.

But my point here was that the addition in posix_decl.c should be on line 78 to end up with this result, or conversely that this should be quite a bit further down.

However, as you mentioned this file feels like you gave it more thought while posix_decl.c feels like it's randomly at the bottom, so logic suggests to adjust posix_decl.c. In any case, they should be the same.

Frenzie: And if the lines don't match up properly then you can't really spot potential mistakes from such an automated result in the diff because nothing matches. :-)

Frenzie: Some historical references: https://github.com/koreader/koreader/pull/4631

zwim:
Not necessary any more, as the ffi.cdecl is done in the emulator plugin

PR #1444, opened 2022-01-16
cre.getTextFromPositions(): fix selection of single CJK char

poire-z:
See https://github.com/koreader/koreader/issues/8619#issuecomment-1012506110.
Should allow closing https://github.com/koreader/koreader/issues/8619.

Also:
`FFIutil.runInSubProcess(): return error string on failure`
Couldn't get any failure, but might be useful.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1444)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -1904,13 +1904,46 @@ static int getTextFromPositions(lua_State *L) {
 		if (r.isNull())
 			return 0;
 
+		// A xpointer references a char in a text node; a char has a width, and getNodeByPoint()
+		// has a trick of returning a xpointer to this char when the point is in the left half of
+		// the glyph, or to the next char when the point is in the right half of the glyph, which
+		// feels clumsy but has the effect that it doesn't need to know nor care about whether the
+		// point references the start or the end of the selection.
+		// This is usually hidden by the above code that extends the selection to start and end
+		// of words. But with CJK chars, where each char is a word, this behaviour is noticable.
+		// It's usually OK with multiple chars selection, even if one needs to go over the further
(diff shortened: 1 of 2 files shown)


cre.cpp
```diff
@@ -1904,13 +1904,46 @@ static int getTextFromPositions(lua_State *L) {
 		if (r.isNull())
 			return 0;
 
+		// A xpointer references a char in a text node; a char has a width, and getNodeByPoint()
+		// has a trick of returning a xpointer to this char when the point is in the left half of
+		// the glyph, or to the next char when the point is in the right half of the glyph, which
+		// feels clumsy but has the effect that it doesn't need to know and care about whether the
```
Frenzie, line 1910:
```suggestion
		// feels clumsy but has the effect that it doesn't need to know nor care about whether the
```

cre.cpp
```diff
@@ -1904,13 +1904,46 @@ static int getTextFromPositions(lua_State *L) {
...
+		// point references the start or the end of the selection.
+		// This is usually hidden by the above code that extends the selection to start and end
+		// of words. But with CJK chars, where each char is a word, this behaviour is noticable.
+		// It's usually OK with multiple chars selection, even if one needs to go over the further
+		// 2nd half of the glyph to have it included in the selection.
+		// But it is less OK with an initial long-press on a glyph, where we get 50% chances of
```
Frenzie, line 1916:
```suggestion
		// But it is less OK with an initial long-press on a glyph, where we get a 50% chance of
```

Frenzie (approved):
I've got nothing to say so I left a couple of silly comments. ;-)

PR #1450, opened 2022-01-25
mupdf: strikeout highlight style, line thickness

hius07:
Unlock new strikeout highlight style.
Reduce underline/strikeout line thickness, to look similar to KOReader highlight style.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1450)

### DIFF
--- a/ffi/mupdf.lua
+++ b/ffi/mupdf.lua
@@ -665,7 +665,7 @@ end
 
 mupdf.STRIKE_HEIGHT = 0.375
 mupdf.UNDERLINE_HEIGHT = 0.075
-mupdf.LINE_THICKNESS = 0.07
+mupdf.LINE_THICKNESS = 0.05
 
 function page_mt.__index:addMarkupAnnotation(points, n, type)
     local color = ffi.new("float[3]")
@@ -683,7 +683,7 @@ function page_mt.__index:addMarkupAnnotation(points, n, type)
         color[2] = 1.0
         line_thickness = mupdf.LINE_THICKNESS
         line_height = mupdf.UNDERLINE_HEIGHT
-    elseif type == M.PDF_ANNOT_STRIKEOUT then
+    elseif type == M.PDF_ANNOT_STRIKE_OUT then
         color[0] = 1.0
         color[1] = 0.0
         color[2] = 0.0
--- a/ffi/mupdf.lua
+++ b/ffi/mupdf.lua
@@ -664,7 +664,7 @@ function page_mt.__index:draw_new(draw_context, width, height, offset_x, offset_
 end
 
 mupdf.STRIKE_HEIGHT = 0.375
-mupdf.UNDERLINE_HEIGHT = 0.01
+mupdf.UNDERLINE_HEIGHT = 0
 mupdf.LINE_THICKNESS = 0.05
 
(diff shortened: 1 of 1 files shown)


ffi/mupdf.lua
```diff
@@ -664,7 +664,7 @@ function page_mt.__index:draw_new(draw_context, width, height, offset_x, offset_
 end
 
 mupdf.STRIKE_HEIGHT = 0.375
-mupdf.UNDERLINE_HEIGHT = 0.075
+mupdf.UNDERLINE_HEIGHT = 0.01
```
poire-z, line 667: That's quite a difference :|
What does this value mean exactly, if you figured it out ?
But as long as it looks OK, I don't mind not knowing :)

hius07: It moves the line up and down

hius07: 0.2

poire-z: > 0.2

What is the black and what the grey line ?

What is the reference ? Where it is drawn with mupdf.UNDERLINE_HEIGHT = 0 ?

hius07: 0.05

hius07: > What is the black and what the grey line ?

Black - mupdf, grey - our highlight.

poire-z: Sorry, still unclear: is that MuPDF only or both KOReader + MuPDF drawings ?
Why it feels there is a black line over a gray line with 0.01 and 0.05 ? KOReader dimming some part of MuPDF highlight making it blacker ? Or MuPDF doing it black, but KOReader drawing some gray over it or part of it ?

I'm unsure if 0 feels like it's the bottom of the line box ?
mupdf.UNDERLINE_HEIGHT is really a delta y and not a line height ?

(But 0 looks better no ? only gray, vs bits of black with 0.01 and 0.05).

And just curious: what would be drawn with mupdf.UNDERLINE_HEIGHT = 1 ?

hius07: > MuPDF doing it black, but KOReader drawing some gray over it or part of it ?

MuPDF draws saved (in the document) highlight in black. KOReader draws our highlight in grey.

0.5


1 (the same highlight!)

hius07: Yes, I think it can be just 0.

poire-z: OK, so it's a delta y, 0 is the bottom of the line box, and 1 is the top of the line box.

> Yes, I think it can be just 0.

Was going to ask why you thought 0.1 was better than 0:

image

I guess no black means they do superpose, which is what you wanted :) no ?

poire-z: (But may be with other PDF document with larger line heights, 0 would be the bottom of the taller line box ? No idea :/)
Anyway, let's go with 0 and see then.

poire-z: I think 0 is fine. With some larger interline PDF, with 0:
image

0 is the bottom of the text box, and I guess in PDFs (unline crengine), we have text boxes absolutely positionned, and it's not the previous line box height that positions the next line box.

poire-z:
Let's go with 0 ? Should I merge and bump ?

hius07:
Yes please.

PR #1451, opened 2022-01-25
PocketBook: Handle rendering of interface and books directly with inkview

rjd22:
Needs also the following PR: https://github.com/koreader/koreader/pull/8727

I've been emailing about color support for the pocketbooks directly with Pocketbook. The awesome folks of pocketbook replied with the implementation of rendering with the native inkview library. I've tested this with my Pocketbook 741 and the results are very promising.

This is their answer:

> Support for PB741 (InkPad Color)
> https://github.com/koreader/koreader/issues/7406 
> 
> There are two types of color screens used by PocketBook: Kaleido and Kaleido plus. These two types have slightly different color filter patterns. Thus we need to use different algorithms to display colors. All this functionality is already implemented in inkview library which is already used by KoReader. The easiest way to have colors in your application is not to forward the image directly to framebuffer, but to do it through inkview lib.
​
> Please find attached an example how to do it. You need to unpack archive and replace files of KoReader. We checked on version 2022.01.
> 
> Best regards,
> Customer Support Team

This PR is their implementation. All the credits go to the awesome PocketBook developer that took the time to write this.

Of course this also needs to be tested on other pocketbook devices. But I hope this will solve some of rendering issues with the pocketbooks.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1451)

### DIFF
--- a/ffi/framebuffer_pocketbook.lua
+++ b/ffi/framebuffer_pocketbook.lua
@@ -58,7 +58,7 @@ function framebuffer:init()
         rotate = vinfo.rotate,
         width_mm = vinfo.width,
         height_mm = vinfo.height,
-  })
+    })
 end
 
 function framebuffer:close(reinit)
--- a/ffi/framebuffer_pocketbook.lua
+++ b/ffi/framebuffer_pocketbook.lua
@@ -1,0 +1,112 @@
+local ffi = require("ffi")
+local BB = require("ffi/blitbuffer")
+local inkview = ffi.load("inkview")
+
+require("ffi/inkview_h")
+require("ffi/linux_fb_h")
+
+local framebuffer = {
+
+}
+
+function framebuffer:init()
+    self._finfo = ffi.new("struct fb_fix_screeninfo")
+    self._vinfo = ffi.new("struct fb_var_screeninfo")
+
+    local finfo = self._finfo
+    local vinfo = self._vinfo
+
+    self.debug("Initialize inkview screen")
+    inkview.OpenScreen()
+
+    local pb_fb = inkview.GetTaskFramebuffer(inkview.GetCurrentTask())
+
+    self._finfo.line_length = pb_fb.scanline
+    self._vinfo.xres = pb_fb.width
+    self._vinfo.yres = pb_fb.height
+    self.fb_size = self._finfo.line_length * self._vinfo.yres
+
+    self._vinfo.width = pb_fb.width
+    self._vinfo.height = pb_fb.height
+
+
+    local bpp = pb_fb.depth
+
+    self.data = pb_fb.addr
+    self.bb = BB.new(pb_fb.width, pb_fb.height, BB["TYPE_BB"..bpp] or BB["TYPE_BBRGB"..bpp], self.data, pb_fb.scanline, pb_fb.width)
+
+    self.blitbuffer_rotation_mode = self.bb:getRotation()
+    self.screen_size = self:getRawSize()
+    self.native_rotation_mode = self.forced_rotation and self.forced_rotation.default or self.ORIENTATION_PORTRAIT
+    self.cur_rotation_mode = self.native_rotation_mode
+
+
+    self.debug("FB info (post fixup)", {
+        fb_size = self.fb_size,
+        xres = vinfo.xres,
+        yres = vinfo.yres,
+        xoffset = vinfo.xoffset,
+        yoffset = vinfo.yoffset,
+        bpp = bpp,
+        xres_virtual = vinfo.xres,
+        yres_virtual = vinfo.yres,
+        line_length = finfo.line_length,
+        stride_pixels = finfo.line_length,
+        smem_len = finfo.smem_len,
+        type = finfo.type,
+        mmio_len = finfo.mmio_len,
+        rotate = vinfo.rotate,
+        width_mm = vinfo.width,
+        height_mm = vinfo.height,
+  })
+end
+
+function framebuffer:close(reinit)
+    if self.bb ~= nil then
+        self.bb:free()
+        self.bb = nil
+    end
+end
+
+--[[ framebuffer API ]]--
+
+function framebuffer:refreshPartialImp(x, y, w, h, dither)
+    self.debug("refresh: inkview partial", x, y, w, h)
+    inkview.PartialUpdate(x, y, w, h)
+end
+
+function framebuffer:refreshFlashPartialImp(x, y, w, h, dither)
+    self.debug("refresh: inkview partial", x, y, w, h)
+    inkview.PartialUpdate(x, y, w, h)
+end
+
+function framebuffer:refreshUIImp(x, y, w, h, dither)
+    self.debug("refresh: inkview partial", x, y, w, h)
+    inkview.PartialUpdate(x, y, w, h)
+end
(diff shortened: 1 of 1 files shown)


ffi/framebuffer_pocketbook.lua
```diff
@@ -0,0 +1,112 @@
...
+        yres_virtual = vinfo.yres,
+        line_length = finfo.line_length,
+        stride_pixels = finfo.line_length,
+        smem_len = finfo.smem_len,
+        type = finfo.type,
+        mmio_len = finfo.mmio_len,
+        rotate = vinfo.rotate,
+        width_mm = vinfo.width,
+        height_mm = vinfo.height,
+  })
```
Frenzie, line 61: nit

Frenzie (commented):
Pinging @NiLuJe; looks quite sane to me.

NiLuJe:
Quick skim on mobile, we obviously lose some control, but gain not having to care about device specific quirks ;).

pazos:
@NiLuJe: I guess the cleanup of https://github.com/koreader/koreader-base/blob/master/ffi/framebuffer_mxcfb.lua can wait a bit and be merged after this PR. Is that true?

NiLuJe:
@pazos: Err, which cleanup? ^^.

pazos:
> @pazos: Err, which cleanup? ^^.

all pocketbook related code that no longer used after this PR ;)

NiLuJe:
Oh, I'd definitely like to keep it, because it might be interesting to either be able to make that configurable, or just that we might actually need it on some devices ;).

(Also, just to document PB's weird mxcfb-on-top-of-sunxi shim ;)).

rjd22:
> Oh, I'd definitely like to keep it, because it might be interesting to either be able to make that configurable, or just that we might still need it on some devices ;).
> 
> (Also, just to document PB's weird mxcfb-on-top-of-sunxi shim ;)).

My suggestion is to still remove it. Pocketbooks are so different each iteration that the code can easily become a lie.

Also having multiple ways of rendering will confuse other new maintainers.

I would remove it in a separate PR tho. This way we will be able to find it back when the need arises.

NiLuJe:
Fair enough ;).

(there's likely more cruft in framebuffer_linux than framebuffer_mxcfb, FWIW).

rjd22:
@NiLuJe how would you approach releasing this change? I don't have any other PB devices to test this on.

One way it merge and release and see if the issue tickets stream in 🙈. Another way would be to somehow find some people with older devices.

poire-z:
May be make a single zip with just your changed files that PB users could unzip on their device to get it all (they would reset it by just unzipping a nightly or stable release). (If unzipping is as easy on PB as it is on Kobo).
And summon/ping by nickname the PB users that took part in some recent PB issues for them to test that zip.

rjd22:
@poire-z good Idea! I have a zip that you indeed can apply over the application on the device. Write a request in the koreader/koreader PR.

hombretranquilo:
I'm running a Koreader version 2022.01-46 with a modified device.lua file according to #8727, and I also created a framebuffer_pocketbook.lua file according to this PR. I believe that nothing more needs to be done to test. Right? 

After a couple of hours of testing, there is no crashes or other unwanted effects. It looks like keyboard responsiveness is improved in a really impressive way (no delay at all between taps).

And the more interesting thing: this PR solves #8187. Now I can use again KOreader on my InkPad3 with new firmware 6.5.1379. Great!!!!

As soon as I can, I will try to test the modifications suggested in #8730 #8729.

Huge thanks @rjd22

rjd22:
@NiLuJe @poire-z This has been tested on quite some devices and I think it is okay to merge this as is. I might adjust the `FullUpdate` with `SoftUpdate` later because it improves the detail of the render one some devices but am waiting to hear back from PocketBook what the disadvantages are (since I guess the had a good reason to choose FullUpdate).

Is that ok with you?

poire-z:
You can now bump base in your frontend PR.

rjd22:
@poire-z I've updated the frontend PR :)

poire-z:
(Mentionning this here to not pollute the frontend PR with git technicalities :)
You might need to rebase your PR to current master. For now, your branch says:
`This branch is 2 commits ahead, 5 commits behind koreader:master. `
So, you might bring stuff that is already merged - which might not be an issue, but for safety, please rebase to master so your PR shows only what has changed.

rjd22:
@poire-z Yep, I saw it has merge conflicts. I rebased it on the latest master now and it should be better.

poire-z:
Yep, looks clean now !

NiLuJe:
I'm also mildly curious as to how this interacts with @ezdiy's hwrota stuff on the few devices where that's supported.... (I'm guessing badly ;p).

(There's also the waveform level stuff that should probably go).

Biep:
Does this mean KOReader will work well, with colour, on the PocketBook InkPad Color 3?
My Kobo Aura One is breaking down (battery, and the Micro-USB port no longer connects unless I torque the cable strongly) and I am looking around for a next reader.

Willing to test for you on whatever machine I'll end up buying - but hesitate to spend an extra € 100+ unless I know I can uae the colour feature on KOReader.

Frenzie:
I think @ElimGarak1 should be able to say more about what it runs and looks like.

NiLuJe:
I would strongly recommend *against* a PB, especially if you have easy access to a Kobo, but that's just me ;).

Frenzie:
There'll probably be some interesting Kobo devices this year, although presumably not with color.

ElimGarak1:
> Does this mean KOReader will work well, with colour, on the PocketBook InkPad Color 3?

IMG_20231113_221533_073

At least not bad ;).

Biep:
Looks decent.

Biep:
> I would strongly recommend _against_ a PB, especially if you have easy access to a Kobo, but that's just me ;).

Why that?  My Kobo Aura One is running toward its end - the battery deteriorating, and most importantly - the Micro-USB port becoming unfunctional.  I have one single cable that can still make contact - if I twist it enough to exert considerable torque.

For non-colour, the PocketBook Era Sunset Copper looks very attractive - and seeing them actively helping you guys getting KOReader running on it makes me more positive.  So I'd be interested in hearing your negatives.

NiLuJe:
c.f., https://github.com/koreader/koreader/issues/11338

Biep:
Thanks.  Somewhere (I can't find it back now) there was mention of the problem that the CPU went in sleep mode after a few seconds, breaking some `sleep` calls.  If this happens after more than a given time (say, 1 sec), then a cheap solution would be a soft sleep: if the sleep time is >1s, sleep 1s and recurse, otherwise sleep the requested amount.
But I probably misunderstand the issue.

Still falling for the 64G of the PocketBook Era Copper Sunset - though that name is somewhat offputting.

rjd22:
> Yes, autostandby is technically broken, too.

This actually can be fixed by not using the pocketbook auto-standby. I still need to make time to send a PR to fix it ^^°.

Biep:
I just ordered a PocketBook Era, so obviously any fix or enhancement to KOReader that improves its behaviour on that machine will incur an immense gratefulness from my side.  😉

PR #1453, opened 2022-02-12
Handle HW nightmode with inkview

rjd22:
This implementation enables night mode with inkview. Opening pictures in night mode will result in inverted pictures. But reading works fine.

IMG_20220212_192513
IMG_20220212_192523
IMG_20220212_192546

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1453)

### DIFF
--- a/ffi/framebuffer_pocketbook.lua
+++ b/ffi/framebuffer_pocketbook.lua
@@ -9,6 +9,33 @@ local framebuffer = {
 
 }
 
+local function _toggleNightModeWhenEnabled(fb, x, y, w, h)
+    if fb.night_mode and fb.device:canHWInvert() then
+        inkview.InvertArea(x, y, w, h);
+    end
+end
+
+local function _updatePartial(fb, x, y, w, h)
+    fb.debug("refresh: inkview partial", x, y, w, h)
+
+    _toggleNightModeWhenEnabled(fb, x, y, w, h)
+    inkview.PartialUpdate(x, y, w, h)
+end
+
+local function _updateFull(fb, x, y, w, h)
+    fb.debug("refresh: inkview full", x, y, w, h)
+
+    _toggleNightModeWhenEnabled(fb, x, y, w, h)
+    inkview.FullUpdate()
+end
+
+local function _updateFast(fb, x, y, w, h)
+    fb.debug("refresh: inkview fast", x, y, w, h)
+
+    _toggleNightModeWhenEnabled(fb, x, y, w, h)
+    inkview.DynamicUpdate(x, y, w, h)
+end
+
 function framebuffer:init()
     self._finfo = ffi.new("struct fb_fix_screeninfo")
     self._vinfo = ffi.new("struct fb_var_screeninfo")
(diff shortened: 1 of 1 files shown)


Frenzie:
The images shouldn't be inverted (or rather they should be pre-inverted and thereby set right). Did you test this by simply forcing framebuffer.night_mode to true without the associated frontend handling? Because otherwise something's broken.

rjd22:
@Frenzie I tested it by enabling night mode in the front-end and opening the image. So it might be that full page images are not pre-inverted. It seems that the images in the list are shown properly.

Frenzie:
So you're not talking about a unique issue on Pocketbook but simply correct behavior that is 100% identical to the emulator with regard to which pictures are and aren't inverted?

NiLuJe:
If this is a simple sw inversion of the IV buffer, and doesn't actually toggle the `EPDC_FLAG_ENABLE_INVERSION` in the (next?), then it's very much not worth it, our own software inversion is probably more efficient and/or better integrated ;).

Check with `strace` whether this actually affects the refresh ioctls?

rjd22:
@Frenzie at this moment it might be a pocketbook only issue. I will need to test why the image is not inverted when selecting a image. It might be caused by inkview itself. I will check this later.

@NiLuJe you might be right. I will check this later on.

NiLuJe:
The name sure makes it sound like our own https://github.com/koreader/koreader-base/blob/64bdeda85efe127357c9b969a8cc72b46521ad87/ffi/blitbuffer.lua#L1379-L1387 ;).

rjd22:
@Frenzie I don't have access to an emulator yet since I've been testing the changes only on my PocketBook till now (since I'm only doing PocketBook specific changes and fixes). But I've tested opening a .jpg picture on the PocketBook with software night mode on.

Navigating the pictures in the list work fine. They are inverted and then inverted back. But then opening the picture, it doesn't seem to be pre-inverted causing it to look inverted on the screen in night mode.

It might be that this is a PocketBook only issue but atm. I have no way to confirm that. The version of KoReader I'm using is 2021.12.1.

Frenzie:
@rjd22 If they're fine in CoverBrowser you're probably seeing nothing unexpected. However, like @NiLuJe said if it's just doing the same thing as our own software inversion function there's no point to it. :-)

NiLuJe:
You'll need shell access ;). It's just a simple binary in a tarball, which you can use to trace KOReader with a call that might look like `strace -fittTvy -e trace=ioctl -p $(pidof reader.lua)`.

That said, given that some/most PBs run on a sunxi SoC with no actual HW inversion support, I wouldn't necessarily spend more time on this, and just go w/ flipping `cxanHWInvert` ;).

rjd22:
@NiLuJe okay, let's do that. I'm reopening the other PR and leaving this one open for PocketBook to take a look at (I linked this PR in the message I send them and they seem to be watching us 😉). After that I will start making some cleanup PR's to clean up some broken/dead code.

rjd22:
I got the following response from the Pocketbook developers today:

> As far as we understand, you have already implemented the inversion on application level, unfortunately, we do not support inversion on hardware level.
We also noticed, that in KoReader the pictures which are images in jpg, png and so on, are displayed inverted in night mode, but at the same time books illustrations are displayed correctly in night mode during reading, but that is a bug in KoReader, the same bug is in KoReader for black and white devices.

@Frenzie do you know if not inverting images is a bug or supposed to happen?

Frenzie:
@rjd22 At least it's definitely not a bug if it meets these conditions: https://github.com/koreader/koreader-base/pull/1453#issuecomment-1037490623

Frenzie:
Night mode is supposed to look like this. Pictures pre-inverted so that they are inverted back into being not being inverted in the end result.
Screenshot_2022-03-09_18-19-06

rjd22:
@Frenzie could you check in your emulator what happens if you open a .jpg file in night mode? Does that look correct in the emulator?

Frenzie:
No, those are inverted.

rjd22:
Most likely that is the behavior the developers of pocketbook are talking about. But in that case it might not be a pocketbook specific issue but something that happens to multiple devices.

NiLuJe:
PicDocument possibly doesn't do the double invert trick, yeah. Check if opening one of those w/ mupdf instead works to confirm ;).

Frenzie:
MuPDF doesn't do that either, so not sure what you expect. :-P

NiLuJe:
Oops, quite right, we only do that in CRe ^^.

rjd22:
Closing this. As we discussed some time ago it's better to let KoReader handle the night mode than moving it to Inkview.

PR #1454, opened 2022-02-13
Fix screen being partially refreshed when screen was rotated 90deg.

rjd22:
Fixes: https://github.com/koreader/koreader/issues/8773

When the screen was rotated 90deg. The screen was only updated
partially.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1454)

### DIFF
--- a/ffi/framebuffer_pocketbook.lua
+++ b/ffi/framebuffer_pocketbook.lua
@@ -9,6 +9,32 @@ local framebuffer = {
 
 }
 
+local function _getPhysicalRect(fb, x, y, w, h)
+    local bb = fb.full_bb or fb.bb
+    w, x = BB.checkBounds(w or bb:getWidth(), x or 0, 0, bb:getWidth(), 0xFFFF)
+    h, y = BB.checkBounds(h or bb:getHeight(), y or 0, 0, bb:getHeight(), 0xFFFF)
+    return bb:getPhysicalRect(x, y, w, h)
+end
+
+local function _updatePartial(fb, x, y, w, h)
+    x, y, w, h = _getPhysicalRect(fb, x, y, w, h)
+
+    fb.debug("refresh: inkview partial", x, y, w, h)
+    inkview.PartialUpdate(x, y, w, h)
+end
+
+local function _updateFull(fb, x, y, w, h)
+    fb.debug("refresh: inkview full", x, y, w, h)
+    inkview.FullUpdate()
+end
+
+local function _updateFast(fb, x, y, w, h)
+    x, y, w, h = _getPhysicalRect(fb, x, y, w, h)
+
+    fb.debug("refresh: inkview fast", x, y, w, h)
+    inkview.DynamicUpdate(x, y, w, h)
+end
+
 function framebuffer:init()
     self._finfo = ffi.new("struct fb_fix_screeninfo")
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
I assume this is simply copied more or less verbatim from another file, but let us know if not 'cause then I'll pay closer attention. ;-)

rjd22:
@Frenzie only line 13-16 are exact copies of the other implementation. The rest was done by me ;)

ffi/framebuffer_pocketbook.lua
```diff
@@ -9,6 +9,32 @@ local framebuffer = {
...
+    x, y, w, h = _getPhysicalRect(fb, x, y, w, h)
+
+    fb.debug("refresh: inkview partial", x, y, w, h)
+    inkview.PartialUpdate(x, y, w, h)
+end
+
+local function _updateFull(fb, x, y, w, h)
+    fb.debug("refresh: inkview full", x, y, w, h)
+    inkview.FullUpdate()
+end
```
NiLuJe, line 29: I'd do it here, too.

Full her means the `UPDATE_FULL` mxcfb flag (i.e., flash). It doesn't necessarily imply full-screen ;).

rjd22: @NiLuJe there is nothing to feed it to yet. That is why I didn't add it yet. 

When we need it or you want have it for documentation we can add it ofcourse.

NiLuJe: I'd do it regardless, otherwise it's either weird or begging for being forgotten and being a source of weird bugs later on ;).

NiLuJe: Oh, wait. You meant to say that the IV function doesn't take any arguments ;).

That's... a problem, because it doesn't match with our uses of this call, where we do expect it to honor the rect...

NiLuJe: A minor one, true, but annoying nonetheless ;).

rjd22: @NiLuJe it doesn't take arguments. I think it just updates the Full display. It's the function of PocketBook that results in the best quality result. When asking PocketBook they send me the following explanation:

> The FullUpdate has the highest quality of update but it works only for full screen. If you call the FullUpdate, the full screen will be updated through the black, we do not recommend using it often because these often blinkings can annoy user.
> 
> As for the SoftUpdate – actually, it is the same to Partial update. It is recommended to use it only for b/w area, it has a good quality and doesn’t blink the screen, in addition to it - we recommend using FullUpdate from time to time. But this Soft update will do some ghosting for color areas.
> 
> The PartialUpdateHQ - can be used if there are color elements on the screen to reduce ghosting on them. To use it properly – you should know the coordinates of color area and call this update only for this area. It will do good quality of colored image and will not annoy user as much as the FullUpdate does. Please note, that PartialUdateHQ has sence only for devices with color screens.

rjd22: Yep thus seems to be the limitation of the inview library. I've been running it for some time now but it does seems to work quite nicely so not all is bad ;)

NiLuJe (approved):

PR #1457, opened 2022-02-27
Implement MTK support on Kindle

NiLuJe:
As found on the PW5 ;).

Thankfully, the driver's API is designed to match the mxcfb API, so this was hilariously trivial compared to the sunxi insanity ;).

----

* On mxcfb-like devices with hardware inversion support, toggle nightmode by switching the grayscale fb vinfo flag when we're running at 8bpp. All the mxcfb-like drivers check this to toggle PxP/MDP inversion, but the MTK driver also does it in other places to make better decisions in nightmode.
* Implement the refresh machinery for the MTK driver found on Kindles. Its API is thankfully based on the mxcfb API, so there's nothing drastic happening ;).
* Resync eink headers w/ FBInk.
* Update FBInk.
* Random lj-sqlite3 luacheck cleanup ;).

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1457)

### DIFF
--- a/ffi/mxcfb_kindle_h.lua
+++ b/ffi/mxcfb_kindle_h.lua
@@ -56,62 +56,62 @@ enum mxcfb_dithering_mode {
   EPDC_FLAG_USE_DITHERING_MAX = 5,
 };
 struct mxcfb_rect {
-  unsigned int top;
-  unsigned int left;
-  unsigned int width;
-  unsigned int height;
+  uint32_t top;
+  uint32_t left;
+  uint32_t width;
+  uint32_t height;
 };
 struct mxcfb_alt_buffer_data {
-  unsigned int phys_addr;
-  unsigned int width;
-  unsigned int height;
+  uint32_t phys_addr;
+  uint32_t width;
+  uint32_t height;
   struct mxcfb_rect alt_update_region;
 };
 struct mxcfb_update_data {
   struct mxcfb_rect update_region;
-  unsigned int waveform_mode;
-  unsigned int update_mode;
-  unsigned int update_marker;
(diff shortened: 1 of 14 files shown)


Frenzie (approved):
Looks pretty simple/sane

NiLuJe (commented):
> Reviewed 14 of 14 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1457)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

NiLuJe:
Okay, finally tested on both Kobo & Kindle, this should be ready ;).

ffi/mxcfb_kindle_h.lua
```diff
@@ -56,62 +56,62 @@ enum mxcfb_dithering_mode {
...
 };
 struct mxcfb_rect {
-  unsigned int top;
-  unsigned int left;
-  unsigned int width;
-  unsigned int height;
+  uint32_t top;
+  uint32_t left;
+  uint32_t width;
+  uint32_t height;
```
Frenzie, line 62: Is this a "real" change or some side effect from ffi-cdecl btw?

Frenzie (approved):
Seems to look fairly innocuous. ;-P

ffi/mxcfb_kindle_h.lua
NiLuJe: That was me switching to C99 types in the headers instead of Linux typedefs, which ffi-cdecl just resolved to base types instead of using the matching C99 ones ;).

PR #1458, opened 2022-03-06
Utf8Proc: add a dumb lowercase method

zwim:
Do the lower casing of a string like crengine does.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1458)

### DIFF
--- a/ffi/utf8proc.lua
+++ b/ffi/utf8proc.lua
@@ -23,8 +23,22 @@ end
 
 local Utf8Proc = {}
 
+--- Lowercases an utf8-encoded string
+--- @string str string to lowercase
+--- @bool normalize normalizes the string during operation
+--- @treturn string the lowercased string
+function Utf8Proc.lowercase(str, normalize)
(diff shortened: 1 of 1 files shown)


poire-z:
Feels a bit.. dumb to have to use/see in frontend `Utf8Proc.lowercase_dumb`.
May be you could wrap it here as a single function, which may do NFKC_Casefold by default (if that's what the frontend code needs most), and you would use in the rare places where you need the dumb variant:
`Utf8Proc.lowercase(word, false)`
(or `Utf8Proc.lowercase(word, true)`, but it sounds better to say `better=false` than `dumber=true` :)

```lua
function Utf8Proc.lowercase(str, normalize)
    if normalize == nil then normalize = true end
```

poire-z:
You need this merged, with the crengine bumps, for your user hyph frontend PR ?

zwim:
Yes, please.

zwim:
Thank you.

PR #1459, opened 2022-03-07
Integrate ezdiy changes

roygbyte:
This doesn't work, and maybe it's not helpful at all, but it's as far as I got integrating @ezdiy's changes. I'm really out of my element here, much more than I thought. Anyways.... :') I'm not even sure I went about the right way of integrating the changes... I opted for a rebase instead of a merge, which may have been a better direction to go. 

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1459)


Frenzie:
A rebase is perfectly fine but it seems something's not quite right about this one. There are hundreds of commits listed that shouldn't be there.

For future reference, given that we're only talking about 2 commits rather than 20 or 200, cherry-pick may make a lot more sense than rebase regardless.

poire-z:
Why did you close this ?
Because it's not valid/working or has issues?
Or just because you got no feedback and it feels like lack of interest ?
Or because it feels like a too heavy responsability to be responsible for this update (it would, to me :)

roygbyte:
Hey, great question and thanks for asking :) I closed it because I really had no idea what I was doing! (Although I did learn quite a bit about CMake, etc. (Having known nothing at all about it prior to this PR)) I would love to be the person who makes it happen, but it will probably be a year or so until I have the technical prowess...

PR #1462, opened 2022-03-16
bump crengine: rendering, CSS and hyphenation tweaks

poire-z:
Includes:
- Fix sizing of float/inline-block image containers https://github.com/koreader/crengine/pull/467
- Fix body background drawing and other clipping issues
- Drawing: don't use `setHidePartialGlyphs()`
- CSS: ignore the `<body>` parent of `<DocFragment>`
- CSS colors: parse `rgb()/rgba()` and 4/8 digits hex values
- LVString: update/fix char props and `lStr_findWordBounds()` https://github.com/koreader/crengine/pull/468
- `UserHyphDict::getHyphenation()`: trim word with `lStr_findWordBounds()` https://github.com/koreader/crengine/pull/469

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1462)

PR #1465, opened 2022-03-17
Purge Github Action

Frenzie:
This reverts commit ab2fa6f3aedb3e9a799a10df39b3bd5346de3f14.

Fixes https://github.com/koreader/koreader-base/pull/1462#issuecomment-1070807378

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1465)


Frenzie:
Merging because this is too annoying and I don't have time. It can be restored later if desired. See https://github.com/koreader/koreader-base/pull/1466 for what's likely to be the basic method to go about it, except not *exactly* like that.

PR #1466, opened 2022-03-17
[CI] Fix GitHub Action

Frenzie:
Fixes https://github.com/koreader/koreader-base/pull/1462#issuecomment-1070807378.

Alternative to #1465. Maintenance-wise something has to give though.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1466)

PR #1471, opened 2022-03-26
bump crengine: rework char categorization and word selection

poire-z:
Includes https://github.com/koreader/crengine/pull/472 :
- LVString: fix char props, add lStr_isCJK() and lStr_isRTL()
- Rework is/prev/next-VisibleWord-Start/End()
- getCursorRect(): fix bottom point check

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1471)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -1914,7 +1914,10 @@ static int getTextFromPositions(lua_State *L) {
 		// It's usually OK with multiple chars selection, even if one needs to go over the further
 		// 2nd half of the glyph to have it included in the selection.
 		// But it is less OK with an initial long-press on a glyph, where we get a 50% chance of
-		// having the next char selected. Try to handle this case better:
+		// having the next char selected. Try to handle this case better.
+		// It turns out this also happens when some CJK/symbol char is near a multi-alpha word,
+		// so, when grabbing prev or next below, we use prevVisibleWordStart()/nextVisibleWordEnd()
+		// instead of just moving offset by -1/1.
 		if (r.getStart() == r.getEnd()) { // for single CJK character
 			bool grab_prev = false;
 			lvRect glyph_pt;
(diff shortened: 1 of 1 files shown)


cre.cpp
```diff
@@ -1914,7 +1914,10 @@ static int getTextFromPositions(lua_State *L) {
 		// It's usually OK with multiple chars selection, even if one needs to go over the further
 		// 2nd half of the glyph to have it included in the selection.
 		// But it is less OK with an initial long-press on a glyph, where we get a 50% chance of
-		// having the next char selected. Try to handle this case better:
+		// having the next char selected. Try to handle this case better.
+		// It turns out this also happens when some CJK/symbol char is near a multi-alpha word,
```
Frenzie, line 1918: multi-apha word?

poire-z: How you would name a word with more than one alphanumerical chars ? vs. CJK and symbols &copy; which are technically words in this context of text selection.

poire-z:
@Frenzie : is there something similar to https://github.com/koreader/crengine/pull/471 (and the other stuff you fixed recently) to update in koreader-base ? We get here both:

```
Cloning into 'busted'...
fatal: remote error: 
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.
```

and

```
E: Failed to fetch http://security.ubuntu.com/ubuntu/pool/main/g/gcc-9/gcc-9-multilib_9.3.0-17ubuntu1~20.04_amd64.deb  404  Not Found [IP: 52.154.174.208 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Error: Process completed with exit code 100.
```

Frenzie:
I didn't fix anything in base yet but GHA can be safely ignored.

Frenzie:
#1465 is a working fix, #1466 for some reason is not

cre.cpp
Frenzie: Not complaining, just not sure what it meant.

Maybe multicharacter?

PR #1473, opened 2022-04-03
Update to KoboUSBMS 1.3.1

NiLuJe:
* Allows connecting to ports negotiated as CDP (which *should* be most PD-aware and not misbehaving and/or nerfed USB 3.1 (or is that 3.2 now?)/4.0 (i.e., Thunderbolt 3/4) ports).

USB is fun.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1473)


NiLuJe:
Should fix https://www.mobileread.com/forums/showthread.php?t=345259

Frenzie (approved):

PR #1474, opened 2022-04-04
cmake 3.23

yparitcher:
on archlinux at least this file was moved in 3.23

DO NOT MERGE!

@NiLuJe @Frenzie 

What would be the correct way to fix /futureproof this


Frenzie:
ExternalProject has had a number of improvements over the years. Someone would have to investigate (I don't have time right now) but we may no longer need what is effectively an improved adaptation of it.

I suspect the thing I wrote where it first tries a fairly shallow clone isn't part of that regardless though.

yparitcher:
Since I don't know cmake & no one is interested in this so far, i implemented a backwards compatible band-aid fix for now.

Ready for review

NiLuJe:
Possibly ìf (EXISTS path)` branch would work, but here be dragons, because CMake ;).

yparitcher:
> Possibly an `if (EXISTS path)` branch would work, but here be dragons, because CMake ;).

But the current method makes it clear what causes the change and when it can be obsoleted

Frenzie:
If I can find the time (NB definitely not today, quite unlikely the next couple of weeks), I'll see if we can pull off another https://github.com/koreader/koreader-base/pull/868, just getting rid of a ton of stuff in favor of vanilla CMake. Most of this stuff was originally written to work with CMake 2.8 or possibly older.

Frenzie (approved):
In the meantime, no objections to merging this.

Frenzie:
PS I think I wrote it down in more detail somewhere already, but CMake-wise:

* Back in the day: 2.8+
* Since 2019: 3.5.1+

* If reasonably possible without too many compromises, 3.10+
* If not, ideally try for 3.13+
* Next step would be 3.16+
* Newer is *possible* but undesirable unless really necessary

NiLuJe:
Problem: `VERSION_GREATER_EQUAL` was introduced in CMake 3.7. The buildbots currently run 3.5 :s.

NiLuJe:
A quick google search seems to give an idiotic workaround: splitting the check in two (VERSION_EQUAL OR VERSION_GREATER).

Frenzie:
PS wrt https://github.com/koreader/koreader/pull/8987#issuecomment-1097204995

>  The buildbots currently run 3.5 :s.

No they don't! *Exclusively* the one in front. Nothing else.

PR #1478, opened 2022-04-11
Input: Support getting/setting key repeat rates

NiLuJe:
Needed by https://github.com/koreader/koreader/pull/8987

----

This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1478)


Frenzie (approved):
Hm, ok

PR #1485, opened 2022-05-30
Makefile.defs: fix regex to handle luarocks paths with numbers

a-m-joseph:
When using nixpkgs, the path to luarocks will frequently contain
[0-9].[0-9] subpatterns in the path; for example:

  $ luarocks --version
  /nix/store/f7qgsj4zckar8jjfl3pmzklcz5wa7mz7-luarocks-3.8.0/bin/.luarocks-wrapped 3.8.0

Let's update the `LR_VER_NUM` regex to only match the number at the
*end* of the line.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1485)

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -19,7 +19,7 @@ NDK?=$(TOOLCHAIN_DIR)/android-ndk-r15c
 NDKABI?=14
 
 # Detect LuaRocks 3
-LR_VER_NUM := $(shell luarocks --version | grep -o [0-9].[0-9])
+LR_VER_NUM := $(shell luarocks --version | grep -o [0-9].[0-9]\$)
 LR_GT_3_0 := $(shell awk -v LR_VER_NUM=$(LR_VER_NUM) 'BEGIN { print (LR_VER_NUM >= 3.0); }')
 
 # Some CMake flags


a-m-joseph:
Can anybody explain to me why there is an advertisement for some company called "reviewable" appended to my PR message?

I've never seen this before.  They exclaim that they are "not affiliated with github" so clearly github didn't do this.

Frenzie:
It's an integration we have. We did that. ^_^

Makefile.defs
```diff
@@ -19,7 +19,7 @@ NDK?=$(TOOLCHAIN_DIR)/android-ndk-r15c
 NDKABI?=14
 
 # Detect LuaRocks 3
-LR_VER_NUM := $(shell luarocks --version | grep -o [0-9].[0-9])
+LR_VER_NUM := $(shell luarocks --version | grep -o [0-9].[0-9]\$)
```
Frenzie, line 22: This isn't right.

It's meant to do something like:

2.4.2 → 2.4
3.8.0 → 3.8

This way that becomes:
2.4.2 → 4.2
3.8.0 → 8.0

I'm not immediately sure otoh what the best way to work around this would be. Perhaps we need awk instead of grep.

But given your OP, it might suffice to do something fairly dumb looking, like:
```
luarocks --version | grep -o " [0-9].[0-9]" | grep -o "[0-9].[0-9]"
```
(Or alternatively with -P to be able to use non-matching groups, though I doubt that'd be more legible.)

Frenzie (changes requested):

Frenzie:
Was this issue resolved? Wouldn't you still need a `| head -1` at the end of the line even if the subpatterns are always identical?

PR #1487, opened 2022-06-03
[CI] Update all but Android to new CMake 3.16 image

Frenzie:
See https://github.com/koreader/virdevenv/pull/80

~~For https://github.com/koreader/koreader-base/issues/1486~~

Android is for next week.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1487)

PR #1489, opened 2022-06-06
Make luacheck >= 0.26 happy

NiLuJe:
Re: #1487

### DIFF
--- a/ffi/mupdf.lua
+++ b/ffi/mupdf.lua
@@ -108,7 +108,7 @@ function mupdf.openDocument(filename)
 
     setmetatable(mupdf_doc, document_mt)
 
-    if not (mupdf_doc:getPages() > 0) then
+    if mupdf_doc:getPages() <= 0 then
         merror("MuPDF found no pages in file.")
     end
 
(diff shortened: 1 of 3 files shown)


ffi/mupdf.lua
```diff
@@ -108,7 +108,7 @@ function mupdf.openDocument(filename)
 
     setmetatable(mupdf_doc, document_mt)
 
-    if not (mupdf_doc:getPages() > 0) then
+    if mupdf_doc:getPages() <= 0 then
```
Frenzie, line 111: I probably do prefer this.

Frenzie (approved):
I do prefer all of these; it's possible I should reevaluate https://github.com/koreader/koreader/pull/9174#issuecomment-1147012782 slightly.

ffi/mupdf.lua
NiLuJe: FWIW, that's the only one that gave me pause (mostly because I don't quite recall what happens on `merror` if `getPages` fail anyway :D).

Frenzie: I would assume that it'll fail ungracefully with both writing styles. ^_^

PR #1490, opened 2022-06-10
bump crengine: text selection tweaks and helpers

poire-z:
Includes https://github.com/koreader/crengine/pull/484 : 
- lvtext: fix `m_kerning_mode` type
- XML: let '`gb2312`' (Chinese) encoding be known
- Add `ldomXPointer::getChar()`
- `LVDocView::getNodeByPoint()`: tweak for text selection
- Fonts: allow fetching the list of embedded fonts

cre.cpp:
- `getTextFromPositions()`: use `getNodeByPoint(forTextSelection=true)` to allow panning in margins and get the nearest text instead of nothing. See https://github.com/koreader/koreader/issues/9169#issuecomment-1146625083.
- add `extendXPointersToSentenceSegment()` to allow extending some xpointer range to include punctuations at start or end (with for now a quite rudimentary implementation). See https://github.com/koreader/koreader/issues/8857#issuecomment-1151198923.
- add `getEmbeddedFontList()` to allow fetching the list of embedded fonts in the current EPUB book. May help with https://github.com/koreader/koreader/issues/9190.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1490)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -699,7 +699,7 @@ static int closeDocument(lua_State *L) {
 
 	/* should be safe if called twice */
 	if(doc->text_view != NULL) {
-		// Call close() to have the cache explicitely saved now
+		// Call close() to have the cache explicitly saved now
 		// while we still have a callback (to show its progress).
 		doc->text_view->close();
 		// Remove any callback
@@ -989,6 +989,26 @@ static int getFontFace(lua_State *L) {
 	lua_pushstring(L, doc->text_view->getDefaultFontFace().c_str());
 
 	return 1;
+}
+
+static int getEmbeddedFontList(lua_State *L) {
+    CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+
+    lString32Collection registered_list;
+    lString32Collection instantiated_list;
+    fontMan->getRegisteredDocumentFontList(doc->dom_doc->getDocIndex(), registered_list);
+    fontMan->getInstantiatedDocumentFontList(doc->dom_doc->getDocIndex(), instantiated_list);
+
+    lua_createtable(L, 0, registered_list.length());
+    for (int i = 0; i < registered_list.length(); i++) {
+        lString32 name = registered_list[i];
+        bool instantiated = instantiated_list.contains(name);
+        lua_pushstring(L, UnicodeToLocal(name).c_str());
+        lua_pushboolean(L, instantiated);
+        lua_rawset(L, -3);
+    }
+
+    return 1;
 }
 
 /*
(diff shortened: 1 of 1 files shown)


cre.cpp
```diff
@@ -2004,6 +2024,104 @@ static int getTextFromPositions(lua_State *L) {
...
+            break;
+        }
+        // Otherwise, it follows some other kind of chars and it is not a sentence
+        // segment start, unless we grabbed some opening punctuation.
+        if ( !grabbed_opening ) {
+            isSentenceSegment = false;
+        }
+        break;
+    }
+    // We explicitely don't want to grab anything at the end if the start is not
```
Frenzie, line 2078: explicitly

Frenzie (approved):

poire-z:
Does the grammarian in you know the right word for my "sequence segment" ? Or is that ok ?

Frenzie:
You mean sentence fragment? There's a [phrase](https://en.wikipedia.org/wiki/Phrase) but that's not just a random bunch of words. What was it you were looking for specifically?

poire-z:
`This is sample, it has 3+1 (four) segments, and this is it.`
My four "sentence segments", that I may grab start and end punctuations, if the first and last word of each is the start and end of the selected text:
```
This is sample,
it has 3+1 (four) segments,
(four)
and this is it.
```
And any overlap/concatenation of these.
Their might be a word for these units of text making a multi part sentence. Would that be "phrase" (or is it clear enough for indicating this and the difference with a full sentence (from first Uppercase letter to ending fullstop/period) ?

In French, phrase == sentence.

Frenzie:
I think trying to hammer grammatical terminology into that which isn't particularly syntactical in nature would obscure that these are merely what you dubbed segments. Anyway, the computer linguistics term for this kind of thing is more or less sentence and word segmentation (cf. [this text about nltk](https://www.guru99.com/tokenize-words-sentences-nltk.html)), so I figure we can simply say it's Close Enough™ to give it that name (as you already did).

hius07:
It seems with this PR we can now long-press on the empty space of the page (eg at the end of a chapter) and have a word highlighted

poire-z:
Yes.
But it already happened before when you long-press in some blank part of the page (inside the margins), in the yellow parts. With this PR, it applies also in the blue parts:
image

I don't really feel it is not good behaviour. And you can't deny it's consistent :)
Whereever you long-press, you'll grab the nearest word. (Actually, wordS - when long-pressing on the yellow parts, the code often grab the previous word at end of the previous paragraph - and the next word, the first of the next paragraph...)

Would long-press in the blue parts have any other usage that this would prevent ?
(I think that before, we could occasionally have some text selected that is many many pages aways, usually the title or first words at the start of the DocFragment - which was quite bothering when you long-press by accident in the margin when you tried to hit a first small word in a line.)

hius07:
My page has only two lines (end of a chapter) and long-pressing in the bottom highights a word in the top.
I was just surprised as I had missed this changes while being in my intercontinental moving.
Okay, the reasons for these changes are important.

PR #1494, opened 2022-06-23
[PW5 MTK] Fix swipe animations with a tiny refresh area

yparitcher:
koreader/koreader#9245 NiLuJe/FBInk#69

The simplest way to trigger this is `fbink -K direction=LEFT,steps=12 -s width=11`

The MTK Driver will crash if given a area smaller than the number of steps. If direction is L/R and w is smaller or if it is U/D and h is smaller Being as one generally will only want animations on a larger area, and I am too lazy to test for direction, disable animations when w or h is less than steps.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1494)


Frenzie (approved):
Alright

PR #1431, opened 2021-11-14
sdcv: update to include multiple results fix

cyphar:
Needed to finally fix koreader/koreader#5168.

Signed-off-by: Aleksa Sarai <cyphar@cyphar.com>

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1431)

PR #1495, opened 2022-06-24
thirdparty/sdcv v0.5.4

Frenzie:
https://github.com/Dushistov/sdcv/compare/v0.5.3...v0.5.4

Except of course we were on a slightly more recent commit, cf. <https://github.com/koreader/koreader-base/pull/1431>.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1495)

### DIFF
--- a/thirdparty/sdcv/CMakeLists.txt
+++ b/thirdparty/sdcv/CMakeLists.txt
@@ -78,7 +78,7 @@ set(WITH_READLINE False CACHE BOOL "")
 # Force utf8 command line parsing, and accept not-found -u dictnames
 set(PATCH_CMD1 "${KO_PATCH_SH} ${CMAKE_CURRENT_SOURCE_DIR}/sdcv-locale-hack.patch")
 
-set(SDCV_GIT_COMMIT d054adb37c635ececabc31b147c968a480d1891a)
+set(SDCV_GIT_COMMIT tags/v0.5.4)
 
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME


pazos:
this breaks dicts on android 😢

Frenzie:
Huh, it does? Hold on, in that case let me post the fully accurate range of commits since our previous version.

https://github.com/Dushistov/sdcv/compare/d054adb37c635ececabc31b147c968a480d1891a...v0.5.4

Frenzie:
@pazos It seems that it doesn't actually, but something about the act of upgrading itself does.

Frenzie:
That is,

Fresh KOReader 2022.06 install → dictionary (sdcv ?) works

Upgrade from 2022.05.1 (or any other version presumably) to 2022.06, dictionary/sdcv broken

So the problem doesn't lie in this PR/repo.

pazos:
o_O

No real idea.

Frenzie:
PS I haven't checked if it's something unique to that upgrade. I shouldn't imagine it is.

pazos:
I will have a look in a couple of days. Too much device janitor for today :smile:

ilyats:
It seems so:
v2022.06 - works
v2022.06-9 - works if installed after removal of app or over v2022.06, does not work if installed over v2022.06-9 or v22.06-10
v2022.06-10 - does not work

ilyats:
After update from release 2022.05.1 to either release 2022.06 or to nightbuild 2022.06-9 it works, after update from 2022.05.1 to nightbuild 2022.06.10 it doesn't

Frenzie:
In that case, I would guess 2021.10 to 2021.11 should also fail.

ilyats:
v2022.06-12 does not work no matter how installed (just like v2022.06-10).
Downgrading from v2022.06-12 to v2022.06-9 is not possible. v2022.06-.11 is
absent on ota.koreader.rocks.

вс, 3 июл. 2022 г. в 10:00, Frans de Jonge ***@***.***>:

ilyats:
It is possible to install v2022.06-09 over itself, and it breaks search, (while v2022.06-9 works when installed after removal of the app or over v2022.06), so the problem is not different libraries.

Frenzie:
There's never a build for every single commit, only one per day.

Anyway, I'd overlooked that the buried not actually duplicate mentioned a change specifically post-stable. So then it is something in the commit diff above.

PR #1507, opened 2022-07-24
thirdparty/harfbuzz 5.0.1

Frenzie:
It says it's fixed on GCC < 7.

https://github.com/harfbuzz/harfbuzz/releases/tag/5.0.0
https://github.com/harfbuzz/harfbuzz/releases/tag/5.0.1

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1507)


Frenzie:
That changing `-j${PARALLEL_JOBS}` to `j2` works is rather strange, given that they should be the same on CircleCI.

Frenzie:
I was wrong, it'll evaluate to 3 due to the +1:
https://github.com/koreader/koreader-base/blob/e3e0e3f38bcf1984d5666bb66bf0bced25e7eb85/thirdparty/cmake_modules/koreader_thirdparty_common.cmake#L14-L16

PR #1508, opened 2022-07-29
{Fake Events} for charging

zwim:
Kobo Sage is not always recognizing the  plug in of an usb cable  This PR should fix this.

The mass of the changes are due to tab - spaces.

The notable changes start at  line # 98.

Not quite read yet.

This is connected to the quirks from https://github.com/koreader/koreader/pull/8934

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1508)

### DIFF
--- a/input/input-kobo.h
+++ b/input/input-kobo.h
@@ -22,7 +22,23 @@
 #define USBPLUG_DEVPATH "/devices/platform/usb_plug"
 #define USBHOST_DEVPATH "/devices/platform/usb_host"
 
+#define USB_CONNECTED_FILE "/sys/class/misc/cilix/usb_conn"
+#define CHARGER_PRESENT_FILE "/sys/class/power_supply/battery/device/charger_type"
+
 #include "libue.h"
+
+static size_t read_file(char const* path, char* buf, size_t buflen)
+{
+    FILE* file = fopen(path, "r");
+    if (!file) {
+        fprintf(stderr, "Could not open '%s'\n", path);
+        return 0;
+    }
+    size_t n = fread(buf, 1, buflen - 1, file);
+    buf[n] = '\0';
+    fclose(file);
+    return n;
+}
 
 static void sendEvent(int fd, struct input_event* ev)
 {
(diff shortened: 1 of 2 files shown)


NiLuJe:
IIRC, there's a clang-format config file for the formatting ('cuz thus is indeed unreviewable as is on mobile ;)).

NiLuJe:
IIRC, the sysfs entries are not necessarily up-to-date at the time the uevents are fired...

pazos:
> IIRC, the sysfs entries are not necessarily up-to-date at the time the uevents are fired...
> 
> (The uevent feed gets notified _significantly_ earlier than everything else, IIRC).

It is the opposite.

Udev identifies devices using attributes exported by the kernel via sysfs. Then it is able to create device nodes following certain rules.

So it is `KERNEL add` (sysfs) -> `UDEV add` (userspace) for each node.

NiLuJe:
I'm not talking about udev directly, I'm talking about the kernel's uevent feed (which is what both udev and us listen to).

e.g., the irq handler calls `usb_plug_handler` which is what calls `kobject_uevent` on `usb_plug`/`usb_host`/`usb_state`.

Usually, by the time this is sent, the same variable that the sysfs io handler reads *should* be up-to-date, too, but I've seen some weird shit on that front when there's a bit of jitter and you get a slew on conflicting events ;).

NiLuJe:
EDIT for email readers: Completed and clarified above response ;).

pazos:
ok, I don't know how it behaves on kobos, also I didn't review the code and didn't try to understand what's trying to do.

But as a rule of thumb: the kernel message always include a `DEVPATH` that **is** the sys path.

NiLuJe:
Another PowerCover quirk to lookout for is [this](https://github.com/koreader/KoboUSBMS/blob/f4adbbb2909c96b9af4a3223ffa1501e8224498d/usbms.c#L1420-L1422).

It's been a while, but, IIRC, *that* caused spurious `CHANGE` events ;).

zwim:
@pazos and @NiLuJe 
Thanks for your hints.

zwim:
Thanks for your help and thoughts, but I think I found an easy solution in frontend. So no need for changes here.

Will really need some time  to verify the solution. Device sleep ... wait  ... plug in/ out (with different start conditions ...). But I am definitly on the way.

PR #1511, opened 2022-08-02
Add thirdparty/libwebp and support for WebP images

poire-z:
#### Add thirdparty/libwebp and ffi/webp.lua wrapper

This adds libwebp.so and libwebpdemux.so, and should allow viewing webp (possibly animated) images in KOReader, and a prerequisite to add support for webp images in crengine and MuPDF.
ffi/pic.lua: also don't load other image libraries on startup, but only when needed.

#### MuPDF: add support for webp images

Includes code from patch submitted at upstream issue https://bugs.ghostscript.com/show_bug.cgi?id=697749#c4
Tested with CBZ containing webp images, and when using MuPDF to render images.
Not tested: PDF containing embedded webp images (if that is even possible).
Should allow closing https://github.com/koreader/koreader/issues/3016

#### bump crengine: add support for webp images

Will need a crengine bump for https://github.com/koreader/crengine/pull/487.
Should allow closing https://github.com/koreader/koreader/issues/9240

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1511)


poire-z:
About the MuPDF build:
I tried making LIBWEBP stuff into the various Makeblah similarly to LIBJPEG, but I didn't manage to make it compile (not enough and it doesnt link to the libweb.so, too much and it tries to link them into its hexdump and namedump early binaries, and it can't find them.

It turned out that if we don't add anything, it builds - and it needs us to provide the libwebp.so and libwebpdemux.so in the final command (I assume) in our CMakefile for them to be linked by libmupdf.so:
`set(SHARED_BUILD_STR sh -c "${CC} -shared ${CFLAGS} ${LINK_OPTS} ${AES_LIB} ${ZLIB} ${JPEG_LIB} ${FREETYPE_LIB} ${HARFBUZZ_LIB} ${LIBWEBP_LIB} ${LIBWEBPDEMUX_LIB} -o ${MUPDF_LIB} -lm")`

I hope that's ok enough - because I have this kind of work into the base CMake/MakeMeAngry stuff.

What I was playing with and did not include in the webp.patch for MuPDF:
```diff
diff --git a/Makefile b/Makefile
index 8e47e3a7e..73731459c 100644
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ LIBS += $(JPEGXR_LIB)
 LIBS += $(LCMS2_LIBS)
 LIBS += $(LIBCRYPTO_LIBS)
 LIBS += $(LIBJPEG_LIBS)
+#LIBS += $(LIBWEBP_LIBS)
 LIBS += $(LURATECH_LIBS)
 LIBS += $(MUJS_LIBS)
 LIBS += $(OPENJPEG_LIBS)
diff --git a/Makerules b/Makerules
index 57065f323..0edcdebcc 100644
--- a/Makerules
+++ b/Makerules
@@ -139,6 +139,7 @@ endif
 SYS_JBIG2DEC_LIBS := -ljbig2dec
 SYS_LIBJPEG_LIBS := -ljpeg
 SYS_ZLIB_LIBS := -lz
+#SYS_LIBWEBP_LIBS := -lwebp -lwebpdemux

 endif

diff --git a/Makethird b/Makethird
index 0ba7c85ac..8f5c34fa5 100644
--- a/Makethird
+++ b/Makethird
@@ -529,6 +529,11 @@ ZLIB_CFLAGS := $(SYS_ZLIB_CFLAGS)
 ZLIB_LIBS := $(SYS_ZLIB_LIBS)
 endif

+# --- LIBWEBP ---
+
+#LIBWEBP_CFLAGS := $(SYS_LIBWEBP_CFLAGS)
+#LIBWEBP_LIBS := $(SYS_LIBWEBP_LIBS)
+
 # --- LCMS2 ---

 ifneq "$(wildcard $(LCMS2_DIR)/INSTALL)" ""
```

ffi/webp.lua
```diff
@@ -0,0 +1,139 @@
...
+    if anim_info[0].frame_count > 1 and type(data) ~= "string" then
+        -- When called from CreDocument:getImageFromPosition(), data is a "userdata" wrapping
+        -- a buffer malloc()'ed by cre.cpp, which is freed as soon as getImageFromPosition()
+        -- returns, as we usually don't need it after having rendered the image data to
+        -- a blitbuffer.
+        -- With animated multiframes images, we keep the decoder object: GifLib is fine with
+        -- data being gone, but libwebp is not.
+        -- We could have renderImageData() return a flag to state that the data should not
+        -- be free()d, but it would complexify the return values and their handling.
+        -- So, do simple: make a copy of the external buffer into an interned Lua string,
```
NiLuJe, line 72: nit: `keep it simple` ;).

Frenzie: Heh, at first I thought you were talking about the code.

Frenzie (approved):
I'm a bit surprised you didn't just bother with WebP but bothered so comprehensively. Anyway, lgtm!

ffi/webp.lua
```diff
@@ -0,0 +1,139 @@
...
+
+function Webp:getFrameImage(number, no_copy)
+    if number < 1 then number = 1 end
+    if number > self.nb_frames then number = self.nb_frames end
+
+    if not self.webp_image then
+        self.webp_image = ffi.new("unsigned char*[1]")
+        -- WebPAnimDecoderGetNext() will update this pointer with the
+        -- address of one of its two internal bitmap data buffers.
+        -- It is so only valid until the next call.
```
NiLuJe, line 103: nit: `As such, it's only valid`

ffi/webp.lua
```diff
@@ -0,0 +1,139 @@
...
+    end
+    if not self.cur_frame then
+        self.cur_frame = 0
+    end
+    if number < self.cur_frame then
+        libwebpdemux.WebPAnimDecoderReset(self.webp_decoder) -- Rewind to start
+        self.cur_frame = 0
+        -- Note: browsing frames forward is fast, but because of this, browsing
+        -- late frames backward can be noticably slower.
```
NiLuJe, line 112: nit: `noticeably`

ffi/webp.lua
```diff
@@ -0,0 +1,139 @@
...
+    end
+    -- Render all frames from cur_frame to the one requested (this is needed as
+    -- webp frames can be partial and need to be blended over the previous frame)
+    local timestamp = ffi.new("int[1]") -- not used by us
+    while self.cur_frame < number do
+        local ret = libwebpdemux.WebPAnimDecoderGetNext(self.webp_decoder, self.webp_image, timestamp)
```
NiLuJe, line 118: I assume this means the blending of previous frames happens automagically, making the first buffer in `self.webp_image` valid?

thirdparty/libwebp/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+assert_var_defined(CPPFLAGS)
+assert_var_defined(LDFLAGS)
+assert_var_defined(CHOST)
+
+ep_get_source_dir(SOURCE_DIR)
+ep_get_binary_dir(BINARY_DIR)
+
+set(CFG_ENV_VAR "CC=\"${CC}\" CXX=\"${CXX}\" CPPFLAGS=\"${CPPFLAGS}\" LDFLAGS=\"${LDFLAGS}\" ")
+# libwebpdecoder.so is enough to decode single frame webp images. To decode animated webp,
+# we'd need the twice larger libwebp.so and libwebpdemux.so.
```
NiLuJe, line 21: I assume that comment is outdated, since you did go the extra mile ;).

NiLuJe:
What @Frenzie said ;).

I'm currently fighting a covid infection, so enjoy my possibly nonsensical grammar nits, yay \o/.

NiLuJe:
@Frenzie: Yeah, looks like HarfBuff murdered CI's RAM again ^^.

ffi/webp.lua
poire-z: Yep, WebPAnimDecoder does it all for us. It uses other public objects in the library that deal with the lower level stuff and the real partial/alpha frames that are in the webp.
At least, that's what I understood from the documentation in the include/ files.

NiLuJe: Oh, that's cool (this sort of stuff is notoriously easy to get wrong when handling animated GIFs & PNGs, or so I heard ;p).

thirdparty/libwebp/CMakeLists.txt
poire-z: Yes, but it's a comment on the next commented line.
Actually, I originally went with libwebp.so. Then, so the thing about `--enable-libwebpdecoder` in configure --help, and thought "oh, smaller lib ! let's go with that" - removed all the libwebp stuff, build, tested - and realized if failed on an animated webp. So, put the initial stuff back.
So, let's keep it as a kind of documentation of why we can't go with libwebpdecoder.so :)

NiLuJe: Oh, yeah, definitely, I just sort of implied it only needed a tiny tense tweak (say, something like "we need the full set of libs").

PR #1513, opened 2022-09-02
Stop using -static-libstdc++

NiLuJe:
Opting instead on shipping the TC's shared STL, as it's bound to be much, much saner.

Modified how Lua modules are linked to avoid passing absolute paths to shared libraries, because, depending on binutils version, this matters.

Also stopped lniking sdcv to a static libz outside of Android (where I'm not even sure why we do that for glib...).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1513)


NiLuJe:
Depends on https://github.com/koreader/libk2pdfopt/pull/40 & https://github.com/koreader/crengine/pull/488

NiLuJe:
> Also stopped lniking sdcv to a static libz outside of Android (where I'm not even sure why we do that for glib...).

Checking the history leads to https://github.com/koreader/koreader-base/pull/342, which limited the static mess to Android, and https://github.com/koreader/koreader-base/pull/317, which *introduced* it...

I'm pretty sure it's completely unnecessary nowadays, but, who knows...

NiLuJe:
Note to self: don't forget to bump crengine & libk2pdfopt once the sister PRs are merged ;).

NiLuJe:
Hey, it almost worked first try on Android ;o).

NiLuJe:
There, fixed Android.

Because it historically offered a choice of STL, everything is terrible:

* `libstdc++.so` exists, but is NOT a STL. It basically just implements `new` & `delete` (which appears to be enough for the nnsvg & xtext modules).
* So, you have to *choose* a STL. In practice, that's libc++ on Clang, and libstd++ on GCC; but, as stated above, they *both* have custom names, and since we're using a custom build-system, that means extra shenanigans to set that up.
* And, of course, the STL is *never* a system lib, it always has to be vendored (which sorta makes sense on Android, actually).

The downside to that last point is that it's probably built w/ debug symbols because it's ginormous (5.5M, compared to my ~1.5M on plain Linux).

NiLuJe:
> > Also stopped lniking sdcv to a static libz outside of Android (where I'm not even sure why we do that for glib...).
> 
> Checking the history leads to #342, which limited the static mess to Android, and #317, which _introduced_ it...
> 
> I'm pretty sure it's completely unnecessary nowadays, but, who knows...

Oh, wait, I know why it was done this way. It's to deal with broken early (-ish) API levels where the dynamic loader is dumber than a bag of rocks (c.f., https://github.com/koreader/android-luajit-launcher/pull/283).

Frenzie:
> It's to deal with broken early (-ish) API levels where the dynamic loader is dumber than a bag of rocks

Bionic is my mortal enemy.

poire-z (approved):

NiLuJe:
Made sure to link xtext, cre.cpp & CRe to the STL on Android, juuust in case.

(It doesn't get thrown out by --as-needed, so I'm assuming it's useful).

Also implemented the PRIVATE/PUBLIC switch for CRe linking as discussed earlier ;).

thirdparty/kpvcrlib/CMakeLists.txt
```diff
@@ -200,4 +201,8 @@ add_library(crengine SHARED ${CRENGINE_SOURCES})
 # Make sure we'll get C++11 support
 target_compile_features(crengine PRIVATE cxx_constexpr)
 
-target_link_libraries(crengine chmlib antiword ${THIRDPARTY_LIBS})
+target_link_libraries(crengine PRIVATE chmlib antiword)
+target_link_libraries(crengine PUBLIC ${THIRDPARTY_LIBS})
```
poire-z, line 205: Should these thirdparty libs be PUBLIC ?
I thought earler that by using PRIVATE, cre.cpp linking to libcrengine would not know that libcrengine would be linked to use liblunasvg and would then get less change to be linked against libstdc++ - but well, it didn't work.

NiLuJe: As usual, CMake's doc is terrible, so this mail looks like a much, much, much better description: https://cmake.org/pipermail/cmake/2016-May/063400.html ;).

NiLuJe: TL;DR: I guess some of 'em could be private, but I'm way too lazy to sort it out, and the point is moslty moot (on sane binutils versions) as `--as-needed` will avoid overlinking stuff that pulls in libcrengine.so anyway.

thirdparty/kpvcrlib/CMakeLists.txt
```diff
@@ -202,7 +202,7 @@ add_library(crengine SHARED ${CRENGINE_SOURCES})
 target_compile_features(crengine PRIVATE cxx_constexpr)
 
 target_link_libraries(crengine PRIVATE chmlib antiword)
-target_link_libraries(crengine PUBLIC ${THIRDPARTY_LIBS})
+target_link_libraries(crengine PRIVATE ${THIRDPARTY_LIBS})
```
poire-z, line 205: When playing with it, I was a bit unsure it was allowed to have 2 target_link_libraries.
So, just be sure you can open a png and a chm after the build :)

> Unfamiliar with CRe's API 

cre.cpp only uses bits declared in crengine include/* - no direct call to things in the libraries used by crengine.

NiLuJe: It'd have failed to link ;).

NiLuJe: And everything is indeed there:

```
┌─(niluje@illyria:pts/14)───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────(/var/tmp/niluje/koreader)─┐
└─(0.12:78%:16:41:%)── readelf -d koreader-android-arm-linux-androideabi-debug/koreader/libs/libcrengine.so                                                                                                                                                                                                                                            ──(Sat, Sep 03)─┘

Dynamic section at offset 0x21d600 contains 42 entries:
  Tag        Type                         Name/Value
 0x00000003 (PLTGOT)                     0x21ec14
 0x00000002 (PLTRELSZ)                   1984 (bytes)
 0x00000017 (JMPREL)                     0x591ec
 0x00000014 (PLTREL)                     REL
 0x00000011 (REL)                        0x44514
 0x00000012 (RELSZ)                      85208 (bytes)
 0x00000013 (RELENT)                     8 (bytes)
 0x6ffffffa (RELCOUNT)                   10253
 0x00000006 (SYMTAB)                     0x148
 0x0000000b (SYMENT)                     16 (bytes)
 0x00000005 (STRTAB)                     0x12528
 0x0000000a (STRSZ)                      160197 (bytes)
 0x00000004 (HASH)                       0x396f0
 0x00000001 (NEEDED)                     Shared library: [libz.so.1]
 0x00000001 (NEEDED)                     Shared library: [libzstd.so.1]
 0x00000001 (NEEDED)                     Shared library: [libfreetype.so.6]
 0x00000001 (NEEDED)                     Shared library: [libharfbuzz.so.0]
 0x00000001 (NEEDED)                     Shared library: [libfribidi.so.0]
 0x00000001 (NEEDED)                     Shared library: [libunibreak.so.3]
 0x00000001 (NEEDED)                     Shared library: [libutf8proc.so.2]
 0x00000001 (NEEDED)                     Shared library: [libjpeg.so.8]
 0x00000001 (NEEDED)                     Shared library: [libpng16.so.16]
 0x00000001 (NEEDED)                     Shared library: [libwebp.so.7]
 0x00000001 (NEEDED)                     Shared library: [libwebpdemux.so.2]
 0x00000001 (NEEDED)                     Shared library: [libgnustl_shared.so]
 0x00000001 (NEEDED)                     Shared library: [libm.so]
 0x00000001 (NEEDED)                     Shared library: [libc.so]
 0x0000000e (SONAME)                     Library soname: [libcrengine.so]
 0x0000001a (FINI_ARRAY)                 0x21a774
 0x0000001c (FINI_ARRAYSZ)               8 (bytes)
 0x00000019 (INIT_ARRAY)                 0x21a77c
 0x0000001b (INIT_ARRAYSZ)               32 (bytes)
 0x0000000f (RPATH)                      Library rpath: [$ORIGIN]
 0x00000010 (SYMBOLIC)                   0x0
 0x0000001e (FLAGS)                      SYMBOLIC BIND_NOW
 0x6ffffffb (FLAGS_1)                    Flags: NOW
 0x6ffffff0 (VERSYM)                     0x41ffc
 0x6ffffffc (VERDEF)                     0x44478
 0x6ffffffd (VERDEFNUM)                  1
 0x6ffffffe (VERNEED)                    0x44494
 0x6fffffff (VERNEEDNUM)                 4
 0x00000000 (NULL)                       0x0
```

NiLuJe (commented):
> Reviewed 2 of 6 files at r1, 3 of 4 files at r5, 1 of 1 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1513)* status: all files reviewed, 2 unresolved discussions (waiting on @poire-z)

PR #1514, opened 2022-09-10
bump crengine: better SVG support with extended LunaSVG 

poire-z:
#### Add thirdparty/lunasvg, patched and extended

To be used by crengine for better SVG support than with NanoSVG.
Will use the help of crengine to draw text and images.
Build from upstream https://github.com/sammycage/lunasvg
Add patch to upstream code, which should hopefully be stable.
Add our own files in a `xtended/` subdir, so we can follow their history in here.
[Got help with images](https://github.com/sammycage/lunasvg/issues/97) from @sammycage, thank you.
More details along https://github.com/koreader/koreader/issues/9298.

#### bump crengine: better SVG support with extended LunaSVG

Build crengine against LunaSVG instead of NanoSVG.
Includes https://github.com/koreader/crengine/pull/489 : 
- SerialBuf: allow serializing longer strings
- Support `` and `<embed>` as images
- Handle the `<svg>` element as an SVG image
- Support SVG images as first class documents

cre.cpp:
- `getImageDataFromPosition()`: allow getting a `creimage` Lua userdata object, so `ImageViewer` can request it to be rendered smoothed scaled to the requested size
- add `renderImageData()`, so frontend can use crengine to render SVG images (ie. for book covers)
- fix small memory leak

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1514)


Frenzie (approved):

poire-z:
What a mess this std:: stuff:
https://www.google.com/search?hl=en&num=25&q=%22round%27+is+not+a+member+of+%27std%27%22 on kindle & co
https://www.google.com/search?hl=en&num=25&q=%22to_string%27+is+not+a+member+of+%27std%27%22 on android

I may understand why crengine decided to use its own LV* implementations of all this somehow standard stuff.

poire-z:
Phew, it builds.
2 things I had to hack around (see the above google search urls for details):
- `std::round()` (used in upstream code) seems to not be available in our kobo/kindle/... toolchains. I had to switch it to `::round()`.
- `std::to_string()` (used in upstream code, and my code) seems to not be available in our old android gcc toolchain: I had to implement it (in xtended/xlunasvg.h, which is included in all sources, including upstream code, to keep it simple).

Is that fine with you @NiLuJe ?

NiLuJe:
I've definitely seen the `round` thing happen before, and I'm not *quite* sure why it's messed up, but, yeah, that's what [I do on my end](https://svn.ak-team.com/svn/Configs/trunk/Kindle/Misc/gdb-10.1-k3-log2-fix.patch), too (well, okay, it's noit exactly `round` here, but basically libm symbols ;p).

The string stuff is probably a guard in libstdc++ against broken libcs, I found similar old issues for MingW, so, yeah, makes sense to fix it on our end.

NiLuJe:
Random possibly unrelated thing I've noticed when testing on Android today: CRe cache saves appear to be *much* slower than before, does that ring a bell?

(I have yet to test on an actual device. Keep in mind that's a debug build, too).

poire-z:
Nope :/
The only thing related to cache is https://github.com/koreader/crengine/pull/489/commits/8dee18c1f3f11c73110fa79840f3e9cf58594e57  , now using the 2nd function instead of the 1st in https://github.com/koreader/crengine/blob/75ec3f9943e3b2a5fb273053cc026b66b0ee291e/crengine/src/lvstring.cpp#L5629-L5646. You could try reverting this just to see, but it feels like just 2 more peanuts.

I'm not noticing any slowdown on my Android phone with this morning nightly (exiting a big Bible which saves its cache for 15 seconds on my Kobo, takes less than 1 sec on my 6 years old android phone).

NiLuJe:
Yeah, possibly unrelated Android fuckery, as I'm also seeing 5s stalls on what's apparently building the font list on startup :?.

(And, ofc, it's random).

poire-z:
I've been also seeing some random 5-10s pause on emulator launch over the last days/week :/ Don't remember exactly on what it was pausing (possibly the font list), nor did I think about looking if cpu was at 100%...

NiLuJe:
Okay, the stall on close is unrelated, it's a quirk of DocCache that I'll be squishing.

(If you open a PDF, scroll a bit, close it, open an ePub, close it; it'll serialize stuff from the PDF).

NiLuJe:
Fixed the DocCache thing in https://github.com/koreader/koreader/pull/9513

And, of course, I couldn't reproduce the stalls on startup ;p.

poire-z:
Just putting here for reference some link I stumbled on while searching for other things, where somebody did the same kind of test files I had to, and asked some of the same questions, when adding this SVG support:
https://wolthera.info/2022/10/svg-text-layout-for-krita/
https://planet.kde.org/wolthera-van-hovell-tot-westerflier-2022-10-30-svg-text-layout-for-krita/ (same article, more readable here)

poire-z:
> Phew, it builds. 2 things I had to hack around (see the above google search urls for details):
> 
> `std::round()` (used in upstream code) seems to not be available in our kobo/kindle/... toolchains. I had to switch it to `::round()`.

Currently picking stuff from previous patch against LunaSVG master.
With our latest toolchain/build updates, do you think I still have to do the above tweaks for our kobo/kindle.... toolchains - that is, have they been updated?

Frenzie:
Since when are we talking about?

poire-z:
Sep 11 2022

Frenzie:
No, see https://github.com/koreader/koxtoolchain (and by extension in virdevenv)

But GCC 11.2 isn't exactly old?

Frenzie:
> std::round() (used in upstream code) seems to not be available in our kobo/kindle/... toolchains. I had to switch it to ::round().

This doesn't sound like something that makes sense. :-/ Maybe just open a PR here so we can see what happens?

poire-z:
Will do.
But I have learned to live with build stuff that doesn't make sense. And so will you :)

NiLuJe:
The lm math namespace thing is unlikely to have changed, as I haven't actually updated the toolchains yet ;).

poire-z:
Nonsense ! as expected:
```
/home/ko/base/thirdparty/lunasvg/build/arm-kindle5-linux-gnueabi/lunasvg-prefix/src/lunasvg/source/parser.cpp: In static member function 'static bool lunasvg::Parser::parseColorComponent(const char*&, const char*, int&)':
/home/ko/base/thirdparty/lunasvg/build/arm-kindle5-linux-gnueabi/lunasvg-prefix/src/lunasvg/source/parser.cpp:887:39: error: 'round' is not a member of 'std'; did you mean 'round'?
  887 |     component = static_cast<int>(std::round(value));
      |                                       ^~~~~
```

Frenzie:
In which one was that specifically?

poire-z:
Can't find the CI run with the failure in my history (the latest one may have got the same url) - but from memory, kobo, kindle, pocketboot - and may be 1 or 2 others.

Frenzie:
Ah, well, never mind then.

If it were the 18.04 image then I was planning image maintenance anyway. I'd originally planned that for July 2023 but I'm holding off for now to keep a clear view on potential Android TC update issues. But all the images you mentioned aren't relevantly outdated unless CMake is causing it somehow.

NiLuJe:
Nah, I've seen this when building other stuff on the same TCs (e.g., gdb).

PR #1516, opened 2022-09-14
Bump sdcv to 0.5.4 + a couple of PRs to fix it

NiLuJe:
I'm not sure if it was actually broken for me before, but at least this one works on both Kobo & Android ;).

@pazos's hunch was nearly perfectly spot-on (my working theory is that it's the `sizeof` comparisons that were broken, quite likely because off64_t vs. ulong, and not the fstat call itself).

Split our existing patch in two, and submitted the sanest part of it upstream, because apparently I'd forgotten about it, and it happened to have been reported upstream.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1516)

### DIFF
--- a/thirdparty/sdcv/CMakeLists.txt
+++ b/thirdparty/sdcv/CMakeLists.txt
@@ -77,10 +77,14 @@ set(CMAKE_SKIP_BUILD_RPATH True)
 set(ENABLE_NLS False CACHE BOOL "")
 set(WITH_READLINE False CACHE BOOL "")
 
-# Force utf8 command line parsing, and accept not-found -u dictnames
-set(PATCH_CMD1 "${KO_PATCH_SH} ${CMAKE_CURRENT_SOURCE_DIR}/sdcv-locale-hack.patch")
+# Don't crash when passing unknown dicts to -u
+set(PATCH_CMD1 "${KO_PATCH_SH} ${CMAKE_CURRENT_SOURCE_DIR}/sdcv-pr91.patch")
+# Use proper types for off_t stuff so as not to upset large file support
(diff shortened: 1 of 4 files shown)


NiLuJe:
```
OnePlus6:/data/user/0/org.koreader.launcher.debug/files $ LD_LIBRARY_PATH=./libs ./sdcv -v                                                                                                                                                                                                                                                                              
Console version of Stardict, version 0.5.4
```

Frenzie (approved):

poire-z:
Possibly posted in not the right sdcv PR, but I just noticing this today.
When I lookup some word, on the emulator or my Kobo, I see this logged (so, even when not in debug mode):
```
Unknown dictionary: Dictionnaire de lâAcadÃ©mie FranÃ§aise, 8Ã¨me Ã©dition (1935).
Unknown dictionary: Petit Robert 2007
Unknown dictionary: XMLittre
Unknown dictionary: Wiktionary FR-FR
Unknown dictionary: Arramooz
Unknown dictionary: English-Arabic dictionary
[...]
```
It's some bit you get added upstream in https://github.com/Dushistov/sdcv/commit/6e36e7730caf07b6cd0bfa265cdf9b5e31e7acad
and may be it is logged because of our option to specify which dicts to use or ignore.

Everything seems to work, it's just noise in the logs, but still :)

NiLuJe:
@poire-z: What's the commandline look like in this instance?

NiLuJe:
Okay, yeah, it's an allow list, so we're stuck handling this on our side...

I'd be inclined to leave it as-is, because logging stderr makes sense, but if not, we'd "just" need to throw stderr to /dev/null in the shell command ;).

NiLuJe:
Or... we could just patch that out in sdcv, because, while the warnings might make sense upstream, they don't really for us ;).

PR #1517, opened 2022-09-14
ffiUtil: Make orderedPairs behave with mixed key types

NiLuJe:
Usually involves sorting numbers with strings, but could theoretically be more esoteric ;).

Still assumes that matching types supports comparisons, which may not always be true for custom types without an overloaded operator...

Should be fine for our current needs, though.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1517)

### DIFF
--- a/ffi/util.lua
+++ b/ffi/util.lua
@@ -677,7 +677,7 @@ local function orderedNext(t, state)
     -- We use a temporary ordered key table that is stored in the table being iterated.
 
     local key = nil
-    --print("orderedNext: state = "..tostring(state))
+    -- print("orderedNext: state = "..tostring(state))
     if state == nil then
         -- the first time, generate the index
         t.__orderedIndex = __genOrderedIndex(t)
--- a/ffi/util.lua
+++ b/ffi/util.lua
@@ -655,12 +655,20 @@ end
 -- pairs(), but with *keys* sorted alphabetically.
 -- c.f., http://lua-users.org/wiki/SortedIteration
 -- See also http://lua-users.org/wiki/SortedIterationSimple
-local function __genOrderedIndex( t )
+local function __genOrderedIndex(t)
     local orderedIndex = {}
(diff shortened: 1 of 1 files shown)


NiLuJe:
Hasn't broken any of my LuaSettings for far ;p.

NiLuJe:
And because I'm a genius, only *now* do I think about checking the references in that function's comment.

And, yeah, I didn't reinvent the wheel, so this looks sound ;).

ffi/util.lua
```diff
@@ -669,10 +677,10 @@ local function orderedNext(t, state)
     -- We use a temporary ordered key table that is stored in the table being iterated.
 
     local key = nil
-    --print("orderedNext: state = "..tostring(state) )
+    --print("orderedNext: state = "..tostring(state))
```
Frenzie, line 680: `--print` or `-- print`?

NiLuJe: That's a good question, do we have a consensus for that in the code?

(I rarely do that in Lua, but I tend to prefer "sticky" comments for those in other languages, so I'd keep it as-is myself, FWIW).

Frenzie: I prefer with a space because I consider it more legible, so I suppose it's a stalemate.

Anyway, it's one of those things I'd have happily ignored for decades but you had to touch the line… :-P

poire-z: (I also prefer with a space after `--`, which is how 99% of our comments are.)

NiLuJe: That's perfectly fine by me ;o).

NiLuJe (commented):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1517)* status: all files reviewed, 1 unresolved discussion

PR #1524, opened 2022-09-28
FB: Misc waveform mode shenanigans

NiLuJe:
* framebuffer: Misc waveform mode shenanigans on mxcfb & sunxi:
  * Support asking for A2 explicitly, since we no longer use it for fast anywhere
  * Unify the logic behind FORCE_MONOCHROME usage (i.e., only use it for DU when it's hardware-accelerated), because it makes sense to make text highlights look better (especially on high density screens). Added comments explaining the behavior a bit.
  * Support toggling NO_MERGE on a per-update basis, because of course I managed to find a case where the collision handling issues in the Sage kernel led to visible issues.
* rtc: Also log the rtc clock when validating an alarm (and switch to threadsafe functions to ensure the pointers we keep around point to our own data, and not the libc's statis storage).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1524)


NiLuJe (commented):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1524)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

NiLuJe (commented):
> Reviewed 3 of 4 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1524)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

PR #1529, opened 2022-10-05
Makefile.defs: Use path agnostic bash location for $SHELL

Baitinq:
This fixes koreader not being built on platforms where bash isnt at /bin/bash

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1529)

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -1,4 +1,4 @@
-SHELL:=/bin/bash
+SHELL:=/usr/bin/env bash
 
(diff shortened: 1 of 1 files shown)


NiLuJe (approved):
:+1:

Frenzie:
Do we even use any bashisms btw?

Baitinq:
> Do we even use any bashisms btw?

I dont think so, I have set an updated pr (#1530)

PR #1532, opened 2022-10-08
sdcv: Don't warn on unknown dictionaries

NiLuJe:
We handle dictionary whitelisting and sorting ourselves, so this isn't helpful information.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1532)


NiLuJe:
re: https://github.com/koreader/koreader-base/pull/1516#issuecomment-1272282346 ;).

poire-z (approved):

PR #1535, opened 2022-10-11
Build LuaJIT w/ table.pack support

NiLuJe:
It's unfortunately behind the Lua 5.2 extensions ifdef in a standard build (possibly because it's so common that it's usually implemented in Lua and monkey-patched over the table metatable?).

But it's trivial and doesn't actually poses an API issue, and it's helpful for a number of our use-cases.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1535)

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -6,6 +6,8 @@ self = false
 
 globals = {
     "G_defaults",
+    "table.pack",
+    "table.unpack",
 }
 
 exclude_files = {
--- a/ffi/util.lua
+++ b/ffi/util.lua
@@ -265,7 +265,7 @@ function util.execute(...)
         local pid = C.fork()
         if pid == 0 then
             local args = table.pack(...)
-            os.exit(C.execl(args[1], table.unpack(args, 1, #args+1))) -- Last arg must be a NULL pointer
+            os.exit(C.execl(args[1], unpack(args, 1, #args+1))) -- Last arg must be a NULL pointer
         end
         local status = ffi.new('int[1]')
         C.waitpid(pid, status, 0)
--- a/ffi/util.lua
+++ b/ffi/util.lua
@@ -264,8 +264,8 @@ function util.execute(...)
     else
         local pid = C.fork()
         if pid == 0 then
-            local args = {...}
-            os.exit(C.execl(args[1], unpack(args, 1, #args+1)))
+            local args = table.pack(...)
+            os.exit(C.execl(args[1], table.unpack(args, 1, #args+1))) -- Last arg must be a NULL pointer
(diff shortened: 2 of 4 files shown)


NiLuJe:
`ipairs` will still stop at the first nil, so a sparse table packed this ways needs to be walked via a for loop.

e.g.,

```
> t = table.pack(1, nil, 3)
> for i, v in ipairs(t) do print(i, v) end
1       1
> for i = 1, #t do print(i, t[i]) end
1       1
2       nil
3       3
```

NiLuJe:
LifeHack: there's nothing magical about the `n` field being used here (it's a remnant of old Lua semantics, IIRC).

```
> t = { n = 3, 1, nil, 3 }
> for i = 1, #t do print(i, t[i]) end
1       1
2       nil
3       3
```

Frenzie:
> It's unfortunately behind the Lua 5.2 extensions ifdef in a standard build (possibly because it's so common that it's usually implemented in Lua and monkey-patched over the table metatable?).

Should we just enable that? The stuff with break statements and whatnot sounds useful.

NiLuJe:
> Should we just enable that? The stuff with break statements and whatnot sounds useful.

A few of the other things are mildly annoying and would break existing code, though, so, ultimately, maybe, but not right now ;).

NiLuJe (commented):
> Reviewed 4 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1535)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

NiLuJe:
> LifeHack: there's nothing magical about the `n` field being used here (it's a remnant of old Lua semantics, IIRC).

Well, there *is* a bit of weird Lua clunk involved nonetheless ;p.

```
> t = { 1, 2, 3, nil, 4 }
> t.n = 5
> print(unpack(t))
1       2       3
> t = { n = 5, 1, 2, 3, nil, 4 }
> print(unpack(t))
1       2       3       nil     4
> t = { 1, 2, 3, nil, 4, n = 5 }
> print(unpack(t))
1       2       3
```

NiLuJe:
As discovered in https://github.com/koreader/koreader/pull/10346, Shit gets interesting when you throw more `nil`s in there...

```
> t = table.pack(1, nil, 3)
> print(#t)
3
> t = table.pack(1, nil, 3, nil)
> print(#t)
1
```

Sooo, yeah, better rely on the `n` , field instead of the `#` syntactic sugar for `len`...

PR #1536, opened 2022-10-13
xtext: Replace soft hyphens with hyphen at eol

zwim:
Replaces a soft-hyphen with a hyphen, if it is at the end of a line.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1536)


poire-z:
Superseded by #1543.

PR #1540, opened 2022-10-16
Support Android AArch64

NiLuJe:
~~Draft because this requires further trickery on the Kotlin side of things, AFAICT (i.e., this passes `kodev build`, but not `kodev release`).~~

NB: This passes *release* builds only. Debug builds will fail to link because of a weird ass assembler bug.

Re: https://github.com/koreader/koreader/issues/9632#issuecomment-1280073258

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1540)


toolchain/Makefile
```diff
@@ -19,7 +19,7 @@ android-sdk:
 	pushd $(SDK_DIR)/tools/bin && \
 		yes | ./sdkmanager --update && \
 		yes | ./sdkmanager --licenses && \
-		./sdkmanager "platform-tools" "build-tools;30.0.2" "platforms;android-30" "patcher;v4" && \
+		./sdkmanager "platform-tools" "build-tools;30.0.3" "platforms;android-30" "patcher;v4" && \
```
Frenzie, line 22: Does this require a Docker image update or is it just random?

NiLuJe: I have no idea how this works in the Docker image, because this is currently handled at runtime, AFAICT?

NiLuJe (commented):
> Reviewed 10 of 11 files at r1, 2 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1540)* status: all files reviewed, 1 unresolved discussion (waiting on @Frenzie)

toolchain/Makefile
Frenzie: The Docker image will proceed to download the, what is it, 1.5+ GB tools if they're not found but a minor detail like this probably doesn't trigger it. You can tell by what CircleCI does. ;-)

Frenzie: But the Docker image comes with them included in a stripped down manner (still at least 600 MB iirc) so it doesn't do that.

Makefile.defs
```diff
@@ -16,7 +16,11 @@ TOOLCHAIN_DIR=$(MAKEFILE_DIR)toolchain
 ANDROID_ARCH?=arm
 ANDROID_TOOLCHAIN=$(TOOLCHAIN_DIR)/android-toolchain-$(ANDROID_ARCH)
 NDK?=$(TOOLCHAIN_DIR)/android-ndk-r15c
-NDKABI?=14
+ifeq ($(ANDROID_ARCH), arm64)
```
pazos, line 19: x86_64?

Oh boy, I miss the android emulator :)

Frenzie: The newer ones don't run 32-bit x86 or what do you mean?

pazos: Nope :/

See https://github.com/koreader/koreader/issues/9446#issuecomment-1219345550

Frenzie: Holy @#$@#$.

NiLuJe (commented):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1540)* status: all files reviewed, 2 unresolved discussions (waiting on @Frenzie and @pazos)

Frenzie:
@NiLuJe I just noticed you didn't add an aarch64 build to the CI even though there's a fair bit of difference. Was that on purpose or just an oversight?

NiLuJe:
I was half waiting to deal with Clang first, but I haven't even begun dealing with that yet, so one as-is ought to do? It just can't be debug, as that won't link ;).

PR #1541, opened 2022-10-24
SDL: Fix a bunch of input timeout edge-cases

NiLuJe:
One being a failure to actually round properly, which lead to potentially passing a float to a C function that takes an int; and the other being that SDL's input timeout only has ms precision, which is much coarser than us, which meant that anything > 0 and < 1 ms was asking for an immediate return for as long as that fractional ms ticked. Instead, we now round up to 1ms in these cases.

Goes hand in hand w/ https://github.com/koreader/koreader/commit/d034e1a71955b45ce02587fdcd0a58b9f491de50

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1541)

### DIFF
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -270,15 +270,10 @@ local SDL_BUTTON_LEFT = 1
 local is_in_touch = false
 
 function S.waitForEvent(sec, usec)
-    -- If we're asking to sleep for a non-zero amount of time, but for less than 1ms (i.e., below our precision),
-    -- round *up*, to avoid a bunch of roundtrips during that fractional ms.
-    if sec == 0 and usec ~= 0 and usec < 1000 then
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
SDL has such low precision? I never realized.

NiLuJe:
IIRC, they discourage from using that model (but this one fits very nicely into our own ;p). FWIW, `poll` has the same precision, though (which is sorta why I kept using `select` in our C module ;), despite poll's API being much nicer to deal with).

NiLuJe (commented):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1541)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

PR #1543, opened 2022-10-26
cre & xtext: more soft hypheness

poire-z:
#### cre.cpp: add `cre.softHyphenateText(lang, text)`

Allow getting any text soft-hyphenated according to crengine hyphenation dicts. May be used by frontend's textboxwidget to hyphenate text before wrapping and rendering it with xtext. See https://github.com/koreader/koreader/pull/9630#issuecomment-1278158193.
Bump crengine for https://github.com/koreader/crengine/pull/492 :
- TextLang: add `TextLangCfg->softHyphenateText(lang, text)`

#### xtext: properly support soft-hyphens

If soft-hyphens were present in the input text, it was considered (by libunibreak) as a candidate for breaking a line, but no real hyphen would be shown when the line is rendered.
So, be sure we properly draw a real hyphen when shaping a line ending with a soft-hyphen.
(Refactor the ellipsis substitution code so we can reuse it with hyphen substitution.)
See https://github.com/koreader/koreader/pull/9630#issuecomment-1279339219 and followups.

#### xtext: option to allow limiting usage of soft-hyphens

Add `makeLine(..., expansion_pct_rather_than_hyphen)`, which can be provided non-zero if text is going to be justified, to avoid small hyphenated words. Ie, with 100:
when about to wrap on a soft hyphen, if a wrap on a previous non-hyphen candidate would get justification to expand the spaces by less than 100% (so, at most double width spaces), consider this expansion admissible and better than this hyphen (which would probably make a small part of a word hyphenated at end of this line).

#### xtext: better ellipsis support with Arabic text

Subtituting a char with an ellipsis on Arabic text could change the "form" (initial, medial, final) of that char, possibly making the meaning of the truncated word ambiguous, but also longer and overflowing the targeted width.
Fix this by substituting with Zero-width-joinder and an ellipsis when needed, which should hopefully keep the previous char's original form.
See https://github.com/koreader/koreader/issues/5359#issuecomment-1287904250 and previous posts/attemps/discussions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1543)

### DIFF
--- a/xtext.cpp
+++ b/xtext.cpp
@@ -66,6 +66,7 @@ extern "C"
 #define NOT_MEASURED INT_MIN
 #define REPLACEMENT_CHAR 0xFFFD
 #define ELLIPSIS_CHAR 0x2026
+#define ZERO_WIDTH_JOINER_CHAR 0x200D
 #define SOFTHYPHEN_CHAR 0x00AD
 #define REALHYPHEN_CHAR 0x002D
 
@@ -248,6 +249,8 @@ int Utf8ToUnicode(const char * src,  int srclen, uint32_t * dst, int dstlen, boo
 #define CHAR_IS_PARA_END                 0x0200
 #define CHAR_PARA_IS_RTL                 0x0400 /// to know the line with this char is part
                                                 /// of a paragraph with main dir RTL
+#define CHAR_IS_UNSAFE_TO_BREAK_BEFORE   0x0800 /// from HarfBuzz (set when kerning or on arabic when
+                                                /// initial/medial/final char forms are involved)
 #define CHAR_IS_TAB                      0x1000 /// char is '\t'
 
 // Info, after measure(), about each m_text char
@@ -351,11 +354,15 @@ public:
     }
 
     void allocate() {
-        m_charinfo = (xtext_charinfo_t *)calloc(m_length, sizeof(*m_charinfo)); // set all flags to 0
+        // We allocate one slot more than m_length, for the case we would have to
+        // truncate with an ellipsis the last char and we need to actually use a
+        // ZERO_WIDTH_JOINER before the ELLIPSIS (which might be needed with Arabic)
+        size_t size = m_length + 1;
+        m_charinfo = (xtext_charinfo_t *)calloc(size, sizeof(*m_charinfo)); // set all flags to 0
         if ( m_has_rtl ) {
-            m_bidi_ctypes = (FriBidiCharType *)malloc(m_length * sizeof(*m_bidi_ctypes));
-            m_bidi_btypes = (FriBidiBracketType *)malloc(m_length * sizeof(*m_bidi_btypes));
-            m_bidi_levels = (FriBidiLevel *)malloc(m_length * sizeof(*m_bidi_levels));
+            m_bidi_ctypes = (FriBidiCharType *)malloc(size * sizeof(*m_bidi_ctypes));
+            m_bidi_btypes = (FriBidiBracketType *)malloc(size * sizeof(*m_bidi_btypes));
+            m_bidi_levels = (FriBidiLevel *)malloc(size * sizeof(*m_bidi_levels));
         }
     }
     void deallocate() {
@@ -380,7 +387,8 @@ public:
         // count the number of unicode codepoints, before allocating m_text,
         // and a 2nd to actually do the conversion and fill m_text.
         m_length = Utf8ToUnicode(utf8_text, utf8_len, NULL, 0, m_is_valid, m_has_rtl);
-        m_text = (uint32_t *)malloc(m_length * sizeof(*m_text));
+        size_t size = m_length + 1; // (one slot more, see above)
+        m_text = (uint32_t *)malloc(size * sizeof(*m_text));
 
         // m_has_rtl is only detected in the 2nd phase.
         // If m_para_direction_rtl is true, set m_has_rtl=true in all case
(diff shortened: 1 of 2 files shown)


xtext.cpp
```diff
@@ -347,11 +354,15 @@ class XText {
     }
 
     void allocate() {
-        m_charinfo = (xtext_charinfo_t *)calloc(m_length, sizeof(*m_charinfo)); // set all flags to 0
+        // We allocate one slot more than m_length, for the case we would have to
+        // truncate with an ellipsis the last char and we need to actually use a
+        // ZERO_WIDTH_JOINER before the ELLIPSIS (which might be needed with Arabic)
+        int size = m_length + 1;
```
NiLuJe, line 360: `size_t ` ;o).

xtext.cpp
```diff
@@ -376,7 +387,8 @@ class XText {
         // count the number of unicode codepoints, before allocating m_text,
         // and a 2nd to actually do the conversion and fill m_text.
         m_length = Utf8ToUnicode(utf8_text, utf8_len, NULL, 0, m_is_valid, m_has_rtl);
-        m_text = (uint32_t *)malloc(m_length * sizeof(*m_text));
+        int size = m_length + 1; // (one slot more, see above)
```
NiLuJe, line 390: Ditto

xtext.cpp
```diff
@@ -402,7 +414,8 @@ class XText {
     // a single good one).
     void setTextFromUTF8CharsLuaArray(lua_State * L, int n) {
         m_length = (int) lua_objlen(L, n); // NOTE: size_t -> int, as that's what both FriBidi & HarfBuzz expect.
-        m_text = (uint32_t *)malloc(m_length * sizeof(*m_text));
+        int size = m_length + 1; // (one slot more, see above)
```
NiLuJe, line 417: Ditto

xtext.cpp
```diff
@@ -1094,6 +1133,25 @@ class XText {
                     candidate_end = i > start ? i-1 : start;
                     next_line_start_offset = i+1;
                 }
+                else if ( m_text[i] == SOFTHYPHEN_CHAR ) {
+                    // A soft-hyphen has a width of 0. But if we end this line on it,
+                    // it should be rendered as a real hyphen with a width.
+                    // So, to consider it a candidate for end, be sure we still won't
+                    // exceeed the targeted width when it is replaced.
```
NiLuJe, line 1140: nit: `exceed`

xtext.cpp
NiLuJe: That was on purpose: that's the signature of malloc and sizeof ;).

NiLuJe (approved):

poire-z:
I forgot something I wanted to check/add to avoid having small parts of hyphenated words at end of line (see screenshots at bottom of https://github.com/koreader/koreader/pull/9630#issuecomment-1279584226), so added something in the spirit of how crengine does that:

> #### xtext: option to allow limiting usage of soft hyphens
> Add makeLine(..., avoid_soft_hyphen_when_filled_percent), allowing not breaking on a soft-hyphen if the line is already filling the provided percentage of the target width.
Ie. with 90 %: if the last normal break candidate (a space) is at 93%, no need to consider next soft-hyphens candidates: best to expand the spaces for justification than having a small part of a hyphenated word at end.
With 100%: any soft-hyphen break will be used.
With 0%: no soft-hyphen break will be used.

crengine would use 95%, but as xtext doesn't expand/shrink spaces widths when computing, we won't do as nice, and we may need to use 90% or 92% to avoid small parts.
I'm sure we could do better, this is just a quick addition so we can test on frontend side if we need it, and/or if we need something better (so, made as an independant commit that we can quickly revert).

Any idea for a better and shorter naming than my awkward `avoid_soft_hyphen_when_filled_percent = 95`  :)

NiLuJe:
`stop_hyphenation_at`?

poire-z:
Reworked my above quick and dirty evening commit to do that in a more sensible and less ugly manner:

> #### xtext: option to allow limiting usage of soft hyphens
> Add tunable global setting `expansion_pct_rather_than_hyphen` to avoid soft-hyphens, with a default of 100%:
When about to wrap on a soft hyphen, if a wrap on a previous non-hyphen candidate would get justification to expand the spaces by less than 100% (so, at most double-width spaces), consider this expansion admissible and better than this hyphen (which would probably make a small part of a word hyphenated at end of this line).

This gives much better results, and feels less language dependant that my previous `avoid_soft_hyphen_when_filled_percent` which I thought would need different values per language.

My default of 100% can be tuned on frontend with:
```diff
--- a/frontend/ui/bidi.lua
+++ b/frontend/ui/bidi.lua
@@ -58,6 +58,7 @@ function Bidi.setup(lang)
     -- Xtext default language and direction
     if G_reader_settings:nilOrTrue("use_xtext") then
         local xtext = require("libs/libkoreader-xtext")
+        xtext.setExpansionPercentageRatherThanHyphen(200)

         -- Text direction should normally not follow ui mirroring
         -- lang override (so that Arabic is still right aligned
```
(set `0` to get this feature disabled).

poire-z:
> Reworked my above quick and dirty evening commit 

Too quick morning commit... We don't want that setting to always apply: only when text is justified.
And my previous "when filled percent" would work better when text left-aligned (but given the limited taste for left aligned+hyphenation, no need to really care about it).

PR #1544, opened 2022-10-27
ffi: Add fflush to posix_h

NiLuJe:
Necessary for `fsync` shenanigans ;).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1544)


poire-z:
(Should I bring this one with my bump for xtext and other libs ?)

NiLuJe:
Nah, I'll do it later tonight ;).

(I have other unpushed changes behind).

NiLuJe:
Actually, no, it's good, the rest will be in the KB PR, so this can go in ;).

PR #1547, opened 2022-10-30
Unbreak Luarocks on macOS (maybe)

NiLuJe:
No idea why it was trying to be a special snowflake: we build our own LuaJIT, use it!

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1547)


NiLuJe:
w00t!

PR #1548, opened 2022-10-30
BB: Add color constants for the full eInk palette

NiLuJe:
And rename a couple of confusing/dumb ones.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1548)


Frenzie (approved):
You agreed when I proposed to lift the names from X11. ;-P

NiLuJe:
Oh, that's where it's coming from ;p.

Yeah, I guess it was better than nothing when it was on its own, but now it just feels weird among all the hex ones ;).

PR #1553, opened 2022-11-17
mupdf: add patch to support sections in comicbooks

mroethke:
I wrote a patch to add support for sections, based on folders in the archive, in the comic book formats (cbz,cbt) in mupdf.
I have not tried to get this upstream because upstream does not seem interested in outside patches, judging by what is written on the mupdf website. But I still would like to see this functionality in koreader.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1553)

### DIFF
--- a/thirdparty/mupdf/mupdf_cbz_chapter_support.patch
+++ b/thirdparty/mupdf/mupdf_cbz_chapter_support.patch
@@ -1,0 +1,126 @@
+commit b15ca8974ab7a686c09f70a0025ddda8097d0194
+Author: Marcel Röthke <marcel@roethke.info>
+Date:   Sun Nov 13 22:10:06 2022 +0100
+
+    add support for chapters in comic book (cbz/cbt) archive
+    
+    Chapters can be created in a comic book archive by putting the images in
+    directories. The directory names will be the chapter names. They can be
+    nested up to 10 times. This not a technical limitation, but it does not
+    really make sense to nest chapters arbitrarily deep. Ordering of
+    chapters will abide by the same rules as the ordering for other files in
+    a comic book archive.
+
+diff --git a/source/cbz/mucbz.c b/source/cbz/mucbz.c
+index 0c93aa458..413205fdf 100644
+--- a/source/cbz/mucbz.c
++++ b/source/cbz/mucbz.c
+@@ -233,6 +233,100 @@ cbz_lookup_metadata(fz_context *ctx, fz_document *doc_, const char *key, char *b
+ 	return -1;
+ }
+ 
++static fz_outline *
++cbz_load_outline(fz_context *ctx, fz_document *doc_)
++{
++	cbz_document *doc = (cbz_document *) doc_;
++	fz_outline *outline = fz_new_outline(ctx);
++	size_t max_chapter_stack_depth = 10;
++	fz_outline *chapter_stack[max_chapter_stack_depth];
++	size_t chapter_stack_depth = 0;
++
++	for (int i = 0; i < doc->page_count; i++)
++	{
++		const char *pagename = doc->page[i];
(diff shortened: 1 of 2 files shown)


Frenzie:
What's written on the MuPDF website?

thirdparty/mupdf/mupdf_cbz_chapter_support.patch
```diff
@@ -0,0 +1,120 @@
...
++	cbz_document *doc = (cbz_document *) doc_;
++	fz_outline *outline = fz_new_outline(ctx);
++	size_t max_chapter_stack_depth = 10;
++	fz_outline *chapter_stack[max_chapter_stack_depth];
++	size_t chapter_stack_depth = 0;
++
++	for (int i = 0; i < doc->page_count; i++)
++	{
++		const char *pagename = doc->page[i];
++		const char *next_seperator;
```
NiLuJe, line 27: nit: sep**a**rator

thirdparty/mupdf/mupdf_cbz_chapter_support.patch
```diff
@@ -0,0 +1,120 @@
+commit bfae5958ba7ebeb7bb58c0a0e56bb6e521ef6d9f
+Author: Marcel Röthke <marcel@roethke.info>
+Date:   Sun Nov 13 22:10:06 2022 +0100
+
+    add support for chapters in cbz
```
NiLuJe, line 5: An actual description of how this actually works and how it expects CBZ to be organized would be helpful.

mroethke:
@Frenzie 
I mean this:
https://mupdf.com/licensing/index.html
> The vast majority of all MuPDF development is done by Artifex engineers, and on rare occasions, bug fixes accepted from outside contributors (under our contribution agreement).

That would stop me from even trying to get this upstream even if did not have an issue with CLAs.

@NiLuJe thanks for the feedback, all fixed!

NiLuJe:
Sounds good to me, and it's nice and self-contained and all ;p.

mroethke:
Awesome! Thanks for merging!

poire-z:
No risk having the mupdf patch in the coming up v2022.11 release ?
It works with flat cbz ? No noticable slow down loading and navigating books ? (not reading cbz books, so I have no reference feeling about that :)

Frenzie:
Unless there are thousands of pages I can't see how it would affect much of anything, though I admit I didn't bother making a test document with thousands of pages.

I seem to recall writing and pasting a script to do something like that a couple of years ago…

NiLuJe:
Yeah, I'm actually mostly concerned about what happens on a flat document (which I assume is "nothing" ;p).

Frenzie:
No noticeable impact on a regular ~100 page flat document. (I don't think I have anything else if I don't create it…)

mroethke:
I would not expect a noticeable change for navigation, but loading might take a little longer for large documents, because it does add an additional pass over the already sorted file list.

I can currently only test in the linux emulator where I do not notice a difference, but this is probably not representative.

KaMyKaSii:
@mroethke thank you so much for this! I have a Kindle Paperwhite 10th and I didn't notice any delay to open the file and I usually make cbz files with around 900 images. I just noticed that letters like "í" don't render correctly, maybe it needs better unicode support?
Reader_2022-12-11_205358

NiLuJe:
Encoding and archives is (usually) a rather horrible mess, but I have no idea how MuPDF deals with it in this instance ;).

mroethke:
I do not know for sure either, but utf8 seems to work tough the sorting could be improved.
@KaMyKaSii can you try converting the folder names to utf8?

window-2022-12-15-17_39_21

PR #1554, opened 2022-11-18
mupdf: fix crash with flat comicbook archives

mroethke:
In a flat comicbook archive, where there are no chapters, `cbz_load_outline` would return an empty `struct fz_outline*`. This causes the ffi wrapper to crash, because it tries to call `strlen` on a string that is `NULL`.
This is fixed by returning NULL instead of the empty `struct fz_outline *`.

I'm sorry I should have tested this, but obviously did not :(

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1554)


Frenzie:
Hm, I thought I'd tested a flat comic book archive, and for sure with your patch applied. I just shut down my PC, but perhaps it was a single folder inside the zip file?

Frenzie (approved):

mroethke:
I thought the same, but then I generated a large flat file for a little performance check …

NiLuJe (approved):

PR #1555, opened 2022-11-27
libunibreak: patch to speed up line break property lookup

bbshelper:
Currently it uses 2-level linear probe. Use LUT for BMP, and binary search for higher codepoints.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1555)


bbshelper:
Previously discussed in https://github.com/koreader/crengine/issues/494. Patch generated from https://github.com/bbshelper/libunibreak/commit/aee6dbc4d2d262709af92483f619316b40cd30db

poire-z:
Is it possible to make the patch smaller, so it is also easier to apply next time upstream libunibreak is bumped ?
ie.
- by commenting with `/*` and `*/` on standalone lines around the big section you removed, instead of removing it?
- by `#include "the_file_your_patch_would_generate.h"` instead of the big section you are adding?

poire-z:
And include your `generate.py` (with a more explicite name) here (so we're standalone, in case your github repo disappears), and mention in it where to get `LineBreak.txt` - and which version of Unicode you got it for, and if different from the one used by upstream libunibreak.

Frenzie:
Also note that I wasn't necessarily objecting to making Python a build requirement, just stating that running a Python script would require more action than just sticking it in there.

bbshelper:
Thanks you both for inputs. Regarding the commenting thing, I should've remembered that :(.

My impression is that the build environment is already somewhat complex, and since the patch can't really be made small without adding Python and downloading LineBreak.txt. I'm trying upstreaming first.

bbshelper:
Upstream has merged the change. Closing.

poire-z:
Will you do the libunibreak bump, or should we ?
I guess you have checked it compiles and works and is still speedy with the other additional upstream commits - and we can go at that blindly ? :)

bbshelper:
> Will you do the libunibreak bump, or should we ?

With upstream release 5.1, it seems to be trivial so I submitted https://github.com/koreader/koreader-base/pull/1563

> I guess you have checked it compiles and works and is still speedy with the other additional upstream commits - and we can go at that blindly ? :)

It compiles and works on my side with expected performance.  In fact, the only other code change in the 5.1 release is data file update with Unicode 15.0, which should be safe.

poire-z:
Thanks for the work, and the work with upstream.
Wanna do the base bump into frontend (you'll bring 2 minor pocketbook commits with it, no pb) and get frontend fame ? :)

bbshelper:
Thanks for the kind suggestion :) I've submitted https://github.com/koreader/koreader/pull/9906

PR #1558, opened 2022-12-03
mupdf: Fix corrupt pdf metadata info if longer than 255 bytes

mroethke:
The function fz_lookup_metadata returns the length the metadata item actually has, not how much was written into the supplied buffer which is what the caller expects. This matters if the item is longer than the supplied fixed size buffer because ffi.string will read out of bounds and include the garbage it reads in the resulting lua string.

This is fixed by checking the return value and calling fz_lookup_metadata again with an appropriately sized buffer if the metadata item is longer than the initial buffer.

Another option would be to call the function first with a zero sized buffer, but than every call would pay the cost of two calls instead of only those calls that actually need a larger buffer.

koreader/koreader#9855

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1558)

### DIFF
--- a/ffi/mupdf.lua
+++ b/ffi/mupdf.lua
@@ -276,8 +276,14 @@ function document_mt.__index:openPage(number)
 end
 
 local function getMetadataInfo(doc, info)
-    local buf = ffi.new("char[?]", 255)
-    local res = M.fz_lookup_metadata(context(), doc, info, buf, 255)
+    local bufsize = 255
+    local buf = ffi.new("char[?]", bufsize)
+    local res = M.fz_lookup_metadata(context(), doc, info, buf, bufsize)
+    if res > bufsize then
+        bufsize = res + 1
+        buf = ffi.new("char[?]", bufsize)
+        res = M.fz_lookup_metadata(context(), doc, info, buf, bufsize)
+    end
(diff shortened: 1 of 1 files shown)


Frenzie (approved):

Frenzie:
> Another option would be to call the function first with a zero sized buffer, but than every call would pay the cost of two calls instead of only those calls that actually need a larger buffer.

I very much doubt that could ever matter, but sure, let's put a good example in the code for when it might. :smile:

mroethke:
Probably not, but it is post hoc rationalization anyway :D
I wrote it this way and while writing the commit message realized that another way exists and proceeded to search for a good reason not to change it.

Frenzie:
I just remembered you can do this:


Of course if the description is even stored by that (probably not, don't know otoh) it still won't make a difference in the grand scheme of things. ;-)

NiLuJe:
That's `snprintf`-like behavior, so, yeah, both approaches work (and I would tend to stick to defaulting to a sensible buffer size first instead of always assuming the worst).

The "always ask" approach makes sense when you really have no idea how large the result will be.

NiLuJe (approved):

PR #1559, opened 2022-12-04
Enable SSH server for pocketbook devices

rjd22:
Pocketbooks have a internal dropbear SSH server that is not used (I'm not sure why but most likely they use it for development). I was able to boot this SSH server so I was wondering if I it would be possible to use KoReaders dropbear server implementation with the no password improvements.

After making my own build of them it seems that this is actually working fine with some restrictions. You can only login with the `reader` user and don't have root access. But it will make debugging Pocketbooks a lot easier!

The command to login into pocketbooks is:

```
ssh -T reader@10.0.0.251 -p 2222
```

For some reason there is no tty access even when `lsof /dev/pts/*` gives results do the `-T` flag is needed else the following error will be thrown:

```
debug2: channel_input_status_confirm: type 100 id 0
PTY allocation request failed on channel 0
debug3: receive packet: type 100
debug2: channel_input_status_confirm: type 100 id 0
shell request failed on channel 0
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1559)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,8 +30,8 @@ all: $(OUTPUT_DIR)/libs $(if $(ANDROID),,$(LUAJIT)) \
 		$(if $(WIN32),,$(OUTPUT_DIR)/sdcv) \
 		$(if $(MACOS),$(OUTPUT_DIR)/koreader,) \
 		$(if $(MACOS),$(SDL2_LIB),) \
-		$(if $(or $(CERVANTES),$(KINDLE),$(KOBO)),$(OUTPUT_DIR)/dropbear,) \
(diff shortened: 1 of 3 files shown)


Frenzie (approved):

NiLuJe (approved):

NiLuJe:
The tty issue *might* need to be documented somewhere. There might be a wiki page for the SSH stuff already?

Frenzie:
Yes, it's basically a c/p of a quick post I once wrote in an issue somewhere.
https://github.com/koreader/koreader/wiki/SSH

Frenzie:
And your issue's already on it I think: https://github.com/koreader/koreader/wiki/SSH#ssh-fails-with-error-message-packet_write_wait-connection-to--port-2222-broken-pipe

PR #1561, opened 2022-12-05
Third-party bumps

NiLuJe:
* FBInk 1.25.0 (look, a tag! :D).
* KoboUSBMS 1.3.5
* SQLite 3.40.0
* LuaJIT 20221201
* luafilesystem w/ precission loss patch (re: https://github.com/koreader/koreader/issues/9860)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1561)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -111,7 +111,10 @@ $(OUTPUT_DIR)/libs/libkoreader-input.so: input/*.c input/*.h $(if $(or $(KINDLE)
 $(OUTPUT_DIR)/libs/libkoreader-lfs.so: \
 			$(if $(USE_LUAJIT_LIB),$(LUAJIT_LIB),) \
 			luafilesystem/src/lfs.c
+	# Avoid precision loss on 32-bit arches (LFS is always built w/ LARGEFILE support, but lua_Integer is always a ptrdiff_t, which is not wide enough).
+	-patch -d luafilesystem -t -N --no-backup-if-mismatch -r - -p1 < patches/lfs-pushnumber-for-wide-types.patch
 	$(CC) $(DYNLIB_CFLAGS) -o $@ luafilesystem/src/lfs.c $(LUAJIT_LIB_LINK_FLAG)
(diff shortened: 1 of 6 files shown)


Frenzie (approved):
What about upstream lfs?

Makefile
```diff
@@ -111,7 +111,10 @@ $(OUTPUT_DIR)/libs/libkoreader-input.so: input/*.c input/*.h $(if $(or $(KINDLE)
 $(OUTPUT_DIR)/libs/libkoreader-lfs.so: \
 			$(if $(USE_LUAJIT_LIB),$(LUAJIT_LIB),) \
 			luafilesystem/src/lfs.c
+	# Avoid precision loss on 32-bit arches (LFS is always built w/ LARGEFILE support, but lua_Integer is always a ptrdiff_t, which is not wide enough).
+	-patch -d luafilesystem -t -N --no-backup-if-mismatch -r - -p1 < patches/lfs-pushnumber-for-wide-types.patch
 	$(CC) $(DYNLIB_CFLAGS) -o $@ luafilesystem/src/lfs.c $(LUAJIT_LIB_LINK_FLAG)
+	-patch -d luafilesystem -t -R --no-backup-if-mismatch -r - -p1 < patches/lfs-pushnumber-for-wide-types.patch
```
poire-z, line 117: Twice ?
(I guess the leading dash is a Makefile thing, and not a typo, or a command line option that would be missing a continuation `\` on the previous line.)

NiLuJe: The second one is to revert it so as not to pollute the git status.

The leading dot is to allow failure in case shit goes wrong somewhere and we attempt to apply it again while it's already applied ;).

NiLuJe:
> What about upstream lfs?

I'd like to see how it fares in the wild first ;).

(Plus, there's the fact that lua_Integer does point to a real integer type on Lua 5.3+ for upstream).

PR #1565, opened 2022-12-15
djvulibre: don't mangle locales

NiLuJe:
It's generally a bad idea, because this runs in our main thread, and locales in POSIX are a mess.

May help w/ https://github.com/koreader/koreader/issues/9918

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1565)


NiLuJe:
Picked a few upstream fixes while I was in there ;o).

thirdparty/djvulibre/CMakeLists.txt
```diff
@@ -38,6 +38,13 @@ if(DEFINED ENV{ANDROID})
     set(PATCH_CMD3 "${KO_PATCH} ${CMAKE_CURRENT_SOURCE_DIR}/android.patch")
 endif()
 
+# Pick a few fixes from upstream
+set(PATCH_CMD4 "${KO_PATCH} ${CMAKE_CURRENT_SOURCE_DIR}/djvulibre-3.5.28-to-6a1e5.patch")
```
Frenzie, line 42: Umm… why?

Frenzie: I mean, why not just write `6a1e5ba1` some 5 lines down instead of some convoluted patch.

NiLuJe: Honestly? To make sure the diff applied cleanly? ;o)

NiLuJe: Oh, that :D. Yeah, generated the patch before checking that we used the git repo and not a tarball :D.

thirdparty/djvulibre/djvulibre-no-locale-mangling.patch
```diff
@@ -0,0 +1,62 @@
...
+--- a/tools/common.h
++++ b/tools/common.h
+@@ -76,8 +76,6 @@
+        djvu_programname(argv[0]); } while (0)
+ #else
+ # define DJVU_LOCALE do {          \
+-       setlocale(LC_ALL,"");       \
+-       setlocale(LC_NUMERIC,"C");  \
+        djvu_programname(argv[0]); } while (0)
+ #endif
```
Frenzie, line 61: This one does seem a bit curious and in need of patching.

thirdparty/djvulibre/djvulibre-no-locale-mangling.patch
```diff
@@ -0,0 +1,62 @@
...
+@@ -88,14 +88,7 @@
+ #endif
+ #include <ctype.h>
+ 
+-#ifndef LC_NUMERIC          //MingW
+-# undef DO_CHANGELOCALE
+-# define LC_NUMERIC 0
+-#endif
+-#ifndef DO_CHANGELOCALE
+-# define DO_CHANGELOCALE 0
```
Frenzie, line 14: Might be better to just `-DDO_CHANGELOCALE=0` or whatever instead of doing a patch that needs to be maintained? (Although this particular dependency is fairly static.)

Frenzie (approved):
lgtm, but I think the patch to libdjvu/GString.cpp may not be necessary as stated in the comment

thirdparty/djvulibre/djvulibre-no-locale-mangling.patch
NiLuJe: Yeah, I went with the nuclear option just to be safe (and because I already had a patch to generate anyway).

thirdparty/djvulibre/djvulibre-no-locale-mangling.patch
NiLuJe: It should only affect tools, though, which we probably doesn't use ;o).

The likely culprit is likely the context creation once, I would imagine.

(Again, I did a very wide sweep just to be safe).

thirdparty/djvulibre/djvulibre-no-locale-mangling.patch
Frenzie: Alright.

thirdparty/djvulibre/djvulibre-no-locale-mangling.patch
Frenzie: I didn't look at the context in detail, common.h does sound quite, ahem, common though xp

PR #1566, opened 2022-12-18
Enable gsensor EVT_ORIENTATION events on modern Pocketbook devices with InkView version 6.7

rjd22:
This enabled the GSensor for modern PocketBook devices using the still unreleased firmware 6.7 and later.

This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1566)

### DIFF
--- a/ffi/input_pocketbook.lua
+++ b/ffi/input_pocketbook.lua
@@ -163,7 +163,7 @@ function input:open()
         return
     end
 
-    -- This enables the inkview orientation events to be send to translateEvent
+    -- This enables the inkview orientation events to be sent to translateEvent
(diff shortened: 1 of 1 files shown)


ffi/input_pocketbook.lua
```diff
@@ -147,6 +147,10 @@ function input:open()
         inkview.iv_setup_gsensor()
         return
     end
+
+    -- This enables the inkview orientation events to be send to translateEvent
```
NiLuJe, line 151: nit: sen***t*** ;).

rjd22:
This should be good to go now. Fix for koreader is also incoming.

NiLuJe (approved):

PR #1568, opened 2022-12-20
Groundwork for a refactor of gyro handling

NiLuJe:
Namely, we only handle a *single* protocol in Input, a custom one (as there's no real standard), based on a custom `EV_MSC` event code, `MSC_GYRO`.

The translation is left to Device implementations, keeping platform-specific code out of Input.

Since that happens to happen in base for PocketBook, do it here ;). (re: https://github.com/koreader/koreader/pull/9873)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1568)


ffi/custom_input_h.lua
```diff
@@ -0,0 +1,19 @@
...
+-- (i.e., this is how the device is *physically* oriented).
+ffi.cdef[[
+static const int EV_SDL = 83;
+
+static const int MSC_GYRO = 71;
+
+static const int DEVICE_ORIENTATION_UPRIGHT = 0;
+static const int DEVICE_ORIENTATION_CLOCKWISE = 1;
+static const int DEVICE_ORIENTATION_UPSIDE_DOWN = 2;
+static const int DEVICE_ORIENTATION_COUNTER_CLOCKWISE = 3;
```
rjd22, line 18: Question: is there a reason we name these differently from the ones that are in screen? Naming them differently requires extra mental mapping IMO.

rjd22 (approved):

ffi/custom_input_h.lua
NiLuJe: Mainly because I'm not entirely happy with the framebuffer ones ;p.

But,byeah, probably too late for that, will tweak this evening.

NiLuJe: I went with the nuclear option and actually renamed the Screen ones, mwahahahaha :}.

Frenzie: Is there some kind of internal kernel name btw or just the 0, 1, 2, 3?

NiLuJe: Yup, in `<linux/fb.h>`:

https://github.com/torvalds/linux/blob/b6bb9676f2165d518b35ba3bea5f1fcfc0d969bf/include/uapi/linux/fb.h#L231-L237

NiLuJe: Which is why you see me peppering UR, CW, UD, CCW comments everywhere ;).

rjd22: > I went with the nuclear option and actually renamed the Screen ones, mwahahahaha :}.

Good God... The man has gone insane... :P

NiLuJe (commented):
> Reviewed 2 of 4 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1568)* status: all files reviewed, 1 unresolved discussion (waiting on @Frenzie)

rjd22 (approved):

PR #1572, opened 2023-01-12
Move natural sorting algorithm to ffi/util

NiLuJe:
And implemented the hardened variant by Egor Skriptunoff

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1572)

### DIFF
--- a/ffi/util.lua
+++ b/ffi/util.lua
@@ -705,6 +705,42 @@ function util.orderedPairs(t)
     return orderedNext, t, nil
 end
 
+
+--[[--
+Natural sorting functions, for use with table.sort
+<http://notebook.kulchenko.com/algorithms/alphanumeric-natural-sorting-for-humans-in-lua>
+--]]
+-- Original implementation by Paul Kulchenko
+--[[
+local function addLeadingZeroes(d)
+    local dec, n = string.match(d, "(%.?)0*(.+)")
+    return #dec > 0 and ("%.12f"):format(d) or ("%s%03d%s"):format(dec, #n, n)
+end
+function util.natsort(a, b)
+    return tostring(a):gsub("%.?%d+", addLeadingZeroes)..("%3d"):format(#b)
+            < tostring(b):gsub("%.?%d+", addLeadingZeroes)..("%3d"):format(#a)
+end
+--]]
+-- Hardened (but more expensive) implementation by Egor Skriptunoff, with an UTF-8 tweak by Paul Kulchenko
+local function natsort_conv(s)
+    local res, dot = "", ""
+    for n, m, c in tostring(s):gmatch("(0*(%d*))(.?)") do
+        if n == "" then
+            dot, c = "", dot..c
+        else
+            res = res..(dot == "" and ("%03d%s"):format(#m, m)
+                                   or "."..n)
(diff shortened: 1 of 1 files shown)


NiLuJe:
RFC: This implementation fixes minor inconsistencies (quoting Kulchenko's example):

> when a number in one string is in a position where a number is absent in some other string. For example: when sorting `{"dir\\file.lua", "dir2\\file.lua", "dir3\\file.lua", "dir3\\file2.lua"}` using the algorithm above, you'd get `{"dir2\\file.lua", "dir3\\file.lua", "dir3\\file2.lua", "dir\\file.lua"}`, which is probably not what you'd expect.

> If you sort the same table using this function, you'll get the sorted table you'd likely expect: `{"dir\\file.lua", "dir2\\file.lua", "dir3\\file.lua", "dir3\\file2.lua"}`.

The drawback being that it is roughly twice as expensive. Now, I don't actually use this myself in the FM, but when trying it on a H2O, the delay is noticeable.
Should we care?
Should we keep both imps and use the faster one in the FM?

poire-z:
Any reason for this to be in ffi/util.lua if it relies on nothing C/ffi ? Could be in frontend/util.lua, no ? Or is it used in base/ ?

NiLuJe:
I waffled a bit on that one; the `strcoll` stuff is in base, but it makes sense, because it's actually C (... sometimes ;p).

So, absolutely no objection about moving this to front if you guys prefer it that way ;).

poire-z:
> the strcoll stuff is in base

What uses that strcoll stuff? And shouldn't that "what" then use this new natsort stuff?

NiLuJe:
> What uses that strcoll stuff? And shouldn't that "what" then use this new natsort stuff?

The FM (via FileChooser, which is where the sorting method selection is implemented).

And ReaderDictionary, for... something :D

poire-z:
> And ReaderDictionary, for... something :D

I guess for the initial sorting of dictionary names (a metadata string in the .ifo).

No idea how strcol does different from your natural sorting, so I too can't answer what you didn't answer :):

> And shouldn't that "what" then use this new natsort stuff?

NiLuJe:
> And shouldn't that "what" then use this new natsort stuff?

Well, it already does?

poire-z:
> Well, it already does?

Uh, but it doesn't if it uses strcol ? :)

But strcol is different than natural sorting (which just deal with numbers), as it's about the locale:
https://www.man7.org/linux/man-pages/man3/strcoll.3.html
```
       The comparison is based on strings interpreted
       as appropriate for the program's current locale for category
       LC_COLLATE.  
```

So, probably not replaceable by your naturalsort stuff.

poire-z:
> The reasoning was that it miiiight be used in base, theoretically, but, yeaaaah ;p.
So, absolutely no objection about moving this to front if you guys prefer it that way ;).

But yes, if it does not need any C/ffi, and is somehow unrelated in purpose to strcol, and we may only need natural-sorting stuff for UI display, better in frontend/util.lua.

NiLuJe:
Everything's now in front @ https://github.com/koreader/koreader/pull/10023 ;).

PR #1576, opened 2023-02-04
mxcfb: Implement fenced updates modes ([partial] & [ui])

NiLuJe:
There's no per-update flag on mxcfb (instead there are global no-merge UPDATE_SCHEME modes). Since we don't want to mess with the global mode, we fake the fencing via explicit wait_for ioctl usage. To account for devices with broken wait_for ioctl (since those happen to be the target for testing this implementation), we force a much larger sleep than usual for the pre- fence (250 instead of 2.5ms).

Re: https://github.com/koreader/koreader/issues/9806#issuecomment-1416827447

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1576)

### DIFF
--- a/ffi/framebuffer_mxcfb.lua
+++ b/ffi/framebuffer_mxcfb.lua
@@ -227,11 +227,11 @@ local function kindle_mxc_wait_for_update_submission(fb, marker)
     return C.ioctl(fb.fd, C.MXCFB_WAIT_FOR_UPDATE_SUBMISSION, fb.submission_data)
 end
 
--- Stub version that simply sleeps for 1ms
--- This is roughly five times the amount of time a real *NOP* WAIT_FOR_UPDATE_COMPLETE would take.
+-- Stub version that simply sleeps for 2.5ms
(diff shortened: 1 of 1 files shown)


poire-z:
I don't understand any of this, and that's fine :)
I always wondered, but never thought of asking, when seeing these "mech_": what does "mech" means ? Mechanical (meaning hardware ? but why not hw_ then ? :) or something else ?

NiLuJe:
> what does "mech" means ?

That's a very good question :D. No idea, it was like that when I came in ;p.

(I would assume it means "mechanism" though, but that's admittedly not great ;p. Never really paid any attention to it, though, I'm used to it by now ^^).

Frenzie:
I think it's semaphore terminology, or something along those lines. But I don't really know much about those things.

https://www.guru99.com/semaphore-in-operating-system.html ("signaling mechanism")
https://docs.kernel.org/scheduler/completion.html ("synchronization mechanism")

PR #1578, opened 2023-02-05
reduce disk usage of thirdparty build directories

benoit-pierre:
- shallower clones
- sparse checkouts for the initial repos
- support not fetching/setting up any submodules
- disable all OpenSSL submodules (not used)
- only fetch/checkout the MuPDF submodules that are needed

Stats after building 3 versions (emulator debug & release versions, kindlepw2 release version):

|                                                  | master |     PR |
| -------------------------                        | ------:|-------:|
| `thirdparty/*/build/git_checkout`                |   2.6G |   296M |
| `thirdparty/*/build/x86_64-pc-linux-gnu`         |   3.2G |   1.5G |
| `thirdparty/*/build/x86_64-pc-linux-gnu-debug`   |   3.6G |   2.0G |
| `thirdparty/*/build/arm-kindlepw2-linux-gnueabi` |   3.5G |   1.8G |
| `thirdparty/*/build`                             |  13.0G |   5.5G |

Still high, but much better.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1578)

### DIFF
--- a/thirdparty/cmake_modules/koreader_thirdparty_git.cmake
+++ b/thirdparty/cmake_modules/koreader_thirdparty_git.cmake
@@ -1,24 +1,28 @@
 find_package(Git)
 
 function(_ko_write_gitclone_script script_filename git_EXECUTABLE git_repository git_tag git_submodules src_name work_dir build_source_dir gitclone_infofile gitclone_stampfile)
+
+# Default depth
+set(git_clone_depth 50)
+
+set(clone_checkout "${work_dir}/${src_name}")
+
   file(WRITE ${script_filename}
 "if(\"${git_tag}\" STREQUAL \"\")
   message(FATAL_ERROR \"Tag for git checkout should not be empty.\")
 endif()
 
-# Default depth
-set(git_clone_depth 50)
+######################################################################
+# 1. if not cloned before, do it
+# 2. try to checkout the requested revision 3 times:
+#   - if the 1st try fails: try to fetch only the requested revision
+#   - if the 2nd try fails: try to unshallow the repository
+#   - if the 3rd try fails: game over…
+# 3. update the requested sub-modules
+# 4. copy everything over to source directory
+######################################################################
 
 set(run 0)
-
-######################################################################
-# 1. if not cloned before, do a git clone
-# 2. are we already at the given commit hash?
-#   2.1. checkout to the given commit hash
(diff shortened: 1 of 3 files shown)


thirdparty/cmake_modules/koreader_thirdparty_git.cmake
```diff
@@ -69,79 +60,76 @@ if(should_clone)
   if(error_code)
     message(FATAL_ERROR \"Failed to clone repository: '${git_repository}'\")
   endif()
+  execute_process(
+    COMMAND \"${git_EXECUTABLE}\" -C \"${clone_checkout}\" sparse-checkout init
```
Frenzie, line 64: This isn't supported yet. But thanks for providing the update.

Frenzie (changes requested):
Not about this PR which is exactly according to plan but blocked until the Android image is updated.

thirdparty/cmake_modules/koreader_thirdparty_git.cmake
benoit-pierre: Note: the code should be robust to `sparse-checkout` not being supported (exit codes are ignored for that reason).

Frenzie:
It's a bit unclear to me how you managed to make CircleCI not run. Do you have any idea?

Frenzie:
Eh well, I think it's fine with a quick try, though it's lacking the peace of mind of having a multitude of configurations checked.

Now running on master: https://github.com/koreader/koreader-base/runs/11450672853

benoit-pierre:
> It's a bit unclear to me how you managed to make CircleCI not run. Do you have any idea?

No idea, but from past experience Circle CI can be a bit funky.

benoit-pierre:
Have you considered converting those jobs to use Github Actions?

benoit-pierre:
No looking good on master...

benoit-pierre:
I'm looking into it, seems like mupdf is missing some variables (e.g. `SYS_LIBJPEG_LIBS`).

Frenzie:
> Have you considered converting those jobs to use Github Actions?

GitHub Actions is unfortunately not fit for purpose. See https://github.com/koreader/koreader-base/pull/1442#issuecomment-1004197219 and https://github.com/koreader/koreader-base/pull/1464.

PR #1583, opened 2023-02-16
ffiUtil: Don't branch on ffi.os at runtime

NiLuJe:
We're unlikely to change OS at runtime ;o).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1583)


NiLuJe:
There are a few drive-by simplifications (no need to cast a `char[]` to `char*`, that's only needed for Lua strings; no need for ffi.copy for a Lua -> C string conversion, `ffi.new` handles constructors).

NiLuJe (commented):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1583)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

PR #1585, opened 2023-02-19
SDL: miscellaneous fixes

benoit-pierre:
* fix a heap buffer overflow
* fix some (minor) memory leaks

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1585)

### DIFF
--- a/ffi/SDL2_0_h.lua
+++ b/ffi/SDL2_0_h.lua
@@ -966,4 +966,5 @@ struct SDL_version {
   Uint8 patch;
 };
 void SDL_GetVersion(struct SDL_version *) __attribute__((visibility("default")));
+void SDL_free(void *) __attribute__((visibility("default")));
 ]]
--- a/ffi-cdecl/SDL2_0_decl.c
+++ b/ffi-cdecl/SDL2_0_decl.c
@@ -258,3 +258,5 @@ cdecl_const(SDL_PIXELFORMAT_ABGR32)
 
 cdecl_struct(SDL_version)
 cdecl_func(SDL_GetVersion)
+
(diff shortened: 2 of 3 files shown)


ffi/SDL2_0_h.lua
```diff
@@ -966,4 +966,5 @@ struct SDL_version {
   Uint8 patch;
 };
 void SDL_GetVersion(struct SDL_version *) __attribute__((visibility("default")));
+void SDL_free(void *);
```
Frenzie, line 969: Please also add this to https://github.com/koreader/koreader-base/blob/96b2a16e685a75de1b7fcfe30b9dd19d54f5f306/ffi-cdecl/SDL2_0_decl.c

Also I'm inclined to think stdinc should go at the very top. What do you think @NiLuJe?

benoit-pierre: Are those `…_decl.c` files really useful? I looked at it briefly and it seems to rely on plugins for an old GCC version (4.x), so I've just been copy pasting the relevant declarations when needed.

Frenzie: Yes, otherwise your lines are likely to be accidentally lost. See https://github.com/koreader/ffi-cdecl

Frenzie: That is, the idea is that the _h files are auto-generated end results. We edit them manually for development because that's easier, and conversely maintenance is easier when automated.

benoit-pierre: OK, done.

NiLuJe (approved):

PR #1586, opened 2023-02-19
MuPDF: fix build

benoit-pierre:
Patch the makefile link command so the build process does not try to link with system libraries (e.g. `-lpeg`), even when building the generator helpers (like `cnamedump.exe`). Since mupdf's build system is only in charge of building a static library, we can just drop the `$(LIBS)` part of the link command.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1586)

### DIFF
--- a/thirdparty/mupdf/CMakeLists.txt
+++ b/thirdparty/mupdf/CMakeLists.txt
@@ -136,6 +136,9 @@ set(PATCH_CMD9 "${KO_PATCH} ${CMAKE_CURRENT_SOURCE_DIR}/webp-upstream-697749.pat
 # add support for cbz chapters
 set(PATCH_CMD10 "${KO_PATCH} ${CMAKE_CURRENT_SOURCE_DIR}/mupdf_cbz_chapter_support.patch")
 
+# don't try link with libraries when building generator helpers…
+set(PATCH_CMD11 "${KO_PATCH} ${CMAKE_CURRENT_SOURCE_DIR}/fix_link_cmd.patch")
+
 # TODO: ignore shared git submodules built outside of mupdf by ourselves
 # https://git.ghostscript.com/mupdf.git is slow, so we use the official mirror on GitHub
 ep_get_source_dir(SOURCE_DIR)
@@ -152,7 +155,7 @@ ExternalProject_Add(
     ${PROJECT_NAME}
     DOWNLOAD_COMMAND ${CMAKE_COMMAND} -P ${GIT_CLONE_SCRIPT_FILENAME}
     BUILD_IN_SOURCE 1
-    PATCH_COMMAND COMMAND ${PATCH_CMD1} COMMAND ${PATCH_CMD2} COMMAND ${PATCH_CMD3} COMMAND ${PATCH_CMD4} COMMAND ${PATCH_CMD5} COMMAND ${PATCH_CMD6} COMMAND ${PATCH_CMD7} COMMAND ${PATCH_CMD8} COMMAND ${PATCH_CMD9} COMMAND ${PATCH_CMD10}
+    PATCH_COMMAND COMMAND ${PATCH_CMD1} COMMAND ${PATCH_CMD2} COMMAND ${PATCH_CMD3} COMMAND ${PATCH_CMD4} COMMAND ${PATCH_CMD5} COMMAND ${PATCH_CMD6} COMMAND ${PATCH_CMD7} COMMAND ${PATCH_CMD8} COMMAND ${PATCH_CMD9} COMMAND ${PATCH_CMD10} COMMAND ${PATCH_CMD11}
     # skip configure
     CONFIGURE_COMMAND ""
     BUILD_COMMAND ${BUILD_CMD_GENERATE} COMMAND ${STATIC_BUILD_CMD} COMMAND ${SHARED_BUILD_CMD}
--- a/thirdparty/mupdf/fix_link_cmd.patch
+++ b/thirdparty/mupdf/fix_link_cmd.patch
@@ -1,0 +1,11 @@
+--- a/Makefile	2023-02-19 19:58:09.894692877 +0100
++++ b/Makefile	2023-02-19 19:58:17.818158811 +0100
+@@ -75,7 +75,7 @@
+ CC_CMD = $(QUIET_CC) $(CC) $(CFLAGS) -o $@ -c $<
+ CXX_CMD = $(QUIET_CXX) $(CXX) $(filter-out -Wdeclaration-after-statement,$(CFLAGS)) -o $@ -c $<
+ AR_CMD = $(QUIET_AR) $(AR) cr $@ $^
+-LINK_CMD = $(QUIET_LINK) $(CC) $(LDFLAGS) -o $@ $^ $(LIBS)
++LINK_CMD = $(QUIET_LINK) $(CC) $(LDFLAGS) -o $@ $^
+ MKDIR_CMD = $(QUIET_MKDIR) mkdir -p $@
+ RM_CMD = $(QUIET_RM) rm -f $@
+ TAGS_CMD = $(QUIET_TAGS) ctags $^


benoit-pierre:
I could reproduce the CI failure locally after moving my system JPEG library out of the way.

benoit-pierre:
On a semi-related note, because cmake is hell, I've got a meson POC build system, if there's some interest. It does make composing things and experimenting much easier.

NiLuJe:
Now that meson finally hit 1.0, and that we don't really care about native win32 anymore... I really, really, really wouldn't be opposed to it, despite the potential disruption...

benoit-pierre:
Current state: compile the emulator on Linux, cross compile the emulator for Windows (`win64`, which I managed to get running with some issues, but not anymore with the recent `mmap` dependency in the partial rendering code), and the kindlepw2 build. Includes release & debug variants, and a mode where most libraries are merged into one big `libkoreader`. Also included, MuPDF updated to the latest version (I was investigating why the Windows version PDF support does not work).

poire-z:
That's impressive ! and a bit frightening as I don't like my living room to be that revamped :)

> and a mode where most libraries are merged into one big libkoreader

Part of my dev worklow, after testing on the emulator, is building _only_ libcrengine.so + libkoreader-cre.so for arm (+ their deps), and move them (only these 2 libs) over a recent koreader-kobo nightly on my device to test.
Given my old & slow & space limited build system, having to rebuild the whole libraries for arm each time there is a unrelated thirdparty bump would be quite bothering to me :/

Frenzie:
> On a semi-related note, because cmake is hell, I've got a meson POC build system, if there's some interest. It does make composing things and experimenting much easier.

Yes, absolutely.

> cross compile the emulator for Windows (win64, which I managed to get running with some issues

Cool!

benoit-pierre:
> That's impressive ! and a bit frightening as I don't like my living room to be that revamped :)
> 
> > and a mode where most libraries are merged into one big libkoreader
> 
> Part of my dev worklow, after testing on the emulator, is building _only_ libcrengine.so + libkoreader-cre.so for arm (+ their deps), and move them (only these 2 libs) over a recent koreader-kobo nightly on my device to test. Given my old & slow & space limited build system, having to rebuild the whole libraries for arm each time there is a unrelated thirdparty bump would be quite bothering to me :/

Yeah, it does increase the base memory use, even if it potentially decreases the maximum memory needed when all the shared libraries are loaded. It does help taking advantage of LTO to further optimize and trim unused symbols.

Anyway, it's optional ;)

poire-z:
> Anyway, it's optional ;)

Well, if the performance improvment is noticable, we'll have to decide how to go with building nightlies & stable. And we would loose quite a bit of practicality with a monolithic libkoreader.so (like shipping debug/test libs to users).
Also, how would we handle stuff like `ffi.load("libs/libgif.so.7")` ? We'd have many new codepaths for when monolibtic mode is enabled vs not?
And if we need that many tweaks to frontend and base/ffi to handle both modes, it will be quite bothering and the need to update & test both modes when adding new stuff :/ so, we'll end up killing one of these, and it won't be optionnal :)

thirdparty/mupdf/CMakeLists.txt
```diff
@@ -136,6 +136,9 @@ set(PATCH_CMD9 "${KO_PATCH} ${CMAKE_CURRENT_SOURCE_DIR}/webp-upstream-697749.pat
 # add support for cbz chapters
 set(PATCH_CMD10 "${KO_PATCH} ${CMAKE_CURRENT_SOURCE_DIR}/mupdf_cbz_chapter_support.patch")
 
+# don't try link with libraries when building generator helpers…
+set(PATCH_CMD10 "${KO_PATCH} ${CMAKE_CURRENT_SOURCE_DIR}/fix_link_cmd.patch")
```
poire-z, line 140: You probably meant `PATCH_CMD11` on line 140, that you use below.

benoit-pierre: Damn... Good catch.

benoit-pierre:
Meson POC available here: [koreader](https://github.com/koreader/koreader/compare/master...benoit-pierre:koreader:meson_support), [koreader-base](https://github.com/koreader/koreader-base/compare/master...benoit-pierre:koreader-base:meson_support), and [crengine](https://github.com/koreader/crengine/compare/master...benoit-pierre:crengine:meson_support).

Notes:

- only tested on Linux 
- support `./kodev build` & `./kodev run` to build/run the emulator
- support `./kodev build kindlepw2` (`curl`, `sftp-server` & `zsync2` not yet included, `sscache` use is forced, see `meson/cross_kindlepw2.txt`)
- support `./koded build win64` & `./kodev run win64` to cross-compile for Windows (wine is used for running, see `meson/cross_emulator_win64.txt`)
- subprojects are used for thirdparty dependencies: patches and overlay trees are in `subprojects/packagefiles`, download cache in `subprojects/packagecache`
- thirdparty dependency source trees are under `subprojects`, removed on `make dist-clean`
- build output is stored under `./build`
- install directories in top dir (e.g.: `koreader-emulator_linux-debug`)
- when building the emulator on Linux, some directories are symlinked to the install directory, otherwise everything is copied
- no more `rocks` directory, everything is under `common`
- `amalgam` mode (one big `libkoreader` library) can be enabled by setting `KOAMALGAM=1` in the environment (a different build/install directory will automatically be used)
- build options are in `./meson`:
  * `options.txt`: common options
  * `options_amalgam.txt`: override for `amalgam` mode
  * `options_debug.txt`: debug build options
  * `options_release.txt`: ditto for release
  * `target_xxx.txt`: target specific options
  * `cross_xxx.txt`: cross-compiling options

TODO:
- add `curl`, `openssh`, `zsync`
- see if an Android build if feasible? I've not even tried it with the standard build system
- rename `amalgam` mode to `monolibtic` ;)
- switch from archives to git clones (the former was faster for developing/testing the build system, but using
  checkouts would make patch management easier)
- …

> Well, if the performance improvment is noticable, we'll have to decide how to go with building nightlies & stable. And we would loose quite a bit of practicality with a monolithic libkoreader.so (like shipping debug/test libs to users).
Also, how would we handle stuff like ffi.load("libs/libgif.so.7") ? We'd have many new codepaths for when monolibtic mode is enabled vs not?
And if we need that many tweaks to frontend and base/ffi to handle both modes, it will be quite bothering and the need to update & test both modes when adding new stuff :/ so, we'll end up killing one of these, and it won't be optional :)

I've added a [`ffi.loadlib` function](https://github.com/benoit-pierre/koreader/blob/9421e0a36c2ec072c9d3d8531dc5b63ad5da1861/setupkoenv.lua#L62) to simplify loading libraries (see [SDL
example](https://github.com/benoit-pierre/koreader-base/blob/279ea93ac93b2f8dd53f20a22afb1f27145b03cb/ffi/SDL2_0.lua#L23)): it takes a series of (library name, optional version) pair to look for, and the [`loadlib` helper](https://github.com/benoit-pierre/koreader/blob/9421e0a36c2ec072c9d3d8531dc5b63ad5da1861/setupkoenv.lua#L42) handle looking for the right library and abstracting the details of the platform specifics prefix/suffix.

When building in `monolibtic` mode, a small lua file is generated and used by `setupkoenv` to handle overrides: redirect some requires and `ffi.loadlib` calls to `libkoreader`. Here is the content of the generated file:
```lua
return {
    library = 'libs/libkoreader.so',
    preloads = {
        ["libs/libkoreader-cre"] = "luaopen_cre",
        ["libs/libkoreader-djvu"] = "luaopen_djvu",
        ["libs/libkoreader-nnsvg"] = "luaopen_nnsvg",
        ["libs/libkoreader-xtext"] = "luaopen_xtext",
    },
    redirects = {
        ["czmq.1"] = true,
        ["freetype.6"] = true,
        ["harfbuzz.0"] = true,
        ["k2pdfopt.2"] = true,
        ["lept.5"] = true,
        ["lodepng"] = true,
        ["mupdf"] = true,
        ["sqlite3.0"] = true,
        ["utf8proc.2"] = true,
        ["z"] = true,
        ["zmq.4"] = true,
        ["zstd.1"] = true,
        ["SDL2-2.0.0"] = true,
        ["wrap-mupdf"] = true,
        ["blitbuffer"] = true,
    },
}
```

NiLuJe:
At a very quick glance, that looks rather sexy ;).

I can certainly see the appeal of mesonifying everything to sidestep annoying upstream buildsystems, too... Is there a tool that helps generate meson.build files or was that a lot of hand-made tweaking?

eli-schwartz:
Oh neat.

> despite my hangups about its approach to cross compilation

@NiLuJe BTW -- nifty little trick, you can actually trigger a cross compilation like this: `CC=my-cross-gcc CXX=my-cross-g++ meson setup builddir --cross-file=/dev/null`.

Practically everything from a cross file is not actually necessary, if that's what you were referring to.

poire-z:
I can't judge about the build system change and pros/cons, I'm not competent enough - but it indeed looks more readable than CMake.
A bit confused about what gets in koreader frontend and what in koreader-base - and what controls what - but I guess I'll learn. Would we still need the 2 repos koreader + koreader-base ?

Some practical questions, just to get an idea how complicated it would get for me :/:
- could we still use our (possibly old, still using gcc-arm from debian10 on my debian11) crosscompiling toolchain ? Or does meson (and/or your upgraded stuff) requires up to date crosstools ? And would it work on current debian 11 tools, or would you require us to have to build newer versions of most of the build tools ? :/
- could I still build all of it on my small headless linux with 4GB or RAM (3GB free) ? How much memory would the monolibtic build need ?
- (not really happy with having Windows support: now, when we add a feature using Linux specific stuff, we would be required to code the Windows counterpart and know about Windows API... which I promise I _will not_ learn ! :)

poire-z:
And btw... _who are you ?!_
Out of nowhere, you pop in, and:
- bump MuPDF, that no one really dared doing for years (didn't look like it needed much work though :)
- kill the dragon CMake
- bring Windows support

What next ? I'm a bit fearing you might rewrite crengine in rust :/
Why such interest in KOReader ? Long time user ? Do you have time to read ? :)

NiLuJe:
> @NiLuJe BTW -- nifty little trick, you can actually trigger a cross compilation like this: `CC=my-cross-gcc CXX=my-cross-g++ meson setup builddir --cross-file=/dev/null`.
> 
> Practically everything from a cross file is not actually necessary, if that's what you were referring to. (And the one thing you officially do "need" you can still do without.)

Oh, I like that, thanks!

(Last I checked, there was a very strong assumption that what's in the env var should only ever affect the *host*, not the *target*, so, e.g., setting CFLAGS with arm-specific flags in the env would break pretty much anything that required running the host compiler during the build (e.g., code generation tools)).

eli-schwartz:
https://mesonbuild.com/Reference-tables.html#environment-variables-per-machine

We support autotools-style `$XXX_FOR_BUILD` variant variables, though anything in a machine file always trumps the environment. Old versions of Meson may have been a bit messy, though. :(

Frenzie:
@poire-z 

> And btw... who are you ?!

On the one hand I don't know either, on the other hand a quick glance comes up with https://github.com/openstenoproject/plover which is an extremely impressive piece of software I've known about for years.

@benoit-pierre I think the bigger question is… where'd you find all the time? :laughing: Either that or you can do in hours what would take me at least a full working week. ;-)

yparitcher:
If you are updating MuPdf please show a little love to [k2pdfopt](https://www.willus.com/k2pdfopt/) :)
Our fork with patches is [https://github.com/koreader/libk2pdfopt](https://github.com/koreader/libk2pdfopt) and it and its dependent libraries can only be bumped in tandem with mupdf

Frenzie:
You probably already saw btw, but https://github.com/koreader/koreader-base/pull/1203 has pretty much all of the work required in it. Specifically see https://github.com/ezdiy/libk2pdfopt

poire-z:
Should we merge this PR quickly, for master to build ? (that is, if you approve the changes)

benoit-pierre:
> I can certainly see the appeal of mesonifying everything to sidestep annoying upstream buildsystems, too... Is there a tool that helps generate meson.build files or was that a lot of hand-made tweaking?

100% organic made.

> Part of my dev worklow, after testing on the emulator, is building only libcrengine.so + libkoreader-cre.so for arm (+ their deps), and move them (only these 2 libs) over a recent koreader-kobo nightly on my device to test.
> Given my old & slow & space limited build system, having to rebuild the whole libraries for arm each time there is a unrelated thirdparty bump would be quite bothering to me :/

```bash
# manually setup for `remarkable` (no `monolibtic`) without building
▸ make KODEBUG=1 TARGET=remarkable setup
# rebuild only libkoreader-cre.so and dependencies (about ~500 targets).
▸ ninja -C build/remarkable-debug libkoreader-cre.so
# copy to target device
▸ scp build/remarkable-debug/libkoreader-cre.so …
```

There's no `libcrengine.so` anymore: it's always statically built, since `libkoreader-cre` is the only user. Same with some of its dependencies: `antiword`, `chmlib` and `lunasvg`.

This take about 140s on my (10 years old) machine, starting from a cold compile cache. As for space, compared to the current [stats on master](https://github.com/koreader/koreader-base/pull/1578#issue-1571593980), with (almost) the same builds:

|                               |      |
|-------------------------------|-----:|
| `subprojects/`                | 987M |
| `build/emulator_linux/`       | 316M |
| `build/emulator_linux-debug/` | 628M |
| `build/kindlepw2/`            | 292M |
| `build/`                      | 1.3G |
| TOTAL                         | 2.2G |

> I can't judge about the build system change and pros/cons, I'm not competent enough - but it indeed looks more readable than CMake.
> A bit confused about what gets in koreader frontend and what in koreader-base - and what controls what - but I guess I'll learn. Would we still need the 2 repos koreader + koreader-base ?

Meson impose the location of sub-projects, so `base/thirdparty` would be gone, as well as:

- `base/Makefile`
- `base/Makefile.defs`
- `base/Makefile.third`
- `base/luafilesystem/` & `base/patches/` (the later contains one patch for `luafilesystem`)

That still leaves quite a few things:

- the source for all helper libraries: `libblitbuffer`, `libkoreader-cre`, `libkoreader-djvu`, …
- FFI stuff (`ffi/` & `ffi-cdecl/`)
- and some miscellaneous stuff: PDF helpers (`extr` & `utils/`), `utils/` (Android toolchain install helper), …

I certainly would welcome less git submodules (always a PITA to deal with).

> Some practical questions, just to get an idea how complicated it would get for me :/:
> 
>     could we still use our (possibly old, still using gcc-arm from debian10 on my debian11) crosscompiling toolchain ?

I don't see why not.

>     Or does meson (and/or your upgraded stuff) requires up to date crosstools ? 

I don't think so.

>     And would it work on current debian 11 tools, or would you require us to have to build newer versions of most of the build tools ? :/
>
>     could I still build all of it on my small headless linux with 4GB or RAM (3GB free) ? How much memory would the monolibtic build need ?

Again, "monolibtic" mode is optional, and the LTO stuff can be disabled in `meson/options_release.txt`. I don't know about memory usage, but linking the monolibtic `libkoreader` for `kindlepw2` with LTO does take a good 5 minutes on my machine… Which I think is probably a combination of older GCC version and non-optimized build: I'm using the latest koxtoolchain release, I haven't managed to rebuild one with `gen-tc.sh`, yet alone attempted to update the GCC version.

>     (not really happy with having Windows support: now, when we add a feature using Linux specific stuff, we would be required to code the Windows counterpart and know about Windows API... which I promise I will not learn ! :)

That's was really more of quick way to test cross-compiling for another platform (and the `ffi.loadlib` stuff). It's broken anyway (mmap, PDF support, subprocess handling). There's also a subtle bug with CRE documents, see if you can spot it:

2023-02-21-200900_540x720_scrot

I also have no love for Windows, so I'm going to drop it.

>  Why such interest in KOReader ? Long time user ? Do you have time to read ? :)

User, yes. Reading times have recently plummeted..

> If you are updating MuPdf please show a little love to k2pdfopt :) […]

I've just done the minimum for sanity's sake (see `subprojects/packagefiles/libk2pdfopt/less_fugly.patch`) for now.

> […] its dependent libraries can only be bumped in tandem with mupdf

1. `mupdf` depends on `freetype2`, `harfbuzz`, `libjpeg`, `libwebpdemux`, and some modified thirdparty libraries: `gumbo-parser`, `jbig2dec`, `mujs` & `openjp2`
2. `libk2pdfopt` depends on `leptonica` & `tessact`
3. `tesseract` depends on `leptonica`
4. and `leptonica` optionally depends on `libopenjp2`, `libjpeg`, `libpng`, `libtiff-4` & `libwebp`: right now I'm only enabling PNG support as analyzing the list of [symbols required by Koreader](https://github.com/benoit-pierre/koreader-base/blob/279ea93ac93b2f8dd53f20a22afb1f27145b03cb/libkoreader.exports/lept) does not show the need for anything else.

> You probably already saw btw, but #1203 has pretty much all of the work required in it. Specifically see https://github.com/ezdiy/libk2pdfopt

Yeah, I've not looked into it yet. What would be helpful is a checklist of tests to ensure everything is indeed working as expected: with so many options, it's hard to know if I've hit all the code paths (like the support for highlighting).

Tentative list:

- PDF support:
  * opening a PDF
  * selecting some text (check multi-lines)
  * highlighting selection
  * changing an highlight style
  * deleting an highlight
  * saving highlights? I've noticed Koreader ask if changes must be saved, but even after declining, there's still some state saved (even on master).
  * what else? encryption support, external fonts, …
- CBZ support:
  * opening a comic
  * try zooming on a panel (long press)
  * …
- HTML support:
  * check the results of a dictionary lookup
- …

eli-schwartz:
> still using gcc-arm from debian10 on my debian11) crosscompiling toolchain ? Or does meson (and/or your upgraded stuff) requires up to date crosstools ? And would it work on current debian 11 tools, or would you require us to have to build newer versions of most of the build tools ? :/

@poire-z,

Meson's CI regularly tests Ubuntu Bionic, which is older than those, and has GCC 7.4; in general, there's no need to worry about using the environment of a still supported, much less quite *recent*, distro.

...

As a generic rule?

Meson tries to avoid arbitrarily restricting the version of your compiler, but in order to integrate well it does sometimes need to know how to activate features. Meson usually version guards this when added, and doesn't gratuitously drop support for old compilers (and we literally cannot drop support for 7.4, since, again, we target that in our CI). There is support code as old as 4.8, which is the first version of GCC released after Meson was started (it handles warning arguments). I do not know whether anyone still relies on it, but there's no plan to drop that either, since it already exists and isn't a burden to leave in place.

Any given user probably does not need GCC 2.x support. If you did, Meson would probably be a bad choice, because Meson makes no effort to support systems or toolsets that were ancient and obsolete before it was started. :P

NiLuJe:
> yet alone attempted to update the GCC version.

That's near the top of my TODO list, FWIW (I'll hopefully get to it in the next couple of months...).

Not that it'll help much with LTO, the only really faster alternatives I know of are Clang in thin-lto mode, coupled with lld or even mold ;).

NiLuJe:
> I do not know whether anyone still relies on it

(re: gcc 4.8)

That's (nearly) the last GCC version licensed under GPL v2, so, that's what the GCC version available in Android NDKs is based on (and we haven't made the move to Clang yet there). Or, even, some really old OS X toolchains, if you're feeling mad ;D).

eli-schwartz:
Sounds painful, but if you do need it and you run into issues with it, I am happy to merge untested PRs on the honor system, that support such old GCCs better.

> some really old OS X toolchains, if you're feeling mad ;D.

... for which definition of mad? :eyes:

NiLuJe:
> ... for which definition of mad? eyes

I can't even remember when Apple dropped GCC, so you'd have to be into some *really* kinky stuff with old Apple OSes ;p.

Frenzie:
> That's (nearly? can't remember the exact cutoff) the last GCC version licensed under GPL v2, so, that's what the GCC version available in (old) Android NDKs is based on

4.9, for the record. But yes, it doesn't make a whole lot of difference.

poire-z:
> There's no `libcrengine.so` anymore: it's always statically built, since `libkoreader-cre` is the only user. Same with some of its dependencies: `antiword`, `chmlib` and `lunasvg`.

OK, so that would be in a non-monolibtic mode, and I'd be ok with a single libkoreader-cre.so to build and move to the device. (Still some aggregations, so, minilibstic :)
(Ideally, we'd have a monolibtic libkoreader.so, and your ffiload would check for sublibs/binary-overlays like libkoreader-cre.so, load it and use its code instead of the equivalent bits already in libkoreader.so :) but I guess there's nothing like that existing :/)

> There's also a subtle bug with CRE documents, see if you can spot it:

Had to use a mirror, to see it's the full cre bb that is flipped (and not the font glyphs).
So, either some win32 codepaths in crengine, or some bug in some bb formats blitting combination that you met on win32, and that we never met with our actual platforms? Either in crengine itself, or in our ffi/blitbuffer.lua/.c ?

poire-z:
Keeping the discussion here, given the new input about your wip in https://github.com/koreader/koreader/issues/4245#issuecomment-1535313307.

It feels enormous, and will be hard to review - and it's a bit unconfortable feeling we'll just be put in front of the final work where you had already taken many decisions and we may have no say (or a single say: accepted or rejected - and rejected valuable new stuff will be hard - or many more work to adapt what's been done to what we may prefer :/

Just had a quick look, and 2 things pain me immensely:
- moving crengine into frontend, and apparently losing all its history and git blame usefulness - which will make any further dev very painful for me
- (dropping all its "old" other frontends like android and qt - even if it hasn't been updated, they can sometimes give some hints about why such function/feature exists, how it is/was used, what we're not using that uses some stuff we do use and that we may break if we change it - other forks do sometimes pick stuff from us, I try to keep things not impacting such old/other using code. So, this will be a hard diverge, and we would break all links with that.)
- and with that, crengine commits will be mixed with all the frontend (and other subprojects) stuff - so, a bit harder to follow, allthough `git log subproject/crengine` may be enough to follow this sub-dev - the benefit being no submodule chained bump.

benoit-pierre:
> Keeping the discussion here, given the new input about your wip in [koreader/koreader#4245 (comment)](https://github.com/koreader/koreader/issues/4245#issuecomment-1535313307).
> 
> It feels enormous, and will be hard to review - and it's a bit unconfortable feeling we'll just be put in front of the final work where you had already taken many decisions and we may have no say (or a single say: accepted or rejected - and rejected valuable new stuff will be hard - or many more work to adapt what's been done to what we may prefer :/

Nothing is set in stone, and I tried to use separate commits so it's easier to merge only some things (like the ffi-cdecl cleanups, for example).

> Just had a quick look, and 2 things pain me immensely:
> 
>     * moving crengine into frontend, and apparently losing all its history and git blame usefulness - which will make any further dev very painful for me

No, the subtree add was done without squashing, so all the history is there. The problem is the git UI/cmdline, particularly rename handling with respect to the log, as always. So blame works ([see in github](https://github.com/benoit-pierre/koreader/blame/meson%2Bmonorepo-ish/subprojects/crengine/crengine/src/lvstring.cpp)), but the log is cut off at the subtree add:

```bash
▸ git log --oneline -n 5 -- subprojects/crengine/crengine/src/lvstring.cpp
1c865e9e2 crengine: fix clang build
1b2aa292e crengine: add as subtree
▸ git show 1b2aa292e
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
commit 1b2aa292e614dbeba358b1bfdbc037c8bc1a35e6 ┃
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┻━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Merge: 16418f6fb 63ca29437
[...]
▸ git log --oneline -n 5 63ca29437\^ -- crengine/src/lvstring.cpp
508f3bdc6 (crengine-ng) Various small picks: unused variables, cast & types
1c299cb71 SerialBuf: allow serializing longer strings
fd47e9374 LVString: fix char props, add lStr_isCJK() and lStr_isRTL()
ed8fed649 LVString: update/fix char props and lStr_findWordBounds() (#468)
9e975ddff lString32Collection::erase(): fix bounds checking
```

>     * (dropping all its "old" other frontends like android and qt - even if it hasn't been updated, they can sometimes give some hints about why such function/feature exists, how it is/was used, what we're not using that uses some stuff we do use and that we may break if we change it - other forks do sometimes pick stuff from us, I try to keep things not impacting such old/other using code. So, this will be a hard diverge, and we would break all links with that.)

It's a separate commit, can be dropped. Both things were done to make it easier than the PITA of submodules (particularly when rebasing), and so one single preview PR could be done, without all the necessary baggage, or having to split it in 3 (base, cre, koreader).

poire-z:
Well, git log (-p, --name-status...) is useful.
I hope that in the end, subproject/crengine/ can be made a submodule referencing the current koreader/crengine, so nothing is lost/change.
For the other smaller subprojects from base, it will be sad to lose their history - but for my other contributions (xtext, lunasvg...), they have been less active, so it's not as essential (but still nice to have. (Dunno if it can be useful to save the output of `git log --name-status xtext.cpp` alongside each of them, for easier browsing/grasping/greping history.)

> Merge: 16418f6fb 63ca29437
> `git log --oneline -n 5 63ca29437\^ -- crengine/src/lvstring.cpp`

Dunno what this says :) You mean we could still access the former history with that second step and manually picking another hash from the last commit message ? Why doesn't git do it by itself? Would making a shell script wrapper to `git log` doing this all transparently possible?

NiLuJe:
I'm ambivalent about the whole submodules thing. I *understand* the reasoning, because, yeah, they can be weird to deal with (although it's gotten better with modern git versions), but I've grown to "like" dealing with the clunk, FWIW ;o). (And the clear separation *can* be helpful, I think).

benoit-pierre:
> > Merge: 16418f6fb 63ca29437
> > `git log --oneline -n 5 63ca29437\^ -- crengine/src/lvstring.cpp`
> 
> Dunno what this says :) You mean we could still access the former history with that second step and manually picking another hash from the last commit message ?

Yes, the whole history of both former submodules is there.

> Why doesn't git do it by itself?

¯\\_(ツ)_/¯

> Would making a shell script wrapper to `git log` doing this all transparently possible?

Another workaround (from [there](https://stackoverflow.com/questions/10918244/git-subtree-without-squash-view-log):
```bash
▸ git log --oneline -n5 '*crengine/src/lvstring.cpp'
1c865e9e2 crengine: fix clang build
1b2aa292e crengine: add as subtree
508f3bdc6 (crengine-ng) Various small picks: unused variables, cast & types
1c299cb71 SerialBuf: allow serializing longer strings
fd47e9374 LVString: fix char props, add lStr_isCJK() and lStr_isRTL()
```

Frenzie:
I take back what I said about git subtree. I misunderstood them as basically nicer submodules but they don't seem to be that at all.

moben:
Just wanted to report that I tested the meson branch successfully as a user who occasionally builds koreader.
The built went a lot smoother than the last time I tried with cmake.

I hit all of these points, except for encryption:

> Tentative list:
> * PDF support:
>   * opening a PDF
>   * selecting some text (check multi-lines)
>   * highlighting selection
>   * changing an highlight style
>   * deleting an highlight
>   * saving highlights? I've noticed Koreader ask if changes must be saved, but even after declining, there's still some state saved (even on master).
>   * what else? encryption support, external fonts, …
> * CBZ support:
>   * opening a comic
>   * try zooming on a panel (long press)
> * HTML support:
>   * check the results of a dictionary lookup

Could also confirm that some files that are displayed incorrectly in master are fixed by the new mupdf (I was hitting #5422 I think)

Frenzie:
What do you mean by smoother?

moben:
I mean it essentially built first try, where in the cmake build I hit some weird build errors in third party dependencies last time I built from master. That might have been something with my setup, but given how many dependencies the build pulls in I found it impressive that the meson build just worked once I had all dependencies in place.

Frenzie:
If that was anywhere near normal I'd have made an effort to switch away from CMake years ago. ;-)

Put another way, you're simply describing the way the CMake system has always worked.

There are some potential pitfalls too btw. Sometimes building something when it shouldn't is a bug and not a feature. (I.e., it should be properly optimized for the target device.)

poire-z:
(It happened to me a few times in the past that some thirdparty bump made CMake confused and stop - and I could only proceed after `rm -rf thirdparty/complainingmodule/build`, so probably some cached stuff from a previous build not getting really reset).

NiLuJe:
I always build in a tmpfs, so I barely ever do incremental builds ;o).

poire-z:
@benoit-pierre : Had a try with your latest meson monolibtic kobo build found at https://github.com/benoit-pierre/koreader/actions/runs/5537996696.
It failed at startup with:
```
has libkoreader? yes
lib_search_path: libs/?
lib_basic_format: lib%s.so
lib_version_format: lib%s.so.%s
 [*] Version: v2023.04-8083-g0aa8affd3_2023-07-13

ffi.load: rt.so.1
ffi.findlib: utf8proc.2
ffi.load: libs/libkoreader.so
ffi.findlib: blitbuffer
ffi.load: libs/libkoreader.so
./luajit: frontend/ui/time.lua:194: ./luajit: undefined symbol: clock_getres
stack traceback:
    [C]: in function '__index'
    frontend/ui/time.lua:194: in main chunk
    [C]: in function 'require'
    frontend/device/generic/device.lua:13: in main chunk
    [C]: in function 'require'
    frontend/device/kobo/device.lua:1: in main chunk
    [C]: in function 'probeDevice'
    frontend/device.lua:61: in main chunk
    [C]: in function 'require'
    ./reader.lua:156: in main chunk
    [C]: at 0x0001a023
!!!!
```
I had to re-add into setupkoenv.lua:
```lua
require("ffi/posix_h")
local C = ffi.C
-- and just for me, make my own modules available (dunno where it is set in your build)
package.path = package.path .. ";/mnt/onboard/.adds/mykoreader/modules/?.lua"
```
and it works. (No noticable + or - noticed :)

NiLuJe:
I haven't looked at the meson branch, but that bit of trickery (clock_* being in librt and not libc on glibc < 2.17; which affects you and your crazy old FW version ;))) is handled in ffi/posix_h, which time obviously require, so i'm a tad confused ;).

benoit-pierre:
> @benoit-pierre : Had a try with your latest meson monolibtic kobo build found at https://github.com/benoit-pierre/koreader/actions/runs/5537996696.

Thanks for testing!

> It failed at startup with:
> […]
> I had to re-add into setupkoenv.lua:
> 
> ```lua
> require("ffi/posix_h")
> local C = ffi.C
> -- and just for me, make my own modules available (dunno where it is set in your build)
> package.path = package.path .. ";/mnt/onboard/.adds/mykoreader/modules/?.lua"
> ```
> and it works. (No noticable + or - noticed :)

I can certainly add back that require line (the `ffi.C` lines is not required, right?), but why, how? As NiLuJe already said, `ffi/posix_h` is already required in `frontend/ui/time.lua`… So what is happening?

poire-z:
> the ffi.C lines is not required, right?

Right. Just `require("ffi/posix_h")` is needed.

> but why, how? 
> So what is happening?

No idea :/ But @NiLuJe probably has one.

NiLuJe:
> I can certainly add back that require line (the `ffi.C` lines is not required, right?)

FWIW, it's only there in upstream because it needs it for the ffi `C._putenv` call on Windoze...

benoit-pierre:
> > I can certainly add back that require line (the `ffi.C` lines is not required, right?)
> 
> FWIW, it's only there in upstream because it needs it for the ffi `C._putenv` call on Windoze...

Exactly, without that call, no need for `…ffi.C…`, and without that, why require `ffi/posix_h`?

NiLuJe:
Yup, no clue why loading it early would help; that's... strange.

... unless the ffi.load overload somehow *breaks* the fancy DT_GLOBAL load... because it doesn't take enough arguments.

NiLuJe:
```diff
diff --git a/setupkoenv.lua b/setupkoenv.lua
index be1b2e228..37af263d9 100644
--- a/setupkoenv.lua
+++ b/setupkoenv.lua
@@ -15,7 +15,7 @@ if ffi.os == "Windows" then
 end
 local ffi_load = ffi.load
 -- patch ffi.load for thirdparty luajit libraries
-ffi.load = function(lib)
+ffi.load = function(lib, global)
     io.write("ffi.load: ", lib, "\n")
     local loaded, re = pcall(ffi_load, lib)
     if loaded then return re end
@@ -27,6 +27,6 @@ ffi.load = function(lib)
         error("Not able to load dynamic library: " .. lib)
     else
         io.write("ffi.load (assisted searchpath): ", lib_path, "\n")
-        return ffi_load(lib_path)
+        return ffi_load(lib_path, global)
     end
 end
```

*should* fix it.

benoit-pierre:
Of course! That must be it.

NiLuJe:
Added that to https://github.com/koreader/koreader/pull/10705 because why not ;D.

benoit-pierre:
> ```lua
> […]
> -- and just for me, make my own modules available (dunno where it is set in your build)
> package.path = package.path .. ";/mnt/onboard/.adds
/mykoreader/modules/?.lua"
> ```

In the build options, e.g:
- [default](https://github.com/benoit-pierre/koreader/blob/bpierre/meson/options.ini#L129)
- [kindlepw2](https://github.com/benoit-pierre/koreader/blob/bpierre/meson/target_kindlepw2.ini#L19)

NiLuJe:
@poire-z:

c.f., `ld(1)`

```
        RTLD_GLOBAL
               The symbols defined by this shared object will be made available for symbol resolution of subsequently loaded shared objects.
```

NiLuJe:
In a way, thanks for confirming that we did indeed need it in the global namespace, because I never actually tested this ;D.

poire-z:
Btw, which explicite ffi.load() _where_ then benefits from this fix so that `C.clock_getres` is then found?
Or is it `require("ffi/posix_h")` which implicitely `ffi.load()`s libc.so or librt.so ?

NiLuJe:
This one:

https://github.com/koreader/koreader-base/blob/a938d082ba3f39a9d5cc39492637ce8711b9dd3f/ffi/posix_h.lua#L434-L440

If you require'ed  `ffi/posix.h` *before* (or early in) `setupkoenv`, it was using the stock `ffi.load` function (which honored `global`), if not, it was using our overload, which ignored `global`.

NiLuJe:
(I double-checked, that's our *only* ffi.load that sets `global` ^^).

poire-z:
Btw, no speedup noticed in my usual benchmark test (a French Bible EPUB with > 1000 fragments), same times (30s loading + 12s rendering with no cache, 3s loading + 23s rerendering if cache rendering hash mistmatch) whether latest nightly or this monolibtic on my Kobo.

benoit-pierre:
> Btw, no speedup noticed in my usual benchmark test (a French Bible EPUB with > 1000 fragments), same times (30s loading + 12s rendering with no cache, 3s loading + 23s rerendering if cache rendering hash mistmatch) whether latest nightly or this monolibtic on my Kobo.

Note that the CI kindlepw2 and kobo build use `buildtype = 'minsize'` (`-Os`), so what about memory use?

NiLuJe:
I imagine you'll probably want to open both an ePub and a PDF to make the test fair?

poire-z:
Given that PDF gets BB of page cached in memory, it's not easy to distinguish the memory used by the app vs. what is cached.
Broswing and siwtching EPUBs is the real test for memory leaks :)

poire-z:
Just a note, as I don't think there's anything we can do as it feels upstream, but newer MuPDF seems worse with some HTML than our current one:

image
(In the MuPDF 1.22 view, some stuff logically after the pink box is drawn above it. Possibly some block element among inlines that get pushed after them.)

This is the nice rendering I currently get with KOReader (including the 2nd definition that is missing in the MuPDF 1.13 above):

image

Frenzie:
What's the CSS? Could be nicer by accident. :)

poire-z:
No CSS involved. Just the html result (which has some style= attributes):
[diabolique.html.txt](https://github.com/koreader/koreader-base/files/12084172/diabolique.html.txt)
(to which I have a dict.lua that does some massage, but it doesn't change anything to the result: 
[outdict.html.txt](https://github.com/koreader/koreader-base/files/12084177/outdict.html.txt) )
Anyway, I'll think about massaging it more in that dict.lua once we'll have this newer MuPDF :)

Frenzie:
It does look like the older MuPDF may be listening to display: inline and the newer isn't? Curious.
```
<div style="margin-top: 15pt; margin-left: 20pt; margin-bottom: 15pt; text-indent: -15pt; display: inline"><span style="background-color: rgb(69,86,130); color: red; border: solid 1px gray; zzz: Verdana; font-size: xx-small; font-weight: bold"><span>Â 1Â </span></span>Â </div>
```

poire-z:
Wanted to "fix" this dict.lua, so downloaded the AppImage from https://github.com/benoit-pierre/koreader/actions/runs/5578752636 v2023.04-8083-g36de7a303_2023-07-17.AppImage
When extracted and run on my Linux (old hardware, maybe you put cpu optimisations/flags my `Intel(R) Celeron(R) CPU 847 @ 1.10GHz` doesn't support?):

```
has libkoreader? yes
lib_search_path: libs/?
lib_basic_format: lib%s.so
lib_version_format: lib%s.so.%s
Illegal instruction

under gdb:
Program received signal SIGILL, Illegal instruction.
0x00007ffff65950b9 in ?? () from libs/libkoreader.so
(gdb) p
The history is empty.
(gdb) bt
#0  0x00007ffff65950b9 in ?? () from libs/libkoreader.so
#1  0x00007ffff7fcfabe in call_init (env=0x7fffffffdc90, argv=0x7fffffffdc78, argc=2, l=<optimized out>) at ./elf/dl-init.c:70
#2  call_init (l=<optimized out>, argc=2, argv=0x7fffffffdc78, env=0x7fffffffdc90) at ./elf/dl-init.c:26
#3  0x00007ffff7fcfba4 in _dl_init (main_map=0x555555615a50, argc=2, argv=0x7fffffffdc78, env=0x7fffffffdc90) at ./elf/dl-init.c:117
#4  0x00007ffff7e47e44 in __GI__dl_catch_exception (exception=<optimized out>, operate=<optimized out>, args=<optimized out>) at ./elf/dl-error-skeleton.c:182
#5  0x00007ffff7fd630e in dl_open_worker (a=a@entry=0x7fffffffd4b0) at ./elf/dl-open.c:808
#6  0x00007ffff7e47dea in __GI__dl_catch_exception (exception=<optimized out>, operate=<optimized out>, args=<optimized out>) at ./elf/dl-error-skeleton.c:208
#7  0x00007ffff7fd66a8 in _dl_open (file=0x7ffff7cb9818 "libs/libkoreader.so", mode=<optimized out>, caller_dlopen=0x5555555ee82a, nsid=<optimized out>, argc=2, argv=0x7fffffffdc78,
    env=0x7fffffffdc90) at ./elf/dl-open.c:884
#8  0x00007ffff7d7e2d8 in dlopen_doit (a=a@entry=0x7fffffffd720) at ./dlfcn/dlopen.c:56
#9  0x00007ffff7e47dea in __GI__dl_catch_exception (exception=exception@entry=0x7fffffffd680, operate=<optimized out>, args=<optimized out>) at ./elf/dl-error-skeleton.c:208
#10 0x00007ffff7e47e9f in __GI__dl_catch_error (objname=0x7fffffffd6d8, errstring=0x7fffffffd6e0, mallocedp=0x7fffffffd6d7, operate=<optimized out>, args=<optimized out>)
    at ./elf/dl-error-skeleton.c:227
#11 0x00007ffff7d7ddc7 in _dlerror_run (operate=operate@entry=0x7ffff7d7e280 <dlopen_doit>, args=args@entry=0x7fffffffd720) at ./dlfcn/dlerror.c:138
#12 0x00007ffff7d7e389 in dlopen_implementation (dl_caller=<optimized out>, mode=<optimized out>, file=<optimized out>) at ./dlfcn/dlopen.c:71
#13 ___dlopen (file=<optimized out>, mode=<optimized out>) at ./dlfcn/dlopen.c:81
#14 0x00005555555ee82a in ?? ()
#15 0x00005555555d3d02 in ?? ()
#16 0x00005555555f38a6 in ?? ()
#17 0x00005555555ded63 in ?? ()
#18 0x00005555555f38a6 in ?? ()
#19 0x00005555555df0c8 in ?? ()
#20 0x00005555555f38a6 in ?? ()
#21 0x0000555555577a73 in lua_pcall ()
#22 0x0000555555563203 in ?? ()
#23 0x00005555555685c4 in ?? ()
#24 0x00005555555f38a6 in ?? ()
#25 0x0000555555577ac1 in lua_cpcall ()
#26 0x000055555555c812 in main ()

```

Frenzie:
You confused me there — you're not talking about "my" AppImage build but some experimental Meson build. :-) We use `-mtune=generic -march=x86-64` which I occasionally test on my old Intel Atom netbook.

poire-z:
As expected with the MuPDF update, HTML dict users can now meet the crappier HTML rendering mentionned above at https://github.com/koreader/koreader-base/pull/1586#issuecomment-1640637412 :/

I guess it can only be worked around by massaging the HTML in each dict specific .lua tweak.

PR #1594, opened 2023-04-02
Fix brew not found on non-Mac platforms

Frenzie:
Follow-up to #1593

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1594)

### DIFF
--- a/ffi/util.lua
+++ b/ffi/util.lua
@@ -80,9 +80,7 @@ require("ffi/posix_h")
 
 local util = {}
 
-if ffi.os == "OSX" then
-    util.KO_DYLD_PREFIX = getlibprefix()
-end
+util.KO_DYLD_PREFIX = ffi.os == "OSX" and getlibprefix() or ""
 
 if ffi.os == "Windows" then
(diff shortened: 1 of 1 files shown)


NiLuJe:
That'll lead to a concat with a nil, won't it? (i.e., this should have an else that sets it to the empty string).

NiLuJe:
Oh, hey, it does. Never mind me :D.

NiLuJe:
Oh, right, I was looking at the diff from the first PR, I'm not going entirely barmy, that's reassuring :D.

Frenzie:
Yeah, I wasn't paying close enough attention and I was also doing some experiments with Podman.

The tl;dr is you need something like this which I don't find very convenient (but I guess it's supposedly more secure, at least on a server):
```
podman unshare chown -R 1000:1000 koreader-for-building-android
podman run -v $(pwd)/koreader-for-building-android:/home/ko/koreader -it localhost/frans/koandroid:0.5.3 bash
```

PR #1595, opened 2023-04-08
Add C declarations needed for zmq http server

bneo99:
Required by [koreader to create a zmq http server](https://github.com/koreader/koreader/pull/10223/)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1595)


ffi/zeromq_h.lua
```diff
@@ -109,4 +113,6 @@ int zpoller_add(zpoller_t *, void *);
 void *zpoller_wait(zpoller_t *, int);
 bool zpoller_expired(zpoller_t *);
 bool zpoller_terminated(zpoller_t *);
+int zmq_send(void *, void *, size_t, int);
+int zframe_send(zframe_t **, void *, int);
```
Frenzie, line 117: Not the biggest deal but these seem to be in the "wrong" location compared to their definition?

Frenzie (approved):
@NiLuJe ?

NiLuJe:
LGTM besides the ordering nit mentioned above (which will make diff'inf further updates confusing, so, yeah, please fix it ;)).

bneo99:
Whoops, should be fixed now

ffi-cdecl/zeromq_cdecl.c
```diff
@@ -112,7 +119,6 @@ cdecl_func(zhash_cursor)
 cdecl_func(zhash_lookup)
 cdecl_func(zmsg_destroy)
 cdecl_func(zhash_destroy)
-cdecl_func(zframe_destroy)
```
Frenzie, line 115: It looks like all the destroys are purposefully together. You prefer them as all the zframes together?

bneo99: There is other destroys that are grouped with their own class functions. I'll move the other destroys to their own  groups. So all functions to be grouped by their class (zhash, zctx, zframe etc.)

Frenzie (approved):
Fine by me, waiting for @NiLuJe ^_^

NiLuJe (approved):

PR #1596, opened 2023-04-08
Fix build on Mac OS X 13.3+ (missing ifr_netmask)

ptrm:
https://github.com/koreader/koreader/issues/8797#issuecomment-1499554576

Mac OS X 13.3 (and 13.3.1) seems to have dropped ifr_netmask exposition in the kernel. These changes attempt to automatically omit ifr_netmask patch for czmq for Darwin 22.4+ kernels.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1596)

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -510,12 +510,6 @@ endif
 ifdef DARWIN_AARCH64_HOST
 # At the moment of writing, apple-m1 is the lowest common denominator for Apple silicon chips. Should work for M2 as well
 	HOST_ARCH:=-mcpu=apple-m1
-
-	# Darwin 22.4 (Ventura 13.3), and presumably later kernels don't support ifr_netmask
-	ifneq ("$(shell echo -e `uname -r`"\n22.3" | sort -Vr | head -1)", "22.3")
-		export DARWIN_AARCH64_HOST_NO_IFR_NETMASK:=1
-	endif
-
 	BREW_PREFIX:=$(shell command brew --prefix 2>/dev/null)
 	ifneq (,$(BREW_PREFIX))
 		export DARWIN_AARCH64_HOST_CFLAG:=-I$(BREW_PREFIX)/include
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -510,6 +510,12 @@ endif
 ifdef DARWIN_AARCH64_HOST
 # At the moment of writing, apple-m1 is the lowest common denominator for Apple silicon chips. Should work for M2 as well
 	HOST_ARCH:=-mcpu=apple-m1
+
+	# Darwin 22.4 (Ventura 13.3), and presumably later kernels don't support ifr_netmask
+	ifneq ("$(shell echo -e `uname -r`"\n22.3" | sort -Vr | head -1)", "22.3")
+		export DARWIN_AARCH64_HOST_NO_IFR_NETMASK:=1
+	endif
+
 	BREW_PREFIX:=$(shell command brew --prefix 2>/dev/null)
 	ifneq (,$(BREW_PREFIX))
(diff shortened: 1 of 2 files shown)


Frenzie (approved):
I don't love it but of course I suggested this myself because I couldn't think of something obviously better.

Although I'm not entirely sure when the situation the patch covers occurs; a potential alternative could be to drop it completely?

NiLuJe:
Yeah, that patch doesn't make a lick of sense on macOS, it's specifically for glibc systems.

Just dropping it entirely on Darwin should be enough.

NiLuJe:
Oh, wait, I was looking at the wrong patch, rofl.

NiLuJe:
In any case, yeah, probably easier to just drop it entirely on Darwin and hope for the best ;o).

ptrm:
I wished to somehow avoid passing unneeded PATCH_COMMAND, but I could not place a working conditional inside `ExternalProject_Add()`. `PATCH_COMMAND ./autogen.sh ...` on the other hand seems to be generating a redundant run, and `PATCH_COMMAND ${UNDEFINED_VAR}` impeded the `CONFIGURE` step and broke the build.

```cmake
ExternalProject_Add(
    ${PROJECT_NAME}
 ➡️ PATCH_COMMAND ./autogen.sh COMMAND ${PATCH_CMD1} COMMAND ${PATCH_CMD2}
    CONFIGURE_COMMAND COMMAND ${CFG_CMD} COMMAND ${SED_CMD1} COMMAND ${SED_CMD2}
    INSTALL_COMMAND ${KO_MAKE_RECURSIVE} installdirs uninstall install
)
```

Frenzie:
Do you mean you tried the new in CMake 3.8 `$<IF:condition,true_string,false_string>`?

https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:IF

ptrm:
Nope, 

> Do you mean you tried the new in CMake 3.8 `$<IF:condition,true_string,false_string>`?
> 
> https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:IF

@Frenzie nope, not this one :D but even removing the PATCH_COMMAND line altogether results in
```
[ 50%] No patch step for 'czmq'
[ 62%] Performing configure step for 'czmq'
sh: /Users/piotr/Source/koreader/base/thirdparty/czmq/build/arm64-apple-darwin22.4.0-debug/czmq-prefix/src/czmq/configure: No such file or directory
make[4]: *** [czmq-prefix/src/czmq-stamp/czmq-configure] Error 127
make[3]: *** [CMakeFiles/czmq.dir/all] Error 2
make[2]: *** [all] Error 2
make[1]: *** [build/arm64-apple-darwin22.4.0-debug/libs/libczmq.1.dylib] Error 2
make: *** [all] Error 2
```

As if the only place autogen.sh was invoked was the PATCH step :o

ptrm:
Oh, so the patch step is sometimes used for invoking autogen :o
image

Frenzie:
> As if the only place autogen.sh was invoked was the PATCH_COMMAND step :o

Well of course it is, why would you invoke it multiple times. ;-) I'm not sure why it's done there and not in CONFIGURE_COMMAND but you have to do it somewhere after all. ^_^

Frenzie:
But to be clear, I was in no shape or form objecting to the PR, it's likely to be much clearer this way than inline anyway. I was just trying to think along.

ptrm:
Yeah, if autogen.sh needs to be invoked, I consider this fix finished, unless more feedback comes :)

NiLuJe (approved):

PR #1597, opened 2023-04-08
[CI] Switch Android Docker image from 18.04 to 22.04

Frenzie:
~~Maybe it will magically compile.~~
Cf. https://github.com/koreader/virdevenv/pull/84

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1597)


Frenzie:
@#$#@ whatever happened to our Android x86 build here. It's only the ARM one that works.

PR #1599, opened 2023-04-15
SDL: add multitouch support

Frenzie:
Partially inspired by the Android implementation.

After I added it, I suddenly realized we could've had fake multitouch on desktop for years. There's a right button on the mouse, after all. A possible alternative would be to simply fake right mouse button as being two fingers at once, as opposed to having to truly press left and right at the same time.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1599)

### DIFF
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -54,6 +54,10 @@ local function SDL_Linked_Version_AtLeast(x, y, z)
     return SDL_VersionNum(getSDLVersion()) >= SDL_VersionNum(x, y, z)
 end
 
+-- Used as the device ID for mouse events simulated with touch input.
+-- Defined in SDL_touch.h as #define SDL_TOUCH_MOUSEID ((Uint32)-1)
+local SDL_TOUCH_MOUSEID = ffi.cast('uint32_t', -1);
+
 local S = {
     w = 0, h = 0,
     screen = nil,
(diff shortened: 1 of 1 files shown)


Frenzie:
@ptrm Btw, could you check if removing these lines works as expected in combination with the check for `SDL_TOUCH_MOUSEID`? It might require a "new" version of SDL in case you still happen to be running one from, say, five years ago.

https://github.com/koreader/koreader-base/blob/263b2113c10719cda3fcf6f5981bc7848de32b31/ffi/SDL2_0.lua#L295-L300

NiLuJe:
Untested, but nothing jumps out at a quick glance ;).

Frenzie:
Rest assured I'd have noticed if anything didn't work. ;-)

@ptrm Have you been able to try the Apple touchpad? I noticed you left a thumbs up :-P

Frenzie:
I tested macOS 13.3.1 myself. It seems to be fine.

ffi/SDL2_0.lua
```diff
@@ -293,64 +348,74 @@ function S.waitForEvent(sec, usec)
...
+            slot = getFingerSlot(event)
+        elseif event.button.button == SDL_BUTTON_RIGHT then
+            slot = 1
+        else -- SDL_BUTTON_LEFT
+            slot = 0
+        end
+        local x = is_finger and event.tfinger.x * S.w or event.button.x
+        local y = is_finger and event.tfinger.y * S.h or event.button.y
+        setPointerDownState(slot, true)
+        genTouchDownEvent(event, slot, x, y)
```
NiLuJe, line 415: This may be called for !is_finger, while it *will* access `event.tfinger.fingerId`, I'm not sure what it points to in the SDL magic union in those cases ;).

Frenzie: Oh right, a minor oversight for the mouse, though it doesn't really matter in practice.

Frenzie: The most obvious would probably be:
```diff
diff --git a/ffi/SDL2_0.lua b/ffi/SDL2_0.lua
index 4823dcac..ee5a767d 100644
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -235,8 +235,9 @@ local function getFingerSlot(event)
 end
 
 local function genTouchDownEvent(event, slot, x, y)
+    local is_finger = event.type == SDL.SDL_FINGERDOWN
     genEmuEvent(C.EV_ABS, C.ABS_MT_SLOT, slot)
-    genEmuEvent(C.EV_ABS, C.ABS_MT_TRACKING_ID, tonumber(event.tfinger.fingerId))
+    genEmuEvent(C.EV_ABS, C.ABS_MT_TRACKING_ID, is_finger and tonumber(event.tfinger.fingerId) or slot)
     genEmuEvent(C.EV_ABS, C.ABS_MT_POSITION_X, x)
     genEmuEvent(C.EV_ABS, C.ABS_MT_POSITION_Y, y)
     genEmuEvent(C.EV_SYN, C.SYN_REPORT, 0)
```

I prefer tracking id to not be constantly 0 and 1 'cause if nothing else it's clearer to read. :sweat_smile:

ptrm:
> Rest assured I'd have noticed if anything didn't work. ;-)
> 
> @ptrm Have you been able to try the Apple touchpad? I noticed you left a thumbs up :-P

Sorry, that was meant as a quick sign of queueing it ;) will try it today, I don't have a touchpad on my mac mini though, so it'll be just mouse.

Frenzie:
Additional tests are welcome, but I did test the touchpad on a Macbook Air M1 already. It didn't seem to generate finger events.

Looking again that used to be the behavior at some point in the past (and clearly it was back ca. 2014 when that code was written), but it looks like more recent SDL has made it configurable:

https://github.com/libsdl-org/SDL/commit/bdc7f958fda50305744dcefdccac29373da97d01

Going back a little further, it was four years ago that the touchpad behavior was changed for Mac, see https://github.com/libsdl-org/SDL/commit/e841b066fd76aa1eb7cf2be9a46cc82798ab3c33 and a little further back https://github.com/libsdl-org/SDL/commit/d9a2eff26f1088535c2f6e608be042b524498e7b

So I think we might want to set `SDL_HINT_TRACKPAD_IS_TOUCH_ONLY` for best results in SDL 2.23+ (?) but the rest of the code should work fine either way regardless on any SDL from at least the past four years. And if there are any lingering issues with synthesized touch or mouse events they'll probably occur across platforms.

PR #1601, opened 2023-04-17
SDL: enable `SDL_HINT_TRACKPAD_IS_TOUCH_ONLY` for multitouch on supported trackpads

Frenzie:
macOS only for now, in SDL itself.

Follow-up to <https://github.com/koreader/koreader-base/pull/1599>.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1601)


Frenzie:
This doesn't actually seem to do anything because I'm just getting mousewheel events, which is perhaps more expected anyway. I opened the PR mainly for archiving purposes. :-)

Issue #1300, opened 2021-02-10
SDL: do something with SDL_GL_GetDrawableSize


Frenzie:
For Wayland type stuff where they seem to have the odd concept that a window will think it's 800x600 even though it's displayed at 1600x1200 (or whatever). Unless I'm overlooking something it's not just as simple as declaring that you'll resize nicely on any size (which I thought we already did with the allow hidpi flag).
https://wiki.libsdl.org/SDL_GL_GetDrawableSize

PR #1602, opened 2023-04-18
SDL: add idiotic workaround to be less blurry on HiDPI

Frenzie:
Fixes https://github.com/koreader/koreader-base/issues/1300

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1602)


Frenzie:
NB I had the displeasure of looking at macOS bad blurry scaling for the past half hour and I'd like to take this opportunity to state to the person I don't recall who told me that macOS is sharp that they are completely and utterly bloody wrong.

a. It sizes things up to 2x so it will *never not be blurry*.
b. If you go beyond 2x it *doesn't go to 3x so it's blurry in the **bad** way*

Frenzie:
Mac resolutions.

The way it works is:

* 2160 x 3840 is 1x
* everything else is 2x

Frenzie:
Anyway, on Windows things have been more or less good since Windows 8 (although they were better in 7) and there's a glimmer of light in XWayland of all places, and I understand Wayland itself will also start to behave sanely at some point in the future, so we can go back to the world of 2015 where everything "HiDPI" worked the way it should:
image

ffi/framebuffer_SDL2_0.lua
```diff
@@ -29,6 +29,17 @@ end
 function framebuffer:resize(w, h)
     w = w or SDL.w
     h = h or SDL.h
+    local output_w = ffi.new("int[1]", 0)
+    local output_h = ffi.new("int[1]", 0)
+    if SDL.SDL.SDL_GetRendererOutputSize(SDL.renderer, output_w, output_h) == 0 and tonumber(output_w[0]) ~= w then
```
NiLuJe, line 34: I assume the double SDL is an unfortunate side-effect of our own ffi wrapper being named SDL, and the C namespace for the actual SDL lib being named SDL too?

NiLuJe:
I think I grok the intent, but I'm unfamiliar with both SDL internals and HiDPI woes in general to say much about it, but I imagine it's at least less worse than before ;).

ffi/framebuffer_SDL2_0.lua
Frenzie: Yes, we could change one of them if desired for clarity. The construct occurs a fair bit throughout the code dealing with SDL I'm sure. ;-)

The one called `S` in `SDL2_0.lua` is called `SDL` in framebuffer here.
https://github.com/koreader/koreader-base/blob/df96630c21ff9c5ededa1a864da2c048e4c542ed/ffi/SDL2_0.lua#L61-L67

Frenzie:
Afaik it should be perfect on Windows… potentially with some additional hacks to rub Windows' nose in the fact that yes, we can do HiDPI. ;-P

PR #1608, opened 2023-05-10
Distinguish between touch and non-touch pointer events

hrdl-github:
Fixes koreader/koreader#10417.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1608)

### DIFF
--- a/ffi/SDL2_0.lua
+++ b/ffi/SDL2_0.lua
@@ -218,7 +218,8 @@ end
 
 -- Keep track of active pointers so we can feed ABS_MT_SLOT 0 and 1 to the frontend for multitouch to work.
 -- down, a boolean denoting whether the pointer is currently down for tracking mouse button status.
--- is_finger is currently only used to work around a bug in sway, cf. <https://github.com/koreader/koreader/issues/10417>
+-- is_finger is currently only used to work around a bug in sway,
+-- cf. <https://github.com/koreader/koreader/issues/10417#issuecomment-1542781926>
 local pointers = {}
 local function setPointerDownState(slot, down, is_finger)
     if not pointers[slot] then


Frenzie:
As written wouldn't the slot 0 is_finger state simply swap between true and false a bit?

ffi/SDL2_0.lua
```diff
@@ -423,7 +424,7 @@ function S.waitForEvent(sec, usec)
         end
         local x = is_finger and event.tfinger.x * S.w or event.button.x
         local y = is_finger and event.tfinger.y * S.h or event.button.y
-        setPointerDownState(slot, true)
+        setPointerDownState(slot, true, is_finger)
```
Frenzie, line 427: Oh right, in the specific sway bug scenario this one will only trigger once. But in that case this would need a lot of comments to explain what it's doing since this won't solve the generic scenario that a naive reader would expect to be solved.

ffi/SDL2_0.lua
```diff
@@ -219,11 +219,12 @@ end
 -- Keep track of active pointers so we can feed ABS_MT_SLOT 0 and 1 to the frontend for multitouch to work.
 -- down, a boolean denoting whether the pointer is currently down for tracking mouse button status.
```
Frenzie, line 220: If you make it something like this I'll probably merge it in a few days.
```suggestion
-- down, a boolean denoting whether the pointer is currently down for tracking mouse button status.
-- is_finger is currently only used to work around a bug in sway, cf. <https://github.com/koreader/koreader/issues/10417>
```

Frenzie (approved):
@NiLuJe Workaround fine by you?

ffi/SDL2_0.lua
NiLuJe: Yeah, more comments would be good; because that MOUSEMOTION event in the middle of the sandwich is hella weird ;o).

(I wouldn't mind a link to https://github.com/koreader/koreader/issues/10417#issuecomment-1542781926 in said comment, as this clearly showcases the issue).

NiLuJe:
Yeah, that looks sensible to me ;).

Frenzie:
Alright, let's wait and see a bit.

PR #1612, opened 2023-05-17
Return size of PDF page document

hrdl-github:
This is a rough approximation to be used by the caching mechanism. I assume one byte per character and a small constant size for coordinates. I don't know how much overhead the table and individual structures require.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1612)

### DIFF
--- a/ffi/mupdf.lua
+++ b/ffi/mupdf.lua
@@ -496,6 +496,7 @@ function page_mt.__index:getPageText()
     -- now we analyze the data returned by the device and bring it
     -- into the format we want to return
     local lines = {}
+    local size = 0
 
     local block = text_page.first_block
     while block ~= nil do
@@ -550,6 +551,7 @@ function page_mt.__index:getPageText()
                             x0 = word_bbox[0].x0, y0 = word_bbox[0].y0,
                             x1 = word_bbox[0].x1, y1 = word_bbox[0].y1,
                         })
+                        size = size + 5 * 8 + textlen
 
                         if ch == nil then
                             break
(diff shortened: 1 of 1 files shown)


Frenzie:
Hm, I guess that change causes this test to fail?
https://github.com/koreader/koreader-base/blob/e2778f3863136c5b885c8ba560ea603bb699abc7/spec/unit/koptcontext_spec.lua#L203

benoit-pierre:
That test is flaky (because of possibly uninitialized memory).

NiLuJe:
> That test is flaky (because of possibly uninitialized memory).

Did you pinpoint exactly where the issue lies?

Frenzie:
I told the tests to run again and they seem to have passed, so that would indeed seem to be slightly flaky. I'm not sure I've ever noticed it flaking before, or perhaps I just tune it out when I know it can't possibly be related which may not necessarily be the case here although it's unlikely.

benoit-pierre:
> > That test is flaky (because of possibly uninitialized memory).
> 
> Did you pinpoint exactly where the issue lies?

Not yet.

NiLuJe:
> I told the tests to run again and they seem to have passed, so that would indeed seem to be slightly flaky. I'm not sure I've ever noticed it flaking before

It's been borked for years ;).

It used to be way way worse because at least one of the copies used to be borked (not null terminated), but it's still random today (and I think that's the one "random" test failure that I could actualy reproduce locally; but it's been a while).

So, yeah, something's fishy with that one, for sure.

benoit-pierre:
See https://github.com/koreader/koreader-base/pull/1613.

PR #1614, opened 2023-06-02
mxcfb_kindle: add temp ioctl constants

yparitcher:
My PW5 is experiencing black -> white ghosting esp. on resume from suspend. It does not happen every time, and i suspect it is temperature dependent (it hapens most when switching climate indoors <-> outdoors) and my kindle's thermometer does not work so well :)
I adapted @poire-z footer patch from [here](https://github.com/koreader/koreader/issues/10231#issuecomment-1477138340) and indeed the temp is ~10°F above ambient, it is annoying to `ffi.cdef` the ioctl constants everywhere i need them, so let us just make the public here. also add the `mxcfb_panel_info` while i am at it.
I plan on integrating this into my SS plugin to auto adjust the temp a few degrees to compensate, and being as we pass `TEMP_USE_AMBIENT` everywhere it will go back to that with no long term effect. This approach appears to be working so far, but i need some more field testing

```lua
local ReaderFooter = require("apps/reader/modules/readerfooter")
local Screen = require("device").screen
local logger = require("logger")
local util = require("util")
local ffi = require("ffi")
local C = ffi.C
require("ffi/posix_h")
ffi.cdef[[
static const int MXCFB_SET_TEMPERATURE = 1074021932;
static const int MXCFB_GET_TEMPERATURE = 2147763768;
]]

TEMP_COUNT = "custom_text"
ReaderFooter.textGeneratorMap[TEMP_COUNT] = function(footer)
    local temp_c = ffi.new("uint32_t[1]")
    C.ioctl(Screen.fd, C.MXCFB_GET_TEMPERATURE, temp_c)
    local temp_f = (temp_c[0] * 1.8) + 32
    return temp_f .. "°"
end
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1614)

### DIFF
--- a/ffi-cdecl/mxcfb_kindle_decl.c
+++ b/ffi-cdecl/mxcfb_kindle_decl.c
@@ -46,6 +46,9 @@ cdecl_const(TEMP_USE_AUTO)
 
 cdecl_const(TEMP_USE_ZELDA_AUTO)
 
+cdecl_const(MXCFB_SET_TEMPERATURE)
+cdecl_const(MXCFB_GET_TEMPERATURE)
+
 cdecl_const(EPDC_FLAG_ENABLE_INVERSION)
 cdecl_const(EPDC_FLAG_FORCE_MONOCHROME)
 cdecl_const(EPDC_FLAG_USE_CMAP)
@@ -120,6 +123,7 @@ cdecl_enum(MTK_SWIPE_DIRECTION_ENUM)
 cdecl_struct(mxcfb_swipe_data)
 
 cdecl_struct(mxcfb_update_data_mtk)
+cdecl_struct(mxcfb_panel_info)
 cdecl_struct(mxcfb_halftone_data)
 cdecl_type(mxcfb_markers_data)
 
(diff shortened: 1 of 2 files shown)


NiLuJe:
FWIW, it *should* be above ambient ;).

It's been a while since I checked, but, indoors, after a bit of use, I seem to remember something like 4 to 5°C above ambient there.

As for your actual experiments, 24°C is *usually* a threshold in the waveforms, but I can't vouch for that on Kindle ;).

(IIRC, there might also be a way to set how often the auto/ambient temperature is updated by the driver).

NiLuJe (approved):

yparitcher:
> FWIW, it *should* be above ambient ;).
> 
> It's been a while since I checked, but, indoors, after a bit of use, I seem to remember something like 4 to 5°C above ambient there.
> 
> As for your actual experiments, 24°C is *usually* a threshold in the waveforms, but I can't vouch for that on Kindle ;).
> 
> (IIRC, there might also be a way to set how often the auto/ambient temperature is updated by the driver).

Ty for this esoteric knowledge :)

Being as there is no documentation, it is hard to figure out what exactly is happening

PR #1616, opened 2023-06-11
fix fribidi includes

benoit-pierre:
```bash
▸ pkg-config --cflags fribidi
-I/usr/include/fribidi
```

Otherwise, using `fribidi/fribidi.h` may result in the wrong (system) include being picked up when building the emulator.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1616)


benoit-pierre:
This depends on https://github.com/koreader/crengine/pull/519.

poire-z:
Fine with me, but not as familiar with build stuff as @NiLuJe and @Frenzie .
Btw, how is it that picking system include does not happen for other stuff ? Is it just because I added this single occurence of an include with a subdir ?

benoit-pierre:
It does not happen right now, but if you're messing with the build system, it's easy to get the wrong include because of it.

NiLuJe (approved):

poire-z (approved):

poire-z:
No longer needs to be "Draft" ?

benoit-pierre:
Yeah, I was waiting for the macOS check.

Frenzie (approved):

PR #1618, opened 2023-06-11
fix most freetype/harfbuzz related memory leaks

benoit-pierre:
There's still one small leak remaining due to using `FT_Done_Library` instead of `FT_Done_FreeType` to handle the Freetype's library lifecycle: its memory object is not freed on exit.

This is also the precursor to some [memory optimizations](https://github.com/koreader/koreader/commit/59f982c65bb9e11dde50fe8b729e7433aa037d8c).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1618)

### DIFF
--- a/ffi-cdecl/freetype2_decl.c
+++ b/ffi-cdecl/freetype2_decl.c
@@ -3,12 +3,13 @@
 #include "ffi-cdecl.h"
 
 #include <freetype/freetype.h>
-#include <freetype/ftsynth.h>
+#include <freetype/ftmodapi.h>
 #include <freetype/ftoutln.h>
 #include <freetype/ftsnames.h>
+#include <freetype/ftsynth.h>
+#include <freetype/ttnameid.h>
 #include <freetype/tttables.h>
 #include <freetype/tttags.h>
-#include <freetype/ttnameid.h>
 
 cdecl_type(FT_String)
 cdecl_type(FT_Byte)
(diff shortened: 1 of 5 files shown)


poire-z:
> fix most freetype/harfbuzz related memory leaks 

Btw, how much of that amount to the well known KOReader memory leaks (cf https://github.com/koreader/koreader/issues/5998#issuecomment-602180998) ?
Does your mesonic monolibtic build still get its memory usage increasing as obviously as ours ?

poire-z (approved):
A bit over my head, so it must be ok :)

benoit-pierre:
Good question, memory does increase as more features get used, menus opened, but I'm not sure it's leaking. Honestly, I don't have to restart on my Kindle anymore thanks to:
- those changes coupled with the aforementioned fonts related memory optimizations: it shaves 30 MB from Koreader's memory use after browsing in the file manager, opening a (previously opened) EPUB, and triggering the various bottom settings panels
- using my meson' size optimized build (`buildtype = 'minsize'`), combined with a pass on all libraries to ensure GNU visibility support is used, as well as trimming unnecessary features and statically linking when applicable: it shaves about 10 MB from the combined BSS+data+code sections

monolibtic is currently broken (because of the visibility patches), but technically fixable.

Frenzie (approved):

NiLuJe (approved):
I'm not up to snuff on the ft/hb API, but this looks sound ;).

PR #1619, opened 2023-06-11
fix FFI cdecls

benoit-pierre:
Nothing major, I just re-run `ffi-cdecl` to ensure prototypes are up-to-date, and removed some ZLIB related duplicates.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1619)


Frenzie (approved):

ffi/leptonica_h.lua
```diff
@@ -33,7 +42,7 @@ struct Numa {
   l_float32 *array;
 };
 typedef struct Numa NUMA;
-static const int L_INSERT = 0;
+const l_int32 L_INSERT;
```
NiLuJe, line 45: I'm not sure what LuaJIT assumes for those (and can't check what the C code says right now).

benoit-pierre: What's your concern?

Note:

- there's also `L_NOCOPY`, which is actually part of the enum:
```c
/*! Access and storage flags */
enum {
    L_NOCOPY = 0,     /*!< do not copy the object; do not delete the ptr  */
    L_COPY = 1,       /*!< make/use a copy of the object                  */
    L_CLONE = 2,      /*!< make/use clone (ref count) of the object       */
    L_COPY_CLONE = 3  /*!< make a new object and fill each object in the  */
                      /*!< array(s) with clones                           */
};
static const l_int32  L_INSERT = 0;  /*!< stuff it in; no copy or clone   */
```
- we don't currently use it anyway, so we could just remove the definition (for monolibtic mode I use the FFI cdecl files to determine which symbols are needed and must be exported and trim the rest when not used).

NiLuJe: I'd prefer `static const l_int32 L_INSERT = 0;`, FWIW ;).

(I don't recall if that's one of the "hand tweaked" ones, IIRC, there's a few of them in the kpdf2opt stuff).

benoit-pierre: I've replaced `L_INSERT` by `L_NOCOPY`, which is the better name anyway.

poire-z:
@NiLuJe : good for you ?

NiLuJe (approved):
Oops, missed the update, my bad ;).

PR #1620, opened 2023-06-24
android: update to use NDK 23c

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1620)

PR #1624, opened 2023-06-26
update minimum C++ standard requirement

benoit-pierre:
At least C11 & C++17.

Will fix warnings when bumping crengine to include #520. 

For the Android toolchain, #1620 is required.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1624)


benoit-pierre:
Oh great, outdated make versions all around… ;(

Frenzie:
> Oh great, outdated make versions all around… ;(

Ftr, I like force pushing a rebase at the end when all is said and done but less so when it means I don't know what you're talking about. ;-) Most notably what did you change in response and what kind of versions are we talking about? I might prefer to tackle the problem from the other angle for example.

benoit-pierre:
> > Oh great, outdated make versions all around… ;(
> 
> Ftr, I like force pushing a rebase at the end when all is said and done but less so when it means I don't know what you're talking about. ;-) Most notably what did you change in response and what kind of versions are we talking about? I might prefer to tackle the problem from the other angle for example.

```bash
▸ make3 --version
```
```
GNU Make 3.81
[…]
```
```bash
▸ make --version
```
```
GNU Make 4.4.1
[…]
```
```bash
▸ make3 -f - <<\EOF
test := $(shell echo '#')
$(info $(call test))
all:
EOF
```
```
/tmp/GmqHWYeT:1: *** unterminated call to function `shell': missing `)'.  Stop.
```
```bash
▸ make -f - <<\EOF
test := $(shell echo '#')
$(info $(call test))
all:
EOF
```
```
#
make: Nothing to be done for 'all'.
```
```bash
▸ make3 -f - <<\EOF
test := $(shell echo '\#')
$(info $(call test))
all:
EOF
```
```
#
make: Nothing to be done for 'all'.
```
```bash
▸ make -f - <<\EOF
test := $(shell echo '\#')
$(info $(call test))
all:
EOF
```
```
\#
make: Nothing to be done for 'all'.
```
```bash
▸ make3 -f - <<\EOF
$(info $(intcmp 0,1,LT,EQ,GT))
all:
EOF
```
```

make3: Nothing to be done for `all'.
```
```bash
▸ make -f - <<\EOF
$(info $(intcmp 0,1,LT,EQ,GT))
all:
EOF
```
```
LT
make: Nothing to be done for 'all'.
```

benoit-pierre:
The `android-arm` CI failure was expected, but I have no idea why the `pocketbook` and `sony-prstux` jobs failed (`kindle` is another transient network error when fetching the certificates bundle during OpenSSL build).

Frenzie:
I intend to update that to 4.2.1 (that is, Ubuntu 20.04) in the not overly distant future given that 18.04 is at end of standard support and the new Debian just came out and we could always include a newer one regardless, but given how minor this is yeah, let's just go with this as written here now.

benoit-pierre:
I'm still not sure why `pocketbook` and `sony-prstux` are failing, but this is going to be a problem:
```
-- The C compiler identification is GNU 7.5.0
```
Note that another user of `if constexpr (…)` is going to be tesseract 5.3.1, when updating to k2pdfopt 2.54.

Frenzie:
I thought GCC 7 already supported that, but in any case don't worry about it, see above. I already prepared 20.04 a long time ago. (Alternative GCC options are also simple enough.)

NiLuJe:
Err, why the hell is it using GCC 7? o_O

The koxtoolchain pocketbook target should basically be identical to everything else (the closest match being kindle, FWIW).

Wrong/outdated docker image?

NiLuJe:
We can probably drop prstux entirely, it's abandoned, and the docs/code are missing some key peices making the base platform actually impossible to reproduce (and as such install).

Frenzie:
I assumed that was only referring to the desktop and "Sony" builds, the latter of which can be dropped.

Frenzie:
Actually we use GCC 8 on desktop, so problem solved already. ;-)

Not using 8 for the prstux build may be an oversight on my end, but as stated, drop it.

Frenzie:
Fingers crossed the 22.04 image will Just Work™ since I'm not sure I ever tested it yet. But if not it I need to fix it up regardless.

benoit-pierre:
> Fingers crossed the 22.04 image will Just Work™ since I'm not sure I ever tested it yet. But if not it I need to fix it up regardless.

Looks like it worked. That leaves the issue of pocketbook.

NiLuJe:
> That leaves the issue of pocketbook.

c.f., https://github.com/koreader/koreader-base/pull/1624#issuecomment-1608307447

Do you remember something that could have happened during the last image updates after the last major koxtoolchain updates, @Frenzie ?

benoit-pierre:
It looks like both the `kindle` and `pocketbook` toolchains are not up-to-date, see [`gen-tc.sh`](https://github.com/koreader/koxtoolchain/blob/master/gen-tc.sh#L133):
```bash
	kindle)
		# NOTE: Don't swap away from the 1.23-kindle branch,
		#       this TC currently fails to build on 1.24-kindle...
		Build_CT-NG \
			https://github.com/NiLuJe/crosstool-ng.git \
			9954e26ad2a0781a12f0066411bc5ee5cd8b2829 \
			"arm-${1}-linux-gnueabi"
		;;
[…]
	pocketbook)
		# NOTE: Don't swap away from the 1.23-kindle branch,
		#       this TC currently fails to build on 1.24-kindle...
		Build_CT-NG \
			https://github.com/NiLuJe/crosstool-ng.git \
			9954e26ad2a0781a12f0066411bc5ee5cd8b2829 \
			"arm-${1}-linux-gnueabi"
		;;
```

NiLuJe:
Oh, crap, I remembered the kindle bit, but not the PB one -_-".

NiLuJe:
Which means the kindle-legacy target is also quite likely to fail in the same way, it's just not in the CI matrix...

NiLuJe:
Do you actively need C++17 or would C++14 do the job?

i.e., should we keep GCC 7.5 at its default, which ought to be gnu++14; or do I need to see if we can update those two ancient crappy TC to something between 7.5 and 11.2?

benoit-pierre:
C++17 is needed for `if constexpr (…)`. Beside the crengine cleanup patch, tesseract 5.3.1 will need it too.

NiLuJe:
Okay then, I'll throw stuff at the wall and see what sticks then...

NiLuJe:
So far:

* The entire GCC-11 branch is similarly fucked, so it's not a weird bug that was fixed later.
* Downgrading to the same binutils version as the older branch (2.36.1) doesn't help, so, unlikely to be a weird linker issue, either.

Currently trying every GCC branch, as that's the only other variable we can control anyway (we need softfp, and we need the crazy old glibc version)...

NiLuJe:
Well, what stuck is "nothing" ;D.

The semi good news is that the same GCC version as the previous branch *also* fails, which means it's related to an internal ct-ng change somewhere.

I'll see if it magically works again once I'm done updating everything, but that'll take a while (and I'm waiting on binutils 2.41, as it fixes a binary bloat issue on arm32), soooo, no good news, really.

The last-ditch attempt for right now is to move to GCC 8 on the old branch and see where that gets us (but it'll be unpatched, and that may be broken on arm softfp, yay?).

NiLuJe:
> The last-ditch attempt for right now is to move to GCC 8 on the old branch and see where that gets us (but it'll be unpatched, and that may be broken on arm softfp, yay?).

Nope, because GCC 8 was when the arm port started being pissy about mixing -march w/ -mcpu, and old ct-ng cannot deal with it sanely.

benoit-pierre:
OK, so the `-std=gnu++17` is supported by gcc 7.5.0, and crengine compiles fine. Looking more closely at the zsync error, it looks like the linker is confused by the additional `main` present in `haslib.a`'s `digest.o`, even though `-Wl,--as-needed` is used.

So let's just drop that file:

```diff
 thirdparty/zsync2/CMakeLists.txt         |  4 +++-
 thirdparty/zsync2/fix-old-tc-build.patch | 15 +++++++++++++++
 2 files changed, 18 insertions(+), 1 deletion(-)

diff --git c/thirdparty/zsync2/CMakeLists.txt i/thirdparty/zsync2/CMakeLists.txt
index e7663cae..680c4be6 100644
--- c/thirdparty/zsync2/CMakeLists.txt
+++ i/thirdparty/zsync2/CMakeLists.txt
@@ -55,6 +55,8 @@ else()
 endif()
 #list(APPEND CMAKE_ARGS "-DCURL_INCLUDE_DIRS=${CURL_DIR}/include")
 
+set(PATCH_CMD1 "${KO_PATCH} ${CMAKE_CURRENT_SOURCE_DIR}/fix-old-tc-build.patch")
+
 # NOTE: We're currently using our own fork instead of upstream's (https://github.com/AppImage/zsync2):
 ## * Enable range requests optimizations on the CLI
 ## * Re-implement the -u flag for compatibility w/ the old zsync CLI
@@ -80,7 +82,7 @@ include(ExternalProject)
 ExternalProject_Add(
     zsync2
     DOWNLOAD_COMMAND ${CMAKE_COMMAND} -P ${GIT_CLONE_SCRIPT_FILENAME}
-    PATCH_COMMAND ""
+    PATCH_COMMAND COMMAND ${PATCH_CMD1}
     CMAKE_ARGS "${CMAKE_ARGS}"
     CMAKE_GENERATOR "Unix Makefiles"
     BUILD_COMMAND ${KO_MAKE_RECURSIVE} -j${PARALLEL_JOBS}
diff --git c/thirdparty/zsync2/fix-old-tc-build.patch i/thirdparty/zsync2/fix-old-tc-build.patch
new file mode 100644
index 00000000..db614100
--- /dev/null
+++ i/thirdparty/zsync2/fix-old-tc-build.patch
@@ -0,0 +1,15 @@
+ lib/hashlib/CMakeLists.txt | 1 -
+ 1 file changed, 1 deletion(-)
+
+diff --git i/lib/hashlib/CMakeLists.txt w/lib/hashlib/CMakeLists.txt
+index 63d3852..2f7880c 100644
+--- i/lib/hashlib/CMakeLists.txt
++++ w/lib/hashlib/CMakeLists.txt
+@@ -11,7 +11,6 @@ include_directories(${PROJECT_SOURCE_DIR}/include)
+ 
+ set(HASHLIB_SRCS
+     ${PROJECT_SOURCE_DIR}/src/crc32.cpp
+-    ${PROJECT_SOURCE_DIR}/src/digest.cpp
+     ${PROJECT_SOURCE_DIR}/src/keccak.cpp
+     ${PROJECT_SOURCE_DIR}/src/md5.cpp
+     ${PROJECT_SOURCE_DIR}/src/sha1.cpp
```

With that patch I can compile locally.

NiLuJe:
I also got a possibly not broken GCC 9.5 TC going over here, FWIW ;).

NiLuJe:
> OK, so the `-std=gnu++17` is supported by gcc 7.5.0

Cool, I have *vastly* more confidence in the current TC (based on the final Linaro GCC release) than in whatever monstrosity I just cobbled together.

(Given that it's the Linaro branch, it may actually be fancier than what one might find in the FSF branch, which could explain why it didn't fall on its face like the supposedly newer Android GCC version did, for instance).

Anyway, at least we'll have the option of something newer and maybe working in case we need it down the line ;).

NiLuJe:
> I also got a possibly not broken GCC 9.5 TC going over here, FWIW ;).

That would be the last couple of commits in the [1.23-kindle branch](https://github.com/NiLuJe/crosstool-ng/tree/1.23-kindle).

NiLuJe (approved):
> Reviewed 12 of 16 files at r1, 5 of 6 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1624)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @benoit-pierre)

poire-z:
Good to go ? Or best to have a nightly with NDK stuff but without that?

benoit-pierre:
I think we might as well merge everything.

PR #1617, opened 2023-06-11
cre: fix and enable teardown

benoit-pierre:
Depends on https://github.com/koreader/crengine/pull/520.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1617)

PR #1625, opened 2023-06-26
decrease fonts memory use

benoit-pierre:
Instead of opening the same font multiple times for each different size (multiple face instances), share one face instance and create multiple size instances.

Running the emulator, and going through those steps:
1. start in the filemanager, navigate (3 pages) to and open a cached EPUB
2. trigger all bottom configuration panels, and the menu top-level entries

Here's the memory use (as reported by KOReader itself):

| After step  | Master | PR |
|--|--:|--:|
| 1 | 115 | 90 |
| 2 | 148 | 107 |

Mind you, one of the entry in the filemanager listing is a collection of novels whose long list of authors result in a ridiculously small label. When going straight to the same previously opened EPUB, the results are:

| After step  | Master | PR |
|--|--:|--:|
| 1 | 95 | 85 |
| 2 | 128 | 107 |

So YMMV. I get similar gains on my Kindle.

Note, this need a small patch to the frontend:
```diff
 frontend/fontlist.lua | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git i/frontend/fontlist.lua w/frontend/fontlist.lua
index b28f843a8..2e516041b 100644
--- i/frontend/fontlist.lua
+++ w/frontend/fontlist.lua
@@ -115,9 +115,9 @@ local function collectFaceInfo(path)
         end
 
         -- If family_name is missing, it's probably too broken to be useful
-        if face.family_name ~= nil then
+        if face.face.family_name ~= nil then
             local fres = face:getInfo()
-            local hbface = HB.hb_ft_face_create_referenced(face)
+            local hbface = HB.hb_ft_face_create_referenced(face.face)
             fres.names = hbface:getNames()
             fres.scripts, fres.langs = hbface:getCoverage()
             fres.path = path
```

Possible improvements, make the size more face like, and "absorb" the face helpers? E.g., instead of having:
```lua
function FTSize_mt.__index:getEmboldenHalfStrength(factor)
    assert(ft2.FT_Activate_Size(self) == 0, "failed to activate font size")
    return self.face:getEmboldenHalfStrength(factor)
end
```
Inline the code for `face:getEmboldenHalfStrength`, and reduce the `face` implementation to the bare minimum (alloc/gc helpers)?

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1625)


NiLuJe:
> Possible improvements

If it makes the API more straight-forward, sure.

(With that said, it's been a while since I've used this, so, without much context, the current approach doesn't strike me as terrible either ;p).

NiLuJe:
> the current approach doesn't strike me as terrible either ;p

Besides the fact that a function named `newFace` returns an `FTSize`, that is ;p

NiLuJe:
The only other nit I have is the lack of our usual module `local C = ffi.C` ;).

xtext.cpp
```diff
@@ -8,6 +8,8 @@
 // For many links and notes about the concepts and libraries used,
 // see: https://github.com/koreader/crengine/issues/307
 
+#include<assert.h>
```
NiLuJe, line 11: nit: missing a space

xtext.cpp
```diff
@@ -577,7 +581,7 @@ class XText {
             luaL_typerror(m_L, -1, "cdata");
         }
         // Get the usable (for Harfbuzz) FT_Face object
-        FT_Face face = *(FT_Face *)lua_topointer(m_L, -1);
+        FT_Size size = *(FT_Size *)lua_topointer(m_L, -1);
```
poire-z, line 584: The comment above should be adapted for the change.
I wouldn't mine in here some small comments explaining what FT_Size is and how it relates to FT_Face.

(Also, I would naively expect a variable named "size" to be a number :) while here it is an FT internal object, so dunno if we could give it a better name like `sized_face` ? Not insisting on that, if you prefer the straightforward `FT_Size size` :) but it needs a comment so that we understand early it's not a number.)

ffi/freetype.lua
```diff
@@ -186,18 +215,71 @@ end
 
 local FTFaceType = ffi.metatype("struct FT_FaceRec_", FTFace_mt) -- luacheck: ignore 211
 
+local FTSize_mt = {__index={}}
+
+function FTSize_mt.__index:checkGlyph(char)
+    assert(ft2.FT_Activate_Size(self) == 0, "failed to activate font size")
```
poire-z, line 221: Sure `FT_Activate_Size()` is really cheap when it is already the active FT_Size, as we'll be calling it often ? And that we don't need to store on the Lua side which is active to avoid calling it ?

ffi/freetype.lua
```diff
@@ -186,18 +215,71 @@ end
...
-    if ft2.FT_Set_Pixel_Sizes(face, 0, pxsize) ~= 0 then
-        face:done()
+    local cache_key = string.format('%s:%d', filename, faceindex)
+    local face = faces_cache[cache_key]
+    if not face then
+        face = new_face(filename, faceindex)
+        faces_cache[cache_key] = face
+    end
+    local size = new_size(face)
+    if ft2.FT_Activate_Size(size) ~= 0 or ft2.FT_Set_Pixel_Sizes(face, 0, pxsize) ~= 0 then
```
poire-z, line 275: I assume this means that we first try to activate it, and if this fails, it means it has not yet been "created/sized" (?) and that we need to call FT_Set_Pixel_Sizes() to "create/size" it and that this implcitely activate it ?
Some small comment to explain that gymnastic (and/or 2 distincts branches to make that clearer?) would be nice.

And yes, as my review comment on xtext.cpp, some small comment explaining what "size" is here too would be nice.

xtext.cpp
benoit-pierre: Amended.

ffi/freetype.lua
benoit-pierre: Yes:
```C
  FT_EXPORT_DEF( FT_Error )
  FT_Activate_Size( FT_Size  size )
  {
    FT_Face  face;

    if ( !size )
      return FT_THROW( Invalid_Size_Handle );

    face = size->face;
    if ( !face || !face->driver )
      return FT_THROW( Invalid_Face_Handle );

    /* we don't need anything more complex than that; all size objects */
    /* are already listed by the face                                  */
    face->size = size;

    return FT_Err_Ok;
  }
```

The alternative would be to not do it automatically, and add an explicit call to `ftsize:activate()` before using any other methods.

xtext.cpp
benoit-pierre: Amended.

benoit-pierre:
Updated:
- no more `FTFace_mt`, code has been inlined in each `FTSize_mt` method
- the frontend code now uses `ftsize`, instead of `face` / `ftface` (see https://github.com/koreader/koreader/pull/10618)

Note:
- `hasKerning()` is gone: it was unused
- wouldn't it make more sense for `checkGlyph(char)` to return a boolean? (Instead of `1` if the character is present in the font, and `0` otherwise).

poire-z:
> wouldn't it make more sense for checkGlyph(char) to return a boolean? (Instead of 1 if the character is present in the font, and 0 otherwise).

Sure !

ffi/freetype.lua
```diff
@@ -184,28 +203,41 @@ function FTFace_mt.__index:getInfo()
...
+function FTSize_mt.__index:done()
+    done_size(self)
+end
+
+local FTSizeType = ffi.metatype("struct FT_SizeRec_", FTSize_mt) -- luacheck: ignore 211
 
-function FT.newFace(filename, pxsize, faceindex)
+local faces_cache = setmetatable({}, {__mode="v"})
+
+function FT.newSize(filename, pxsize, faceindex)
```
poire-z, line 214: What do you think of naming this one `newFaceSize` ? Would read more explicitely in frontend that it's a face size object, rather than a "size" ?

```lua
ok, ftsize = pcall(Freetype.newFaceSize, _v, size, faceindex)
ok, ftsize = pcall(Freetype.newSizedFace, _v, size, faceindex) -- too strange?
```

poire-z (approved):
Given your numerous ongoing PRs, and that they go in pairs with a frontend one, you'll have to tell us when you want each of them merged :)

ffi/freetype.lua
benoit-pierre: Done.

benoit-pierre:
> > wouldn't it make more sense for checkGlyph(char) to return a boolean? (Instead of 1 if the character is present in the font, and 0 otherwise).
> 
> Sure !

Done, but now I'm wondering if `hasGlyph` would not be a better name… :P

poire-z (approved):

poire-z:
> Done, but now I'm wondering if hasGlyph would not be a better name… :P

Would work for me :)

benoit-pierre:
> > Done, but now I'm wondering if hasGlyph would not be a better name… :P
> 
> Would work for me :)

Done!

poire-z:
> Given your numerous ongoing PRs, and that they go in pairs with a frontend one, you'll have to tell us when you want each of them merged :)

Now that v2023.06 is out, we're ready to be copiloted by you.

benoit-pierre:
> > Given your numerous ongoing PRs, and that they go in pairs with a frontend one, you'll have to tell us when you want each of them merged :)
> 
> Now that v2023.06 is out, we're ready to be copiloted by you.

Start with the easy stuff, and ramp-up the difficulty?

- [ ] [fix listing tests with `busted --list`](https://github.com/koreader/koreader/pull/10616)

- [ ] improve ccache support when cross-compiling:
    * [ ] [base](https://github.com/koreader/koreader-base/pull/1623)
    * [ ] *[TODO]* koreader: bump base

- [ ] fix FFI cdecls:
    * [ ] [base](https://github.com/koreader/koreader-base/pull/1619)
    * [ ] *[TODO]* koreader: bump base

- [ ] decrease fonts memory use:
    * [ ] [base](https://github.com/koreader/koreader-base/pull/1625)
    * [ ] [koreader](https://github.com/koreader/koreader/pull/10618)

- [ ] fix uninitialized memory access in libk2pdfopt:
    * [ ] [libk2pdfopt](https://github.com/koreader/libk2pdfopt/pull/41)
    * [ ] *[TODO]* base: bump libk2pdfopt
    * [ ] *[TODO]* koreader: bump base

- [ ] fix and enable crengine teardown:
    * [ ] update to use NDK 23c
        - [ ] [luajit-launcher](https://github.com/koreader/android-luajit-launcher/pull/419)
        - [ ] [base](https://github.com/koreader/koreader-base/pull/1620)
        - [ ] [koreader](https://github.com/koreader/koreader/pull/10610)
    * [ ] [update minimum C++ standard requirement](https://github.com/koreader/koreader-base/pull/1624)
    * [x] [crengine](https://github.com/koreader/crengine/pull/520)
    * [ ] [base](https://github.com/koreader/koreader-base/pull/1617)
    * [ ] *[TODO]* koreader: bump base

poire-z:
I guess we can keep `decrease fonts memory use` for the next round of merge (in case the frontend PRs tests for the ones bumped require a fix).

poire-z:
So, can I merge now the 3 base PRs ?
image
In some preferred logical order ? All the 3 now ? 
Or do you want to do only the NDK23 up to front before the others?

benoit-pierre:
Only the NDK update for now.

PR #1626, opened 2023-06-27
ci: drop `sony-prstux` build

benoit-pierre:
As per [discussion](https://github.com/koreader/koreader-base/pull/1624#issuecomment-1608313118) in #1624.

### DIFF
--- a/.ci/build_script.sh
+++ b/.ci/build_script.sh
@@ -40,12 +40,6 @@ elif [ "$TARGET" = "pocketbook" ]; then
         -v "${HOME}/.ccache:${DOCKER_HOME}/.ccache" \
         -v "$(pwd):${DOCKER_HOME}/base" "${DOCKER_IMG}" \
         /bin/bash -c "source /home/ko/.bashrc && cd /home/ko/base && sudo chown -R ko:ko . && make VERBOSE=1 TARGET=pocketbook all"
-elif [ "$TARGET" = "sony-prstux" ]; then
-    sudo chmod -R 777 "${HOME}/.ccache"
-    docker run -t \
-        -v "${HOME}/.ccache:${DOCKER_HOME}/.ccache" \
-        -v "$(pwd):${DOCKER_HOME}/base" "${DOCKER_IMG}" \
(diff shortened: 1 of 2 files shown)


Frenzie (approved):
Do we keep the remainder for now?

benoit-pierre:
> Do we keep the remainder for now?

Meaning the kodev / Makefile related code and platform support in the frontend?

Frenzie:
That would be implied by removing it from here of course, but I was just talking about this repo in the context of this PR at the moment.

If we don't want to remove it I'd rather provide an updated Docker image with a newer GCC since that's very little effort. Not having it in CI virtually guarantees it'll break. (I.e., most of us only test some combination of desktop/Kobo/Kindle actively, though not being a problem case like Android or Mac does mean it might remain fine for many years.)

Frenzie:
Concretely I mean some things in the Makefiles as well as input/input-sony-prstux.h, some references in the framebuffer implementation and maybe something else I didn't consider otoh.

NiLuJe:
It might serve some documentation purposes for a barebones linuxfb-ish port, so I'd keep it around (e.g., it might be useful reference for a non-SDL pmOS port or something).

Frenzie:
Then I don't want to merge this PR. I'll make sure the GCC is more recent sometime in July 2023 as planned or sooner if necessary.

NiLuJe:
Err, I meant the *code*, I don't see what the CI has to do with that?

Frenzie:
The effort involved in keeping it in (theoretically) working condition and letting it rot in place is more or less identical and in practice letting it rot is quite likely to take more effort imho.

To use a newer GCC, change `DOCKER_IMG: koreader/kobase-18.04:0.3.0` to `kobase-20.04:0.3.0` and done.

Frenzie:
That one (that should've been deleted too in this PR but that side :-) ):
https://github.com/koreader/koreader-base/blob/d3744561601e5c97219899b02a608cd94c2b02e0/.circleci/config.yml#L164-L168

Frenzie:
kobase-22.04:0.3.0 exists too

NiLuJe:
Fair enough ;).

We might still want to drop it from the OTA servers, though?

Frenzie:
I did that nearly a year ago. ;-)

https://gitlab.com/koreader/nightly-builds/-/commit/cf225de75107a9a21c78e21461e6726959790e89

PR #1610, opened 2023-05-12
SDL: distinguish between window size and screen size

hrdl-github:
Pointer input and window resize events use window coordinates. Fixes
koreader/koreader#10420.

I've tested this using SDL's wayland and x11 backends on both sway and plasma-wayland using no scaling, 150% and 200% output scaling. Scaling produces bad results on sway with the x11 backend, but it has done so before. I don't have access to macOS or a Windows system for testing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1610)

PR #1627, opened 2023-06-27
SDL: Unbreak window state save & restore

NiLuJe:
Fix https://github.com/koreader/koreader/issues/10502, regression since #1610

Well, only the size is a regression (plus, it was sortof an ffi/framebuffer API break, even if SDL is quirky in that fb.h / fb.w might not match fb.bb's dimensions because HiDPI shenanigans).

The position one is a bit of a mystery, we get an initial move event no matter what, for some reason, and without the explicit set, it's at the top left despite passing the right stuff to CreateWindow o_O.

Also dropped a duplicate openGameController call, as we handle it via an SDL_CONTROLLERDEVICEADDED event already.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1627)

### DIFF
--- a/ffi-cdecl/SDL2_0_decl.c
+++ b/ffi-cdecl/SDL2_0_decl.c
@@ -142,6 +142,8 @@ cdecl_func(SDL_CreateTexture)
 cdecl_func(SDL_UpdateTexture)
 cdecl_func(SDL_DestroyTexture)
 cdecl_func(SDL_SetWindowTitle)
+cdecl_func(SDL_GetWindowPosition)
+cdecl_func(SDL_SetWindowPosition)
 
 cdecl_type(SDL_blit)
 cdecl_func(SDL_CreateRGBSurface)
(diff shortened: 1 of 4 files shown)


Frenzie:
> Also dropped a duplicate openGameController call, as we handle it via an SDL_CONTROLLERDEVICEADDED event already.

Are you sure? If my memory doesn't betray me (but it might, I wrote that like five years ago?) the event only occurs when you plug in a controller, while normally it's already plugged in.

NiLuJe:
> Are you sure?

Works for me on Linux w/ a Sony DS4v2, FWIW.

Frenzie:
Indeed, I can also confirm that it works on my Hori on SDL 2.26.5.

Frenzie (approved):

Frenzie:
Did you want to rebase it into two commits or just squash the whole thing together?

NiLuJe:
Squash, definitely ;).

PR #1629, opened 2023-07-10
xtext: fix possible linking issue

benoit-pierre:
On some environment, putting the `-l` arguments after `-o $@ xtext.cpp` results in a library that does *not* depend on freetype, harfbuzz, etc…

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1629)


Frenzie:
Apologies, I already committed that straight to master (something we normally avoid).

PR #1631, opened 2023-07-11
Third-party bumps

NiLuJe:
Fingers crossed we can finally update harfbuzz ;o).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1631)


Frenzie (approved):

Frenzie:
Heh, I figured we'd just wait for 8.0.1 in a week or two.

Frenzie:
It sounds like there's an issue with the libjpegturbo thing?

https://app.circleci.com/pipelines/github/koreader/koreader-base/1670/workflows/bb66a69d-6bcc-41ca-9cfa-2ab30bdc6baf/jobs/14997?invite=true#step-105-1073

ffi/jpeg.lua
```diff
@@ -95,9 +95,8 @@ function Jpeg.encodeToFile(filename, source_ptr, w, stride, h, quality, color_ty
     end
 
     turbojpeg.tj3Destroy(handle)
-    if jpeg_image[0] ~= nil then
-        turbojpeg.tj3Free(jpeg_image[0])
-    end
+    -- Handles NULL pointers sanely
+    turbojpeg.tj3Free(jpeg_image[0])
```
Frenzie: This change isn't dangerous?

NiLuJe: Nope, the comment was supposed to help avoid similar questions in the future ^^.

```
 * @param buffer address of the buffer to free.  If the address is NULL, then
 * this function has no effect.
```

c.f., https://rawcdn.githack.com/libjpeg-turbo/libjpeg-turbo/main/doc/html/group___turbo_j_p_e_g.html#gaddb84fb6c81769e9faa0f5a63b296606

PR #1633, opened 2023-07-12
Unbreak LuaJIT

NiLuJe:
One of the few commits in the last bump appears to be causing us trouble...

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1633)


benoit-pierre:
I can confirm it works for me too after reverting that commit.

benoit-pierre:
Cleanup on aisle pocketbook… Another curl failure when fetching certificates.

PR #1637, opened 2023-07-13
RPATH shenanigans

NiLuJe:
While making sure we stick an $ORIGIN rpath in webp, I noticed that curl's made no sense, and that zstd's made no sense (and doesn't matter anyway).

Currently running a few builds of my own to double-check the sanity of all that...

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1637)


NiLuJe:
Seems to have taken:

```
  72   │ Dynamic section at offset 0x2ef0 contains 30 entries:
  73   │   Tag        Type                         Name/Value
  74   │  0x00000001 (NEEDED)                     Shared library: [libwebp.so.7]
  75   │  0x00000001 (NEEDED)                     Shared library: [libsharpyuv.so.0]
  76   │  0x00000001 (NEEDED)                     Shared library: [libm.so.6]
  77   │  0x00000001 (NEEDED)                     Shared library: [libpthread.so.0]
  78   │  0x00000001 (NEEDED)                     Shared library: [libc.so.6]
  79   │  0x0000000e (SONAME)                     Library soname: [libwebpdemux.so.2]
  80   │  0x0000000f (RPATH)                      Library rpath: [/var/tmp/niluje/koreader/base/thirdparty/libwebp/build/arm-kobo-linux-gnueabihf/libwebp-prefix/src/libwebp-build/lib:$ORIGIN]
```

PR #1638, opened 2023-07-13
RPATH & escaping cleanup

NiLuJe:
Drop build-time rpaths from aaaaalmost everything (CMake is being CMake, so sdcv gets both), and fix a few missing $ORIGIN rpaths in the process.

Depends on https://github.com/koreader/libk2pdfopt/pull/42

----

Note that, technically, a few things that depend on the STL *ought* to be using an $ORIGIN rpath, the current list would be:
djvulibre, lunasvg, libzmq... and rapidjson >_<".

The last one is via luarocks, so I can't even imagine the amount of hoop-jumping that might require, so, let's table that for now ;D.

EDIT: Weeeell, handled those, too ;o).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1638)

PR #1639, opened 2023-07-15
ci: don't install SDL2 from brew

benoit-pierre:
Since the macOS build already come with one.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1639)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,7 +27,7 @@ jobs:
 
       - name: Homebrew install dependencies
         # Compared to the README, adds ccache for faster compilation times and removes sh5sum to prevent some conflict with coreutils, and gnu-getopt because we're not using kodev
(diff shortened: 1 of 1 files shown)


benoit-pierre:
@NiLuJe: OK, do you want to add it to your #1638 pile?

NiLuJe:
We're going to need a PR in front to handle the bump and updating the bundle script, so I have no particular attachment at how we go at this.

(I can push to this very branch if necessary anyway).

benoit-pierre:
Note that the build script does not try to bundle brew' SDL, so for that particular issue, there's no need to patch it.

pazos:
Not sure if it is relevant here but the emulator relies on brew's sdl2

benoit-pierre:
How?

pazos:
> How?

:p

Just by checking the history tree I realized the entire workflow exists just to generate macos artifacts.

So, yep, no relevant and indeed not needed since https://github.com/koreader/koreader-base/pull/1143

benoit-pierre:
Ping!

Frenzie:
Force pushing already pings. ;-)

PR #1643, opened 2023-07-22
Build: Don't create the legacy history dir

NiLuJe:
It's no longer in use, and gets deleted at runtime if empty, but its presence here means it makes it into the list of mandatory files in the `package.index` file used by OTAs, which means the *first* OTA after a fresh install will fail.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1643)


NiLuJe:
(Regression since https://github.com/koreader/koreader/pull/10162)

Frenzie (approved):

PR #1645, opened 2023-08-01
Kobo: Allow poking at the EPDC PM directly

NiLuJe:
On devices where it's available (~Mk. 8+), Nickel does that on *touch* events every ~1.5s. For some mysterious reason (it may be a red herring), this seems to help with the mysterious hangs we get on, strangely enough, all the NXP boards where this feature is available...

It does *not* help with the timeouts in the wait__for_update_complete ioctl these boards *also* suffer from, despite Nickel also doing something weird with those, which I tried to replicate for shit'n giggles. Unsurprisingly so, as these timeouts have been reproduced in Nickel (and, in fact, I did just that myself in the course of testing, c.f., 729fa5a9a91ac84da560ce8f58709f3b0165bf34).

Speaking of testing, this was all done thanks to @ezuk, who donated a Clara 2E to try to get to the bottom of this mess. Thanks again ;).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1645)


NiLuJe (commented):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1645)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

ffi/framebuffer_mxcfb.lua
```diff
@@ -271,6 +306,20 @@ local function mxc_update(fb, ioc_cmd, ioc_data, is_flashing, waveform_mode, x,
...
+    --       but to *touch* input: on any registered *touch* input (physical buttons are exempt, for some reason), Nickel will forcibly awaken the EPDC,
+    --       unless that was already done less than roughly 1.5s ago.
+    --       Think of something like some properly implemented & tuned kernels with the interactive cpufreq governor that will boost on touch input, for instance.
+    --       In practice, that still trips *very* close to the ioctl (but may not always come *before* a [wait -> refresh -> wait] sandwich;
+    --       e.g., I mostly see it before the refresh on NXP & sunxi, but I mostly see it in front of everything on MTK).
+    --       TL;DR: Because, on devices flagged !hasReliableMxcWaitFor, we were seeing deadlock issues with refreshes that aren't necessarily tied to input,
+    --       (and for simplicity's sake), we simply unconditionally do this here as early possible.
+    -- NOTE: This might be a gigantic red herring, and simply a case of the very few extra cpu cycles involved throwing off the race...
```
Frenzie, line 318: lol

NiLuJe:
I have yet to validate whether this actually helps with the missed refreshes on wakeup from standby on sunxi, i.e., https://github.com/koreader/koreader/issues/10743#issuecomment-1649432244

Mainly because these are fairly rare, which means I actually need time to read on that device ;p.

ffi/framebuffer_mxcfb.lua
NiLuJe: Wheee \o/

NiLuJe: On the upside, I genuinely haven't managed to hang the Clara 2E with jump markers since ;).

NiLuJe: The downside being I *never* actually ùanaged to hang it with anything else (specifically, Notifications).

My only other hang was *extremely* overenginereed, as it involved [ftrace](https://github.com/NiLuJe/FBInk/blob/master/utils/finger_trace.c), which will positively spam refresh ioctls ;).

PR #1650, opened 2023-08-30
android: forward play and pause to the system

pazos:
See https://www.mobileread.com/forums/showpost.php?p=4351935&postcount=459

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1650)


pazos:
> Is this standalone ? No issue if I bump it with my crengine bump?

Yes. Nope, these are already defined elsewhere: https://github.com/koreader/android-luajit-launcher/blob/master/assets/android.lua#L349

Please go ahead

PR #1659, opened 2023-09-24
ci: do not waste resources on Github Actions

benoit-pierre:
Avoid concurrent workflow runs on the same ref.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1659)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,9 @@
 name: macos
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 on: [push, pull_request]
 
 jobs:


Frenzie:
In which scenario do they occur? Or put another way, why would the same ref be running multiple in the first place? Or is this talking about doing something like canceling older builds if you push another commit before they finished.

benoit-pierre:
Interesting error:
2023-09-24-180207_890x787_scrot

Thankfully, their AI experiment is there to help!

2023-09-24-180242_262x566_scrot

:P 

Anyway, just the usual cURL failure when generating certificates.

benoit-pierre:
> In which scenario do they occur? Or put another way, why would the same ref be running multiple in the first place? Or is this talking about doing something like canceling older builds if you push another commit before they finished.

Yes, exactly: push another commit on the same PR/branch/tag, and the older workflow run is canceled.

Frenzie:
I'm not necessarily sure if that's always desirable, but presumably it generally is.

PR #1660, opened 2023-09-24
tests: tag tests relying on a working internet access

benoit-pierre:
So they can easily be filtered-out when offline.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1660)


Frenzie:
Incidentally, I notice it does something that's no longer necessary, see https://github.com/koreader/koreader/pull/6307

PR #1662, opened 2023-10-05
PocketBook: Rewrite translateEvent

NiLuJe:
* Rewrite it in a style closer to what we do on Android
* Simplify said Android implementation
* Add debug logging like in Input, to dump the full stream of PocketBook events being processed (or not).
* This still doesn't deal with MTSYNC properly though, see the linked issue ;).

Re: https://github.com/koreader/koreader/issues/10950

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1662)


NiLuJe:
(To be merged post-release)

poire-z:
(Might depends on when the release happen? - my custom handmade stuff will have lots of translation, so the next release may be by end of month.)

Frenzie:
I don't have anything specific planned other than trying to prep most of some release notes this weekend if I find the time.

poire-z:
Last eleven times you said that, I had to wait with my PR in my pants for 2 or 3 weeks before it happened :)
So, unless you're sure you can get a release by sunday morning, we could wait for end of month - not much happened since v2023.08 and I don't think we're in a hurry with some fixes.

ffi/input_pocketbook.lua
```diff
@@ -114,15 +254,17 @@ local function translateEvent(t, par1, par2)
...
     return 0
 end
 
 local raw_keymap -- [EV_KEY] = "Action" supplied by frontend
 local poll_fds, poll_fds_count = nil, 0
 function input:open()
+    -- Don't do this at home, kids, these belong to front.
+    DEBUG = require("dbg")
+    logger = require("logger")
+
```
rjd22, line 267: Are these left on purpose?

Frenzie:
> So, unless you're sure you can get a release by sunday morning

Definitely not, depends on all kinds of things like weather, what other people want to do, what I want to do, etc. ;)

ffi/input_pocketbook.lua
NiLuJe: Yeah, I want to guard the logging calls behind verbose debug like Input, so the hackery is necessary so as not to upset the testsuite (I'm assuming this is even covered by the testsuite); but, in any case, not an example to follow; but I didn't want to complicate frontend code just to get a sane logger in here ;).

NiLuJe (commented):
> Reviewed 2 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1662)* status: all files reviewed, 1 unresolved discussion (waiting on @rjd22)

PR #1667, opened 2023-10-11
ffi/sdl: fix leaks on exit

benoit-pierre:
Ensure the renderer is destroyed (which will also destroy its associated texture and window).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1667)


ffi/SDL2_0.lua
```diff
@@ -190,6 +190,7 @@ function S.setWindowFullscreen(full_screen)
 end
 
 function S.destroyTexture(texture)
+    ffi.gc(texture, nil)
```
NiLuJe, line 193: Why is that one necessary? Is there an implicit finalizer on this somehow?

benoit-pierre: Nope, good catch: leftover from before I added a finalizer for the renderer (which takes care of the texture too).

NiLuJe (approved):

PR #1675, opened 2023-10-24
crengine: don't compile unused code

benoit-pierre:
Depends on https://github.com/koreader/crengine/pull/542.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1675)


Frenzie (approved):
Makes sense but I'm not entirely sure how that corresponds to the discussion over there atm. :-)

benoit-pierre:
> Anything else to do here?

Nope, should be ready.

NiLuJe (approved):

PR #1678, opened 2023-10-25
fix a number of `Makefile.third` rules

benoit-pierre:
- using `$@` for some of those multiple-targets rules is a mistake (it's not going be the library name when triggered by a dependency on the include directory)
- add missing targets (include directory, library names)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1678)


Frenzie (approved):

NiLuJe (approved):

Frenzie:
Closing in favor of the larger PR then.

PR #1680, opened 2023-10-26
Add initial RGB painting support to C blitbuffer

smasher816:
Add new BB_fill_rect_color and BB_blend_rect_color functions taking a ColorRGB32.
Use a new is_rgb flag in Color structs for lua wrapper to decide which BB_ function to call.
If a ColorRGB32 is passed to fillRect/lightenRect then the new path will be taken. Otherwise original Color8A/uint8_t code is called.

No changes should occur for existing code paths. Callers must be patched to provide RGB colors if desired.
Lua blitbuffer implementation has not been changed.

### DIFF
--- a/blitbuffer.c
+++ b/blitbuffer.c
@@ -76,6 +76,7 @@ static const char*
 // Helpers to pack pixels manually, without going through the Color structs.
 #define Y8_To_Y8A(v) (0xFFu << 8U | v)
 #define RGB_To_RGB32(r, g, b) ((uint32_t) (0xFFu << 24U) | (uint32_t) (b << 16U) | (uint32_t) (g << 8U) | r)
+#define RGBA_To_RGB32(r, g, b, a) ((uint32_t) (a << 24U) | (uint32_t) (b << 16U) | (uint32_t) (g << 8U) | r)
 #define Y8_To_RGB32(v) ((uint32_t) (0xFFu << 24U) | (uint32_t) (v << 16U) | (uint32_t) (v << 8U) | v)
 
 // __auto_type was introduced in GCC 4.9 (and Clang ~3.8)...
@@ -416,6 +417,65 @@ void BB_fill_rect(BlitBuffer * restrict bb, unsigned int x, unsigned int y, unsi
     }
 }
 
+void BB_fill_rect_color(BlitBuffer * restrict bb, unsigned int x, unsigned int y, unsigned int w, unsigned int h, ColorRGB32 * restrict color) {
+    const int rotation = GET_BB_ROTATION(bb);
+    unsigned int rx, ry, rw, rh;
+    // Compute rotated rectangle coordinates & size
+    switch (rotation) {
+        case 0:
+                rx = x;
+                ry = y;
+                rw = w;
+                rh = h;
+                break;
+        case 1:
+                rx = bb->w - (y + h);
+                ry = x;
+                rw = h;
+                rh = w;
+                break;
+        case 2:
+                rx = bb->w - (x + w);
+                ry = bb->h - (y + h);
+                rw = w;
+                rh = h;
+                break;
+        case 3:
+                rx = y;
+                ry = bb->h - (x + w);
+                rw = h;
+                rh = w;
+                break;
+    }
+
+    // Handle any target pitch properly
+    const int bb_type = GET_BB_TYPE(bb);
+    switch (bb_type) {
+        case TYPE_BBRGB32:
+            // And here either, as we want to preserve the alpha byte
+            if (rx == 0 && rw == bb->w) {
+                // Single step for contiguous scanlines
+                const uint32_t src = (uint32_t) RGBA_To_RGB32(color->r, color->g, color->b, color->alpha);
+                uint32_t * restrict p = (uint32_t *) (bb->data + bb->stride*ry);
+                size_t px_count = bb->pixel_stride*rh;
+                while (px_count--) {
+                    *p++ = src;
+                }
+            } else {
+                // Scanline per scanline
+                const uint32_t src = (uint32_t) RGBA_To_RGB32(color->r, color->g, color->b, color->alpha);
+                for (unsigned int j = ry; j < ry+rh; j++) {
+                    uint32_t * restrict p = (uint32_t *) (bb->data + bb->stride*j) + rx;
+                    size_t px_count = rw;
+                    while (px_count--) {
+                        *p++ = src;
+                    }
+                }
+            }
+            break;
+    }
+}
+
 void BB_blend_rect(BlitBuffer * restrict bb, unsigned int x, unsigned int y, unsigned int w, unsigned int h, Color8A * restrict color) {
     const int bb_type = GET_BB_TYPE(bb);
     const int bb_rotation = GET_BB_ROTATION(bb);
@@ -471,6 +531,25 @@ void BB_blend_rect(BlitBuffer * restrict bb, unsigned int x, unsigned int y, uns
                     dstptr->r = (uint8_t) DIV_255(dstptr->r * ainv + color->a * alpha);
                     dstptr->g = (uint8_t) DIV_255(dstptr->g * ainv + color->a * alpha);
                     dstptr->b = (uint8_t) DIV_255(dstptr->b * ainv + color->a * alpha);
+                }
+            }
+            break;
+    }
+}
+
+void BB_blend_rect_color(BlitBuffer * restrict bb, unsigned int x, unsigned int y, unsigned int w, unsigned int h, ColorRGB32 * restrict color) {
+    const int bb_type = GET_BB_TYPE(bb);
+    const int bb_rotation = GET_BB_ROTATION(bb);
+    switch (bb_type) {
+        case TYPE_BBRGB32:
+            for (unsigned int j = y; j < y + h; j++) {
+                for (unsigned int i = x; i < x + w; i++) {
+                    ColorRGB32 * restrict dstptr;
+                    BB_GET_PIXEL(bb, bb_rotation, ColorRGB32, i, j, &dstptr);
+                    dstptr->r = (uint8_t) MAX(color->r - (255 - dstptr->r), 0);
+                    dstptr->g = (uint8_t) MAX(color->g - (255 - dstptr->g), 0);
+                    dstptr->b = (uint8_t) MAX(color->b - (255 - dstptr->b), 0);
+                    dstptr->alpha = 0xFF;
                 }
             }
             break;
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -118,7 +118,9 @@ typedef struct BlitBufferRGB32 {
 
 void BB_fill(BlitBuffer * restrict bb, uint8_t v);
 void BB_fill_rect(BlitBuffer * restrict bb, unsigned int x, unsigned int y, unsigned int w, unsigned int h, uint8_t v);
+void BB_fill_rect_color(BlitBuffer * restrict bb, unsigned int x, unsigned int y, unsigned int w, unsigned int h, ColorRGB32 * restrict color);
 void BB_blend_rect(BlitBuffer * restrict bb, unsigned int x, unsigned int y, unsigned int w, unsigned int h, Color8A * restrict color);
+void BB_blend_rect_color(BlitBuffer * restrict bb, unsigned int x, unsigned int y, unsigned int w, unsigned int h, ColorRGB32 * restrict color);
 void BB_invert_rect(BlitBuffer * restrict bb, unsigned int x, unsigned int y, unsigned int w, unsigned int h);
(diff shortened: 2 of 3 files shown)


smasher816:
Initial inspiration. https://github.com/koreader/koreader/issues/9024#issuecomment-1778999195

blitbuffer.c
```diff
@@ -416,6 +417,65 @@ void BB_fill_rect(BlitBuffer * restrict bb, unsigned int x, unsigned int y, unsi
...
+    }
+
+    // Handle any target pitch properly
+    const int bb_type = GET_BB_TYPE(bb);
+    switch (bb_type) {
+        case TYPE_BBRGB32:
+            // And here either, as we want to preserve the alpha byte
+            if (rx == 0 && rw == bb->w) {
+                // Single step for contiguous scanlines
+                const uint32_t src = (uint32_t) RGBA_To_RGB32(color->r, color->g, color->b, color->alpha);
```
NiLuJe, line 458: Unnecessary, ColorRGB32 is already an uint32_t in RGBA layout.

blitbuffer.c
```diff
@@ -416,6 +417,65 @@ void BB_fill_rect(BlitBuffer * restrict bb, unsigned int x, unsigned int y, unsi
...
+                for (unsigned int j = ry; j < ry+rh; j++) {
+                    uint32_t * restrict p = (uint32_t *) (bb->data + bb->stride*j) + rx;
+                    size_t px_count = rw;
+                    while (px_count--) {
+                        *p++ = src;
+                    }
+                }
+            }
+            break;
+    }
```
NiLuJe, line 476: Needs RGB565 support

blitbuffer.c
```diff
@@ -477,6 +537,25 @@ void BB_blend_rect(BlitBuffer * restrict bb, unsigned int x, unsigned int y, uns
...
+                    ColorRGB32 * restrict dstptr;
+                    BB_GET_PIXEL(bb, bb_rotation, ColorRGB32, i, j, &dstptr);
+                    dstptr->r = (uint8_t) MAX(color->r - (255 - dstptr->r), 0);
+                    dstptr->g = (uint8_t) MAX(color->g - (255 - dstptr->g), 0);
+                    dstptr->b = (uint8_t) MAX(color->b - (255 - dstptr->b), 0);
+                    dstptr->alpha = 0xFF;
+                }
+            }
+            break;
+    }
```
NiLuJe, line 556: Needs RGB565 support

ffi/blitbuffer.lua
```diff
@@ -170,13 +172,13 @@ local P_BlitBuffer = ffi.typeof("BlitBuffer*")
...
-local ColorRGB16_mt = {__index={}}
-local ColorRGB24_mt = {__index={}}
-local ColorRGB32_mt = {__index={}}
+local Color4L_mt = {__index={ is_rgb=false }}
+local Color4U_mt = {__index={ is_rgb=false }}
+local Color8_mt = {__index={ is_rgb=false }}
+local Color8A_mt = {__index={ is_rgb=false }}
+local ColorRGB16_mt = {__index={ is_rgb=false }}
+local ColorRGB24_mt = {__index={ is_rgb=false }}
+local ColorRGB32_mt = {__index={ is_rgb=true }}
```
NiLuJe, line 181: I'm... not necessarily a fan of this approach.

I get the idea of avoiding duplication like in @Galunid's PoC, but.... eeeh :/.

Possibly something based of `ffi.istype` instead?

smasher816: Initially I wanted to do something like `if color:type() == "ColorRGB32"` or similar but I wasn't sure how that interacts with ffi and I'm not super familiar with lua apis.
Would the expected syntax be `if ffi.istype(ColorRGB32, color)` ?

blitbuffer.c
smasher816: Compiler isn't super happy about using the value as-is.

```
blitbuffer.c:460:28: warning: cast to smaller integer type 'uint32_t' (aka 'unsigned int') from 'ColorRGB32 *' (aka 'struct ColorRGB32 *') [-Wpointer-to-int-cast]
                    *p++ = (uint32_t) color;
```

smasher816: I think I want to do this `*p++ = *((uint32_t*)color);` (treat memory pointer as if it looking at uint32_t, then dereference that value)

smasher816:
Unrelated, but all of a sudden after pushing to forks I am unable to recompile. Any pointers would be helpful since I haven't touched 3rd-party deps.
I would like to tests my fixes on-device before updating the MR.

```
dewarp2.c:144:8: error: call to undeclared function 'dewarpBuildPageModel_ex'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
return(dewarpBuildPageModel_ex(dew,debugfile,2));
       ^
dewarp2.c:144:8: note: did you mean 'dewarpBuildPageModel'?
dewarp2.c:141:1: note: 'dewarpBuildPageModel' declared here
dewarpBuildPageModel(L_DEWARP    *dew,
^
dewarp2.c:230:9: error: call to undeclared function 'dewarpFindVertDisparity_ex'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
    if (dewarpFindVertDisparity_ex(dew, ptaa2, 0, fit_order) != 0) {
        ^
dewarp2.c:230:9: note: did you mean 'dewarpFindVertDisparity'?
./allheaders.h:654:25: note: 'dewarpFindVertDisparity' declared here
LEPT_DLL extern l_int32 dewarpFindVertDisparity ( L_DEWARP *dew, PTAA *ptaa, l_int32 rotflag );
                        ^
dewarp2.c:301:8: error: call to undeclared function 'dewarpFindVertDisparity_ex'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
return(dewarpFindVertDisparity_ex(dew,ptaa,rotflag,2));
       ^
dewarp2.c:1411:8: error: call to undeclared function 'dewarpBuildLineModel_ex'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
return(dewarpBuildLineModel_ex(dew,opensize,debugfile,2));
       ^
4 errors generated.
make[4]: *** [Makefile:823: dewarp2.lo] Error 1
make[4]: *** Waiting for unfinished jobs....
make[4]: Leaving directory '/home/arch/git/koreader/base/thirdparty/leptonica/build/aarch64-unknown-linux-android21/leptonica-prefix/src/leptonica/src'
make[3]: *** [Makefile:470: install-recursive] Error 1
make[3]: Leaving directory '/home/arch/git/koreader/base/thirdparty/leptonica/build/aarch64-unknown-linux-android21/leptonica-prefix/src/leptonica'
make[2]: *** [Makefile:95: liblept.so] Error 2
make[2]: Leaving directory '/home/arch/git/koreader/base/thirdparty/libk2pdfopt/build/aarch64-unknown-linux-android21/libk2pdfopt-prefix/src/libk2pdfopt'
make[1]: *** [Makefile.third:372: build/aarch64-unknown-linux-android21/libs/libk2pdfopt.so.2] Error 1
```

NiLuJe:
> Unrelated

Make sure your submodules are up-to-date and point to the right place (both base's and base itself), then try a `make dist-clean`?

ffi/blitbuffer.lua
NiLuJe: Or `P_ColorRGB32` if it's a pointer, yeah.

blitbuffer.c
NiLuJe: Yup, I'm fairly sure there are other instances of something like this (possibly sandwiched between GCC pragmas, because compilers don't like it).

NiLuJe: Ohkay, maybe not here, but in [FBInk](https://github.com/NiLuJe/FBInk/blob/af1a15b839462c3b19c611d4e3c6007dfc398da1/fbink.c#L235-L238) ;p.

smasher816:
Fixed the above. A fresh clone also had issues so I figured out it was something with the machine and not the koreader repo.
After looking into `env` I found the following unexpectedly set.
```
profile.d/android-ndk.sh
2:export ANDROID_NDK=/opt/android-ndk
4:export ANDROID_NDK_HOME=/opt/android-ndk
6:export ANDROID_NDK_ROOT=/opt/android-ndk
```

It looks like installing a system package added those variables by default.
```
$ pacman -Qo /etc/profile.d/android-ndk.sh
/etc/profile.d/android-ndk.sh is owned by android-ndk r26.b-1
```

Unsetting the above fixed everything and now I can compile again. I'll work on resolving your comments.

blitbuffer.c
smasher816: For 565 (RGB16) I think we can do ColorRGB32_To_Color16, and then use a uint16_t pointer loop.

For 888 (RGB24) there is no uint24_t type (and this is probably not the time/place to add one).
Given that we can't do `memset(p, v, (rw * 3U));`(setting all three bytes to the same value), I think the best way to approch this is setting the values independently then incrementing the uint8_t pointer by 3 bytes.
```
p[0] = src.r;
p[1] = src.g;
p[2] = src.b;
```
 
 Do the above approaches seem reasonable? I don't have a real device with the above bit layouts to test this on.

smasher816: Or maybe something like memcpy would work since it accepts variable lengths along with any pointer type.
```
memcpy(p, &src, 3);
```

smasher816:
New revision should have support for other screentypes and for real color blending (stacking multiple highlights makes the color get stronger and stronger).

smasher816:
Newest revision supports color pdf highlights. If new `bb_color` arg is not passed then the existing defaults are used.

blitbuffer.c
```diff
@@ -477,6 +560,56 @@ void BB_blend_rect(BlitBuffer * restrict bb, unsigned int x, unsigned int y, uns
...
+    const int bb_rotation = GET_BB_ROTATION(bb);
+    const uint8_t alpha = color->alpha;
+    const uint8_t ainv = alpha ^ 0xFF;
+    switch (bb_type) {
+        case TYPE_BBRGB16:
+            for (unsigned int j = y; j < y + h; j++) {
+                for (unsigned int i = x; i < x + w; i++) {
+                    ColorRGB16 * restrict dstptr;
+                    BB_GET_PIXEL(bb, bb_rotation, ColorRGB16, i, j, &dstptr);
+                    // Only edit light colors and leave blacks alone. This avoids whitewashing text.
```
NiLuJe, line 574: Code disagrees with the comment for this specific case ;).

smasher816: Good catch. I'll fix this.

NiLuJe (commented):
I've pushed some minor simplifications, but have *NOT* tested them, I've just checked that it built (and was generating similar machine code before actual logic deviations were implemented ;p).

> Reviewed 1 of 3 files at r2, 3 of 3 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1680#-:-NiBHGY599ajD_njFu1H:bbqbdcu)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @smasher816)

NiLuJe (commented):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1680)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @smasher816)

poire-z:
Is this ok?

NiLuJe:
Needs testing ;).

NiLuJe:
@poire-z: Because I'm super lazy, If you have a random test file with colored text on hand, that could come in handy. Thanks ;).

poire-z:
> Because I'm super lazy, 

I know :)
Here's 2 of my test.html files where I tested various text & bg colors: [test-color-stuff.zip](https://github.com/koreader/koreader-base/files/14000680/test-color-stuff.zip)
Any Wikipedia saved as EPUB will probably have various kinds of colored images if you need more.

NiLuJe (commented):
> Reviewed 4 of 4 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1680)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @smasher816)

NiLuJe:
Hey, look, I've finally gotten around to finish this before the PR's first birthday ;p.

(I basically just want to test a few color changes tomorrow, but other than that, it's ready).

NiLuJe (commented):
> Reviewed 1 of 1 files at r10, 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1680)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @smasher816)

NiLuJe:
Commit history will be archived in my [color-highlights branch](https://github.com/NiLuJe/koreader-base/tree/color-highlights)

PR #1682, opened 2023-10-27
android: update SDK command line tools and tweak NDK install

benoit-pierre:
To allow updating the OpenJDK Java version to 17.

Cf. https://github.com/koreader/android-luajit-launcher/pull/445. Next step will be to try to update the Android Docker image to 22.04 again, and if that does not work, it looks like `openjdk-17-jdk-headless` is available on Bionic, so fallback to that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1682)

### DIFF
--- a/toolchain/Makefile
+++ b/toolchain/Makefile
@@ -1,48 +1,63 @@
-SHELL:=/bin/bash
+.DELETE_ON_ERROR:
+.ONESHELL:
+
+SHELL := /bin/bash
+.SHELLFLAGS := -xec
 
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
For after the release. I should be able to put it out this weekend.

Frenzie:
Sorry, I forgot to merge this. Do you still want it in as is? (I mean besides the minor conflict with current master.)

benoit-pierre:
Yes please.

PR #1683, opened 2023-10-28
mupdf: ensure an HTML document fonts are freed on close

benoit-pierre:
Clear the context's glyph cache on close, otherwise references to fonts are kept that prevent them being freed immediately. And since every new HTML document opened allocate new fonts, those glyph cache entries / fonts are unused and waste memory.

Running the emulator's debug version, with dictionary lookups in an EPUB (lookup word, keep `PageUp`/`PageDown` pressed a few seconds, close window, refresh status bar, rinse and repeat), the memory use stabilizes itself at  between 10 to 20 MB less.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1683)


Frenzie (approved):
They allocate new fonts even though they're tye same fonts? Strange.

NiLuJe (approved):

poire-z (approved):

PR #1685, opened 2023-10-29
Input: Ensure clearTimer won't compute a lower nfds value than one of our opened fds

NiLuJe:
By making sure in openInputDevice that the top member of the array also happens to be the highest fd number.

----

Initial draft below:

Mention that nfds computations assume the inputfds array is sorted in increasing fd numbers

This doesn't necessarily hold true in practice (e.g., if one were to first open an fd during a directory walk with a few dir/file fds open...)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1685)

### DIFF
--- a/input/input.c
+++ b/input/input.c
@@ -172,11 +172,11 @@ static int openInputDevice(lua_State* L)
 
     // NOTE: Make sure the top member has the highest fd number, for clearTimer's sake when it recomputes nfds
     if (fd_idx > 0) {
+        int prev_fd   = inputfds[fd_idx - 1];
         int opened_fd = inputfds[fd_idx];
-        int top_fd    = inputfds[fd_idx - 1];
-        if (opened_fd < top_fd) {
+        if (opened_fd < prev_fd) {
             inputfds[fd_idx - 1] = opened_fd;
-            inputfds[fd_idx]     = top_fd;
+            inputfds[fd_idx]     = prev_fd;
         }
     }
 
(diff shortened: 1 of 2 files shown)


NiLuJe:
Just so that I remember to fix that one day ;).

(Possibly by sorting the `inputfds` array on open/close, to keep the timer code simple and avoid having to walk it).

NiLuJe:
Open random fds between input.open calls in Kobo to repro, e.g.,

```
diff --git a/frontend/device/kobo/device.lua b/frontend/device/kobo/device.lua
index e2df4a95a..dd2f94322 100644
--- a/frontend/device/kobo/device.lua
+++ b/frontend/device/kobo/device.lua
@@ -816,12 +816,16 @@ function Kobo:init()
     -- And then handle the extra shenanigans if necessary.
     self:initEventAdjustHooks()
 
+    local f = io.open("/dev/null", "r")
+    local f2 = io.open("/dev/null", "w")
     -- Various HW Buttons, Switches & Synthetic NTX events
     self.ntx_fd = self.input.open(self.ntx_dev)
     -- Dedicated Power Button input device (if any)
     if self.power_dev then
         self.input.open(self.power_dev)
     end
+    f:close()
+    f2:close()
     -- Touch panel
     self.input.open(self.touch_dev)
     -- NOTE: On devices with a gyro, there may be a dedicated input device outputting the raw accelerometer data
```

Then `strace -fittvTyY -s128 -e trace=_newselect -p $(pidof reader.lua | sed -e 's/ / -p /g')` to watch the horror unfold ;).

Frenzie (approved):

PR #1686, opened 2023-11-01
improve macOS CI build

benoit-pierre:
- fix cache handling: the wrong path was configured, so the cache was not saved. Also tweak use so the cache is still saved on cancellation / failure.
- build checkout phase: update action so a full clone with filtering can be used (so tags are fetched and koreader's version is correct).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1686)


Frenzie:
Thanks!

PR #1687, opened 2023-11-01
minor Makefile cleanups

benoit-pierre:
- `RCP`: use `cp -R` (supported on all systems we care about)
- `ISED`: get rid of it (none of the makefiles use it)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1687)

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -230,14 +230,7 @@ USE_LUAJIT_LIB=$(or $(DARWIN),$(ANDROID),$(WIN32))
 
 SKIP_LUAJIT_BIN=$(or $(ANDROID),$(MACOS))
 
-# handle utility params on different host systems
-ifdef DARWINHOST
-	RCP:=cp -R
-	ISED:=sed -i '' -e
-else
-	RCP:=cp -r
-	ISED:=sed -i -e
-endif
+RCP := cp -R
 
 # unknown device
 ifdef SBOX_UNAME_MACHINE


Frenzie:
As an aside I'd say there are some platforms we support with a half eye. For example I might want to set up a Sailfish build sometime, and while I have little interest in it myself, Windows is mentioned occasionally. But of course that doesn't change the fact that I think `-R` should be supported by everything and if anything `-r` is probably the trickier option with little reason to be used.

> ISED: get rid of it (none of the makefiles use it)

I'm less enthusiastic about that one in the sense that I can see it being added back again tomorrow, which would mean needing to remember or rediscover that Mac(/BSD?) vs. GNU distinction. Not an objection per se, but pinging e.g. @NiLuJe and @poire-z for some thoughts. :-)

Frenzie:
The description in the OP for ISED is slightly misleading. My comment is implicitly addressed in #1689 by moving the logic to CMake.

https://github.com/koreader/koreader-base/pull/1689/files#r1378980226

Frenzie (approved):

NiLuJe:
> I'm less enthusiastic about that one in the sense that I can see it being added back again tomorrow, which would mean needing to remember or rediscover that Mac(/BSD?) vs. GNU distinction. Not an objection per se, but pinging e.g. @NiLuJe and @poire-z for some thoughts. :-)

We don't actually use it in Makefiles anymore, and I tweaked the CMake variant to use a syntax that miraculously work on both variants, so, yeah, let's drop that one to avoid further confusion ;).

NiLuJe (approved):

Frenzie:
> I tweaked the CMake variant to use a syntax that miraculously works on both variants, so, yeah, let's drop that one to avoid further confusion ;).

Well, that's added to my confusion for the moment. ;-)

So https://github.com/koreader/koreader-base/blob/d1516bde5287429e3634b0253fd3c59422a51010/thirdparty/cmake_modules/koreader_thirdparty_common.cmake#L8 does all it needs and that means #1689 doesn't need any of that complexity?

benoit-pierre:
> > I'm less enthusiastic about that one in the sense that I can see it being added back again tomorrow, which would mean needing to remember or rediscover that Mac(/BSD?) vs. GNU distinction. Not an objection per se, but pinging e.g. @NiLuJe and @poire-z for some thoughts. :-)
> 
> We don't actually use it in Makefiles anymore, and I tweaked the CMake variant to use a syntax that miraculously works on both variants, […]

Are you sure it does? It looks like `sed -ie` can result in using `e` as a backup suffix on macOS. (Are we sure the CI is not using GNU version installed by brew?)

And with GNU sed:
```bash
 ▹  sed -ie 'p' -e ';' file
sed: can't read p: No such file or directory
sed: can't read file: No such file or directory
▹  sed -i -e 'p' -e ';' file
sed: can't read file: No such file or directory
```
So it does not work with multiple scripts.

NiLuJe:
I don't recall when/how/if I tested it, but the macOS CI seemed happy enough ;p.

PR #1689, opened 2023-11-01
cleanup thirdparty cmake rules

benoit-pierre:
- drop `KO_PATCH_SH`, always use `KO_PATCH`.
- avoid unnecessary use of `sh -c`, use `list(APPEND PATCH_CMD COMMAND …)` to handle patches and other commands in general, so calls to `ExternalProject_Add` can be simplified and the different commands can be reordered, commented out or uncommented without the need to renumber a bunch of variables or change multiple places. Additionally, this can reduce the number of backslashes in some cases.
- factorize the Android libtool patch.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1689)


thirdparty/djvulibre/CMakeLists.txt
```diff
@@ -25,24 +36,13 @@ set(CXXFLAGS "${CXXFLAGS} -Wno-register -Wno-error=register")
 
 set(CFG_ENV_VAR "CC=${CC}" "CXX=${CXX}" "CFLAGS=${CFLAGS}" "CXXFLAGS=${CXXFLAGS}" "LDFLAGS=${LDFLAGS}")
 set(CFG_OPTS --disable-desktopfiles --disable-static --enable-shared --disable-xmltools --disable-largefile --without-jpeg --without-tiff -host=${CHOST})
-set(CFG_CMD env ${CFG_ENV_VAR} ${SOURCE_DIR}/configure ${CFG_OPTS})
```
Frenzie, line 28: Note to others reading: this was moved up a few lines.

thirdparty/djvulibre/CMakeLists.txt
```diff
@@ -25,24 +36,13 @@ set(CXXFLAGS "${CXXFLAGS} -Wno-register -Wno-error=register")
...
-
-# Never *ever* call setlocale, for any reason.
-# Most of our target platforms have custom/broken/incomplete locales, don't screw with it.
-# And this runs in our main thread, so this is Very Bad(TM).
-set(PATCH_CMD1 "${KO_PATCH} ${CMAKE_CURRENT_SOURCE_DIR}/djvulibre-no-locale-mangling.patch")
```
Frenzie, line 33: Note to others reading: this was moved up a few lines.

thirdparty/cmake_modules/koreader_thirdparty_common.cmake
```diff
@@ -5,7 +5,11 @@ if(NOT DEFINED PROCESSOR_COUNT)
     set(PROCESSOR_COUNT ${N})
 endif()
 
-set(ISED "sed -ie")
+if (DEFINED ENV{DARWIN})
+    set(ISED sed -i "" -e)
+else()
+    set(ISED sed -i -e)
+endif()
```
Frenzie, line 12: Note: this addresses my comment in #1687.

benoit-pierre: Except it does not work ;(, I'm thinking because the empty arguments is not passed.

benoit-pierre: But `sed -ie` is not right, or at least is reported not to work in all macOS versions. (And is wrong with GNU sed when using multiple `-e` script too).

Frenzie: Well, I'm perfectly fine leaving the ISED stuff as it was. ;-)

I was surprised to see this PR actually, given Meson. ^_^

benoit-pierre: I fear I might have to resort to something like:
```cmake
set(ISED sh -c "sed -i '' -e \"$@\"" --)
```

Frenzie: I assume you mean it's not some weird escaping issue like this?

```
set(ISED sed -i \"\" -e)
```

Frenzie: Or perhaps even along these lines?

```
set(ISED "sed -i '' -e")
```

benoit-pierre: > `set(ISED sed -i \"\" -e)`

No, the problem is, quoting the cmake documentation:
>  Each non-empty element is given to the command invocation as an argument.

> `set(ISED "sed -i '' -e")`

That's one argument.

Frenzie (approved):
lgtm, but I figure it's better to do a base bump now (or at least soon) and this one at least one nightly build later

Frenzie:
I think something may not be quite right; my MuPDF seems to be always recompiling. I don't have time to investigate right now; just mentioning it in case anybody else noticed. (Not being able to start in a split second is quite disturbing. :-) )

benoit-pierre:
I don't see anything in the changes to the corresponding `CMakeLists.txt` that would trigger this, and the makefiles are unchanged, so… Try running with `MAKEFLAGS=--debug=why` to get an explanation.

Frenzie:
I do have time to quickly run that, but:

```
export MAKEFLAGS=--debug=why
$ ./kodev run -s=kobo-aura-one
[*] Building KOReader…
make: *** unknown debug level specification 'why'.  Stop.
```

benoit-pierre:
The crengine `JCONFIG_INCLUDE_DIR` setting can just be removed, and with that gone, so is the MuPDF dependency.

Frenzie:
> As for the fix: use order-only prerequisites for directories.

Could you explain that a bit more?

1. They're missing target. (It seems to me they're implicit, not missing. But okay, it's arguably clearer.)
2. But they should be some kind of fake run-once target instead to prevent them from acting like a dependency. (This doesn't matter in practical terms, because they remain implicit.)

That sounds like a lot of obfuscation and indirection to get right back to the beginning. What specifically is the advantage?

benoit-pierre:
Here's the patch for both:
```diff
 Makefile                           |  2 +-
 Makefile.third                     | 28 +++++++---------------------
 thirdparty/kpvcrlib/CMakeLists.txt |  2 --
 3 files changed, 8 insertions(+), 24 deletions(-)

diff --git i/Makefile w/Makefile
index 2d268c52..3b4ef814 100644
--- i/Makefile
+++ w/Makefile
@@ -241,7 +241,7 @@ $(OUTPUT_DIR)/button-listen: button-listen.c
 # ===========================================================================
 # the attachment extraction tool:
 
-$(OUTPUT_DIR)/extr: extr.c $(MUPDF_LIB) $(MUPDF_DIR)/include $(JPEG_LIB) $(FREETYPE_LIB)
+$(OUTPUT_DIR)/extr: extr.c $(MUPDF_LIB) $(JPEG_LIB) $(FREETYPE_LIB) | $(MUPDF_DIR)/include
 	$(CC) -I$(MUPDF_DIR) -I$(MUPDF_DIR)/include \
 		$(CFLAGS) $(LDFLAGS) -o $@ extr.c \
 		-lmupdf $(JPEG_LIB_LINK_FLAG) $(FREETYPE_LIB_LINK_FLAG)
diff --git i/Makefile.third w/Makefile.third
index 9ecd3bfe..55e9aa15 100644
--- i/Makefile.third
+++ w/Makefile.third
@@ -132,11 +132,9 @@ $(AES_LIB): $(THIRDPARTY_DIR)/minizip/*.*
 
 # by default, mupdf compiles to a static library:
 # we generate a dynamic library from the static library:
-$(MUPDF_LIB) $(MUPDF_DIR)/include $(MUPDF_DIR)/scripts: $(JPEG_LIB) \
-		$(FREETYPE_LIB) $(FREETYPE_DIR)/include \
-		$(HARFBUZZ_LIB) $(HARFBUZZ_DIR)/include \
-		$(LIBWEBP_LIB) $(LIBWEBP_DIR)/include \
-		$(ZLIB) $(AES_LIB) $(THIRDPARTY_DIR)/mupdf/*.*
+$(MUPDF_LIB) $(MUPDF_DIR)/include $(MUPDF_DIR)/scripts: $(THIRDPARTY_DIR)/mupdf/*.* \
+	$(AES_LIB) $(FREETYPE_LIB) $(HARFBUZZ_LIB) $(JPEG_LIB) $(LIBWEBP_LIB) $(ZLIB) \
+	| $(FREETYPE_DIR)/include $(HARFBUZZ_DIR)/include $(LIBWEBP_DIR)/include
 	-rm -rf $(MUPDF_BUILD_DIR)
 	install -d $(MUPDF_BUILD_DIR)
 	cd $(MUPDF_BUILD_DIR) && \
@@ -251,26 +249,14 @@ ifdef DARWIN
 endif
 
 # crengine, fetched via GIT as a submodule
-$(CRENGINE_LIB) $(CRENGINE_THIRDPARTY_LIBS): $(CRENGINE_SRC_FILES) \
-		$(FREETYPE_DIR)/include \
-		$(FRIBIDI_DIR)/include \
-		$(HARFBUZZ_DIR)/include \
-		$(JPEG_DIR)/include \
-		$(LIBUNIBREAK_DIR)/include \
-		$(LIBWEBP_DIR)/include \
-		$(LUNASVG_DIR)/include \
-		$(MUPDF_DIR)/scripts \
-		$(PNG_DIR)/include \
-		$(SRELL_INCLUDE_DIR) \
-		$(UTF8PROC_DIR) \
-		$(ZLIB_DIR)/include \
-		$(ZSTD_DESTDIR)/include \
-		$(THIRDPARTY_DIR)/kpvcrlib/*.*
+$(CRENGINE_LIB) $(CRENGINE_THIRDPARTY_LIBS): $(CRENGINE_SRC_FILES) $(THIRDPARTY_DIR)/kpvcrlib/*.* \
+	| $(FREETYPE_DIR)/include $(FRIBIDI_DIR)/include $(HARFBUZZ_DIR)/include $(JPEG_DIR)/include \
+	  $(LIBUNIBREAK_DIR)/include $(LIBWEBP_DIR)/include $(LUNASVG_DIR)/include $(PNG_DIR)/include \
+	  $(SRELL_INCLUDE_DIR) $(UTF8PROC_DIR) $(ZLIB_DIR)/include $(ZSTD_DESTDIR)/include
 	install -d $(CRENGINE_BUILD_DIR)
 	cd $(CRENGINE_BUILD_DIR) && \
 		$(CMAKE) $(CMAKE_FLAGS) \
 		-DJPEGLIB_INCLUDE_DIR='$(JPEG_DIR)/include' \
-		-DJCONFIG_INCLUDE_DIR='$(MUPDF_DIR)/scripts' \
 		-DPNG_INCLUDE_DIR='$(PNG_DIR)/include' \
 		-DLUNASVG_INCLUDE_DIR='$(LUNASVG_DIR)/include' \
 		-DZLIB_INCLUDE_DIR='$(ZLIB_DIR)/include' \
diff --git i/thirdparty/kpvcrlib/CMakeLists.txt w/thirdparty/kpvcrlib/CMakeLists.txt
index 630879bc..b8799eef 100644
--- i/thirdparty/kpvcrlib/CMakeLists.txt
+++ w/thirdparty/kpvcrlib/CMakeLists.txt
@@ -19,7 +19,6 @@ set(ANTIWORD_INCLUDE_DIR ${CR_3RDPARTY_DIR}/antiword)
 set(CHM_INCLUDE_DIRS ${CR_3RDPARTY_DIR}/chmlib/src)
 set(CREGINE_INCLUDE_DIR ${CRE_DIR}/include)
 
-assert_var_defined(JCONFIG_INCLUDE_DIR)
 assert_var_defined(FREETYPE_INCLUDE_DIR)
 assert_var_defined(HARFBUZZ_INCLUDE_DIR)
 assert_var_defined(FRIBIDI_INCLUDE_DIR)
@@ -33,7 +32,6 @@ assert_var_defined(LUNASVG_INCLUDE_DIR)
 assert_var_defined(ZSTD_INCLUDE_DIR)
 
 include_directories(
-    ${JCONFIG_INCLUDE_DIR}
     ${FREETYPE_INCLUDE_DIR}
     ${HARFBUZZ_INCLUDE_DIR}
     ${FRIBIDI_INCLUDE_DIR}
```

benoit-pierre:
> > As for the fix: use order-only prerequisites for directories.
> 
> Could you explain that a bit more?

Order-only requisites don't factor in determining if the target is outdated an must be rebuilt. That's exactly what we want from directories (as you can't rely on their timestamps): just make sure it's created before attempting to update the target.

> 
>     1. They're a missing target. (It seems to me they're implicit, not missing. But okay, it's arguably clearer.)
> 
>     2. But they should be some kind of fake run-once target instead to prevent them from acting like a dependency. (This doesn't matter in practical terms, because they remain implicit.)
> 
> 
> That sounds like a lot of obfuscation and indirection to get right back to the beginning. What specifically is the advantage?

Frenzie:
> Here's the patch for both:

It's still very enthusiastically rebuilding all kinds of stuff at random.
```
     Prerequisite 'build/x86_64-linux-gnu-debug/libs/libharfbuzz.so.0' is newer than target 'build/x86_64-linux-gnu-debug/libs/libkoreader-xtext.so'.
     Prerequisite 'build/x86_64-linux-gnu-debug/libs/libfribidi.so.0' is newer than target 'build/x86_64-linux-gnu-debug/libs/libkoreader-xtext.so'.
    Must remake target 'build/x86_64-linux-gnu-debug/libs/libkoreader-xtext.so'.
```

benoit-pierre:
Nothing random in those traces, what's the full log?

Frenzie:
> Nothing random in those traces

Random = nothing changed, ergo the program should start in less than a second because there's nothing to do instead of doing a ton of work.

[log.txt](https://github.com/koreader/koreader-base/files/13353566/log.txt)

benoit-pierre:
That's the root of the issue:
```
Prerequisite 'thirdparty/fribidi/CMakeLists.txt' is newer than target '/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/x86_64-linux-gnu-debug/fribidi-prefix/src/fribidi-build/include'.
```

Frenzie:
Yes, the issue is that targets like `'/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/x86_64-linux-gnu-debug/fribidi-prefix/src/fribidi-build/include'` were added.

Either the FriBiDi target succeeds and `'/home/frans/src/kobo/koreader/base/thirdparty/fribidi/build/x86_64-linux-gnu-debug/fribidi-prefix/src/fribidi-build/include'` exists or it fails and it doesn't. And if something about that assumption breaks down for some reason (e.g., updated dependency) the downstream target will gracefully say so.

I still fail to understand the rationale behind making them a target, but I'd glanced over the implications which didn't become clear until the inevitable future discrepancy between `CMakeLists.txt` (or rather `*.*`) and `/include`.

In my view a line like this would be a sneaky pitfall for anybody in the future, assuming it made a noticeable difference.

```
| $(FREETYPE_DIR)/include $(HARFBUZZ_DIR)/include $(LIBWEBP_DIR)/include
```

I suppose you could touch the folder at the end but as stated then you have a niche syntax coupled with extra lines to get right back where you started.

PR #1690, opened 2023-11-01
optimize gettext / glib builds

benoit-pierre:
- compile only the part of gettext we need
- compile only a static version of `libglib-2.0`, since that's all sdcv needs

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1690)


Frenzie:
As an aside and I don't know if it matters, but it's the glib becoming a bit old?

benoit-pierre:
It's not too fresh, yeah.

NiLuJe:
Yeah, newer builds switched to meson and I didn't have the will to deal with it for a platform idgaf about ;).

Frenzie:
@benoit-pierre What does the draft status currently indicate? ^_^

benoit-pierre:
I was just waiting for #1689 to be merged.

PR #1691, opened 2023-11-02
fix crengine compilation cache

benoit-pierre:
Use `-include […]/crsetup.h` instead of `--include […]/crsetup.h`, as the later would result in 0% cacheable calls when building with ccache (cf. [ccache issue #1324](https://github.com/ccache/ccache/issues/1324))…

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1691)


NiLuJe (approved):

benoit-pierre:
For the macOS failure, see #1692.

PR #1697, opened 2023-11-14
experimental build system improvements

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1697)


.ci/build_script.sh
```diff
@@ -15,37 +15,37 @@ if [ "$TARGET" = "android" ] && [ -n "${DOCKER_IMG}" ]; then
     docker run -t \
         -v "${HOME}/.ccache:${DOCKER_HOME}/.ccache" \
         -v "$(pwd):${DOCKER_HOME}/base" "${DOCKER_IMG}" \
-        /bin/bash -c 'source /home/ko/.bashrc && cd /home/ko/base && sudo chown -R ko:ko . && make TARGET=android all'
+        /bin/bash -c 'source /home/ko/.bashrc && cd /home/ko/base && sudo chown -R ko:ko . && make TARGET=android -j'
```
Frenzie, line 18: Unlimited if I recall correctly? Is that wise?

thirdparty/CMakeLists.txt
```diff
@@ -0,0 +1,173 @@
+cmake_minimum_required(VERSION 3.5.1)
```
Frenzie, line 1: I'd rather be moving further away from CMake tbh. Didn't someone have a Meson branch? ;-) But with the stuff from Debian cross forcing it into some semblance of sanity it's fine by me. ^_^

thirdparty/curl/CMakeLists.txt
```diff
@@ -1,41 +1,112 @@
 cmake_minimum_required(VERSION 3.5.1)
```
Frenzie, line 1: As an aside and not really relevant here, but after dealing with #1682 we can jump to 3.16. (Otoh I think feature-wise I'd want at least 3.12 but it'd be basically untested.)

Strictly speaking of course that can also be done separately if desired, but it's probably less effort overall that way.

poire-z:
Can't judge, but that's a lot of changes.

Just curious: did you rewrite all this manually ? How many iterations/rebuilds to get it working ?!! :)
Or did you use some tool to make it all consistent ?

Also, why, while happy with your meson branch, would you spend that much time in CMake hell, fixing/improving/tidying all that ?!

.ci/build_script.sh
Frenzie: Change to `$(getconf _NPROCESSORS_ONLN)` in the meantime.

benoit-pierre:
> Also, why, while happy with your meson branch, would you spend that much time in CMake hell, fixing/improving/tidying all that ?!

Morbid curiosity? Refreshing my make-fu, confirming my feelings on CMake (bleh).

benoit-pierre:
Does not work without at least 3.18 anyway, even without using the experimental experimental mode ;).

Frenzie:
> Does not work without at least 3.18 anyway

Ah, I wasn't planning on going quite there just yet — though then again it's been over a year since I explicitly didn't plan that. For cmake_language() huh? In that case I'd probably propose to jump straight to 3.22 (default in Ubuntu 22.04) or maybe even 3.25 (default in Debian 12).

benoit-pierre:
Yes, and for the other mode (`thirdparty/CMakeLists.txt`): 3.19 at least for [CMP0114](https://cmake.org/cmake/help/latest/policy/CMP0114.html#cmp0114) and ([CMP0113](https://cmake.org/cmake/help/latest/policy/CMP0113.html#cmp0113) looks pretty bad too if ninja is not used).

For make, I've stuck to at least 4.0 (2013).

benoit-pierre:
Which if course is still to new (make 4.0) for macOS it seems… Why am I not surprised?

Frenzie:
> 3.19 at least for [CMP0114](https://cmake.org/cmake/help/latest/policy/CMP0114.html#cmp0114)

You read my mind. ;-)

> For make, I've stuck to at least 4.0 (2013).

My guiding principle is simply that I try not to get too much in the way of someone who might randomly decide to try some hacking. In practical terms that comes down to trying to make it so that someone using the previous stable Debian/Ubuntu can just install a few packages and go, without having to figure out how to install a new CMake. It's not that hard, but it's just an extra obstacle.

But on Mac none of that applies. You have to do all manner of weird things regardless. Adding `brew (g?)make` to the list isn't really adding anything I'd say. ^_^

NiLuJe:
Feel free to also PR buildsystem changes from a few of the patches to the projects where we're also upstream (e.g., libk2pdfopt, fbink) ;).

benoit-pierre:
You'll notice a step in that direction: changing the way leptonica / tesseract / k2pdfopt are built.

NiLuJe:
Speaking of FBInk in particular, master has seen some more "improvements" compared to the version currently being pinned ('and a bump would be harmless).

It's still a mess, but there's a semblance of possibly viable make dependencies at play now ;p.

benoit-pierre:
O.o
```
************ NINJA: /usr/bin/ninja -j36 -l36 **********
************ MAKE: make -j36 -l36 **********
```
(on Circle CI)

Frenzie:
> O.o
> 
> ```
> ************ NINJA: /usr/bin/ninja -j36 -l36 **********
> ************ MAKE: make -j36 -l36 **********
> ```
> 
> (on Circle CI)

Oh yeah, I recall running into that in the past, same on Travis CI, possibly better on GitLab CI. The same problem exists with e.g. nproc and /proc since they'll just give info about the machine they're on rather than about imposed limits. Our real limit is 2, though in practical terms I think there's an option of peaks when it's quiet. But again, iirc and if it hasn't changed since.

In short, I think that should just say if CIRCLECI JOBS=2.

Frenzie:
Also something in this general direction might work, probably even on macOS?
```
taskset -cp $BASHPID
```

PR #1706, opened 2023-12-02
cmake: cleanup language handling

benoit-pierre:
Most projects don't need any language (their external project do).


Frenzie (approved):
I have on occasion found it useful but indeed it's mostly just a waste of time.

PR #1708, opened 2023-12-02
circleci: save the cache before running tests

benoit-pierre:
No sense in wasting re-building stuff without cache if the testsuite fails.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1708)


Frenzie (approved):
There was probably a reason for it but I can't think of anything.

PR #1710, opened 2023-12-02
luajit: ensure ccache is used

benoit-pierre:


NiLuJe (approved):
Not familiar with ccache (at all) and it's not like we can see anything without verbose builds, but it looks sound ;).

PR #1711, opened 2023-12-02
add fallback URLs for gettext, libiconv & tar downloads

benoit-pierre:
Note: this does increase the minimum CMake version to 3.7.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1711)


Frenzie:
And it works because I forgot about https://github.com/koreader/virdevenv/commit/45c839446e590a932d9edee20d16096f12b02bdc when I left this comment: <https://github.com/koreader/koreader-base/pull/1697#discussion_r1393324445>

PR #1713, opened 2023-12-10
Fix dowload android sdk

hugleo:
java -version
openjdk version "1.8.0_392"
OpenJDK Runtime Environment (build 1.8.0_392-b08)
OpenJDK 64-Bit Server VM (build 25.392-b08, mixed mode)

./sdkmanager --list --verbose | grep patcher
empty result

Build fails here:

./sdkmanager "platform-tools" "build-tools;30.0.2" "platforms;android-30" "patcher;v4" Warning: Failed to find package patcher;v4

Besides the fact I use arch linux and messed up with my system from time to time is only me facing this problem?

Can someone test whether the patcher is available? ./sdkmanager --list --verbose | grep patcher

I had to remove "patcher;v4" from the Makefile so the build could proceed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1713)

### DIFF
--- a/toolchain/Makefile
+++ b/toolchain/Makefile
@@ -25,7 +25,7 @@ $(TOOLCHAIN_DIR)/$(SDK_DIR):
 	pushd $(SDK_DIR)/tools/bin && \
 		yes | ./sdkmanager --update && \
 		yes | ./sdkmanager --licenses && \
-		./sdkmanager "platform-tools" "build-tools;30.0.2" "platforms;android-30" "patcher;v4" && \
+		./sdkmanager "platform-tools" "build-tools;30.0.2" "platforms;android-30" && \
 		popd
 	rm -f $(SDK_DIR)/$(SDK_TARBALL)
 


Frenzie:
> Can someone test whether the patcher is available? ./sdkmanager --list --verbose | grep patcher

Looks like Google removed it.

The SDK Patcher was added in #930 but I don't know otoh what it was used for.

hugleo:
If we ever need such thing I think the file is here: https://dl.google.com/android/repository/3534162-studio.sdk-patcher.zip

pazos:
> > Can someone test whether the patcher is available? ./sdkmanager --list --verbose | grep patcher
> 
> Looks like Google removed it.
> 
> The SDK Patcher was added in #930 but I don't know otoh what it was used for.

Yup, was required at the time when we did the ant -> gradle & sdk19 -> sdk28 bump.

pazos (approved):

PR #1714, opened 2023-12-13
sunxi: Minor comment updates

NiLuJe:
Finally had time to see if I could find a workaround for the missing post-standby refreshes on sunxi, but that didn't pan out, so we're just left with comment cleanups ;p.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1714)

### DIFF
--- a/ffi/framebuffer_ion.lua
+++ b/ffi/framebuffer_ion.lua
@@ -30,6 +30,8 @@ local framebuffer = {
 
     _finfo = nil,
     _vinfo = nil,
+
+    _just_rotated = nil,
 }
 
 --[[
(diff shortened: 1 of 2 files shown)


NiLuJe:
The issue in question is *probably* related to this report: https://github.com/koreader/koreader/issues/10743

Although my own experience is much less dramatic and completely unrelated to key repeats: once in a blue moon (really, I see this happen like less than once every ~12 hours of use or so), if the device is woken up from *standby* by a button press, the press registers, we turns the page, we request a refresh, but the refresh is never actually done. If unaware of the issue, your next reflex is to press the next page button again, in which case it refreshes, but you realize that you've skipped a page ;).

FWIW, it very very much seems to correlate with "janky" button presses (e.g., short/soft/mushy, or on the edge), but that might just be a coincidence...

PR #1718, opened 2023-12-31
FB: Refactor refresh rectangle bounding checks

NiLuJe:
https://github.com/koreader/koreader/issues/11303 showed that the current alignment trickery for HW dithering on mxcfb was a bit shoddy, so refactor the whole thing to do everything at once, in a function actually designed for that use case, instead of re-purposing a blitting helper...

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1718)

### DIFF
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -1067,7 +1067,7 @@ i.e., make sure that x & y don't go OOB on either side, and that x+w & y+h don't
 
 @return rect strictly bounded inside the bb
 --]]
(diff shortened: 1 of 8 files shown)


NiLuJe (commented):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1718)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

PR #1719, opened 2023-12-31
Fix Kindle swipe_direction in D-orientation in framebuffer_mxcfb.lua

mergen3107:
Closes #11180 (if accepted)

@yparitcher @NiLuJe 
I have been using this solution (https://github.com/koreader/koreader/issues/11180#issuecomment-1869083861) for a few days on Kindle Scribe - no issues! Swipe-direction is correct in both `U` and `D` orientation startups.

I tried to fix Kindle at `U` at the start, but it didn't help with swipe-direction.
I don't know how this behaves on other MTK devices (PW5 only?).

If there is no other better solution, would you guys agree to accept this PR?

Thanks!

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1719)

### DIFF
--- a/ffi/framebuffer_mxcfb.lua
+++ b/ffi/framebuffer_mxcfb.lua
@@ -604,6 +604,10 @@ end
 -- Set the swipe animation direction. This is sticky
 function framebuffer:_MTK_SetSwipeDirection(left)
     local swipe_direction = left and C.MTK_SWIPE_LEFT or C.MTK_SWIPE_RIGHT
+    -- If device was started in D orientation, rotate swipe_direction by 180 deg.
+    if self.native_rotation_mode == self.DEVICE_ROTATED_UPSIDE_DOWN then
+        swipe_direction = bxor(swipe_direction, 1)
+    end
     -- Account for KOReader rotation not matching the FB rotation with some bit trickery.
     -- Alternatively here is a LUT:
     -- local lut = { -- L R D U


yparitcher:
@mergen3107 
On a hunch can you check the values for `supportedOrientation` in the `properties` table of `/var/local/appreg.db`? Specifically for `kual` `kol` `com.lab126.booklet.home` & `com.lab126.booklet.reader`.

It is a sqlite database.

mergen3107:
@yparitcher 
I made a backup copy and got these values (while in KOReader in `D` orientation):
```
/var/local/tmp$ echo $(sqlite3 /var/local/tmp/appreg_bak.db "SELECT value FROM properties WHERE handlerId = 'com.lab126.booklet.reader' AND name = 'supportedOrientation'")
URL
/var/local/tmp$ echo $(sqlite3 /var/local/tmp/appreg_bak.db "SELECT value FROM properties WHERE handlerId = 'com.lab126.booklet.home' AND name = 'supportedOrientation'")
U
```

mergen3107:
It is weird that `reader` says only `URL` - I can definitely start it or rotate it to `D` as well.

yparitcher:
One last thing,
Can you check if the values of `orientation` or `orientationLock` change when the `accelerometer` changes (and or when setting `orientationLock`) , by running (probably via ssh) `lipc-probe com.lab126.winmgr -v` while in a kindle book, and in koreader?

If orientation is updated with orientationLock, then we can use it at startup to determine the software orientation.

mergen3107:
1) Only things that change are `orientation` and `accelerometer` (read-only)
2) `orientationLock` stays the same (`U`), but it is read/write.

image

mergen3107:
When fixing `orientationLock` to `D`, `orientation` and `accelerometer` stay the same.

yparitcher:
I wonder if adding `D` to kual `supportedOrientation` in appreg.db (so `UD`) would convince the framework to allow it to be upside down properly (it might just make everything go crazy). If you feel like trying this make sure the kual does not exit like on the voyage or try it with Kol

mergen3107:
Thanks for the idea! :D 
So I need to `UPDATE value` via sqlite3?

yparitcher:
> Thanks for the idea! :D 
> So I need to `UPDATE value` via sqlite3? 

Yes

mergen3107:
So I added `UD` with this:
```
echo $(sqlite3 /var/local/tmp/appreg_bak.db "UPDATE properties SET value = 'UD' WHERE handlerId LIKE '%.kindlelauncher' AND name = 'supportedOrientation'")
```

Then in KUAL extenstion for KOReader added `"exitmenu": false,`, the same for Voyage.

Then with `koreader.sh` and `kindle/device.lua:KindleScribe:init()` tried all combinations for `self.screen.native_rotation_mode` and `self.screen:setRotationMode()`.

In `koreader.sh` I try to do this:
```
# Store current orientationLock and then force it to U
export KINDLE_ORIENTATION_ORIGINAL="$(lipc-get-prop com.lab126.winmgr accelerometer)"
lipc-set-prop com.lab126.winmgr orientationLock U
export KINDLE_ORIENTATION_CHANGED="$(lipc-get-prop com.lab126.winmgr orientationLock)"
```

or this:
```
# Store current orientationLock and then force it to U
export KINDLE_ORIENTATION_ORIGINAL="$(lipc-get-prop com.lab126.winmgr accelerometer)"
lipc-set-prop com.lab126.winmgr orientationLock $KINDLE_ORIENTATION_ORIGINAL
export KINDLE_ORIENTATION_CHANGED="$(lipc-get-prop com.lab126.winmgr orientationLock)"
```

I end up with three outcomes:
1) The same as before this PR: screen is aligned with touch, shows up the same orientation as KUAL, but `swipe_direction` is opposite.
2) Screen and touch are opposite.
3) The same as (1) but shows up upside down.

Did I miss something?

mergen3107:
Summary, when in `D` in KUAL:
1) Number 1 above is this:
```
            self.screen.native_rotation_mode = 2
            self.screen:setRotationMode(2)
```

2) Number 2 is:
```
```
            self.screen.native_rotation_mode = 0
            self.screen:setRotationMode(0)
```
3) other combinations 2/0, 0/2 are broken in one way or another.

`swipe_direction` isn't a friend to anyone when KOReader starts in `D`. Hence, this PR

mergen3107:
Same reasons as here (cannot make KOReader rotate Kindle's UI upon exit): https://github.com/koreader/koreader/pull/11809#issuecomment-2150755707

I left this as is. I got a Kindle Scribe case that enforces `U` orientation, and so I haven't encountered this issue since then.

PR #1721, opened 2024-01-12
New year's third-party bumps ;p

NiLuJe:
A few of 'em, but nothing particularly scary (famous last words...).


Frenzie (approved):

NiLuJe:
Updated fbink to (hopefully) fix the cervantes & pb build.

(Actual oversight in the Debian packaging script for cervantes, and old gcc issue for pb)

PR #1728, opened 2024-01-15
setHWRotation: nil guard self.forced_rotation

yparitcher:
will be needed to test kindle rotate shenanigans where we do not set forced_rotation

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1728)


NiLuJe (approved):
That had never been used outside of PB, FWIW ;).

PR #1732, opened 2024-01-19
Bump cURL & zsync2

NiLuJe:
Hopefully up-to-date code works properly w/ modern autoconf versions...

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1732)


NiLuJe:
Well, that wasn't *too* painful. prstux had to get axed, but it's a non-issue, as we no logner ship OTA for that anyway.

Frenzie:
> Well, that wasn't _too_ painful. prstux had to get axed, but it's a non-issue, as we no longer ship OTA for that anyway.

Huh? It's using a reasonably up-to-date image.

Frenzie:
> It somehow fails to build there.
We don't even ship OTA on that platform and it's 150% dead,
so, just nuke it to make everybody's life easier.

I will remark that it doesn't necessarily do that. If you just want to quickly play around you can install your distro's cross-compile TC (in Debuntu `sudo apt install gcc-arm-linux-gnueabihf g++-arm-linux-gnueabihf`) and get a sufficiently working build of everything, even if it might not be perfect. The prstux build functioned as a sanity check for that. That one may have already ceased to work on the actual device years ago; that wasn't quite the build's purpose anymore.

That being said, I just noticed that even on master that method seems to break down on dropbear, which goes to show how much I personally use it I suppose. :-) But it was something I appreciated when it worked, perhaps a little for personal use but mainly to make it easier for people to jump in.

All of that is not an objection to removing features from that build, but in doing so it's important to keep in mind that the canary function is removed as well.

PS Here's the dropbear error, which I don't quite see how it relates to the TC but that aside.

```
/home/frans/src/koreader-base/thirdparty/dropbear/build/arm-linux-gnueabihf/dropbear-prefix/src/dropbear/libtomcrypt/../sysoptions.h:237:10: error: #error "DROPBEAR_SVR_PASSWORD_AUTH requires `crypt()'."
  237 |         #error "DROPBEAR_SVR_PASSWORD_AUTH requires `crypt()'."
      |          ^~~~~
```

NiLuJe:
It probably doesn't relate to the TC per-se, but to some weird insane CMake quirk, so I have absolutely no interest to waste any more time on it ;).

NiLuJe:
FWIW, the aforementioned stellar CMake behavior:

```
-- Could NOT find CURL (missing: HTTP HTTPS) (found version "8.5.0-DEV")
-- Could NOT find CURL (missing: HTTP) (found version "8.5.0-DEV")
```

Wheee!

PR #1734, opened 2024-01-25
PB: make sure libinkview-compat actually ends up w/ inkview as a DT_NEEDED

NiLuJe:
The TC is old enough that the order of flags matters, a lot.

Regression since #1638
Re: https://github.com/koreader/koreader/issues/11172

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1734)


Frenzie (approved):
I don't think LDFLAGS would be missed here, but good to know.

PR #1738, opened 2024-02-25
Use Bresenham's algorithm and anti-aliasing for rounded corners

ghost:
A minor optimization for rounded corners: use Bresenhams algorithm.

This PR works as excpected (=is a tad faster and does not need float).

Draft by now, as I want to include an antialiasing feature ;-)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1738)

### DIFF
--- a/blitbuffer.c
+++ b/blitbuffer.c
@@ -182,7 +182,7 @@ static const char*
 })
 
 static inline void BB8_SET_PIXEL_CLAMPED(BlitBuffer * restrict bb, int rotation, unsigned int x, unsigned int y, unsigned int width, unsigned int height, const Color8 * restrict color) {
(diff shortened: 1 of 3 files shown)


Frenzie:
Besides the anti-aliasing they are visually identical?

Frenzie:
I forgot to mention this the other day, but it's not a given that some kind of anti-aliasing will actually look better on eink.

Frenzie:
@NiLuJe Fine by you?

Frenzie:
I think this is intended to be squashed?

Frenzie:
Could you do it? I think I saw something about switching to Alois' algorithm somewhere, so I'd rather not mess it up. ;-)

PR #1741, opened 2024-03-11
thirdparty/turbo: remove some cdefs to prevent duplicates

Frenzie:
Comes with a sense of déjà vu from #1501. In reply to https://github.com/koreader/koreader/issues/11535.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1741)

### DIFF
--- a/thirdparty/turbo/turbo.patch
+++ b/thirdparty/turbo/turbo.patch
@@ -4,6 +4,41 @@ new mode 100644
 index efa69fc..b6ccfff
 --- a/turbo/cdef.lua
 +++ b/turbo/cdef.lua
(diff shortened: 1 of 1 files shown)


thirdparty/turbo/turbo.patch
```diff
@@ -53,11 +88,11 @@ index efa69fc..b6ccfff
              CLOCK_MONOTONIC
          };
 +        */
- 
+
```
Frenzie, line 91: These were an unintentional change caused by the editor stripping spaces, but I guess it doesn't seem to be harmful. Regardless, it shouldn't be in here.

PR #1744, opened 2024-03-21
thirdparty/harfbuzz: update to 8.3.1

Frenzie:
https://github.com/harfbuzz/harfbuzz/releases/tag/8.3.1

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1744)

### DIFF
--- a/thirdparty/harfbuzz/CMakeLists.txt
+++ b/thirdparty/harfbuzz/CMakeLists.txt
@@ -41,7 +41,7 @@ if(DEFINED ENV{ANDROID})
     list(APPEND CFG_CMD COMMAND ${ANDROID_LIBTOOL_FIX_CMD})
 endif()
 
-set(HB_VER 8.3.0)
+set(HB_VER 8.3.1)
 
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME


benoit-pierre:
Harfuzz 9.0 is out, and so is autotools! :P

> Autotools build system have been dropped. Meson is the only supported build system in HarfBuzz going forward.

Which means a hard dependency on meson if we update: and a hard dependency on ninja, so we might as well drop support for the make generator and simplify a few things.

Frenzie:
Yes, I saw that.

> so we might as well drop support for the make generator and simplify a few things.

I suppose that's alright. What kind of things?

benoit-pierre:
I meant simplify the build system code by dropping make generator support.

benoit-pierre:
The cmake build is still there however, which we use, and getting update, so we don't need to switch to meson. But I'd prefer it.

benoit-pierre:
It's [community maintained](https://github.com/harfbuzz/harfbuzz/issues/4613#issuecomment-2006389772).

Frenzie:
I don't necessarily see as strong a connection between dropping some generator flexibility and using meson for harfbuzz, but I'm not opposed. After all I added ninja support in the first place and then made it the default when available on the system.

But also pinging @poire-z @NiLuJe @pazos in case they have a reason to want to keep the Make option around.

benoit-pierre:
Because if there's a hard dependency on ninja (via meson), then we can rely on ninja always being there and there's no point in supporting the (IMHO inferior) cmake generator support for make.

benoit-pierre:
This would have the number of major variants I have to test when hacking the build system: cmake 3.15 + ninja, cmake < 3.20 + make, latest cmake + ninja, latest cmake + make.

poire-z:
No strong opinion, I guess I have accepted @benoit-pierre will shuffle the world around and make my life harder. I'll adapt :)
Hoping it will do better "the right thing", cf the other recent discussion.
(I only care about keeping crengine a submodule/subproject with its own commits and history.)

benoit-pierre:
> No strong opinion, I guess I have accepted @benoit-pierre will shuffle the world around and make my life harder. I'll adapt :)

Common, harder? I don't understand why you insist on using the slowest alternative when you compile on an under-powered machine.

> Hoping it will do better "the right thing", cf the other recent discussion.

I'm working on it, but you still have not defined "the right thing".

>  (I only care about keeping crengine a submodule/subproject with its own commits and history.)

Even if the whole history is intact? A subtree merge with no squashing has the same drawback as a rename. The git interface with respect to rename is not great, so you have to be explicit. Example with my meson branch:

```bash
▹ git blame -L :addInstance subprojects/crengine/crengine/src/lvfntman.cpp
8 months ago    Benoit Pierre   a32b9ed3d13│7427│ void LVFontCache::addInstance( const LVFontDef * def, LVFontRef fnt )
17 years ago    Vadim Lopatin   aac8cb6d918│7428│ {
8 months ago    Benoit Pierre   a32b9ed3d13│7429│     assert( !fnt.isNull() );
17 years ago    Vadim Lopatin   aac8cb6d918│7430│     LVFontCacheItem * item = new LVFontCacheItem(*def);
8 months ago    Benoit Pierre   a32b9ed3d13│7431│     item->_fnt = fnt;
17 years ago    Vadim Lopatin   aac8cb6d918│7432│     _instance_list.add( item );
                                           │7433│ }
                                           │7434│
```
```bash
▹ git sh --oneline --no-patch --stat a32b9ed3d13
a32b9ed3d1 lvfntman: fix possible font leaks
 crengine/include/lvfntman.h |   5 +++++
 crengine/src/lvfntman.cpp   | 113 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++------------------
 2 files changed, 97 insertions(+), 21 deletions(-)
```
```bash
▹ git sh --oneline a32b9ed3d13 crengine/include/lvfntman.h
fatal: ambiguous argument 'crengine/include/lvfntman.h': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```
```bash
▹ git sh --oneline a32b9ed3d13 -- crengine/include/lvfntman.h
a32b9ed3d1 lvfntman: fix possible font leaks
 crengine/include/lvfntman.h | 5 +++++
 1 file changed, 5 insertions(+)

diff --git a/crengine/include/lvfntman.h b/crengine/include/lvfntman.h
index 57f3c98241..3b65429491 100644
--- a/crengine/include/lvfntman.h
+++ b/crengine/include/lvfntman.h
@@ -535,6 +535,11 @@ public:
     // Obsolete:
     // virtual int getKerningOffset(lChar32 ch1, lChar32 ch2, lChar32 def_char) { CR_UNUSED3(ch1,ch2,def_char); return 0; }

+    /// clear internal references to this font pointer
+    virtual void clearFontRefs( const LVFont *ptr ) { }
+    /// clear all internal font references
+    virtual void clearFontRefs() { }
+
     /// set fallback font for this font
     virtual void setFallbackFont( LVFontRef font ) { CR_UNUSED(font); }
     /// get fallback font for this font
```

poire-z:
> Common, harder? I don't understand why you insist on using the slowest alternative when you compile on an under-powered machine.

I insist on nothing - I know nothing about all that, and I'm not really interested in the topic. I'll follow your choices. (You'll just have me spend a few hours on updating my wrapper scripts multiple times to get me back to not thinking about it , which I'll survive.)

> I'm working on it, but you still have not defined "the right thing".

Can't tell more that what I wrote at https://github.com/koreader/koreader-base/pull/1815#issuecomment-2191248120

> > (I only care about keeping crengine a submodule/subproject with its own commits and history.)

> Even if the whole history is intact? 

Yes. Independant, PR and issues of its own, History of its own. No git explicit magick needed.
And being able to merge the 4 PRs like I did earlier today, and still keeping that out of frontend until the next stable, is nice to have.

benoit-pierre:
And base?

poire-z:
I would personally be less bothered if it would happen to base, because I do a lot less work on stuff there (and my work is already mixed and drown among all of yours' works :)).
We previously discussed that around https://github.com/koreader/koreader-base/pull/1586#issuecomment-1535491587

NiLuJe:
Absolutely no attachment to the make generator as far as I'm concerned, FWIW ;).

PR #1746, opened 2024-03-24
tweak `patch-wrapper.sh`

benoit-pierre:
To support multiple patches in argument. Better than ignoring extra arguments after the first (e.g. in case of missing `COMMAND` in an external project patch command list).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1746)


Frenzie (approved):
Thanks!

PR #1748, opened 2024-03-24
djvulibre: drop unused libjpeg dependency

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1748)

### DIFF
--- a/Makefile.third
+++ b/Makefile.third
@@ -232,7 +232,7 @@ ifdef DARWIN
 		$(LUNASVG_LIB)
 endif
 
-$(DJVULIBRE_LIB): $(JPEG_LIB) $(THIRDPARTY_DIR)/djvulibre/*.*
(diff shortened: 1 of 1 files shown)


Frenzie:
Note that the README states the following, so if it's indeed unused my initial reaction is that precisely the reverse should be done. That is, to make sure that it is in fact used to support these files.
> 3.1- PREREQUISITES
> […]
> LIBJPEG/LIBTIFF --
A small number of DjVu files internally use JPEG encoding instead 
of IW44 wavelet encoding.  These files are processed using either 
version 6a or 6b of the IJG libjpeg library <http://www.ijg.org>.  
Some utility programs are able to read TIFF files. These programs
rely on the LIBTIFF library <http://libtiff.org>.
You can also compile DjVuLibre with neither JPEG nor TIFF support.  
You might never notice the difference.

PR #1749, opened 2024-03-24
make: drop unused variables

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1749)

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -263,8 +263,6 @@ ifneq (,$(findstring clang, $(CC)))
 	QFLAGS+=-Wno-parentheses-equality
 	# For dropbear's libtom...
 	QFLAGS+=-Wno-macro-redefined
-	# Note that things appear to be better with Clang >= 6 ;).
-	export CLANG=1
 else
 	ifdef CHOST
 		CC:=$(CHOST)-gcc
@@ -302,8 +300,6 @@ HOSTSTRIP:=strip
 HOSTAR:=ar
 HOSTLD:=ld
 HOSTRANLIB:=ranlib
-LIBTOOL:=$(if $(DARWIN),glibtool,libtool)
-LIBTOOLIZE:=$(if $(DARWIN),glibtoolize,libtoolize)
 
 # Path to the shared STL, because we need to ship it on most platforms.
 # We don't want to rely on -static-libstdc++, as it risks breaking the ODR.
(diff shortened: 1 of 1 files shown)


Makefile.defs
```diff
@@ -512,11 +508,6 @@ endif
 ifdef DARWIN_AARCH64_HOST
 # At the moment of writing, apple-m1 is the lowest common denominator for Apple silicon chips. Should work for M2 as well
 	HOST_ARCH:=-mcpu=apple-m1
-	BREW_PREFIX:=$(shell command brew --prefix 2>/dev/null)
-	ifneq (,$(BREW_PREFIX))
-		export DARWIN_AARCH64_HOST_CFLAG:=-I$(BREW_PREFIX)/include
-		export DARWIN_AARCH64_HOST_LDFLAG:=-L$(BREW_PREFIX)/lib
-	endif
```
Frenzie, line 519: Is this unused in the sense that it's not necessary on M1/M2 and/or newer Brew? That is, this isn't comparable to some internal `CLANG=1` remnant.

Makefile.defs
```diff
@@ -585,13 +576,6 @@ CMAKE_STAGING_DIR=build/staging-$(MACHINE)
 # in that case.
 
 ifdef EMULATE_READER
-	# CMake is difficult about CC and CXX, especially when it's not just a full path
-	CMAKE_C_COMPILER:=$(shell command -v $(strip $(HOSTCC)))
-	CMAKE_CXX_COMPILER:=$(shell command -v $(strip $(HOSTCXX)))
-	ifdef CCACHE
-		CMAKE_C_COMPILER_LAUNCHER:=$(CCACHE)
-		CMAKE_CXX_COMPILER_LAUNCHER:=$(CCACHE)
-	endif
```
Frenzie, line 594: Same question as above except with greater prejudice (against CMake, that is ^_^).

Makefile.defs
benoit-pierre: This was added in 30bf21c674, but not necessary anymore after d1910575bf240a14d53b5649e9301fec21f2d950 (and I should have removed it there).

Makefile.defs
benoit-pierre: It's not used, not exported, and when cross-compiling, already handled as part of the CMake toolchain definitions.

Frenzie: The former I'm not asking about! The latter I'll trust you on.

PR #1750, opened 2024-03-25
build system improvements

benoit-pierre:
Rework the build system to make it faster and better.

TODO:

- [x] PR fbink build system patches
- [x] resolve libk2pfopt situation
- ~[ ] update docker images (chrpath+meson)~

Highlights:

- move most of it to CMake: allow for parallelize each external project build
  steps as much as possible (e.g. all download / patch steps can be executed in
  parallel).
- add convenience targets for building koreader libraries and each external project
  build steps (e.g.: `make koreader-cre` to build only `libkoreader-cre.so` and all
  its dependencies, `make leptonica-re` to rebuild leptonica, etc…). Top level targets
  can be serialized and are forwarded to CMake (so its possible to do `make clean zlib`
  for example).
- change the build tree layout: keep the downloads in `thirdparty/*/build/…`,
  move the external projects build directories under the machine specific build
  directory: reduce the depth of the build tree, and simplify clean rules.
- “install” all external projects to the same staging directory under the build
  directory: simplify dependency lookups in external projects (including CMake).
- always use a CMake toolchain file: including when not cross-compiling, as way
  to pass parameters to CMake and simplify building CMake based external projects.
- use CMake to build external projects when possible (faster).
- use Ninja as default generator (faster).
- add missing `luajson` dependency, don't use luarocks for building/installing LUA
  external projects: faster builds, no need for network, all dependencies explicitly
  accounted for, simpler install tree (no need for the `rocks` directory anymore).
- use `-l` as as well as `-j` when running make / ninja to manage system load.
- rework libk2pdfopt / leptonica / tesseract situation: leptonica & tesseract are
  build by their respective external projects.
- changing an external project source file will automatically trigger its build step
  during the next build.
  

Build times (on my machine: -j8 / -j8 -l8, from scratch: after `make distclean`):

|                                  | master|      PR        |
-----------------------------------|------:|---------------:|
 **emulator-debug** (from scratch) | 11:05 | 4:30 `(÷ 2.5)` |
 **emulator-debug** (from cache)   |  4:04 | 0:38 `(÷ 6.4)` |
 **android-arm** (from scratch)    |  9:58 | 4:27 `(÷ 2.2)` |
 **android-arm** (from cache)      |  3:44 | 0:34 `(÷ 6.6)` |
 **kindlepw2** (from scratch)      | 15:50 | 6:51 `(÷ 2.3)` |
 **kindlepw2** (from cache)        |  5:21 | 0:50 `(÷ 6.4)` |

Disk usage (downloads + build tree(s))

|                   | master             |      PR            |
--------------------|-------------------:|-------------------:|
 **emulator-debug** | 291M + 2.3G = 2.6G | 155M + 1.6G = 1.7G |
 **android-arm**    | 286M + 1.4G = 1.6G | 150M + 1.0G = 1.2G |
 **kindlepw2**      | 326M + 1.5G = 1.9G | 176M + 1.2G = 1.3G |

Notes:

- minimum supported CMake version 3.15 (already mandated by zsync2's cpr submodule).
- using make as a CMake generator is still supported (with some limitations: no proper build by-products support), but since meson does not support it, I think we might as well drop support for it.
- need this patch to the top koreader project:
```diff
 Makefile | 14 +++++++-------
 1 file changed, 7 insertions(+), 7 deletions(-)

diff --git i/Makefile w/Makefile
index e3092b174..bdd6bfcaf 100644
--- i/Makefile
+++ w/Makefile
@@ -77,7 +77,7 @@ DOCKER:=$(shell grep -q docker /proc/1/cgroup 2>/dev/null && echo 1)
 INSTALL_FILES=reader.lua setupkoenv.lua frontend resources defaults.lua datastorage.lua \
 		l10n tools README.md COPYING
 
-all: $(if $(ANDROID),,$(KOR_BASE)/$(OUTPUT_DIR)/luajit)
+all:
 	$(MAKE) -C $(KOR_BASE)
 	install -d $(INSTALL_DIR)/koreader
 	rm -f $(INSTALL_DIR)/koreader/git-rev; echo "$(VERSION)" > $(INSTALL_DIR)/koreader/git-rev
@@ -85,12 +85,12 @@ ifdef ANDROID
 	rm -f android-fdroid-version; echo -e "$(ANDROID_NAME)\n$(ANDROID_VERSION)" > koreader-android-fdroid-latest
 endif
 ifeq ($(IS_RELEASE),1)
-	$(RCP) -fL $(KOR_BASE)/$(OUTPUT_DIR)/. $(INSTALL_DIR)/koreader/.
+	bash -O extglob -c '$(RCP) -fL $(KOR_BASE)/$(OUTPUT_DIR)/!(cache|cmake|history|staging|thirdparty) $(INSTALL_DIR)/koreader/'
 else
 	cp -f $(KOR_BASE)/ev_replay.py $(INSTALL_DIR)/koreader/
 	@echo "[*] create symlink instead of copying files in development mode"
 	cd $(INSTALL_DIR)/koreader && \
-		bash -O extglob -c "ln -sf ../../$(KOR_BASE)/$(OUTPUT_DIR)/!(cache|history) ."
+		bash -O extglob -c 'ln -sf ../../$(KOR_BASE)/$(OUTPUT_DIR)/!(cache|cmake|history|staging|thirdparty) .'
 	@echo "[*] install front spec only for the emulator"
 	cd $(INSTALL_DIR)/koreader/spec && test -e front || \
 		ln -sf ../../../../spec ./front
@@ -130,9 +130,6 @@ ifeq ($(IS_RELEASE),1)
 	rm -rf $(INSTALL_DIR)/koreader/data/{cr3.ini,cr3skin-format.txt,desktop,devices,manual}
 endif
 
-$(KOR_BASE)/$(OUTPUT_DIR)/luajit:
-	$(MAKE) -C $(KOR_BASE)
-
 $(INSTALL_DIR)/koreader/.busted: .busted
 	ln -sf ../../.busted $(INSTALL_DIR)/koreader
 
@@ -143,7 +140,9 @@ $(INSTALL_DIR)/koreader/.luacov:
 testfront: $(INSTALL_DIR)/koreader/.busted
 	# sdr files may have unexpected impact on unit testing
 	-rm -rf spec/unit/data/*.sdr
-	cd $(INSTALL_DIR)/koreader && ./luajit $(shell which busted) \
+	eval "$$($(LUAROCKS_BINARY) path)" && cd $(INSTALL_DIR)/koreader && \
+		env TESSDATA_DIR=$(INSTALL_DIR)/koreader/data \
+		./luajit "$$(which busted)" \
 		--sort-files \
 		--output=gtest \
 		--exclude-tags=notest $(BUSTED_OVERRIDES) $(BUSTED_SPEC_FILE)
@@ -629,4 +628,5 @@ static-check:
 doc:
 	make -C doc
 
+.NOTPARALLEL:
 .PHONY: all clean doc test update
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1750)


benoit-pierre:
For libk2pdfopt, I think the external project should switch to the official source archive + an overlay directory (similarly to how lunarsvg is handled). That's how I handle it on my meson branch too.

benoit-pierre:
Right, emulator builds don't go through docker, so meson is missing…

Frenzie:
fixup commits aren't supported by GH rebase btw (unless they added support in the last few months), so for that a force push will be required.

poire-z:
Not competent and experienced enough to judge.
I just hit <kbd>n</kbd>180 times, and it didn't look to scary (to read, but real scary to write and test).

Some comments and questions:

40b323b614a leptonica: are these new patches ? Taken from libkopt2pdf project? Is there some history?

934627bdcab zlib: drop of patches intentional?

(I've seen here and there some added comments that sounded (wording, vocabulary) like @NiLuJe 's :) I've seen some of these removed from other commits. You did just moved things around? Or are you beginning to sound like @NiLuJe ? :))

About using tarball and specifying `URL_MD5` : will we need to download the tarball and md5sum it ourselves?

733bbbba9 `target_include_directories(crengine PUBLIC` vs. `PRIVATE)` Why/when to use PUBLIC and when PRIVATE ?

8ff389866d8 `add soname to our helper libraries`. Just curious, what does this bring ? (I guess this won't get us libkoreader-cre.so.0, it just adds some string in the lib itself.)

ae753438c1e 5919035c any reason for the reordering of $(LIBXYZ), or just for balance of the text?

ae753438c1e some ellipsis `...`, dunno if we want to keep all that ASCII

043d3c98cf So, what's our official policy ? Do we love CMake or not ? :)
(At least I can understand a bit of Make - CMake has a lot of new vocabulary and needs us to read docs. But anyway, not caring, I'll continue to just copy and paste from existing stuff :))
Having to list all the headers in the BYPRODUCTS stuff feels ugly and like a step back ?

84995247d4 `libiconv: only compile on Android & macOS. On other platforms, rely on glibc' support.`
Just highlighting this to be sure @Frenzie @NiLuJe agree.

669045d27a So, this moves _some_ bits to (and requires?) meson.
Just because these projects propose building with meson?
Or is it a first step to a better meson-only world?

And asking again https://github.com/koreader/koreader-base/pull/1697#issuecomment-1811365720 : 
> > Also, why, while happy with your meson branch, would you spend that much time in CMake hell, fixing/improving/tidying all that ?!

> Morbid curiosity? Refreshing my make-fu, confirming my feelings on CMake (bleh).

Why all the pain of this PR?
Will you keep having your own full-meson branch?
Is this PR just a first step into going full-meson?
Are you convinced we should here stick to CMake?
Are you convinced you won't be able to convince us to switch to meson?
What's the future? :)

> For libk2pdfopt, I think the external project should switch to the official source archive + an overlay directory (similarly to how lunarsvg is handled). That's how I handle it on my meson branch too.

I must say it's quite painful to update LunaSVG. I have a local git clone of the upstream repo (not pushed anywhere) where I keep my history of changes, and just upload the extended.patch here. I'm not certain it's the best way to handle that.

Frenzie:
> [8499524](https://github.com/koreader/koreader-base/commit/84995247d4fa92a2b57200abd47758cb8a13c13b) `libiconv: only compile on Android & macOS. On other platforms, rely on glibc' support.`
> Just highlighting this to be sure @Frenzie @NiLuJe agree.

I believe that's simply a port of the current behavior, not a behavioral change.

> https://github.com/koreader/koreader-base/commit/669045d27aea18c15bf2b2a900e7a4bf193d776b So, this moves some bits to (and requires?) meson.
Just because these projects propose building with meson?
Or is it a first step to a better meson-only world?

The projects themselves have introduced meson. (I didn't double check but I know at least some of them did.)

benoit-pierre:
> [40b323b](https://github.com/koreader/koreader-base/commit/40b323b614a2ab446cab25207a3b54c6ed8978d3) leptonica: are these new patches ? Taken from libkopt2pdf project? Is there some history?

I just extracted the changes to leptonica from the version of k2pdfopt being used. No relevant changes to tesseract (only patches for old Windows versions, e.g. XP, or debug traces).

> [934627b](https://github.com/koreader/koreader-base/commit/934627bdcaba34e3d7d39e5ad19926df461e8dd3) zlib: drop of patches intentional?

Yes: those are autotools specific.

> (I've seen here and there some added comments that sounded (wording, vocabulary) like @NiLuJe 's :) I've seen some of these removed from other commits. You did just moved things around?

Some.

> Or are you beginning to sound like @NiLuJe ? :))

Probably for the same reasons: PTSD from to much CMake :P
 
> About using tarball and specifying `URL_MD5` : will we need to download the tarball and md5sum it ourselves?

You have to fill-in the md5sum manually, yes. Note however that it's optional, so you can:
- update the version
- comment the `URL_MD5` argument
- re-download (e.g.: `make zlib-{clean,download}`)
- and then update the MD5
 
> [733bbbb](https://github.com/koreader/koreader-base/commit/733bbbba90a2648f270b42d69e7f142bbb235887) `target_include_directories(crengine PUBLIC` vs. `PRIVATE)` Why/when to use PUBLIC and when PRIVATE ?

- `PRIVATE`: only used when compiling the library
- `PUBLIC`: also needed by users of said library

> [8ff3898](https://github.com/koreader/koreader-base/commit/8ff389866d83c3bb77d4c74895f4c40f27a2ce41) `add soname to our helper libraries`. Just curious, what does this bring ? (I guess this won't get us libkoreader-cre.so.0, it just adds some string in the lib itself.)

CMake automatically adds it, to this was done to reduce differences with the final version (produce the same binaries as much as possible to avoid regressions, or at least the same `readelf -n` output).

> [ae75343](https://github.com/koreader/koreader-base/commit/ae753438c1e38e3d67e0f590a355a2177731c9aa)[5919035](https://github.com/koreader/koreader-base/commit/5919035c8bdb503caeacf0235d7e44a137762bdd) any reason for the reordering of $(LIBXYZ), or just for balance of the text?

Same as above.

> [ae75343](https://github.com/koreader/koreader-base/commit/ae753438c1e38e3d67e0f590a355a2177731c9aa) some ellipsis `...`, dunno if we want to keep all that ASCII

Wrong commit?

> [043d3c9](https://github.com/koreader/koreader-base/commit/043d3c98cf4b9ce45297bde0001622b9964373fa) So, what's our official policy ? Do we love CMake or not ? :)

I certainly don't, but it still better than make or autotools (for some definition of better).

> (At least I can understand a bit of Make - CMake has a lot of new vocabulary and needs us to read docs. But anyway, not caring, I'll continue to just copy and paste from existing stuff :))

Yeah, it can be hell. But to be fair, the current build system is not doing itself any favor by installing each external project to a dedicated directory, instead of a common staging directory that can be more easily searched by CMake. That simplify things a lot. Even if you still have to contend with the numerous ways a CMake project has to find particular dependency (and the potential differences between CMake versions, as in how the initial call to `cmake_minimum_required(…)` can influence things).

benoit-pierre:
> > [8499524](https://github.com/koreader/koreader-base/commit/84995247d4fa92a2b57200abd47758cb8a13c13b) `libiconv: only compile on Android & macOS. On other platforms, rely on glibc' support.`
> > Just highlighting this to be sure @Frenzie @NiLuJe agree.
> 
> I believe that's simply a port of the current behavior, not a behavioral change.

No, it's a change. Same with not compiling gettext, and then relying on `proxy-libintl` instead. On my meson branch, I always use `proxy-libintl` (because we don't need i18n support in sdcv). I also switched from `libiconv` to `miniconv`, as FWIU, we don't really need full iconv support, if even, with the patches to sdcv (UTF-8 is always used).

> > [669045d](https://github.com/koreader/koreader-base/commit/669045d27aea18c15bf2b2a900e7a4bf193d776b) So, this moves some bits to (and requires?) meson.
> > Just because these projects propose building with meson?
> > Or is it a first step to a better meson-only world?
> 
> The projects themselves have introduced meson. (I didn't double check but I know at least some of them did.)

Yes, meson is the default build system of those projects.

benoit-pierre:
> > For libk2pdfopt, I think the external project should switch to the official source archive + an overlay directory (similarly to how lunarsvg is handled). That's how I handle it on my meson branch too.
> 
> I must say it's quite painful to update LunaSVG. I have a local git clone of the upstream repo (not pushed anywhere) where I keep my history of changes, and just upload the extended.patch here. I'm not certain it's the best way to handle that.

Right, in lunasvg's case, there's some patching to the source involved, but not for libk2pdfopt: the overlay is independent. Of course you still need to update the leptonica / tesseract patches (if any), and the overlay code itself.

benoit-pierre:
> Having to list all the headers in the BYPRODUCTS stuff feels ugly and like a step back ?

I don't like it either, but it's required for proper dependencies handling (particularly on incremental parallel builds).

And there's no dependency file equivalent for by-products.

Initially, I had setup every external project to always build (see `BUILD_ALWAYS` [documentation](https://cmake.org/cmake/help/latest/module/ExternalProject.html#build-step-options)). This was done so changes to an external project sources would get picked up automatically. But this resulted in a slow no-change `make` (about ~2 seconds): because with `BUILD_ALWAYS`, the build and external step of each external project is always executed, and the build steps are never run in parallel (they are the ninja console pool: max 1 job).

But thanks to the sources depfile trick, there's no need for `BUILD_ALWAYS`, a build step will automatically be triggered on source change.

And there might be a way to avoid the need for using by-products by moving the crengine library, koreader libraries, and miscellaneous executables builds to dedicated external projects. I need to investigate.

benoit-pierre:
I also need someone with access to macOS to test. I checked the resulting executable/libraries have the right runtime path / soname information, and that's it. I think all the `install_name` calls in `do_mac_bundle.sh` are not necessary anymore. Same with building the `gettext` and `libsodium` libraries from brew.

poire-z:
> > [ae75343](https://github.com/koreader/koreader-base/commit/ae753438c1e38e3d67e0f590a355a2177731c9aa) some ellipsis `...`, dunno if we want to keep all that ASCII
> 
> Wrong commit?

Indeed :/ Can't find it, and not looking again at these 180 commits :)
Anyway, there are four new `…` and 1 already there moved around, probably from some @NiLuJe comment. So, well, not minding, just mentionning it in case someone else cares.

Frenzie:
> No, it's a change. Same with not compiling gettext, and then relying on proxy-libintl instead. On my meson branch, I always use proxy-libintl (because we don't need i18n support in sdcv). I also switched from libiconv to miniconv, as FWIU, we don't really need full iconv support, if even, with the patches to sdcv (UTF-8 is always used).

Once upon a time we had a GLIB (without libiconv and gettext, except on MacOS) and a GLIB_STATIC (for Android, with libiconv). If that's no longer the case it would require investigation into why that was changed. At a glance looks like the relevant conditions were dropped in https://github.com/koreader/koreader-base/pull/1690, meaning that was the change in behavior, not this, imho.

Frenzie:
> * update docker images (chrpath+meson)

I assume that for chrpath anything will do. That is, it's been at 0.16 for years, unless the new [0.17](https://people.skolelinux.org/pere/blog/New_chrpath_release_0_17.html) offers something we need. But is the same true for Meson?

benoit-pierre:
Those projects that use meson are usually pretty conservative (same with [WrapDB](https://mesonbuild.com/Wrapdb-projects.html)):
```bash
▹ grep meson_version build/x86_64-pc-linux-gnu-debug/cmake/*/source/meson.build
build/x86_64-pc-linux-gnu-debug/cmake/fribidi/source/meson.build
2:  meson_version : '>= 0.54')

build/x86_64-pc-linux-gnu-debug/cmake/freetype2/source/meson.build
27:  meson_version: '>= 0.55.0',

build/x86_64-pc-linux-gnu-debug/cmake/harfbuzz/source/meson.build
2:  meson_version: '>= 0.55.0',

build/x86_64-pc-linux-gnu-debug/cmake/proxy-libintl/source/meson.build
3:  meson_version : '>= 0.46.0',

build/x86_64-pc-linux-gnu-debug/cmake/glib/source/meson.build
3:  meson_version : '>= 0.47.0',
```

benoit-pierre:
> I assume that for chrpath anything will do.

With that last commit, it's actually not needed anymore: no need to list all build by-products for each external projects anymore, each project actual install rules are used. I moved the crengine build to a dedicated external project. Same with the koreader libraries / executable. (But I kept support for things like `make koreader-cre` to only build `libkoreader-cre` and its depedencies).

chrpath was only needed to cleanup meson build rpath before manually "installing" to staging, that's unnecessary now that meson install rules are used.

I'll have to check if the macOS runtime path is still correct however: from past experience with my meson branch, custom rpath entries are stripped on install.

Frenzie:
> pretty conservative

Not conservative enough for the Ubuntu 20.04 repos. :-) But perhaps the time is right to switch to 22.04, if Android wants to participate.

benoit-pierre:
We can always install from pip, even if that will not always give us the last version (e.g. the kindle image uses Python 3.6, so meson 0.61.5 is the latest supported version).

Frenzie:
Of course we can, but I'd like to avoid that if we don't have to.

benoit-pierre:
What about the appimage? What's the oldest target?

Frenzie:
> (e.g. the kindle image uses Python 3.6, so meson 0.61.5 is the latest supported version).

This is close to the same statement ftr. I want them on 20.04 (or maybe 22.04) but the Android build mysteriously crashes in CI. We have yet to test if the update from a while back solves that though. The 0.3.0 image (with 20.04) has been available for immediate use for all but Android for a year. ^_^

benoit-pierre:
OK. One thing I'd like to PR, is the testsuite improvements I made on branch: isolating tests, and then using meson as a test runner (each spec file run is isolated to its own process, tests run in parallel).

It's much faster, and you don't have to work around the lack of real isolation in busted.

Frenzie:
> What about the appimage? What's the oldest target?

That's basically what I was referring to above by saying maybe the time has come for 22.04. For anything else it really doesn't matter all that much, but I consider lower versions (when reasonable) more advantageous since they ensure it's easy for people to start hacking. Keeping them in line with the AppImage is two birds one stone.

benoit-pierre:
Ubuntu 20.04 has meson 0.53.2, 22.04: 0.61.2.

Frenzie:
Right, but FreeType would like 0.55.

benoit-pierre:
Yep, so update to 22.04, or pip install. ¯\\_(ツ)_/¯

benoit-pierre:
What do you not like about using pip?

Frenzie:
Yes, that's why I said what I said in https://github.com/koreader/koreader-base/pull/1750#issuecomment-2029440477 :laughing: 

But note that the Android build didn't work in 22.04 either,[1] or it'd have been on that already years ago. Unfortunately there's a chance it'll require a pretty big time investment to find some kind of solution that works. The most straightforward being not to use any image at all, though that has the annoying side effect that GitLab CI, CircleCI, GHA, etc. would all require something slightly custom.

[1] In CI. The build works fine locally using those same Docker images.

Frenzie:
> What do you not like about using pip?

Unless you're dealing with a combination of really wanting to stay with 20.04 and really wanting Meson >=0.55, which I don't believe we are, why would you want to add an additional complication?

Frenzie:
> With regard to developing, I think it makes sense to mandate the most recent build tools as possible (toolchains, cmake, meson).

That instantaneously turns away almost everyone. That's a very explicit anti-goal to me.

> And using pip can help ensure the same meson version is used for all CI jobs (including macOS), without the need to deal with separate package managers (like brew…).

A not uninteresting point, although ultimately I don't think it relates to much of anything given macOS's status aparte. Simply put, it's easily possible that using pip on macOS is the most dependable solution with the least complications because it's not remotely comparable to the Debian ecosystem. Brew is just a thing, it's not a feature.

Frenzie:
> I also need someone with access to macOS to test. I checked the resulting executable/libraries have the right runtime path / soname information, and that's it. I think all the `install_name` calls in `do_mac_bundle.sh` are not necessary anymore. Same with building the `gettext` and `libsodium` libraries from brew.

I cranked up the ole Macbook Air M1 and the OpenSSL build isn't succeeding on your changes, complaining about `SOURCE_DATE_EPOCH` not being an environment variable or something. This doesn't seem to be an issue on current master, but it's possible I did something wrong.

I'm having some issues with Synergy 3 (probably Mac-exclusive; it connects between various Linux machines), so here's a literal screenshot.

benoit-pierre:
Yeah, looks like `util/mkbuildinf.pl` automatically add the " UTC" part:
```perl
my $date = gmtime($ENV{'SOURCE_DATE_EPOCH'} || time()) . " UTC";
```

poire-z:
> With that last commit, it's actually not needed anymore: no need to list all build by-products for each external projects anymore

Hard to see what changed since I last looked. Do you mean that my worry  `Having to list all the headers in the BYPRODUCTS stuff feels ugly and like a step back ?` is obsolete ? (I don't see the list of .h by quickly browsing the full diff - but I don't see their removal while looking at your last two "force pushes" diff links).

benoit-pierre:
> > With that last commit, it's actually not needed anymore: no need to list all build by-products for each external projects anymore
> 
> Hard to see what changed since I last looked. Do you mean that my worry `Having to list all the headers in the BYPRODUCTS stuff feels ugly and like a step back ?` is obsolete ? 

Yes.

benoit-pierre:
New libk2pdfopt proposal:
- a dedicated [`upstream` branch](https://github.com/benoit-pierre/libk2pdfopt/commits/upstream/):
  - no history in common with master.
  - made by importing k2pdfopt release zips (easily done with a [small script](https://github.com/benoit-pierre/libk2pdfopt/blob/pr/new_build/scripts/update_upstream.sh)).
- in my proposed [update to master](https://github.com/koreader/libk2pdfopt/compare/master...benoit-pierre:libk2pdfopt:pr/new_build):
  - I started by [a zero-changes merge of the right upstream commit](https://github.com/benoit-pierre/libk2pdfopt/commit/b7bc9490abe414bc8e6f6274752f7ddcafcd35c6), so updating to the latest k2pdfopt version is done by merging upstream again.
  - koreader code has been moved to lib (reduce conflicts with upstream), with a CMake solution for building, leptonica & tesseract are not built by the project anymore.
  - another [small script](https://github.com/benoit-pierre/libk2pdfopt/blob/pr/new_build/scripts/generate_upstream_patch.sh) can be used to create updated patches for leptonica / tesseract / … when updating to a new k2pdfopt version.

Thoughts?

Frenzie:
I don't like the upstream branch, or perhaps more to the point I don't like that it's *completely useless* and fattening (i e., zip files).

Maintaining patches without git sounds significantly harder and less desirable to me than maintaining them as a few git commits with git rebase. (Emphasis on a few.)

Moving code to a lib reducing conflicts sounds good of course.

benoit-pierre:
> I don't like the upstream branch, or perhaps more to the point I don't like that it's _completely useless_ and fattening (i e., zip files).

The content of the zip files is imported, not the zip themselves. There's very little fattening. Since there's no official upstream to fetch from, we create it ourselves from the zips.  It's a smarter way to handle updates, taking advantage of git. Nicer history/blame too.

> Maintaining patches without git sounds significantly harder and less desirable to me than maintaining them as a few git commits with git rebase. (Emphasis on a few.)

Which patches are we talking about? The libk2pdfopt repo is still used. This is about the changes to leptonica & tesseract: which right now, consist of:
- leptonica: overwriting `dewarp2.c` with the version from k2pdopt
- tesseract: overwriting `tessdatamanager.cpp` with the version from k2pdfopt, and applying a totally unnecessary 
extra patch (`baseapi.cpp.patch`, already part of the tesseract version used)

IMHO, using patches is nicer, making it possible to do minor version updates.

And being able to easily generate those patches is also a nicer way to find out what is modified in each version of k2pdfopt (and ignore most of it, like the win32 related changes).

Frenzie:
I wrote that on my phone; as it turns out the only problem is that your auto-generated commit messages were very deceptive. ;-)
image

Frenzie:
> IMHO, using patches is nicer, making it possible to do minor version updates.

Could you be a bit more explicit? Where are these "patches"? The word suggests something I find incredibly undesirable and annoying.

Frenzie:
> And being able to easily generate those patches is also a nicer way to find out what is modified in each version of k2pdfopt (and ignore most of it, like the win32 related changes).

To be clear, what I want is a git branch with commits ("patches") on top of said branch that can be rebased and that actually tell a story. Not some kind of disconnected thing that's a total pain making life incredibly difficult, which is what's implied by the word patches.

benoit-pierre:
Look at the 2 `k2pdfopt.patch` for leptonica & tesseract in this PR.

Frenzie:
And that's somehow auto-generated?

benoit-pierre:
Yes, for example to find out what are the changes made to leptonica in k2pdfopt 2.42:
```bash
▹ ./scripts/generate_upstream_patch.sh ~/dev/leptonica 979224f | diffstat
 allheaders.h |    3 +++
 dewarp2.c    |  111 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-------------
 2 files changed, 101 insertions(+), 13 deletions(-)
```

Frenzie:
Yes, that sounds good.

benoit-pierre:
And to update to a new k2pdfopt version:
```bash
# update upstream
▹ ./scripts/update_upstream.sh k2pdfopt_v2.55_src.zip
# merge
▹ git merge upstream
```
And then use `./scripts/generate_upstream_patch.sh` to find out what patch, if any, need to be applied to leptonica, tesseract, mupdf, …

Frenzie:
For the commit message, something like this to strip the .zip?

```
$(archive_base=${archive##*/}; echo ${archive_base%.*})
```

In the other file, I'd probably make:
```
diff --ignore-trailing-space --unified --label "a/${candidates}" "${checkout}/${candidates}" --label "b/${candidates}" <(git show "${upstream}:${file}") || [[ $? -eq 1 ]]
```
into `${candidates[0]}` so it's clear what'll happen to people less well-versed in Bash. (I see there's some counting logic above it but still.)

benoit-pierre:
OK, done.

benoit-pierre:
So how do we go about updating lik2pdfopt? I can't PR the `upstream` branch (no common history), and for the other changes, it's important to preserve the merge of `upsteam`. IMHO the easiest way would be for someone with commit rights to fetch from my fork, push `upstream`, and fast forward `master` from my `pr/new_build` branch and push it.

Frenzie:
Done.

benoit-pierre:
Should be ready now: here's the [diff or recent changes](https://github.com/koreader/koreader-base/compare/09881b8085753fbbbf761c686e312729dc0ee0e5..21e1a499ac6ebcd72980b4c11c8e41ff77aff6fa).

I've removed the meson support for now, so it's everything up-to and including the switch to CMake as the main driver, plus some tweaks to the build progress output.

To recap:
- support CMake `>= 3.15`
- crengine & koreader are handled as external projects, always "built" (so source changes are picked up)
- for other external projects, source changes should trigger the build / install steps:
   - when using the ninja generator
   - when using the make generator and CMake `>= 3.20`
- external projects are installed to a staging directory (`$(OUTPUT_DIR)/staging`)
- change to patch files don't trigger a reconfiguration; rational: it's buggy, since the source directory is not reset to a pristine state, it may not work depending on the patch, and convenience targets can be used to easily rebuild a project (e.g. `make libpng-re`)
- like with the old build system, it's a good idea to do a whole rebuild on `thirdparty/xxxx/*` change(s) (`make re`)
- convenience targets:
  - will be serialized, build & clean targets can be combined
  - `re`: equivalent to `clean all`
  - `setup`: just run the main CMake project configure phase
  - `download`: execute `download` step of all external projects for the target platform
  - `download-all`: execute `download` step of **all** external projects (enabled of not for the target platform)
  - per external project targets:
    - `PROJECT_NAME`, `PROJECT_NAME-clean`, `PROJECT_NAME-re`: build & install, clean, clean + build + install
    - `PROJECT_NAME-download`, `PROJECT_NAME-patch`, `PROJECT_NAME-configure`, `PROJECT_NAME-build`, `PROJECT_NAME-install`: self explanatory
    - `PROJECT_NAME-deps`: build dependencies of `PROJECT_NAME`

Note:
- all projects can be explicitly built (or attempted to), no matter the target platform (e.g. `make fbink` when building the emulator)
- a "no changes" build when using the make generator is much slower (up to ~1s, particularly with CMake `>= 3.20`) then with ninja (~0.2s)

poire-z:
(Not competent and experienced enough to judge - so not spending the time like I did 2 months ago at the start of this PR, trusting you. I'll probably complain as usual on my first rebuild after this is merged :))

NiLuJe:
I'll give it a once over Soon(TM) (seriously, sometimes during the week soon ;p). Thanks ;).

Frenzie (approved):
I'll wait to merge until after @NiLuJe has had some time to look at it in more detail as well, but it's very good in my eyes. The nitpick-looking remarks I made [here](https://github.com/koreader/koreader-base/pull/1750#issuecomment-2120812755) are intended more as open questions or musings than nitpicks per se.

benoit-pierre:
> That's a very impressive piece of work.
> 
> Regarding "djvulibre: drop unused libjpeg dependency", is that an implicit disagreement with what I wrote in [#1748 (comment)](https://github.com/koreader/koreader-base/pull/1748#issuecomment-2016967634)?
> 
> > if it's indeed unused my initial reaction is that precisely the reverse should be done.

Yes, and addressed in 0b717e3 (one of the last commits).

> Regarding "switch to the source distribution," for some particular ones like DjvuLibre and MuPDF I think that will make life harder. It's currently very simple to checkout different tags or branches, create a custom branch with some changes, stash them, etc. But I suppose one might effectively revert [5a9631e](https://github.com/koreader/koreader-base/commit/5a9631ee780252132b381c1b6e3f169757ce9614) for development purposes, or something. I'm not very enthusiastic about it, in any case. I'll grant the ./autogen.sh time savings are a decent argument though.

I don't know, it's still of limited use since the clone is temporary. I did take a look at switching to a worktree, but that approach is not really compatible with submodules.

When using an archive, you can `make mupdf-download` or `make mupdf-patch`, and then `git init build/…/thirdparty/mupdf/source` to work on an existing / new patch.

Frenzie:
> Yes, and addressed in [0b717e3](https://github.com/koreader/koreader-base/commit/0b717e33697a380e8fb43e28c3a7b953675a1fc0) (one of the last commits).

Ah, excellent. :-)

> I don't know, it's still of limited use since the clone is temporary.

It doesn't need to be permanent to be useful, though fwiw I suspect my temporary main one may well have existed for a decade. A quick check shows that's not true, and it's only been there since 2019. Apparently I removed it for some reason, perhaps because it was the quickest way to switch origins.

> When using an archive, you can `make mupdf-download` or `make mupdf-patch`, and then `git init build/…/thirdparty/mupdf/source` to work on an existing / new patch.

That only sounds like a minuscule subset of what's extremely simple with git.

benoit-pierre:
I forgot to mention one thing: [this commit](77fa2037dfcc519b56ab05c5b16810afe89a3238). I could never build the release version of luajit with `-ffast-math`:
```
CC        luajit.o
vm_x64.dasc:394: error: bad integer number `-8':
  |  and PC, -8
vm_x64.dasc:396: error: bad integer number `-8':
  |  lea RA, [RA+PC-8]                  // Rebase RA and prepend one result.
vm_x64.dasc:397: error: bad integer number `-8':
  |  mov PC, [BASE-8]                   // Fetch PC of previous frame.
vm_x64.dasc:403: error: bad integer number `1':
  |  add RDd, 1                         // RD = nresults+1
vm_x64.dasc:417: error: bad integer number `-8':
  |  and PC, -8
vm_x64.dasc:421: error: bad integer number `1':
  |  sub RDd, 1
vm_x64.dasc:423: error: bad integer number `1':
  |1:  // Move results down.
vm_x64.dasc:425: error: bad integer number `-16':
  |  mov [BASE-16], RB
vm_x64.dasc:426: error: bad integer number `8':
  |  add BASE, 8
vm_x64.dasc:427: error: bad integer number `1':
  |  sub RDd, 1
vm_x64.dasc:429: error: bad integer number `2':
  |2:
vm_x64.dasc:430: error: bad integer number `2':
  |  mov L:RB, SAVE_L
vm_x64.dasc:432: error: bad integer number `3':
  |3:
vm_x64.dasc:434: error: bad integer number `2':
  |  mov RAd, SAVE_NRES                 // RA = wanted nresults+1
vm_x64.dasc:435: error: bad integer number `4':
  |4:
vm_x64.dasc:438: error: bad integer number `5':
  |5:
vm_x64.dasc:439: error: bad integer number `16':
  |  sub BASE, 16
vm_x64.dasc:443: error: bad integer number `4':
  |  mov RA, SAVE_CFRAME                // Restore previous C frame.
vm_x64.dasc:448: error: bad integer number `5':
  |  restoreregs
  |    add rsp, CFRAME_SPACE    [MACRO restoreregs (0)]
vm_x64.dasc:451: error: bad integer number `6':
  |6:
vm_x64.dasc:*: warning: too many errors (suppressed further messages).
vm_x64.dasc:*: info: 1243 errors in input file -- no output file generated.
make[4]: *** [Makefile:665: host/buildvm_arch.h] Error 1
```
IMHO, using `-ffast-math` is a bad idea, and especially in the context of luajit, Mike Pall has [a few choice words](https://github.com/LuaJIT/LuaJIT/issues/1184#issuecomment-2065541658) about the option.

Incidentally, `-ffast-math` also breaks recent MuPDF versions.

Looking at a blame of `Makefile.defs`, the option has been there forever, with no clear explanations of the benefits.

We can either:
1. keep using it globally, and disable it selectively in the projects that break
2. or drop it altogether

I'm in favor of option 2.

NiLuJe:
I distinctly remember adding `-fno-fast-math` to the LuaJIT build years ago, FWIW.

So, yeah, it doesn't build with fast-math, I'm aware ;).

I did a bunch of extremely unscientific tests for a few key areas (again, years ago), and it was moderately helpful in some cases (and, more importantly, never harmful except for projects that explicitly know they don't work with it and either don't build or disable it).

So, I'm all for 1. ;).

NiLuJe:
Welp, apparently, that was SQLite ;D.

Still, that rings a bell for LuaJIT, so I must have done that... somewhere ;). And I'm definitely aware of Mike PAll's thoughts on the subject ;).

Frenzie:
> Looking at a blame of `Makefile.defs`, the option has been there forever, with no clear explanations of the benefits.

I suspect the rationale might come from Linaro recommending it (over a decade ago, possibly now but I don't particularly care to find out).

Though there also seems to be some @NiLuJe involvement ;-P https://github.com/koreader/kindlepdfviewer/pull/348

benoit-pierre:
There might also be a third option: weaken it, `-fno-finite-math-only -fno-unsafe-math-optimizations` is apparently enough for both LuaJIT and MuPDF. I'm still in favor of disabling it completely.

NiLuJe:
> I suspect the rationale might come from Linaro recommending it (over a decade ago, possibly now but I don't particularly care to find out).

I don't specifically remember anything about fast-math, FWIW. They *do* (did?) recommend `-O3`, though.

I definitely agree that it's not something one should enable blindly and everywhere (e.g., I don't use it in my base Gentoo CFLAGS).

In our case, we only build a very small subset of stuff though, and it felt pretty safe for *that* at the time.

Frenzie:
> I don't specifically remember anything about fast-math, FWIW. They _do_ (did?) recommend `-O3`, though.

Phrased more accurately, they recommend(ed?) `-Ofast`, which includes `-ffast-math`.

Frenzie:
Or at least that's what [some random guy on a forum](https://www.mobileread.com/forums/showthread.php?p=2240459#post2240459) claims. :angel:

NiLuJe:
Funnily enough, Clang is looking into dropping `Ofast` because of the terrible name and the fact that apparently some people don't understand that it implies `-ffast-math` ;o).

Frenzie:
To be fair, in a computing context "fast math" kind of implies "take shortcuts to be faster," though it could certainly also mean something like "use newer, faster math instructions."

benoit-pierre:
One possible way to check if `-ffast-math` is safe for a project is to check for use of `isnan`: MuPDF uses it, so does harfbuzz (in the subset code, which AFAIK we don't use), sqlite, tesseract, zsync2 maybe (`std::isnan` is used).

I still think it's more trouble than it's worth.

Frenzie:
Going once, going twice… *looks at @NiLuJe*

NiLuJe:
I mean, we haven't explicitly seen it break anything at runtime yet, but it's not a hill I'm willing to die on, so, whatever ;).

Frenzie:
True, I've never noticed any issues but of course it's stated that one of the issues is overloading the cache, which is one of those issues you probably won't notice even if you do notice them (since of course things are a bit slower on a poor little arm processor).

NiLuJe:
I believe the cache comment is mostly aimed at `-O3` and code size bloat (particularly because of more aggressive unrolling), which becomes a somewhat different beast altogether on armv7, because thumb mode instructions are literally smaller ;).

So, yes, on paper, again -O3 by default is not necessarily a great idea (depending on your target. Modern processors have *significantly* more cache and make much better use of it than in the past. But it's still the name of the game, I/O, at any level, is always going to be your bottleneck). But armv7 is a different beast on that front, thanks to the thumb instruction set.

The issue with `-ffast-math` is much simpler: it's one of accuracy (or even just pain correctness) with some types of fp maths ;).

Frenzie:
@NiLuJe annoying reminder ping

NiLuJe:
Might happen this week-end, I've had family visiting from Quebec, so it's been a bit more hectic than anticipated ;).

NiLuJe:
I'm slowly working through it, it looks super neat so far, kudos and thanks ;).

Frenzie:
@NiLuJe Please rebase and merge when ready.

(I note there have been some minor changes since I last checked, but I assume they're all positive. ;-)

NiLuJe:
> rebase and merge

Just to confirm, an actual rebase: we want the full history?

Frenzie:
Don't you?

NiLuJe:
Might I suggest a merge commit then, in order to have a clear "this is the merge" at the head of the numerous commits?

Frenzie:
If you prefer.

NiLuJe (commented):
> Reviewed 119 of 119 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1750)* status: all files reviewed, 6 unresolved discussions (waiting on @benoit-pierre)

___
*[`thirdparty/cmake_modules/koreader_targets.cmake` line 38 at r1](https://reviewable.io/reviews/koreader/koreader-base/1750#-NyvLRZu5F-HTQ3q4sZ-:-NyvLRZu5F-HTQ3q4sZ0:brr1eqs) ([raw file](https://github.com/koreader/koreader-base/blob/0eb92faa38f39c6bba27e2993239283bc0e47443/thirdparty/cmake_modules/koreader_targets.cmake#L38)):*
> ```CMake
>             set_target_properties(${NAME} PROPERTIES VISIBILITY_INLINES_HIDDEN TRUE)
>             set(VISIBILITY "hidden")
>         endif()
> ```

If I ignore the fact that it's unclear to me, even after reading the docs and the howto, whether `-fvisbility-inlines-hidden` does anything more when already using `-fvisbility=hidden`, one thing that *is* clear to me is that it's only meaningful for C++.

So, can we automate that somehow, e.g., if visibility is hidden and lang is cxx?

(Instead of having to manually flag CXX projects differently).



_Code quote:_
```CMake
        if(VISIBILITY STREQUAL "inline-hidden")
            set_target_properties(${NAME} PROPERTIES VISIBILITY_INLINES_HIDDEN TRUE)
            set(VISIBILITY "hidden")
        endif()
```
___
*[`thirdparty/cmake_modules/koreader_thirdparty_common.cmake` line 71 at r1](https://reviewable.io/reviews/koreader/koreader-base/1750#-NyvOlaI3gsj8wo2RXHX:-NyvOlaI3gsj8wo2RXHY:broigmw) ([raw file](https://github.com/koreader/koreader-base/blob/0eb92faa38f39c6bba27e2993239283bc0e47443/thirdparty/cmake_modules/koreader_thirdparty_common.cmake#L71)):*
> ```CMake
> function(append_install_commands CMD_LIST)
>     cmake_parse_arguments("" "" "DESTINATION" "" ${ARGN})
>     if(_DESTINATION)
> ```

Any reason why this one is not using `DEFINED` compared to `append_binary_install_command`?
___
*[`thirdparty/luasec/CMakeLists.txt` line 18 at r1](https://reviewable.io/reviews/koreader/koreader-base/1750#-Nyw2tDTBYTJFzwXEtrt:-Nyw2tDTBYTJFzwXEtru:b-rdovsz) ([raw file](https://github.com/koreader/koreader-base/blob/0eb92faa38f39c6bba27e2993239283bc0e47443/thirdparty/luasec/CMakeLists.txt#L18)):*
> ```CMake
>     CC=${CC} CFLAGS=${CFLAGS}
>     LDFLAGS=${DYNLIB_LDFLAGS}
>     LIBS=${LIBS} EXTRA=${LIBS}
> ```

Couldn't this use `append_autotools_vars`?



_Code quote:_
```CMake
set(MAKE_CMD
    ${KO_MAKE_RECURSIVE}
    AR=${AR} RANLIB=${RANLIB}
    CC=${CC} CFLAGS=${CFLAGS}
    LDFLAGS=${DYNLIB_LDFLAGS}
    LIBS=${LIBS} EXTRA=${LIBS}
```
___
*[`thirdparty/mupdf/shared-libmupdf.patch` line 9 at r1](https://reviewable.io/reviews/koreader/koreader-base/1750#-Nyw5ezI0ZtCBG3GxL9n:-Nyw5ezI0ZtCBG3GxL9o:b-vah331) ([raw file](https://github.com/koreader/koreader-base/blob/0eb92faa38f39c6bba27e2993239283bc0e47443/thirdparty/mupdf/shared-libmupdf.patch#L9)):*
> ```Diff
> +MUPDF_SHLIB = $(OUT)/libmupdf.so
> +$(MUPDF_SHLIB): $(MUPDF_LIB) $(THIRD_LIB) $(LIBS)
> +	$(QUIET_LINK) $(CC) $(LINK_OPTS) $(LIBS) -o $@
> ```

Does this include `$(CFLAGS)` (or whatever that makefile uses for that)?

(My line of thinking being having CFLAGS set when linking for LTO).
___
*[`thirdparty/openssl/CMakeLists.txt` line 92 at r1](https://reviewable.io/reviews/koreader/koreader-base/1750#-Nyw6ILJ-qYidwpC4GM7:-Nyw6ILJ-qYidwpC4GM8:b6rocy0) ([raw file](https://github.com/koreader/koreader-base/blob/0eb92faa38f39c6bba27e2993239283bc0e47443/thirdparty/openssl/CMakeLists.txt#L92)):*
> ```CMake
>     # Reproducible builds.
>     SOURCE_DATE_EPOCH=1694383200
> )
> ```

Same comment as in luasec about `append_autotools_var`, at least for parts of this ;).



_Code quote:_
```CMake
set(MAKE_CMD
    ${KO_MAKE_RECURSIVE}
    AR=${AR}
    CC=${CC}
    CXX=${CXX}
    LD=${LD}
    MAKEDEPEND=${CC}
    RANLIB=${RANLIB}
    # Reproducible builds.
    SOURCE_DATE_EPOCH=1694383200
)
```
___
*[`thirdparty/zsync2/CMakeLists.txt` line 9 at r1](https://reviewable.io/reviews/koreader/koreader-base/1750#-Nyw8zQ_DRvJ6fl-9No2:-Nyw8zQa8UL6w3GZtpmR:b1lai8m) ([raw file](https://github.com/koreader/koreader-base/blob/0eb92faa38f39c6bba27e2993239283bc0e47443/thirdparty/zsync2/CMakeLists.txt#L9)):*
> ```CMake
>     ${CMAKE_CURRENT_SOURCE_DIR}/cpr_need_pthread.patch
>     # Need another zlib (beside the internal patched one) for compiling the client.
>     ${CMAKE_CURRENT_SOURCE_DIR}/need_zlib.patch
> ```

We control this fork, so feel free to PR the two non-cpr patches (and ping me, because I'm not sure I get email notifications from new PRs/issues, which is how I missed the FBInk one ;p).



_Code quote:_
```CMake
    # Reproducible builds (don't set BUILD_DATE to configure time).
    ${CMAKE_CURRENT_SOURCE_DIR}/reproducible-builds.patch
    # CPR need `-pthread`.
    ${CMAKE_CURRENT_SOURCE_DIR}/cpr_need_pthread.patch
    # Need another zlib (beside the internal patched one) for compiling the client.
    ${CMAKE_CURRENT_SOURCE_DIR}/need_zlib.patch
```

yparitcher:
@benoit-pierre @NiLuJe 
This Pr is removing use / support of luarocks.

I have a custom [plugin](https://github.com/yparitcher/zmanim.koplugin), that relies on a C library. Until now i have been using luarocks (i made a [rockspec](https://luarocks.org/modules/yparitcher/libzmanim/1.0-1)) to allow easy support for cross compiling and installing (just point luarocks to the koreader rocks tree).

This looks to be going away, so i will have to move my `.so` to `common/` which i can do manually for now, however how do you recommend injecting a 3rd party lib into koreader without patching the build system?

Asking as i have a few users of said plugin whom until now did not need a cross compiler as i have uploaded a kindle build to luarocks so they can install directly. This does not appear to be as easy anymore.

Don't let this hold off progress, but as always the bug is always somebody else's feature ;)

benoit-pierre:
> We control this fork, so feel free to PR the two non-cpr patches (and ping me, because I'm not sure I get email notifications from new PRs/issues, which is how I missed the FBInk one ;p).

@NiLuJe: ping!

- https://github.com/NiLuJe/zsync2/pull/2
- https://github.com/NiLuJe/zsync2/pull/3
- https://github.com/NiLuJe/zsync2/pull/4

benoit-pierre:
> _[`thirdparty/cmake_modules/koreader_targets.cmake` line 38 at r1](https://reviewable.io/reviews/koreader/koreader-base/1750#-NyvLRZu5F-HTQ3q4sZ-:-NyvLRZu5F-HTQ3q4sZ0:brr1eqs) ([raw file](https://github.com/koreader/koreader-base/blob/0eb92faa38f39c6bba27e2993239283bc0e47443/thirdparty/cmake_modules/koreader_targets.cmake#L38)):_
> 
> > ```cmake
> >             set_target_properties(${NAME} PROPERTIES VISIBILITY_INLINES_HIDDEN TRUE)
> >             set(VISIBILITY "hidden")
> >         endif()
> > ```
> 
> If I ignore the fact that it's unclear to me, even after reading the docs and the howto, whether `-fvisbility-inlines-hidden` does anything more when already using `-fvisbility=hidden`, one thing that _is_ clear to me is that it's only meaningful for C++.
> 
> So, can we automate that somehow, e.g., if visibility is hidden and lang is cxx?
> 
> (Instead of having to manually flag CXX projects differently).

Right, no point in complicating things if it's not required. Will amend.

benoit-pierre:
> _[`thirdparty/mupdf/shared-libmupdf.patch` line 9 at r1](https://reviewable.io/reviews/koreader/koreader-base/1750#-Nyw5ezI0ZtCBG3GxL9n:-Nyw5ezI0ZtCBG3GxL9o:b-vah331) ([raw file](https://github.com/koreader/koreader-base/blob/0eb92faa38f39c6bba27e2993239283bc0e47443/thirdparty/mupdf/shared-libmupdf.patch#L9)):_
> 
> > ```diff
> > +MUPDF_SHLIB = $(OUT)/libmupdf.so
> > +$(MUPDF_SHLIB): $(MUPDF_LIB) $(THIRD_LIB) $(LIBS)
> > +	$(QUIET_LINK) $(CC) $(LINK_OPTS) $(LIBS) -o $@
> > ```
> 
> Does this include `$(CFLAGS)` (or whatever that makefile uses for that)?
> 
> (My line of thinking being having CFLAGS set when linking for LTO).

It does not! Fixed.

yparitcher:
> You can simply ship the binary and pass the full relative path to it to `ffi.load`, it doesn't have to live in `common` (e.g., `ffi.load("foo.koplugin/libs/libbar.so.0")`. This can easily be extended to load something different based on the platform if need be.

Now i remember that i did that originally :)
However i use it in 2 plugins (i know), so that didn't scale well. I guess `common/` it is.

benoit-pierre:
> You can simply ship the binary and pass the full relative path to it to `ffi.load`, it doesn't have to live in `common` (e.g., `ffi.load("foo.koplugin/libs/libbar.so.0")`. This can easily be extended to load something different based on the platform if need be.

Honestly, the luarocks approach does not seem too user friendly for kindle users: they have to follow the link, intuit that `kindlepw2` is the one they want (easy), then they end up with a `rock`, and somehow know that it's a zip archive. Wouldn't it be easier to distribute one zip with everything, your plugin and the `libzmanim` library, for those kindle users?

benoit-pierre:
Also: we could still keep those `package.path` & `package.cpath` luarocks entries, which I planed to remove in koreader too:
```diff
diff --git a/setupkoenv.lua b/setupkoenv.lua
index 5a55a5d7a..2bd1aad9b 100644
--- a/setupkoenv.lua
+++ b/setupkoenv.lua
@@ -1,9 +1,9 @@
 -- set search path for 'require()'
 package.path =
-    "common/?.lua;rocks/share/lua/5.1/?.lua;frontend/?.lua;" ..
+    "common/?.lua;frontend/?.lua;" ..
     package.path
 package.cpath =
-    "common/?.so;common/?.dll;/usr/lib/lua/?.so;rocks/lib/lua/5.1/?.so;" ..
+    "common/?.so;common/?.dll;/usr/lib/lua/?.so;" ..
     package.cpath
 
 -- set search path for 'ffi.load()'
@@ -20,7 +20,7 @@ ffi.load = function(lib, global)
     local loaded, re = pcall(ffi_load, lib)
     if loaded then return re end
 
-    local lib_path = package.searchpath(lib, "./lib?.so;./libs/lib?.so;./libs/lib?.so.1;./rocks/lib/lua/5.1/lib?.so")
+    local lib_path = package.searchpath(lib, "./lib?.so;./libs/lib?.so;./libs/lib?.so.1")
 
     if not lib_path then
         io.write("ffi.load (warning): ", re, "\n")
diff --git a/spec/unit/commonrequire.lua b/spec/unit/commonrequire.lua
index 68baccb22..4b5db4a79 100644
--- a/spec/unit/commonrequire.lua
+++ b/spec/unit/commonrequire.lua
@@ -34,8 +34,8 @@ if not busted_ok then
     end
 end
 
-package.path = "?.lua;common/?.lua;rocks/share/lua/5.1/?.lua;frontend/?.lua;" .. package.path
-package.cpath = "?.so;common/?.so;/usr/lib/lua/?.so;rocks/lib/lua/5.1/?.so;" .. package.cpath
+package.path = "?.lua;common/?.lua;frontend/?.lua;" .. package.path
+package.cpath = "?.so;common/?.so;/usr/lib/lua/?.so;" .. package.cpath
 
 -- turn off debug by default and set log level to warning
 require("dbg"):turnOff()
diff --git a/spec/unit/dbg_spec.lua b/spec/unit/dbg_spec.lua
index 675761111..25905752e 100644
--- a/spec/unit/dbg_spec.lua
+++ b/spec/unit/dbg_spec.lua
@@ -1,7 +1,7 @@
 describe("Dbg module", function()
     local dbg, dbg_on
     setup(function()
-        package.path = "?.lua;common/?.lua;rocks/share/lua/5.1/?.lua;frontend/?.lua;" .. package.path
+        package.path = "?.lua;common/?.lua;frontend/?.lua;" .. package.path
         dbg = require("dbg")
         dbg_on = dbg.is_on
     end)
diff --git a/tools/benchmark.lua b/tools/benchmark.lua
index f3f1c18c7..77666fa07 100644
--- a/tools/benchmark.lua
+++ b/tools/benchmark.lua
@@ -2,8 +2,8 @@
 -- usage: ./luajit tools/benchmark.lua test/sample.pdf
 
 require "defaults"
-package.path = "common/?.lua;rocks/share/lua/5.1/?.lua;frontend/?.lua;" .. package.path
-package.cpath = "common/?.so;common/?.dll;/usr/lib/lua/?.so;rocks/lib/lua/5.1/?.so;" .. package.cpath
+package.path = "common/?.lua;frontend/?.lua;" .. package.path
+package.cpath = "common/?.so;common/?.dll;/usr/lib/lua/?.so;" .. package.cpath
 
 local DataStorage = require("datastorage")
 --G_reader_settings = require("docsettings"):open(".reader")
diff --git a/tools/kobo_touch_probe.lua b/tools/kobo_touch_probe.lua
index dbebdb15a..f4042e00b 100755
--- a/tools/kobo_touch_probe.lua
+++ b/tools/kobo_touch_probe.lua
@@ -2,8 +2,8 @@
 -- usage: ./luajit tools/kobo_touch_probe.lua
 
 require "defaults"
-package.path = "common/?.lua;rocks/share/lua/5.1/?.lua;frontend/?.lua;" .. package.path
-package.cpath = "common/?.so;common/?.dll;/usr/lib/lua/?.so;rocks/lib/lua/5.1/?.so;" .. package.cpath
+package.path = "common/?.lua;frontend/?.lua;" .. package.path
+package.cpath = "common/?.so;common/?.dll;/usr/lib/lua/?.so;" .. package.cpath
 
 local DataStorage = require("datastorage")
 local _ = require("gettext")
diff --git a/tools/wbuilder.lua b/tools/wbuilder.lua
index 61442553a..2edf1bda0 100755
--- a/tools/wbuilder.lua
+++ b/tools/wbuilder.lua
@@ -2,8 +2,8 @@
 -- usage: ./luajit tools/wtest.lua
 
 print(package.path)
-package.path = "common/?.lua;rocks/share/lua/5.1/?.lua;frontend/?.lua;" .. package.path
-package.cpath = "common/?.so;common/?.dll;/usr/lib/lua/?.so;rocks/lib/lua/5.1/?.so;" .. package.cpath
+package.path = "common/?.lua;frontend/?.lua;" .. package.path
+package.cpath = "common/?.so;common/?.dll;/usr/lib/lua/?.so;" .. package.cpath
 
 -- Load default settings
 G_defaults = require("luadefaults"):open()
```

Frenzie:
> Also: we could still keep those `package.path` & `package.cpath` luarocks entries, which I planed to remove in koreader too:

Perhaps not a bad idea to be able to more easily toss some stuff in there using the LuaRocks ecosystem but I'm fine either way.

benoit-pierre:
> _[`thirdparty/luasec/CMakeLists.txt` line 18 at r1](https://reviewable.io/reviews/koreader/koreader-base/1750#-Nyw2tDTBYTJFzwXEtrt:-Nyw2tDTBYTJFzwXEtru:b-rdovsz) ([raw file](https://github.com/koreader/koreader-base/blob/0eb92faa38f39c6bba27e2993239283bc0e47443/thirdparty/luasec/CMakeLists.txt#L18)):_
> 
> > ```cmake
> >     CC=${CC} CFLAGS=${CFLAGS}
> >     LDFLAGS=${DYNLIB_LDFLAGS}
> >     LIBS=${LIBS} EXTRA=${LIBS}
> > ```
> 
> Couldn't this use `append_autotools_vars`?
> 
> _Code quote:_
> 
> ```cmake
> set(MAKE_CMD
>     ${KO_MAKE_RECURSIVE}
>     AR=${AR} RANLIB=${RANLIB}
>     CC=${CC} CFLAGS=${CFLAGS}
>     LDFLAGS=${DYNLIB_LDFLAGS}
>     LIBS=${LIBS} EXTRA=${LIBS}
> ```

In general, I prefer to avoid indiscriminately overriding too many variables when calling make, as this could result in bypassing all related makefile assignments.

Seeing as `LDFLAGS` is not set to `LDFLAGS`, and the need to set `EXTRA`, I'm also not sure it would leads to much simpler code.

> Same comment as in luasec about `append_autotools_var`, at least for parts of this ;).
> 
> _Code quote:_
> 
> ```cmake
> set(MAKE_CMD
>     ${KO_MAKE_RECURSIVE}
>     AR=${AR}
>     CC=${CC}
>     CXX=${CXX}
>     LD=${LD}
>     MAKEDEPEND=${CC}
>     RANLIB=${RANLIB}
>     # Reproducible builds.
>     SOURCE_DATE_EPOCH=1694383200
> )
> ```

Ditto, we mostly need to undo configure bad choices: ensure ccache is used, and prevent the build system from using `makedepend`.

poire-z:
(If the 162 commits end up as 162 commits in master, could you somehow make them have the same date, and not spanning 7 days, so it's easier to grasp all this happened at the same time in git blame? I usually use: `git rebase --force-rebase --exec "sleep 2" --exec "git commit --amend --no-edit --date=now" master` - I needed `sleep 2` because otherwise, github could show 2 commits in the same second not ordered as they really are.)

yparitcher:
I am fine tossing the luarocks paths, also if we do keep then then the old library versions will be in the users `path` (being as we don't delete old files), which may lead to some mysterious failures down the the line.

benoit-pierre:
> (If the 162 commits end up as 162 commits in master, could you somehow make them have the same date, and not spanning 7 days, so it's easier to grasp all this happened at the same time in git blame? I usually use: `git rebase --force-rebase --exec "sleep 2" --exec "git commit --amend --no-edit --date=now" master` - I needed `sleep 2` because otherwise, github could show 2 commits in the same second not ordered as they really are.)

Done, and it only took 7 minutes! But now branchless it not happy about the mess…

poire-z:
> branchless 

What's that ?

benoit-pierre:
> > branchless
> 
> What's that ?

https://github.com/arxanas/git-branchless

NiLuJe:
> @NiLuJe: ping!

Done, thanks ;). You should be able to jump to the new commit now.

NiLuJe:
> Ditto, we mostly need to undo configure bad choices: ensure ccache is used, and prevent the build system from using `makedepend`.

Cool, I suspected your line of reasoning was something along those lines, makes perfect sense ;).

NiLuJe (commented):
> Reviewed 102 of 102 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1750)* status: all files reviewed, 1 unresolved discussion (waiting on @benoit-pierre)

___
*[`thirdparty/cmake_modules/koreader_targets.cmake` line 38 at r1](https://reviewable.io/reviews/koreader/koreader-base/1750#-NyvLRZu5F-HTQ3q4sZ-:-Nz09Xee4Y-I4qUllXe8:b-9x4brb) ([raw file](https://github.com/koreader/koreader-base/blob/0eb92faa38f39c6bba27e2993239283bc0e47443/thirdparty/cmake_modules/koreader_targets.cmake#L38)):*
<details><summary><i>Previously, benoit-pierre (Benoit Pierre) wrote…</i></summary><blockquote>

Right, no point in complicating things if it's not required. Will amend.
</blockquote></details>

Oh, right, yeah, we can just always use *both*, duh', that's even simpler than what I said ;p.

NiLuJe:
Welp, one zsync2 bump and we're off to the races I guess? Many thanks for taking on this endeavor!

benoit-pierre:
> Welp, one zsync2 bump and we're off to the races I guess? Many thanks for taking on this endeavor!

Done.

Frenzie:
@NiLuJe You said you wanted a merge commit?

NiLuJe (approved):
> Reviewed 2 of 102 files at r11, 2 of 34 files at r12, 99 of 99 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1750)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @benoit-pierre)

PR #1751, opened 2024-03-27
Call adjustAreaDefault library function

dmalinovsky:
This will take care of both saturation and dithering. Saturation will be altered using standard parameters. See https://github.com/koreader/koreader/issues/8872#issuecomment-1096624979 for more details.

Fixes https://github.com/koreader/koreader/issues/10446

### DIFF
--- a/ffi/framebuffer_pocketbook.lua
+++ b/ffi/framebuffer_pocketbook.lua
@@ -16,8 +16,8 @@ local function _getPhysicalRect(fb, x, y, w, h)
 end
 
 local function _adjustAreaColours(fb)
-    if fb.device.hasColorScreen and fb.device.color_saturation then
-        fb.debug("adjusting image color saturation", fb.device.color_saturation)
+    if fb.device.hasColorScreen then
+        fb.debug("adjusting image color saturation")
 
         inkview.adjustAreaDefault(fb.data, fb._finfo.line_length, fb._vinfo.width, fb._vinfo.height)
     end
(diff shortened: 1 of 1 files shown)


Frenzie (approved):

ffi/framebuffer_pocketbook.lua
```diff
@@ -16,10 +16,10 @@ local function _getPhysicalRect(fb, x, y, w, h)
 end
 
 local function _adjustAreaColours(fb)
-    if fb.device.hasColorScreen and fb.device.color_saturation then
-        fb.debug("adjusting image color saturation", fb.device.color_saturation)
```
rjd22, line 20: You might also want to delete the color_saturation settings from all the pocketbook device in:

https://github.com/koreader/koreader/blob/ee7c3ab551f320ff0095c276bdadfac77beeee69/frontend/device/pocketbook/device.lua

rjd22 (changes requested):

ffi/framebuffer_pocketbook.lua
dmalinovsky: Of course, it's in the separate PR: https://github.com/koreader/koreader/pull/11597

rjd22 (approved):

dmalinovsky:
> LGTM, the advantage is that this would safe us from needing to implement saturation setting sliders.

Indeed! Also, whatever optimizations Pocketbook will do in the future firmware updates, will be applied automatically.

PR #1752, opened 2024-03-27
Call `FullUpdateHQ` to eliminate ghosting on Pocketbook

dmalinovsky:
This makes rendering slightly slower but removes all ghosting artifacts.

See my tests below. I have "Always flash on pages with images" setting ON.

| With `FullUpdate()` | With `FullUpdateHQ()` |
|--------|--------|
|  |  |
|  |  | 

Ghosting is especially visible on the light parts of the images.

This should also help with the issues similar to https://github.com/koreader/koreader/issues/11233#issuecomment-1929536127.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1752)

### DIFF
--- a/ffi/framebuffer_pocketbook.lua
+++ b/ffi/framebuffer_pocketbook.lua
@@ -42,7 +42,7 @@ local function _updateFull(fb, x, y, w, h, dither)
         _adjustAreaColours(fb)
     end
 
-    inkview.FullUpdate()
+    inkview.FullUpdateHQ()
 end
 
 local function _updateFast(fb, x, y, w, h, dither)
--- a/ffi/framebuffer_pocketbook.lua
+++ b/ffi/framebuffer_pocketbook.lua
@@ -42,7 +42,11 @@ local function _updateFull(fb, x, y, w, h, dither)
         _adjustAreaColours(fb)
     end
 
-    inkview.FullUpdateHQ()
+    if fb.device.hasColorScreen then
+        inkview.FullUpdateHQ()
+    else
+        inkview.FullUpdate()
+    end
 end
 
 local function _updateFast(fb, x, y, w, h, dither)


Frenzie (approved):
Fine by me.

rjd22 (commented):
This change might break backwards compatibility for older devices that don't have support for the FullUpdateHQ method but it has been available in the inkview Library for 10 years now.

My question would be if it's also slower when reading text? Because if it is I honestly would only run this on the color devices where it is actually useful and avoid running it on the non-color devices.

dmalinovsky:
> My question would be if it's also slower when reading text?

It will be slower only if the user has "Full refresh rate" set to "Every page". For me, I have it set to the pages with the images only, so there's no extra delay for plain text.

I can add an extra check for color device compatibilities, what do you think, @rjd22?

rjd22:
> > My question would be if it's also slower when reading text?
> 
> It will be slower only if the user has "Full refresh rate" set to "Every page". For me, I have it set to the pages with the images only, so there's no extra delay for plain text.
> 
> I can add an extra check for color device compatibilities, what do you think, @rjd22?

I think only running it for color devices now will be a good middle ground. From what I recall the FullUpdateHQ only really adds something for color devices so avoiding it being run on non-color devices will lower the impact of the change.

dmalinovsky:
> I think only running it for color devices now will be a good middle ground. From what I recall the FullUpdateHQ only really adds something for color devices so avoiding it being run on non-color devices will lower the impact of the change.

Fair point, I added a check. I also tested it on my Pocketbook.

rjd22:
I dug in my mail for what Pocketbook told me. Hopefully it's helpful:

> The FullUpdate has the highest quality of update but it works only for full screen. If you call the FullUpdate, the full screen will be updated through the black, we do not recommend using it often because these often blinkings can annoy user.
>  
> As for the SoftUpdate – actually, it is the same to Partial update. It is recommended to use it only for b/w area, it has a good quality and doesn’t blink the screen, in addition to it - we recommend using FullUpdate from time to time. But this Soft update will do some ghosting for color areas.
>  
> The PartialUpdateHQ - can be used if there are color elements on the screen to reduce ghosting on them. To use it properly – you should know the coordinates of color area and call this update only for this area. It will do good quality of colored image and will not annoy user as much as the FullUpdate does. Please note, that PartialUdateHQ has sence only for devices with color screens.

dmalinovsky:
Thanks, @rjd22, it is helpful. My tests confirm this. :)

dmalinovsky:
It's interesting that `PartialUpdateHQ` is still flashing the screen but does it faster than `FullUpdateHQ`.

At first, I tried calling `PartialUpdateHQ` for all partial updates, but then the screen was flashing for every page, and it's too much.

rjd22:
@dmalinovsky Thank you for your research and hard work! I'm really glad to see others working on the pocketbook side of things.

dmalinovsky:
Yes, it was the last small thing I didn't like in KOReader on color Pocketbook vs the built-in software. Now I can read my news articles in peace... :)

rjd22:
> It's interesting that `PartialUpdateHQ` is still flashing the screen but does it faster than `FullUpdateHQ`.
> 
> At first, I tried calling `PartialUpdateHQ` for all partial updates, but then the screen was flashing for every page, and it's too much.

Color changes require a reset of the screen state to be able to show them properly without any ghosting. It requires a balanced approach on refreshing the screen. This is really hard to do in KoReader because we don't have an easy way yet to find out if the part that needs a refresh contains colors or not. As such we have to err on the side of being the least irritating.

Frenzie:
@rjd22 We normally prefer to use the squash or rebase option, since merge commits mostly strike us as noise in the log. Anyway, thanks for merging. :+1:

rjd22:
> @rjd22 We normally prefer to use the squash or rebase option, since merge commits mostly strike us as noise in the log. Anyway, thanks for merging. 👍

Ah sorry I forgotten to switch the setting. My bad :)

ffi/framebuffer_pocketbook.lua
```diff
@@ -42,7 +42,11 @@ local function _updateFull(fb, x, y, w, h, dither)
         _adjustAreaColours(fb)
     end
 
-    inkview.FullUpdate()
+    if fb.device.hasColorScreen then
+        inkview.FullUpdateHQ()
+    else
+        inkview.FullUpdate()
```
ElimGarak1, line 48: During testing, I noticed that this does not result in a screen refresh on PocketBook devices without a color screen.

dmalinovsky: This PR doesn't change anything for B&W Pocketbooks. Do you have full refresh turned on?

ElimGarak1: > Do you have full refresh turned on?

Yes, of course.

> This PR doesn't change anything for B&W Pocketbooks.

Unfortunately, because of lines 45 and 46, B&W PocketBook devices now ignore line 48 for whatever reason.

dmalinovsky: Hmm, maybe we need to explicitly set `hasColorScreen` to false. I've created https://github.com/koreader/koreader/pull/11608 to address this.

dmalinovsky: Okay, now I'm 99% sure https://github.com/koreader/koreader-base/pull/1757 should fix it. :)

PR #1756, opened 2024-04-01
make: fix test rule on Arch Linux

benoit-pierre:
I forgot I had worked around the fact that the `busted` launcher installed on Arch Linux is actually a shell script, not a LUA file.

Said script also messes with the LUA environment (forcibly using lua5.4).

Switch to invoking the busted runner manually.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1756)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -274,11 +274,7 @@ $(OUTPUT_DIR)/spec/base: | $(OUTPUT_DIR)/spec/
 	ln -sf ../../../spec $(OUTPUT_DIR)/spec/base
 
 test: all test-data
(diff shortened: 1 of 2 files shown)


Frenzie:
That's not Arch-specific btw, but a decade ago LuaRocks didn't do that yet. See for example https://github.com/koreader/virdevenv/blob/8268e86f6d8eae9439d81e87e9b3a0c81a4495b3/docker/ubuntu/baseimage-20.04/install_luarocks.sh#L20

Frenzie (approved):
Bit annoying to read. :-)

PR #1757, opened 2024-04-01
Bugfix: device property is a function

dmalinovsky:
Previous code was always returning true, even for B&W Pocketbooks. 

See https://github.com/koreader/koreader/blob/42c93a7623c6d04a2001f943ddc5b0b0638e3a24/frontend/device/pocketbook/device.lua#L16 for `yes` and `no` definitions.

Tested on Pocketbook Inkpad Color 3.

### DIFF
--- a/ffi/framebuffer_pocketbook.lua
+++ b/ffi/framebuffer_pocketbook.lua
@@ -16,7 +16,7 @@ local function _getPhysicalRect(fb, x, y, w, h)
 end
 
 local function _adjustAreaColours(fb)
-    if fb.device.hasColorScreen then
+    if fb.device.hasColorScreen() then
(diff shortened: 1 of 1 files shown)


dmalinovsky:
Thank you, @ElimGarak1, for catching this bug!

dmalinovsky:
Thanks, @Frenzie, for quick turnaround!

ElimGarak1:
Thank you, @dmalinovsky. 

I have PB-IP Color 2&3; PB-IP 4; PB Era; PB Verses Pro; IP Lite; PB-Basic Lux 4; ~ tested and can confirm that a screen update for color and black and white devices is now carried out correctly.

PR #1758, opened 2024-04-05
cre: add identifiers to doc props

baswag:
Depends on https://github.com/koreader/crengine/pull/560

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1758)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -813,6 +813,10 @@ static int getDocumentProps(lua_State *L) {
 	lua_pushstring(L, UnicodeToLocal(doc->text_view->getKeywords()).c_str());
 	lua_rawset(L, -3);
 
+    lua_pushstring(L, "isbn");
+    lua_pushstring(L, UnicodeToLocal(doc->text_view->getIdentifiers()).c_str());
+    lua_rawset(L, -3);
+
 	return 1;
 }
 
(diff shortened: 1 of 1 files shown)


cre.cpp
```diff
@@ -813,6 +813,10 @@ static int getDocumentProps(lua_State *L) {
 	lua_pushstring(L, UnicodeToLocal(doc->text_view->getKeywords()).c_str());
 	lua_rawset(L, -3);
 
+    lua_pushstring(L, "isbn");
+    lua_pushstring(L, UnicodeToLocal(doc->text_view->getIdentifiers()).c_str());
+    lua_rawset(L, -3);
```
baswag, line 818: I don't know why the indentation is shown wrong here, it looks good when I view the file locally :/

cre.cpp
```diff
@@ -813,6 +813,10 @@ static int getDocumentProps(lua_State *L) {
...
+
```
benoit-pierre, line 819: Surrounding code is using tags for indentation:
```suggestion
	lua_pushstring(L, "isbn");
	lua_pushstring(L, UnicodeToLocal(doc->text_view->getIdentifiers()).c_str());
	lua_rawset(L, -3);

```

PR #1760, opened 2024-04-06
harfbuzz: fix ccache 4.9.1 errors

benoit-pierre:
Additionally, fix possible build error if ragel is not installed and some of the ragel source files end up with a more recent timestamp than their pre-generated results.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1760)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -53,7 +53,6 @@ jobs:
             makedepend
             nasm
             pkg-config
-            ragel
             wget
 
       - name: Update PATH
--- a/thirdparty/harfbuzz/CMakeLists.txt
+++ b/thirdparty/harfbuzz/CMakeLists.txt
@@ -17,6 +17,21 @@ assert_var_defined(FREETYPE_DIR)
 ep_get_source_dir(SOURCE_DIR)
 ep_get_binary_dir(BINARY_DIR)
 
+# Get rid of a bunch of `#line` directives that trip ccache 4.9.1.
+# NOTE: this has also the nice effect of "touching" each file, so
+# the build system does not try to regenerate them (and building
+# without ragel is consistently possible).
+list(APPEND PATCH_CMD COMMAND ${ISED} "/^#line/d"
+    src/hb-ot-shaper-khmer-machine.hh
+    src/hb-ot-shaper-use-machine.hh
+    src/hb-ot-shaper-myanmar-machine.hh
+    src/hb-buffer-deserialize-json.hh
+    src/hb-buffer-deserialize-text-glyphs.hh
+    src/hb-number-parser.hh
+    src/hb-ot-shaper-indic-machine.hh
+    src/hb-buffer-deserialize-text-unicode.hh
+)
+
 list(APPEND PATCH_CMD COMMAND env NOCONFIGURE=1 ./autogen.sh)
 
 # No build rpath


Frenzie (approved):
Yes, I'd wanted to do this. :+1:

benoit-pierre:
I don't understand why the cache hit rate is lower with the `emu_gcc_ninja_debug` job (81.47 %) compared to `emu_clang_ninja` (99.20 %).

Frenzie:
I notice that we don't seem to be including `{{ arch }}` in the cache key here, but the specific physical server it happens to run on does play a role. Perhaps it would be more consistent with something like `emu-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-`.

benoit-pierre:
Only for that job? #1750 has a 100% ccache hit rate on all jobs.

Frenzie:
> Only for that job?

No, I mean the same principle for each and every one (that involves some kind of code generation).

benoit-pierre:
No objections to merging? (Needed before bumping the docker images).

poire-z:
Was just about to write "let me merge the cre skip finalizers suff so I don't bring other build related nightmare with it :)

Can I, or will I indeed bring hell with it ?

Frenzie:
This shouldn't result in any nightmares, but in that case let's hold off on merging the Docker bump a little bit longer. :-)

poire-z:
Also no problem if base brings a luajit bump, but luajit-launcher lags ?
(I'm in no hurry merging my cre PR, or it can be merged and bumped later, it's a no-op without the frontend part, I can wait for things to get ISO and clean, if it is best.)

Frenzie:
luajit-launcher should normally be bumped at the same time. Which is not an issue, things are peaceful and quiet over there. ;-)

You can go ahead and merge things if you're ready, but if you're still working on the frontend part I expect the Docker stuff to be sorted out soon. I'd already tested a very similar configuration a year or longer ago. The only thing I'm worried about is the Android build, but that one in particular is quite different now.

benoit-pierre:
> > Only for that job?
> 
> No, I mean the same principle for ~each and every one~both of them. [For the cross-compilation it shouldn't play a role.]

If you look at [this build](https://app.circleci.com/pipelines/github/koreader/koreader-base/2113/workflows/c5efa7e3-d611-4c05-94c1-4aa504d24440/jobs/18944)'s [restore cache step](https://circleci.com/api/v1.1/project/github/koreader/koreader-base/18944/output/102/0?file=true&allocation-id=661d8ac2e635f87210e04cfa-0-build%2FABCDEFGH), part of #1750, you'll see a cache from another unrelated PR was actually restored (from [this build](https://app.circleci.com/pipelines/github/koreader/koreader-base/2110/workflows/16f1b41d-cdfc-46b4-96d7-224f11f2b3cc/jobs/18919))!

On Github Actions, a branch will never use a cache from another branch, except from their ancestor (so a PR can use a cache from the master branch), is CircleCI different?

benoit-pierre:
Apparently not, Cf. CircleCI documentation:

> Caching and open source
>
> If your project is open source/available to be forked and receive PRs from contributors, make note of the following:
>
>    PRs from the same fork repository share a cache (this includes, as previously stated, that PRs in the main repository share a cache with main).
>
>    Two PRs in different fork repositories have different caches.
>
>    Enabling the sharing of environment variables allows cache sharing between the original repository and all forked builds.

vs Github Actions:

> Access restrictions provide cache isolation and security by creating a logical boundary between different branches or tags. Workflow runs can restore caches created in either the current branch or the default branch (usually main). If a workflow run is triggered for a pull request, it can also restore caches created in the base branch, including base branches of forked repositories. For example, if the branch feature-b has the base branch feature-a, a workflow run triggered on a pull request would have access to caches created in the default main branch, the base feature-a branch, and the current feature-b branch.

benoit-pierre:
Which mean we should prepend the branch name to the cache key, with a fallback on master.

PR #1763, opened 2024-04-07
fix building with older make versions

benoit-pierre:
Don't try to forcibly append `$(MAKEFLAGS)` to the make command, as it may only contain single letters only (short options with no dash).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1763)


poire-z (approved):
My approval is not worth more than "Works for me" :)

PR #1764, opened 2024-04-07
factorize cgroups detection (for proper jobs/load settings)

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1764)


Wssfh:
我装最新版koreader 触摸失灵，点哪里都没用，哪位大神知道哪里的问题？

Frenzie:
For bug reports see https://github.com/koreader/koreader/issues/new/choose but just for asking questions https://www.mobileread.com/forums/forumdisplay.php?f=276 is generally better. I understand there are also some Chinese forums where people discuss KOReader.

PR #1765, opened 2024-04-07
Update LuaJIT to 20240310

hugleo:
Update LuaJIT to 20240310

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1765)

### DIFF
--- a/thirdparty/luajit/CMakeLists.txt
+++ b/thirdparty/luajit/CMakeLists.txt
@@ -78,7 +78,7 @@ list(APPEND BUILD_CMD COMMAND ${KO_MAKE_RECURSIVE} ${BUILD_ARGS})
 ko_write_gitclone_script(
     GIT_CLONE_SCRIPT_FILENAME
     https://github.com/LuaJIT/LuaJIT
-    c525bcb9024510cad9e170e12b6209aedb330f83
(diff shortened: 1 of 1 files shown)


Frenzie:
https://github.com/LuaJIT/LuaJIT/compare/c525bcb9024510cad9e170e12b6209aedb330f83...d06beb0480c5d1eb53b3343e78063950275aa281

hugleo:
This one uses luajit -b

https://github.com/kernelsauce/turbo/blob/v2.1.3/Makefile#L147

Frenzie:
I'm not sure what the question is?

hugleo:
Specifically about commit: https://github.com/LuaJIT/LuaJIT/commit/4b90f6c4d7420139c135435e1580acb52ea18436

If we need -W, -X -or d option like:

luajit -bW -g ... 
luajit -bX -g ... 
luajit -bd -g ... 

Or if some load* function in our code needs an optional mode parameter like "bW".
loadstring("XXX", "bt") , loadstring("XXX", "t") , loadstring("XXX", "btW")... PS: Don't know if is configured this way.

But I think is not applicable for us since we don't use bytecode bin format and directly load the file with "./luajit reader.lua".

Frenzie:
Right, we don't use any of that.

PR #1767, opened 2024-04-10
ci: bump (most) docker images (switch to Ubuntu 20.04)

benoit-pierre:
Note: `emu_clang_ninja` was supposed to use clang-16 but the image has clang-14 and the logs show that's indeed what was being used. Our glib version does not like clang-16, so hold off bumping that image.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1767)


Frenzie:
You can delete that. They're some vestigial lines from back when we used Travis CI and/or the early CircleCI days. It's because whichever clang was in 16.04 (or perhaps even 18.04) was too old.

https://github.com/koreader/koreader-base/blob/499148649d1ea95faa70f3c89b07b0194f13a597/.ci/emulator_install.sh#L17-L28

Frenzie:
Actually that entire file is probably no longer necessary. The first part of it is done as part of the image building process.

benoit-pierre:
Yep, will remove. What about the travis configuration? Is there still a point in keeping it around?

Frenzie:
It can be removed. I'll also remove the webhook (which has been returning 503 for a couple of years now).

Frenzie (approved):
Seems fine to me. Good for you?

benoit-pierre:
Yep :+1:

Frenzie:
@poire-z I think you wanted to get a build out for your changes using the current Docker images first?

poire-z:
No, not really caring about getting a built, just wanting to get this bit out of my messy tree :)
You can merge in base all you want, and bump it in frontend later this evening after my https://github.com/koreader/koreader/pull/11658 is merged (no pb for me if you'd get isssues and no nightly tomorrow).

PR #1768, opened 2024-04-12
ffi/framebuffer_mxcfb.lua: Fix Night Mode on Kobo Elipsa 2E

katadelos:
This commit "fixes" reported graphical problems caused by switching to Night Mode on Elipsa 2E.

Mentioned here: https://github.com/koreader/koreader/pull/10719#issuecomment-1693425726

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1768)

### DIFF
--- a/ffi/framebuffer_mxcfb.lua
+++ b/ffi/framebuffer_mxcfb.lua
@@ -935,8 +935,8 @@ function framebuffer:init()
             -- REAGL is *always* available
             self.waveform_partial = C.HWTCON_WAVEFORM_MODE_GLR16
             -- Eclipse waveform modes are *always* available
-            self.waveform_night = C.HWTCON_WAVEFORM_MODE_GLKW16
-            self.waveform_flashnight = C.HWTCON_WAVEFORM_MODE_GCK16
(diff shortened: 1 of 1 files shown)


katadelos:
IMG_6725
Image of this issue for reference

NiLuJe:
IIRC, we've confirmed that Nickel *does* use the Eclipse waveform mode explciitly and... it works there, so i'm still mostly stumped ;p.

Since the slew of new devices run on the same SoC, at least we'll have more opportunity to test stuff soon...

Frenzie:
@NiLuJe Does that mean to merge now?

NiLuJe:
The opposite, I'd rather wait and see how the new devices fare with the current behavior (... assuming the ioctls are compatible ;p).

NiLuJe:
Completely unrelated query, given the MTK stuff happening with the new devices: Aren't gyro-triggered rotations inverted?

NiLuJe:
Note to self: We can also probably preemptively handle the Tolino variants, so that we're ready once they actually make the switch from Android to Nickel (which is... possibly on release for those?).

PR #1770, opened 2024-04-14
make: speedup `fetchthirdparty`

benoit-pierre:
- on CI, always clone with a shallow depth of 1
- use 3 jobs

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1770)

### DIFF
--- a/Makefile.third
+++ b/Makefile.third
@@ -6,7 +6,7 @@ fetchthirdparty:
 	git submodule init
 	git submodule sync
 	git submodule foreach --recursive git reset --hard
-	git submodule update
+	git submodule update --jobs 3 $(if $(CI),--depth 1)
 	@echo "cleaning up crengine checkout..."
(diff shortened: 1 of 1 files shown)


Makefile.third
```diff
@@ -6,7 +6,7 @@ fetchthirdparty:
 	git submodule init
 	git submodule sync
 	git submodule foreach --recursive git reset --hard
-	git submodule update
+	git submodule update --jobs 3 $(if $(CI),--depth 1,$(--recommend-shallow))
```
Frenzie, line 9: Same remark as in https://github.com/koreader/koreader/pull/11677#discussion_r1564802473

It seems to me it's both more readable and friendlier to local git preferences to leave it empty, especially since it's the default.

```suggestion
	git submodule update --jobs 3 $(if $(CI),--depth 1,)
```

PR #1777, opened 2024-05-03
make: saner `fetchthirdparty`

benoit-pierre:
Avoid destructive operations.

The CIs don't rely on it.

It does not help that it can be triggered by `./kodev build` when using a sparse checkout that does not include all submodules (like `platform/ubuntu-touch/ubuntu-touch-sdl`).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1777)

### DIFF
--- a/Makefile.third
+++ b/Makefile.third
@@ -5,13 +5,7 @@
 fetchthirdparty:
 	git submodule init
 	git submodule sync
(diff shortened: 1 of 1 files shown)


Frenzie:
I'm not really sure why that would've done a reset.

PR #1778, opened 2024-05-03
curl: make certificates bundle generation more robust

benoit-pierre:
Don't use the default build system rule to create the certificates bundle: connections to `hg.mozilla.org` from CIs seem to be flaky, resulting in regular failures.

Cf. https://curl.se/docs/caextract.html

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1778)


benoit-pierre:
Updated to fetch the bundle as part of the download step, and to only show a warning if it can't be updated: this was the only remaining thing preventing some platform offline builds after `make download` with the new build system.

PR #1784, opened 2024-05-13
make: fix NEON support detection

benoit-pierre:
NEON support is available when compiling with `TARGET=linux LINUX_ARCH=arm64` (`-march=armv8-a`).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1784)


NiLuJe (approved):
Yup, this one slipped through the cracks ;).

PR #1791, opened 2024-05-20
mupdf: pre update cleanups

benoit-pierre:
There's one minor fix (correct `fz_recognize_image_format`'s prototype), but mostly tweak some patches to make updating easier (e.g. reduce conflicts), and trim FFI cdecls to keep only the declarations we need.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1791)

### DIFF
--- a/ffi-cdecl/mupdf_decl.c
+++ b/ffi-cdecl/mupdf_decl.c
@@ -1,5 +1,6 @@
 // We need to make generate first...
 // CPPFLAGS="-I../koreader-base -I/var/tmp/niluje/mupdf/include"
+
 #include "wrap-mupdf.h"
 
 #include "ffi-cdecl.h"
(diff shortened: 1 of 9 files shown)


Frenzie (approved):
Preferably for after #1750?

benoit-pierre:
Yes (will conflict in `thirdparty/mupdf/CMakeLists.txt`).

Frenzie:
Ready to merge or would you rather wait until there's a nightly with all the other changes out first?

benoit-pierre:
Up to you, I have the MuPDF update and the leptonica+tesseract+libk2pfopt updates ready.

benoit-pierre:
I mean, this can be merged without waiting, but wait for a nightly or 2 before the proper updates?

Frenzie:
I've manually triggered a pipeline run speed up the process a little. If everything seems fine I'll merge some more things here but indeed let's wait with the actual update a couple of days.

benoit-pierre:
Welp, it did not take long: https://gitlab.com/koreader/nightly-builds/-/jobs/6973852572.

benoit-pierre:
OK, fix incoming (need to remove `-ffile-prefix-map=/home/ko/koreader-base/=` like pocketbook).

Frenzie:
> Welp, it did not take long: https://gitlab.com/koreader/nightly-builds/-/jobs/6973852572.

Nobody expects the ~~Spanish~~legacy inquisition! (Or at least I didn't.)

benoit-pierre:
I don't know, but the fact that the page number come first in the page selection is a bug IMHO. Compare with xpdf or evince:

https://github.com/user-attachments/assets/408b6634-3fef-4459-9565-7a4f82038c81


https://github.com/user-attachments/assets/c87bf723-593a-4f59-bfe8-cbda4de543ea

https://github.com/user-attachments/assets/f41d48fd-0bbb-4b63-9f81-6d5a32d27801


https://github.com/user-attachments/assets/7746f20b-512c-4840-baa7-04456e57aedb

And then there's the way the selection is update around the starting point, which is also buggy IMHO:

https://github.com/user-attachments/assets/20e2d648-68c7-41f6-8bd7-cdd8d7b89e67

hius07:
> the fact that the page number come first in the page selection is a bug IMHO.

Sure, but at first I'm concerned about these weird entries.

hius07:
Can it be an error in the base-code?
(Before I try to filter these entries in the frontend)

benoit-pierre:
The default values for empty rectangles with the MuPDF update that followed:
```c
const fz_rect fz_empty_rect = { FZ_MAX_INF_RECT, FZ_MAX_INF_RECT, FZ_MIN_INF_RECT, FZ_MIN_INF_RECT };
```
contrary to what the C documentation still says:
```c
const fz_rect fz_empty_rect = { 0, 0, 0, 0 };
```
So we should just filter them out (can't use `fz_is_empty_rect` because it's an inline function):

```diff
 ffi/mupdf.lua | 11 ++++++-----
 1 file changed, 6 insertions(+), 5 deletions(-)

diff --git i/ffi/mupdf.lua w/ffi/mupdf.lua
index cd198437f..73634d120 100644
--- i/ffi/mupdf.lua
+++ w/ffi/mupdf.lua
@@ -561,11 +561,12 @@ function page_mt.__index:getPageText()
                         ch = ch.next
                     end
 
-                    line.x0, line.y0 = line_bbox.x0, line_bbox.y0
-                    line.x1, line.y1 = line_bbox.x1, line_bbox.y1
-                    size = size + 5 * 8
-
-                    table.insert(lines, line)
+                    if line_bbox.x0 < line_bbox.x1 and line_bbox.y0 < line_bbox.y1 then
+                        line.x0, line.y0 = line_bbox.x0, line_bbox.y0
+                        line.x1, line.y1 = line_bbox.x1, line_bbox.y1
+                        size = size + 5 * 8
+                        table.insert(lines, line)
+                    end
                 end
 
                 mupdf_line = mupdf_line.next
```

hius07:
Thanks, this should be good.

benoit-pierre:
As for the weird behavior when moving around the selection start point, the problem is the gesture detector's pan threshold of 35 pixels, which does not make sense with a mouse. Combined with the fact that panning stays anchored to the initial position, and the selection is not updated when you're in the "dead" zone:

https://github.com/user-attachments/assets/141bc447-8693-4aed-abcf-a6ae6813ac95

PR #1794, opened 2024-05-24
Kobo: Boost saturation on CFA updates

NiLuJe:
Front will expose a dev setting to disable this.

c.f., https://github.com/koreader/koreader/issues/11756#issuecomment-2128613652

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1794)


poire-z:
Can you somewhere write what's the acronym "CFA" means (or what you guessed it means) ?
(Your abusing of abbreviations is already hard to follow, so a little reference somewhere would be nice :))

Frenzie:
I believe it's the "Colo(u?)r Filter Array".

poire-z:
TY, TLs will be aπ .
(But even in base, some introduction to the technical terms may be nice - but I understand you want to keep all that crpytic so newcomers don't go messing with it :))

NiLuJe (commented):
> Reviewed 5 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1794)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

PR #1795, opened 2024-05-29
crengine: fix `libcrengine.a` symbols visibility

benoit-pierre:
We want both `-fvisibility-inlines-hidden` **and** `-fvisibility=hidden`.

This reduce `koreader-cre`'s final code size by ~174-220 KB (and potentially improve performance).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1795)


Frenzie (approved):
Sounds good.

PR #1798, opened 2024-05-29
fix building for `kindle-legacy` target

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1798)

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -585,8 +585,8 @@ ifneq (yes,$(firstword $(call hasminstd,$(CXX),c++,__cplusplus,201700)))
 	CXXSTD_FLAGS := -std=gnu++17
 endif
 
-# Improve reproducibility (disabled on pocketbook due to the ancient toolchain).
-ifneq ($(TARGET),pocketbook)
+# Improve reproducibility (disabled on ancient toolchains).
+ifeq (,$(filter kindle-legacy pocketbook,$(TARGET)))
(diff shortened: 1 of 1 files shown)


Frenzie:
Hold on, did that say:
```
-- The C compiler identification is GNU 7.5.0
-- The CXX compiler identification is GNU 7.5.0
```
I don't think that sounds right (despite the name "legacy"). It should still be GCC 11.

benoit-pierre:
I think this is right: those are the 2 toolchains really lagging behind, Cf. https://github.com/koreader/koxtoolchain/blob/master/gen-tc.sh#L133.

```bash
▹ ~/dev/x-tools/arm-pocketbook-linux-gnueabi/bin/arm-pocketbook-linux-gnueabi-gcc --version
arm-pocketbook-linux-gnueabi-gcc (crosstool-NG 1.23.0.90-9954e26 - NiLuJe) 7.5.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Frenzie:
Yes, I had just spotted the same comment. I probably knew all this very well around the time of the PB troubles. The troubles being that it was stuck on something like 4.6 or 4.7; 7.5 is delightfully new in comparison. ;-)

PR #1799, opened 2024-05-29
luajit: fix library cross-compiling link arguments

benoit-pierre:
No real impact in practice, as we only ship the library on android, and in this case the luajit-launcher version is used.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1799)


benoit-pierre:
Which by the way is a situation I would really like to address by using the base library: it does not make sense to compile 2 versions of luajit, and the less time spent in gradle land, the better.

Frenzie:
Preferably while remaining independently usable, but if you know or want to get to know Gradle better you certainly have my blessing. Pinging @pazos for some tips, thoughts or possibly objections, since he's responsible for the modern Gradle/Kotlin Android renovation.

PR #1800, opened 2024-05-30
update leptonica, tesseract, and libk2pdfopt

benoit-pierre:
Depend on https://github.com/koreader/libk2pdfopt/pull/49.

Thanks to some upstream cleanups in tesseract, the resulting code size (bss+data+text) is reduced by ~1.4-2.3 MB.

Note: tesseract is compiled with legacy engine support, which means the resulting mode after initialization will depend on the type of language data available:
- [tessdata](https://github.com/tesseract-ocr/tessdata) (the one we use for the testsuite): 22.4 MB for English; contains data for both legacy & new engine: run the LSTM recognizer (new), but allow fallback to Tesseract (legacy) when things get difficult.
- [tessdata-fast](https://github.com/tesseract-ocr/tessdata_fast): 3.92 MB for English; best “value for money” in speed vs accuracy (`Integer` models): run the LSTM line recognizer only.
- [tessdata-best](https://github.com/tesseract-ocr/tessdata_best): 14.7 MB for English; best results on Google’s eval data, slower (`Float` models): same mode as above.

AFAIK, there are no legacy only language files.

I tested both modes, English only with `tessdata` & `tessdata-fast`, and it did not seem to make a difference (which might just mean that the fallback on legacy was never triggered). Disabling the legacy engine would save an additional ~500 KB. But I don't know if the new engine is (always) better, especially for other languages.

The version used by Linux distributions seem to vary: for example Arch Linux is using the English `tessdata` variant, while Ubuntu uses `tessdata-fast`. ¯\\_(ツ)_/¯

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1800)


Frenzie:
> I tested both modes, English only with tessdata & tessdata-fast, and it did not seem to make a difference (which might just mean that the fallback on legacy was never triggered).

You mean on a device I presume?

benoit-pierre:
> > I tested both modes, English only with tessdata & tessdata-fast, and it did not seem to make a difference (which might just mean that the fallback on legacy was never triggered).
> 
> You mean on a device I presume?

Mainly with the emulator.

Frenzie (approved):

PR #1808, opened 2024-06-01
MuPDF: Unbreak RGB->BGR swap

NiLuJe:
`fz_default_color_params` is no longer a function, but a plain constant.

Regression since #1805

Re: https://github.com/koreader/koreader/issues/11952

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1808)

### DIFF
--- a/ffi-cdecl/mupdf_decl.c
+++ b/ffi-cdecl/mupdf_decl.c
@@ -104,7 +104,7 @@ cdecl_func(fz_drop_stext_page)
 
 cdecl_type(fz_color_params)
 cdecl_type(fz_default_colorspaces)
-cdecl_func(fz_default_color_params)
+cdecl_const(fz_default_color_params)
 
 /* pixmaps */
 cdecl_func(fz_new_pixmap) // compat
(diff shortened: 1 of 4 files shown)


NiLuJe:
(Untested, mind you ^^).

jospalau:
Commenting the line self.hasBGRFrameBuffer = yes in device.lua for Kobo makes it work. Thanks

NiLuJe:
Another random query, as I'm unfamiliar with the whole thing: `fz_convert_pixmap` can throw, should it be wrapped in one our our `wrap-mupdf.h` wrappers?

NiLuJe:
Eh, wrapped it.

Also, tested it now ;p.

Which means I had my first new buildsystem experience: she goes!

```
- Real: 64.51s User: 551.75s System: 87.05s Percent: 990% Cmd: ./kodev build MajFault: 867 VolContSw: 265299 InvContSw: 176338 ```
(Add ~10s for a clean clone, this was a rebuild so the sources were downloaded already).

 Kudos @benoit-pierre ;).

PR #1814, opened 2024-06-06
feat: Add support for kindles after fw 5.16.3

notmarek:
Adds support for building koreader for kindles after the switch to hardfloats using https://github.com/notmarek/koxtoolchain. This is more of an example for people adventuring into jailbreak on newer firmwares rather than a ready for production toolchain since @NiLuJe mentioned wanting to do multiple modifications to it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1814)

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -146,6 +146,14 @@ else ifeq ($(TARGET), kindlepw2)
 	else
 		CHOST?=arm-linux-gnueabi
 	endif
+else ifeq ($(TARGET), kindlehf)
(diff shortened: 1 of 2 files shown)


Frenzie (approved):
lgtm at a glance

NiLuJe:
Front will need a `cp make/kindlepw2.mk make/kindlehf.mk` but that other than that it should be good enough for now ;).

pazos (approved):

PR #1818, opened 2024-06-11
mupdf: fix shared library permissions

benoit-pierre:
Ensure it's executable, so `bindeps` & `libcheck` targets don't ignore it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1818)


Frenzie:
Didn't we have a discussion about not making them executable? (I'm not trying to state any particular preference.)

Frenzie (approved):

benoit-pierre:
It's not the default on Debian, and [lintian complains about it](https://github.com/benoit-pierre/koreader/blob/master/platform/linux/do_debian_package.sh#L119).

Frenzie:
In principle it doesn't seem to make much sense to set it as executable if the result isn't anything useful, in any case.

PR #1820, opened 2024-06-11
improve linting

benoit-pierre:
The plan is to use the new `./utils/shellcheck.sh` helper in koreader too, to avoid duplication and ensure the same standard is used.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1820)

### DIFF
--- a/.ci/helper_shellchecks.sh
+++ b/.ci/helper_shellchecks.sh
@@ -7,21 +7,4 @@ source "${CI_DIR}/common.sh"
 # shellcheck disable=2016
 mapfile -t shellscript_locations < <({ git grep -lE '^#!(/usr)?/bin/(env )?(bash|sh)' && git submodule --quiet foreach '[ "$path" = "thirdparty/kpvcrlib/crengine" ] && git grep -lE "^#!(/usr)?/bin/(env )?(bash|sh)" | grep .ci | sed "s|^|$path/|" || git grep -lE "^#!(/usr)?/bin/(env )?(bash|sh)" | sed "s|^|$path/|"' && git ls-files ./*.sh; } | sort | uniq)
 
-SHELLSCRIPT_ERROR=0
-
-for shellscript in "${shellscript_locations[@]}"; do
-    echo -e "${ANSI_GREEN}Running shellcheck on ${shellscript}"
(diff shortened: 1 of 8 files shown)


.ci/helper_shellchecks.sh
```diff
@@ -7,21 +7,4 @@ source "${CI_DIR}/common.sh"
...
-        shfmt -i 4 "${shellscript}" || SHELLSCRIPT_ERROR=1
-        continue
-    fi
-    if [ "$(cat "${shellscript}")" != "$(shfmt -i 4 "${shellscript}")" ]; then
-        echo -e "${ANSI_RED}Warning: ${shellscript} does not abide by coding style, diff for expected style:"
-        shfmt -i 4 "${shellscript}" | diff "${shellscript}" - || SHELLSCRIPT_ERROR=1
-    fi
-done
-
-exit "${SHELLSCRIPT_ERROR}"
```
Frenzie, line 27: It's worth noting that if there were any improvements since this was written, it's this file that would've been potentially overlooked and not the one in front.

As it happens that one was changed to use a variable, albeit combined with an ugly shellcheck disable.

https://github.com/koreader/koreader/commit/cb2314d11b6c2c51a88539d3106dab11389eda3d

More important because it affects the output, I added `-ci` (`switch cases will be indented`) in https://github.com/koreader/koreader/commit/668eee97fa4a322d434b5de11df8b9b8d1df3329

Frenzie: Although the commit "utils: lint" at https://github.com/koreader/koreader-base/pull/1820/commits/fc9c8aab47c9ceb3adc63fb6eb9ccac28c632f2a seems to suggest `-ci` may be the default now?

Frenzie: Ah never mind, I somehow accidentally skipped past a commit.

Let me paraphrase in any case as that as structured I find it harder to gauge what are changes and what's merely updating it as per in front, but it looks like it's carrying over the most important changes.

Frenzie (approved):
PS Mostly just thinking aloud, no need to change anything in my book.

PR #1821, opened 2024-06-12
mupdf: fixes

benoit-pierre:
Rework external fonts patch to fix second variant of https://github.com/koreader/koreader/issues/11959.

Fix `mupdf_fz_resolve_link` backward incompatible change when link position cannot be resolved: don't return `(NaN, NaN)`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1821)


Frenzie (approved):
Somewhere in front is also fine by me; I don't know if it needs to be compatible in that way. ^^

PR #1822, opened 2024-06-13
ci: fix shellchecks helper

benoit-pierre:
Quote argument to `git ls-files`: ensure the glob is evaluated by git, not bash.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1822)


Frenzie (approved):

benoit-pierre:
Any objections to merging this now so it can be part of a base bump in https://github.com/koreader/koreader/pull/12023?

PR #1825, opened 2024-06-16
ffi/util: Deal with stupid GNU basename quirks

NiLuJe:
basename("/foo/") yielding "" is very much not what we signed up for...

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1825)

### DIFF
--- a/ffi/util.lua
+++ b/ffi/util.lua
@@ -193,10 +193,15 @@ if ffi.os == "Windows" then
     end
 else
     function util.basename(in_path)
-        -- We have no guarantee of getting a GNU implementation of basename;
-        -- POSIX compliant implementations *will* modify input, so, always make a copy.
-        local path = ffi.new("char[?]", #in_path + 1, in_path)
-        return ffi.string(C.basename(path))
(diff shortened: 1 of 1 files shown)


NiLuJe:
There is thankfully no such madness happening with `dirname`, so we once again have a guarantee that joining the output of dirname with the output of basename yields somethign sensible.

NiLuJe:
See https://github.com/koreader/koreader/pull/12038#issuecomment-2171805576 for more details, because I had completely forgotten we'd witched to (attempt to) prefer the GNU imp, because Android...

mergen3107:
Ah, OK xD You beat me to it, lol

NiLuJe:
> because (ironically) Android...

Specifically, the only guarantee we have on bionic is that the public symbol will be named `basename`. This happens to match the name of the *GNU* implementation on glibc, where the POSIX is instead exported as `__xpg_basename`.

Whether the Android implementation leans more toward a POSIX or GNU behavior is irrelevant (for the record, it leans towards a sane, POSIX behavior, while also not modifying its input ;p).

Frenzie (approved):
Ummmm, well, okay.

PR #1827, opened 2024-06-17
LuaSocket: Bump to 20240524

NiLuJe:
And enforce CLOEXEC on open'ed sockets, to avoid weird crap like https://github.com/koreader/koreader/issues/12043

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1827)

### DIFF
--- a/thirdparty/luasocket/CMakeLists.txt
+++ b/thirdparty/luasocket/CMakeLists.txt
@@ -17,6 +17,11 @@ list(APPEND PATCH_CMD COMMAND sh -c "\"\$@\" src/*" --
 # Don't enforce CFLAGS.
 list(APPEND PATCH_CMD COMMAND ${ISED} "s|-O2 -ggdb3 ||g" src/makefile)
 
+# Make sure socket fds are CLOEXEC on *nix
+list(APPEND PATCH_CMD COMMAND ${KO_PATCH}
+    ${CMAKE_CURRENT_SOURCE_DIR}/luasocket-enforce-cloexec.patch
+)
+
 if(DARWIN)
     set(PLAT macosx)
 elseif(WIN32)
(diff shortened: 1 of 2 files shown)


Frenzie (approved):

Frenzie:
Is it suitable to PR upstream as is?

NiLuJe:
That's a fair question. I have no idea if there's a Windows equivalent, and/or if it's something upstream would even want. (I imagine there *are* use-cases where you want your child processes to inherit sockets).

benoit-pierre:
Wouldn't using `SOCK_CLOEXEC` make more sense?

> SOCK_CLOEXEC    Set  the  close-on-exec  (FD_CLOEXEC)  flag on the new file descriptor.  See the description of the O_CLOEXEC flag in open(2) for reasons why this may be useful.

NiLuJe:
It would, but it might not be supported by the crappy old kernels found on our oldest devices (and it might actually make the call fail instead of silently ignoring the unknown flag, I can't recall).

So, using fcntl is, unfortunately, a portability necessity.

NiLuJe:
If I check my notes from the last time I looked into it, as far as Kobo is concerned, it *would* be safe, the only CLOEXEC compat issue I'm aware of is the lack of `accept4`; open/socket/inotify are OK.

I *know* that open is *NOT* okay on legacy Kindle/PB, though, so I'm assuming that's also true of socket there ;).

thirdparty/luasocket/CMakeLists.txt
```diff
@@ -17,6 +17,11 @@ list(APPEND PATCH_CMD COMMAND sh -c "\"\$@\" src/*" --
 # Don't enforce CFLAGS.
 list(APPEND PATCH_CMD COMMAND ${ISED} "s|-O2 -ggdb3 ||g" src/makefile)
 
+# Make sure socket fds are CLOEXEC on *nix
```
Frenzie, line 20: if(NOT WIN32)

Frenzie: Given your remark in https://github.com/koreader/koreader-base/pull/1827#issuecomment-2174289775 ;-)

NiLuJe: It's already patching a file that's never ever even built on windows, so it's perfectly harmless ;).

pazos:
I would like to revert this on android. See the linked issue. I will check tomorrow that reverting does fix the issue.

NiLuJe:
Could be interesting to try to just skip the CLOEXEC patch on Android?

In case bionic did a bionic...

PR #1828, opened 2024-06-18
Minor third-party bumps

NiLuJe:
Nothing major in LuaJIT
And LPeg & LuaJSON because I noticed the update on Gentoo ;p.

I'm not even, sure we actually have anything that still uses laujson in front...


NiLuJe:
> I'm not even, sure we actually have anything that still uses luajson in front...

Oh, quite a lot of stuff actually...

NiLuJe:
> > I'm not even, sure we actually have anything that still uses luajson in front...
> 
> Oh, quite a lot of stuff actually...

And since we like shipping thirty different JSON parsers, I'll add that `dksjon` looks like a slightly less abandoned variant ;).

NiLuJe:
> And since we like shipping thirty different JSON parsers, I'll add that `dksjon` looks like a slightly less abandoned variant ;).

Snark aside, there's a potential cleanup opportunity to drop luajson in favor of lunajson for the few things using it in main.

(IIRC, we still need lpeg for lua-spore, though).

pazos:
> Snark aside, there's a potential cleanup opportunity to drop luajson in favor of lunajson (or, well, rapidjson even) for the few things using it in main.

Do you mean 
```lua 
require("json")
```

or yet another implementation ?

NiLuJe:
Yeah, that's what tripped me up in saying we weren't using luajson anymore: I was looking for a require("luajson"), but that's indeed  luajson ;).

PR #1832, opened 2024-06-21
mupdf improvements

benoit-pierre:
- only export what we need
- build a static library, and fold it into wrap-mupdf
- enable ICC profile support

To make sure all the required symbols are exported, the wrap-mupdf library is linked with some generated scripts using the symbols referenced in the corresponding FFI cdecl file (`ffi-cdecl/mupdf_decl.c`). This is the same technique I use on my meson branch to cut down the size of the monolibtic library build.

(The plan is to do the same with the k2pdfopt library and a static tesseract library).

Impact on code size (bss+data+text, release builds):
| android-arm64| emulator | kindlepw2 |
| ---: | ---: | ---: |
| +79.6 KB | +135.4 KB | +122.6 KB

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1832)


thirdparty/mupdf/CMakeLists.txt
```diff
@@ -126,21 +102,17 @@ set(MAKE_CMD
     # Build out of tree.
     # NOTE: need "shared" somewhere in the path,
     # or the main makefile will raise an error.
```
NiLuJe, line 99: Outdated comment ;)

PR #1834, opened 2024-06-22
tesseract: disable debug fonts

benoit-pierre:
Avoid spurious error traces:
```
Error in pixReadMemTiff: function not present
Error in pixReadMem: tiff: no pix returned
Error in pixaGenerateFontFromString: pix not made
Error in bmfCreate: font pixa not made
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1834)

### DIFF
--- a/thirdparty/tesseract/no_debug_fonts.patch
+++ b/thirdparty/tesseract/no_debug_fonts.patch
@@ -1,0 +1,11 @@
+--- i/src/ccstruct/debugpixa.h
++++ w/src/ccstruct/debugpixa.h
+@@ -14,7 +14,7 @@
+   // TODO(rays) add another constructor with size control.
+   DebugPixa() {
+     pixa_ = pixaCreate(0);
+-#ifdef TESSERACT_DISABLE_DEBUG_FONTS
++#if 1//def TESSERACT_DISABLE_DEBUG_FONTS
+     fonts_ = NULL;
+ #else
+     fonts_ = bmfCreate(nullptr, 14);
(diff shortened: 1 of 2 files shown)


thirdparty/tesseract/no_debug_fonts.patch
```diff
@@ -0,0 +1,11 @@
+--- i/src/ccstruct/debugpixa.h
++++ w/src/ccstruct/debugpixa.h
+@@ -14,7 +14,7 @@
+   // TODO(rays) add another constructor with size control.
+   DebugPixa() {
+     pixa_ = pixaCreate(0);
+-#ifdef TESSERACT_DISABLE_DEBUG_FONTS
++#if 1//def TESSERACT_DISABLE_DEBUG_FONTS
```
Frenzie, line 8: Patching feels a bit convoluted compared to just setting it?

benoit-pierre: There's no knob in the build system to enable the flag… And it's only used here, so it's actually easier then having to patch `CMakeLists.txt`, or create a dedicated toolchain file to tweak the compilation flags just for tesseract.

(I almost used something like `set(CMAKE_C_FLAGS_INIT "$(strip $(CFLAGS)) $${CPPFLAGS} $${CFLAGS}")` when generating the toolchain file to make it easier to tweak those).

Frenzie: Strange, but alright.

Frenzie (approved):

NiLuJe (approved):

PR #1849, opened 2024-07-14
 cmake: enable `-Wall` when compiling koreader targets

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1849)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -2066,7 +2066,6 @@ static int getTextFromPositions(lua_State *L) {
 	}
 
 	LVDocView *tv = doc->text_view;
-	lvRect margin = tv->getPageMargins();
 
 	lvPoint startpt(x0, y0);
 	lvPoint endpt(x1, y1);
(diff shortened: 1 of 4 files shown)


cre.cpp
```diff
@@ -2724,12 +2721,10 @@ static int getPageLinks(lua_State *L) {
 	lua_newtable(L); // all links (actual entries may be less than links.length(), so, no pre-alloc)
 
 	ldomXRangeList links;
-	ldomXRangeList & sel = doc->text_view->getDocument()->getSelections();
```
poire-z, line 2727: (I was a bit worried these selections() methods may do/have wanted side effects when called, but this one just returns a reference.)
Anyway, keep this line commented out, as below we have (which may be useful):
https://github.com/koreader/koreader-base/blob/adafa0bbf8c39b5f0455c3a2754a924a0b5556fb/cre.cpp#L2801-L2805
and may be line 2732 too (it may be needed to clear past ones if we _need to visually see new ones when debugging_).

Frenzie (approved):

poire-z (approved):

PR #1863, opened 2024-07-24
mupdf: add missing `libm` dependency

benoit-pierre:
Fix loading `libwrap-mupdf` on some (KitKat era?) Android devices.

Cf. https://github.com/koreader/koreader/issues/12207.

### DIFF
--- a/thirdparty/cmake_modules/koreader_thirdparty_libs.cmake
+++ b/thirdparty/cmake_modules/koreader_thirdparty_libs.cmake
@@ -90,7 +90,7 @@ get_target_property(LUAJIT_INC luajit::luajit INTERFACE_INCLUDE_DIRECTORIES)
 declare_dependency(lunasvg::lunasvg SHARED lunasvg)
 
 # mupdf
-set(LIBRARIES)
+set(LIBRARIES m)
 if(ANDROID)
     list(APPEND LIBRARIES log)
 endif()


NiLuJe (approved):

Frenzie:
The problem definitely doesn't occur on 6+, likely not on 5+, so yeah, only KitKat.

thirdparty/cmake_modules/koreader_thirdparty_libs.cmake
```diff
@@ -90,7 +90,7 @@ get_target_property(LUAJIT_INC luajit::luajit INTERFACE_INCLUDE_DIRECTORIES)
 declare_dependency(lunasvg::lunasvg SHARED lunasvg)
 
 # mupdf
-set(LIBRARIES)
+set(LIBRARIES m)
```
Frenzie, line 93: I'd probably add a comment that this can be removed once 4.4 support is dropped.

NiLuJe: Can it? AFAICT, that DT_NEEDED entry is relevant: something *definitely* requires libm in there ;).

The fact that you can get away with in on modern systems is fairly irrelevant in my book ;).

```
┌─(ROOT@ahsoka:pts/1)────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────(~SVN/Bash/trunk)─┐
└─(0.49:19%:06:49:#)── lddtree /usr/lib64/libmupdf.so                                                                                                                                                                                                                                                                                                  ──(Wed, Jul 24)─┘
/usr/lib64/libmupdf.so (interpreter => None)
    libfreetype.so.6 => /usr/lib64/libfreetype.so.6
        libbz2.so.1 => /usr/lib64/libbz2.so.1
        libpng16.so.16 => /usr/lib64/libpng16.so.16
    libgumbo.so.2 => /usr/lib64/libgumbo.so.2
    libharfbuzz.so.0 => /usr/lib64/libharfbuzz.so.0
        libglib-2.0.so.0 => /usr/lib64/libglib-2.0.so.0
            libpcre2-8.so.0 => /usr/lib64/libpcre2-8.so.0
        libgraphite2.so.3 => /usr/lib64/libgraphite2.so.3
    libjpeg.so.62 => /usr/lib64/libjpeg.so.62
        ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2
    libmujs.so.1.3.5 => /usr/lib64/libmujs.so.1.3.5
    libz.so.1 => /usr/lib64/libz.so.1
    libjbig2dec.so.0 => /usr/lib64/libjbig2dec.so.0
    libopenjp2.so.7 => /usr/lib64/libopenjp2.so.7
    libm.so.6 => /usr/lib64/libm.so.6
    libc.so.6 => /usr/lib64/libc.so.6
```

Frenzie: I'm pretty sure it's *linked* by default, isn't it?

Frenzie: Never mind; it's listed here. But the paraphrase of what I wrote is effectively that the PR message and the commit/commit message are telling a somewhat contradictory story.

https://github.com/ArtifexSoftware/mupdf/blob/1e46ab83af1135093cce40bcad0d925d598d8d61/Makefile#L26

benoit-pierre: > mupdf: add missing `libm` dependency
>
> This was not a problem on most systems thanks to another transitive dependency on `libm` via `libharfbuzz`, but would prevent loading on KitKat era Android devices.

Better?

NiLuJe: I'm curious why this passed `make libcheck`, though?

benoit-pierre: https://github.com/koreader/koreader/issues/12098#issuecomment-2191624424

benoit-pierre: Plus the transitive dependency via libharfbuz.

PR #1865, opened 2024-07-26
mxcfb: Refactor CFA handling

NiLuJe:
Namely, drop the explicit `color` & `colortext` refresh modes, and just automatically promote reagl & full to their CFA counterparts.

~~The kernel sources should confirm that, but I have very strong suspicions that GCC16 is just GC16 + HWTCON_FLAG_CFA_MODE_G1 while GLRC16 is just GLR16 + HWTCON_FLAG_CFA_MODE_G1, so we could eventually simplify that further by also dropping those waveform mode aliases (especially since we use another CFA_MODE most of the time anyway).~~
EDIT: They came out the night after, lol. And, nope, it's a real wfm, with a dedicated REGAL_CFA regal mode...

As a bonus, this ensures manual refreshes via diagonal swipes will automatically be CFA-enabled where relevant.

(This also changes the behavior of ImageViewer, which now uses GLRC16 instead of GCC16, offering a less flashy behavior closer to how it behaves on other devices).

Also, update FBInk to enable switching to 8bpp on B&W MTK devices (or when color rendering is disabled).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1865)


Frenzie (approved):
Huh, well, okay :-)

NiLuJe (commented):
> Reviewed 2 of 2 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1865)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

NiLuJe (commented):
> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1865)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @NiLuJe)

PR #1866, opened 2024-07-27
make: improved binaries check/info rules

benoit-pierre:
- check a binary's unresolved symbols are all provided by its direct dependencies (as some dynamic loaders don't support when a symbol is provided by a transitive dependency, e.g. Android KitKat)
- don't check (intermediate) libraries under `build/xxx/cmake`
- move generated stub libraries to the staging directory (so they get cleaned with `make clean`)
- add support for checking macOS binaries
- factorize, rewrite in Python

This would have prevented https://github.com/koreader/koreader/issues/12190 & https://github.com/koreader/koreader/issues/12207, and found 3 additional issues:
- djvulibre: static library need to be linked with `libstdc++`
- djvulibre: static library need to be linker with `CoreFoundation` framework on macOS
- czmq: must be linked with `libpthread`, but is not on older (<=3.25) CMake versions

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1866)


Frenzie (approved):

benoit-pierre:
PRs for czmq & djvulibre: #1867 & #1868.

PR #1869, opened 2024-07-27
crengine: sort source list

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1869)

### DIFF
--- a/thirdparty/kpvcrlib/CMakeLists.txt
+++ b/thirdparty/kpvcrlib/CMakeLists.txt
@@ -28,42 +28,42 @@ add_subdirectory(${CRE_3RDPARTY_DIR}/antiword)
 
 add_library(crengine STATIC
     ${CRE_DIR}/qimagescale/qimagescale.cpp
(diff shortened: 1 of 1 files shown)


benoit-pierre:
The migration to XCode 15.4 (https://github.com/actions/runner-images/issues/10121) is done, the macOS arm64 glib build is broken when not hitting the ccache cache.

benoit-pierre:
Which also mean we should add the Xcode version to the cache key.

benoit-pierre:
> Which also mean we should add the Xcode version to the cache key.

Or pin Xcode.

I'll wait for #1866 to avoid conflicts, and PR a variant of https://github.com/benoit-pierre/koreader-base/commit/f3df4b8201878523db584e475e06e61768980af8.

And once #1858 is merged, I have a PR for meson support, and then an update to glib.

benoit-pierre:
Never mind, if we pin Xcode and bump `CACHE_EPOCH` on change, there's no need to add the version to the cache key, and no conflict. PR incoming.

PR #1872, opened 2024-07-28
freetype2 & harfbuzz & glib: switch to meson + replace gettext with proxy-libintl

benoit-pierre:
Next step: glib update.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1872)


poire-z:
So, that's the first PR making meson a hard requirement ? Are we all ok with that?
(I still have it not installed, not minding, just asking.)

thirdparty/glib/CMakeLists.txt
```diff
@@ -3,65 +3,38 @@ list(APPEND PATCH_FILES
...
-        list(APPEND gen_cache_config ac_cv_func_mkostemp=no)
-    endif()
-    if(ANDROID)
-        # disable getpwuid, getgrgid and timegm for > android-12
-        list(APPEND gen_cache_config ac_cv_func_posix_getpwuid_r=no)
-        list(APPEND gen_cache_config ac_cv_func_posix_getgrgid_r=no)
-        list(APPEND gen_cache_config ac_cv_func_getpwuid_r=no)
-        list(APPEND gen_cache_config ac_cv_func_getgrgid_r=no)
-        list(APPEND gen_cache_config ac_cv_func_timegm=no)
-    endif()
```
benoit-pierre, line 28: That part confused me: the comment about android-12 is from 23985d3, in 2016, so I assume 12 is for the API level? Android 3.1  (HONEYCOMB_MR1, 2011), and not Android 12 (S[now Cone], level 31, 2021). The flags themselves come from 23985d3 and 773db3f66 (via db68822f9 & e579d0eba).

benoit-pierre:
> So, that's the first PR making meson a hard requirement ? Are we all ok with that? (I still have it not installed, not minding, just asking.)

Note:
- recent glib versions only support meson
- harfuzz 9.0 dropped autotools support, and the CMake support is community maintained
- fribidi: still support autotools
- freetype2: support all 3

thirdparty/glib/CMakeLists.txt
Frenzie: Yup, that's API level 12.

https://developer.android.com/guide/topics/manifest/uses-sdk-element.html#ApiLevels

Frenzie:
> So, that's the first PR making meson a hard requirement ? Are we all ok with that?

I am, yes.

thirdparty/glib/CMakeLists.txt
benoit-pierre: On that subject, this page has IMHO a better overview: https://apilevels.com/.

cmake/CMakeLists.txt
```diff
@@ -84,8 +84,44 @@ function(setup_project NAME)
...
+    execute_process(COMMAND ${${VAR}} ${VERSION_ARG} OUTPUT_VARIABLE VERSION OUTPUT_STRIP_TRAILING_WHITESPACE)
+    string(REGEX MATCH "(^|[ \n])([0-9][.0-9]*)([\n ]|$)" VERSION "${VERSION}")
+    message(STATUS "Found ${NAME}: ${${VAR}} (found version \"${CMAKE_MATCH_2}\")")
+endfunction()
+
 # }}}.
 
+# Git. {{{
+
+find_package(Git REQUIRED)
```
Frenzie, line 104: This refers to the plain Git executable, not Git development files or libraries?

benoit-pierre: Yes, this is just to get a trace with the git version on setup, and an early error if git is missing in some corner cases (building in a docker mounted volume).

Frenzie:
I suspect `harfbuzz: switch to meson for building` should be amended with a `git rm thirdparty/harfbuzz/soversion.patch`.

Frenzie (approved):

PR #1878, opened 2024-07-29
gettext & libiconv & tar: add fallback URLs

benoit-pierre:
To another mirror, in case gnu.org is down.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1878)


benoit-pierre:
Yes, done.

PR #1885, opened 2024-08-02
bump crengine

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1885)


benoit-pierre:
Done in 2 commits so there's no need to manually patch stuff when bisecting.

PR #1886, opened 2024-08-02
bump glib

benoit-pierre:
This add ~70-110 KB to sdcv's code size.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1886)


Frenzie (approved):
I'm a bit surprised a patch is needed for Android in 2024.

benoit-pierre:
Actually we might not need it since 2.80.1: https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3987.

Frenzie:
Sounds good. :-)

benoit-pierre:
OK, so I can't reproduce the initial issue: I tried downgrading to 2.80.0 with no patch, and it still works. So let's just assume everything is fine now, and drop the whole patch (the `epoll_create` part of the patch [can be dropped too](https://gitlab.gnome.org/GNOME/glib/-/merge_requests/2868)).

poire-z:
```
FAILED: /space/kobo/koreader/base/build/x86_64-linux-gnu-debug/thirdparty/glib/stamp/configure
[...]
The Meson build system
Version: 1.0.1
Source dir: /space/kobo/koreader/base/build/x86_64-linux-gnu-debug/thirdparty/glib/source
Build dir: /space/kobo/koreader/base/build/x86_64-linux-gnu-debug/thirdparty/glib/build
Build type: native build

../source/meson.build:1:0: ERROR: Meson version is 1.0.1 but project requires >= 1.2.0

A full log can be found at /space/kobo/koreader/base/build/x86_64-linux-gnu-debug/thirdparty/glib/build/meson-logs/meson-log.txt
make: *** [base/Makefile:89: base-all] Error 1
```
Debian 12's meson is meson 1.0.1-5.
Anyway to patch something so we not need to be bleeding edge?

Frenzie:
1.5 is in backports btw. I don't know if you're familiar with backports, so here's a link: https://wiki.debian.org/Backports

poire-z:
OK, got the .deb and did dpkg -i it.

Btw, after make clean, I have previously-downloaded tar.gz (including multiple past versions) in base/thirdparty/*/downloads/ . Any official way to prune these (at least older versions)?

I also see `some` base/thirdparty/*/sources/ (not all have it). What is it ? Can/should it be cleaned ?

benoit-pierre:
> Btw, after make clean, I have previously-downloaded tar.gz (including multiple past versions) in base/thirdparty/*/downloads/ . Any official way to prune these (at least older versions)?

Nothing has changed in that respect, those only get removed when nuking everything with `make distclean`.

> I also see `some` base/thirdparty/*/sources/ (not all have it). What is it ? Can/should it be cleaned ?

That's the original clones for those external projects using git.

PR #1898, opened 2024-08-06
cmake: fix `find_executable(… RESERVED)`

benoit-pierre:
Manually handle `RESERVED` keyword, since `find_program(… RESERVED)` is only supported with CMake >= 3.18.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1898)


Frenzie:
If desired we could also load for example 3.25 onto the image. I think I may have previously said something about preferring to stay with some lower version such as 3.10, but that would've been when Debian Bookworm and/or Ubuntu 24.04 wasn't out yet or only barely.

benoit-pierre:
There's not much to gain now that CMake's `ExternalProject` is not used anymore and we only support ninja.

Aside from this issue with `find_program(…)`, we only have 2 other workarounds / version specific tweaks:
- writing a file and including it instead of using `cmake_language(CALL …)` (CMake >= 3.18)
- using `copy_directory` instead of `copy_directory_if_different` instead of `` for CMake < 3.26

Frenzie:
Alright, works for me.

PR #1899, opened 2024-08-07
luajit: add android luajit-launcher patch

benoit-pierre:
Cf. https://github.com/koreader/koreader/pull/12285.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1899)


Frenzie:
I'm not entirely sure if that patch is still relevant btw. It could be it was mainly a 2.3 problem. But it's hard to test; I couldn't reproduce it myself even on 2.3.

Frenzie:
Also see this somewhat related discussion: https://github.com/koreader/koreader/issues/4137#issuecomment-410677398

NiLuJe:
Oh, it very much still relevant, the virtual address space mapping layout hasn't changed on Android AFAIK ;).

NiLuJe:
Speaking of, this doesn't include the debug patch, which is really helpful in double-checking that it's behaving as expected ;).

NiLuJe (approved):

PR #1900, opened 2024-08-11
drop `-ffast-math` from all compilation flags

benoit-pierre:
Cause too many issues: it makes more sense to selectively enable it for code where it's known to really improve performance than enabling it for all code and having to progressively blacklist it as non-obvious problems crop up.

Cf.:
- [previous discussion](https://github.com/koreader/koreader-base/pull/1750#issuecomment-2122924884)
- [Android issue](https://github.com/koreader/koreader/issues/12332)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1900)


Frenzie (approved):

NiLuJe (approved):
Alas, poor Yorick ;).

PR #1902, opened 2024-08-12
cmake: fix possible git external project issue

benoit-pierre:
When using the `exec $SHELL` directive in the plan for `git rebase --interactive`, the spawned shell will inherit the `$GIT_DIR` environment variable. This breaks most `git` invocations when downloading / checking out a git external project (like enabling a sparse-checkout in base, wiping whole parts of the tree, …).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1902)


benoit-pierre:
And changing the origin remote…

PR #1903, opened 2024-08-12
bump crengine

benoit-pierre:
- https://github.com/koreader/crengine/pull/584
- https://github.com/koreader/crengine/pull/598


poire-z:
Please bump this in front soon.

PR #1905, opened 2024-08-12
make: improved binaries check

benoit-pierre:
Add `--glibc-version-max` to gate the maximum supported GLIBC version: bincheck will fail on libraries that link with more recent symbols.


benoit-pierre:
And another regression caught:
```
/home/ko/project/build/sftp-server
  RPATH     : $ORIGIN:$ORIGIN/libs
  NEEDED    : librt.so.1
  NEEDED    : libc.so.6
  UNRESOLVED:
    utimensat@GLIBC_2.6
```
Looks like tar is not the only potential user.

benoit-pierre:
As well as:
```
/home/ko/project/build/libs/libkoreader-input.so
  SONAME    : libkoreader-input.so
  RPATH     : $ORIGIN:$ORIGIN/libs
  NEEDED    : libdl.so.2
  NEEDED    : libinkview.so
  NEEDED    : libpthread.so.0
  NEEDED    : libc.so.6
  UNRESOLVED:
    pipe2@GLIBC_2.9
```
and
```
libzmq.so.4
  FILE      : /home/ko/project/build/libs/libzmq.so.4
  SONAME    : libzmq.so.4
  RPATH     : $ORIGIN:$ORIGIN/libs
  NEEDED    : libpthread.so.0
  NEEDED    : librt.so.1
  NEEDED    : libstdc++.so.6
  NEEDED    : libgcc_s.so.1
  NEEDED    : libc.so.6
  UNRESOLVED:
    eventfd@GLIBC_2.7
```

benoit-pierre:
That's for pocketbook, no issue with `koreader-input` locally when building for `kindle-legacy`.

Which is missing from Circle CI's matrix, how about adding it?

benoit-pierre:
Tentative patch for the remaining pocketbook issue:
```diff
 input/input.c | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git i/input/input.c w/input/input.c
index 5414d301e..764afa3af 100644
--- i/input/input.c
+++ w/input/input.c
@@ -125,7 +125,7 @@ static int openInputDevice(lua_State* L)
     if (!strcmp("fake_events", inputdevice)) {
         // Special case: the power slider for Kindle and USB events for Kobo.
         int pipefd[2U];
-#if defined(KINDLE_LEGACY)
+#if defined(KINDLE_LEGACY) || defined(POCKETBOOK)
         // pipe2 requires Linux 2.6.27 & glibc 2.9...
         if (pipe(pipefd) == -1) {
             return luaL_error(L, "Cannot create fake event generator communication pipe (pipe(): %s)", strerror(errno));
```
@NiLuJe?

NiLuJe:
> Tentative patch for the remaining pocketbook issue:

Yup, lgtm!

(Note that we no longer *actually* use the Lua/C input module on PB, but keeping it actually functional usually costs us nothing ;p).

NiLuJe:
For context, this sort of stuff *should* mostly only happen on kindle-legacy & pocketbook, because they run glibc 2.5, but the toolchain is built against 2.9 (because good luck building 2.5, especially with even a halfway recent GCC version).

For the rest of our targets, the ct-ng config will target the "proper" glibc version directly. Note that even that is a moving target, because vendors will sometime miraculously update it ;).

e.g., kindle5 & kindlepw2 target 2.12.2, but it was actually updated to 2.19 on FW 5.6.5. (Ad it was probably updated again for the hf float, but I lost track).

Similarly, kobo targets 2.15, but it was updated to 2.19 (surprise!) on FW 4.6. And will be (massively) updated to 2.3x (can't recall the exact one right now) on 5.0.

Frenzie:
> Which is missing from Circle CI's matrix, how about adding it?

Sure, makes sense. It's likely to overlap very strongly with PB but it's not identical.

PR #1272, opened 2020-12-27
Revert "android: unversioning shared libraries"

pazos:
Reverts koreader/koreader-base#1266

Because https://github.com/koreader/koreader/pull/7040#issuecomment-751459350, an unversioned libs in filesDir may clash with system libs.

I will apply it again on the future, once underlying issues were ironed out.

---
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1272)

PR #1907, opened 2024-08-17
android: don't version shared objects

benoit-pierre:
- autotools doesn't by default
- ditto with meson
- gradle will ignore versioned libraries, and not include them in the APK

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1907)

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -590,17 +590,20 @@ CFLAGS:=$(BASE_CFLAGS) $(QFLAGS) $(CSTD_FLAGS)
 CXXFLAGS:=$(BASE_CFLAGS) $(QFLAGS) $(CXXSTD_FLAGS)
 
 ifndef DARWIN
-	# To prefer our shipped libraries over the system's,
-	# we set a couple rpath via an include file to make sure it goes through buildsystem indirection unscathed...
-	# NOTE: We enforce DT_RPATH over DT_RUNPATH because DT_RUNPATH is *not* honored for *transitive* dependencies,
-	#       which can lead to hilarious catastrophe if such a dependency pulls a SONAME identical to something we ship,
-	#       because it will *not* load ours, but still let ours be loaded after that...
-	#       (Assuming no LD_LIBRARY_PATH are set, since the search order is basically RPATH -> LD_LIBRARY_PATH -> RUNPATH).
-	#       c.f., https://github.com/koreader/koreader-base/pull/1638#issuecomment-1636725691
-	#       And https://bugs.launchpad.net/ubuntu/+source/eglibc/+bug/1253638/comments/5 for a nice recap.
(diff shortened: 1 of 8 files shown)


Frenzie:
I believe the reason I made them versioned was because unversioned caused the extremely obvious kinds of problems you might expect, particularly because Bionic doesn't support RPATH in conjunction with issues pertaining to SONAME and NEEDED.

If we're lucky I'm wrong in predicting that this will cause terrible breakage primarily on devices we don't own, but if I'm right it should be possible to alternatively fool idiotic Gradle with filenames like libharfbuzz_0.so.

NiLuJe:
I imagine as long as the actual in-ELF-header SONAME & SOVER match and this doesn't just change the filename, we should mostly be okay?

benoit-pierre:
Yes, and we use our own loader anyway.

pazos:
Came from https://github.com/koreader/koreader/issues/12348#issuecomment-2299638037

Just to point that I've tried it in the past and reverted it in https://github.com/koreader/koreader-base/pull/1272 because https://github.com/koreader/koreader/pull/7040#issuecomment-751459350

I can't state if something above is true or not, but I'm willing to trust my old self.

NiLuJe:
Random stupid idea: does it behave any better if you load *everything* in the global namespace in dl.lua?

benoit-pierre:
Nope, but I'm not sure it's even correctly supported:
```lua
ffi.cdef[[
void *dlopen(const char *filename, int flag);
char *dlerror(void);
const static int RTLD_LOCAL    = 0;
const static int RTLD_LAZY     = 0x00001;
const static int RTLD_NOLOAD   = 0x00004;
const static int RTLD_NODELETE = 0x01000;
]]
-- There's a bit of heinous hackery going on on 32-bit ABIs with RTLD_NOW & RTLD_GLOBAL...
-- c.f., https://android.googlesource.com/platform/bionic/+/refs/heads/master/libc/include/dlfcn.h
if ffi.abi("64bit") then
    ffi.cdef[[
    const static int RTLD_NOW      = 0x00002;
    const static int RTLD_GLOBAL   = 0x00100;
    ]]
else
    ffi.cdef[[
    const static int RTLD_NOW      = 0x00000;
    const static int RTLD_GLOBAL   = 0x00002;
    ]]
end
```

NiLuJe:
It should, that comment was just about bionic mangling the constants, but there's another comment later that mentions old API levels not using it automatically for *some* things that really really should; and that vaguely rings a bell.

NiLuJe:
> mangling the constants depending on the bitness of the arch

Or, as the upstream comment says, "`/* LP32 has historical ABI breakage. */`", lol.

PR #1911, opened 2024-08-23
ffi/util: remove dead code

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1911)


poire-z (approved):
The less traces of "Windows" there are, the less likely one will be tempted to assume or make it work :)

Frenzie:
The rest of the code all uses `if ffi.os == "Windows"` directly.

PR #1912, opened 2024-08-24
ffi: avoid loading 2 different zlib libraries

benoit-pierre:
Avoid loading the system version instead of ours.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1912)


Frenzie:
How does that relate to #1907?

Frenzie (approved):

NiLuJe (approved):

benoit-pierre:
> How does that relate to #1907?

It's unrelated.

PR #1919, opened 2024-08-31
cre.cpp: fix findAllText() context at end of book, bump crengine

poire-z:
Fix https://github.com/koreader/koreader/issues/12435#issuecomment-2322752801

(Tried first to fix the wrap around in crengine, but it was not obvious... So, local fix it is.)

Also bump crengine:
fix xml parser regression (infinite loop) https://github.com/koreader/crengine/pull/602

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1919)


Frenzie (approved):
Hmm, ok.

PR #1930, opened 2024-09-08
luajit: patch out library from pkg-config entry when not used

benoit-pierre:
To prevent Lua modules from linking with it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1930)

### DIFF
--- a/thirdparty/luajit/CMakeLists.txt
+++ b/thirdparty/luajit/CMakeLists.txt
@@ -91,6 +91,9 @@ endif()
 
 if(USE_LUAJIT_LIB)
     append_shared_lib_install_commands(INSTALL_CMD luajit)
+else()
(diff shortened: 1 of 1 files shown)


benoit-pierre:
In preparation for using pkg-config to handle LuaJIT's compilation flags in some external projects (`luajit.pc` is currently not used).

NiLuJe:
Wasn't there a thing where Android (and possibly macOS) needed to be explicitly linked to the luajit lib for... reasons?

benoit-pierre:
Yes, which is why it's only patched out when we don't use the luajit library.

NiLuJe:
Oh, duh'. Long day, sorry ;).

PR #1933, opened 2024-09-19
PocketBook: import adjustAreaWithSaturationBrightnessAndDithering for color devices

dmalinovsky:
My hope is that playing with `dithering` parameter may help with https://github.com/koreader/koreader/issues/12527.

@rjd22, what do you think?

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1933)


rjd22:
I'm okay with it so you can play around.

rjd22 (approved):

Frenzie:
Could you suggest a slightly better commit message? `PocketBook: import adjustAreaWithSaturationBrightnessAndDithering for color devices`?

dmalinovsky:
> Could you suggest a slightly better commit message? `PocketBook: import adjustAreaWithSaturationBrightnessAndDithering for color devices`?

Will the PR title change work?

rjd22:
No you need to change the commit message. You can do so by doing a git commit amend (google it) and force pusing.

NiLuJe:
> Will the PR title change

Since we squash by default, yes, that'll do ;).

rjd22:
> > Will the PR title change work?
> 
> Since we squash by default, yes, that'll do ;).

That works? Sorry then. Ignore my message.

NiLuJe:
> That works? Sorry then. Ignore my message.

I mean, we can choose the exact title & log when squashing, so, sort of ;p.

(There might have been a time where GH actually preferred the PR title over the commits, but that's not the case right now).

Frenzie:
> Will the PR title change work?

A "that's fine" would've sufficed but I needed to know it wasn't incorrect in some way.

PR #1938, opened 2024-09-22
ffi/loadlib: add module

benoit-pierre:
Provide a `ffi.loadlib` helper, replacing `ffi.load` and `util.ffiLoadCandidates` for loading our libraries.

Prerequisite for #1920 & https://github.com/koreader/koreader/pull/12545.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1938)

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -1037,6 +1037,7 @@ eval "$$($(LUAROCKS_BINARY) path)";
 export TESSDATA_DIR="$$PWD/data";
 ./luajit -e 'require "busted.runner" {standalone = false}' /dev/null
 --exclude-tags=notest
+--helper=ffi/loadlib.lua
 --output=gtest
 --sort-files
 "$$@";
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -2676,7 +2676,7 @@ BB.TYPE_TO_BPP = {
 
 BB.has_cblitbuffer = false
 -- Load the C blitter, and default to using it if available
-BB.has_cblitbuffer, cblitbuffer = pcall(ffi.load, "blitbuffer")
+BB.has_cblitbuffer, cblitbuffer = pcall(ffi.loadlib, "blitbuffer")
 if BB.has_cblitbuffer then
     -- If we can, assume we'll want to use it
     use_cblitbuffer = true
--- a/ffi/crypto.lua
+++ b/ffi/crypto.lua
@@ -7,15 +7,7 @@ LuaJIT FFI wrapper for libcrypto (OpenSSL).
 local ffi = require("ffi")
 require("ffi/crypto_h")
 
-local libcrypto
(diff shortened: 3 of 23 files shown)


NiLuJe (commented):
> Reviewed 23 of 23 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1938)* status: all files reviewed, 1 unresolved discussion

___
*[`ffi/jpeg.lua` line 19 at r1](https://reviewable.io/reviews/koreader/koreader-base/1938#-O7TyGmE7bf9Ib8UShWL:-O7TyGmE7bf9Ib8UShWM:bpwns73) ([raw file](https://github.com/koreader/koreader-base/blob/880729772afdaf948f2653f35d972ac204dc2756/ffi/jpeg.lua#L19)):*
> ```Lua
> -- backward compatible: the major & patch numbers are always 0, and when
> -- a new API version is made available, the minor number is incremented.
> local turbojpeg = ffi.loadlib("turbojpeg", "0.3.0", "turbojpeg")
> ```

The idea behind the third argument is to fall-back to an unversioned library, right?

I wouldn't mind proper documentation about that in `ffi/loadlib.lua` itself ;).

(Specifically as far as this sort of stuff is concerned, as well as the `util.loadSDL2` workflow where you can just pass a list of `name, version` tuples to a *single* call and they'll get chunked automatically).

benoit-pierre:
> > Reviewed 23 of 23 files at r1, all commit messages.
> > _[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1938)_ status: all files reviewed, 1 unresolved discussion
> 
> _[`ffi/jpeg.lua` line 19 at r1](https://reviewable.io/reviews/koreader/koreader-base/1938#-O7TyGmE7bf9Ib8UShWL:-O7TyGmE7bf9Ib8UShWM:bpwns73) ([raw file](https://github.com/koreader/koreader-base/blob/880729772afdaf948f2653f35d972ac204dc2756/ffi/jpeg.lua#L19)):_
> 
> > ```lua
> > -- backward compatible: the major & patch numbers are always 0, and when
> > -- a new API version is made available, the minor number is incremented.
> > local turbojpeg = ffi.loadlib("turbojpeg", "0.3.0", "turbojpeg")
> > ```
> 
> The idea behind the third argument is to fall-back to an unversioned library, right?

Actually, if the goal is to maybe not ship our libraries, for projects we don't patch (e.g. to use system libraries when building Linux packages), then the code should be amended to `ffi.loadlib("turbojpeg", "0.3.0") or ffi.load("turbojpeg")`, so we **do fallback** to the unversioned system library if we're not shipping ours and the system version is different: if the version is too old, the symbols we need will be unresolved, otherwise, any newer version should be backward compatible.
 
> I wouldn't mind proper documentation about that in `ffi/loadlib.lua` itself ;).

See #1939.

PR #1939, opened 2024-09-23
ffi/loadlib: add some documentation

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1939)

### DIFF
--- a/ffi/loadlib.lua
+++ b/ffi/loadlib.lua
@@ -1,3 +1,21 @@
+--[[--
+Helper for loading native libraries.
+
+**Example:**
+
+    local sdl = ffi.loadlib("SDL2-2.0", 0, "SDL2-2.0", nil, "SDL2", nil)
+
+Will search `lib_search_path` for the following candidates on Linux:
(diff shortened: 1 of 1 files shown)


ffi/loadlib.lua
```diff
@@ -31,6 +36,14 @@ ffi.load = function(lib, global)
     return ffi_load(lib, global)
 end
 
+-- Example usage: `ffi.loadlib("SDL2-2.0", 0, "SDL2-2.0",nil, "SDL2", nil)`,
```
Frenzie, line 39: If you put an `@module ffi.loadlib` at the top you could also write this as an LDoc function comment. See https://github.com/koreader/koreader/blob/ca569c1355d14ad023eb1bcad7b995ae027e423d/frontend/gettext.lua#L295-L323 for an example.

Frenzie: (Or we can just merge it as is, but I find it nice to have; it also shows up in IDEs.)

benoit-pierre: Done.

PR #1940, opened 2024-09-23
ffi/jpeg: fix turbojpeg library load fallback

benoit-pierre:
We want to fallback on the unversioned system library if we're not shipping ours and the system version is different: if the version is too old, the symbols we need will be unresolved, otherwise, any newer version should be backward compatible.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1940)

### DIFF
--- a/ffi/jpeg.lua
+++ b/ffi/jpeg.lua
@@ -16,7 +16,7 @@ require("ffi/turbojpeg_h")
 -- The turbojpeg library symbols are versioned, so it should always be
 -- backward compatible: the major & patch numbers are always 0, and when
 -- a new API version is made available, the minor number is incremented.
-local turbojpeg = ffi.loadlib("turbojpeg", "0.3.0", "turbojpeg")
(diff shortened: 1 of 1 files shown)


NiLuJe:
This isn't a protected call, wouldn't the first loadlib incantation failing just throw?

benoit-pierre:
D'oh! Proving once again that untested code is broken code…

NiLuJe:
There's probably a similar issue for the loadSDL2 chain, I imagine?

benoit-pierre:
I don't think so. For SDL2, the multiple candidates is to account for the fact that we don't always ship our library, and need to look into the brew prefix on macOS. So on macZOS: look for `libSDL2-2.0.0.dylib` in our libs / brew, then `libSDL2-2.0.dylib`, and finally `libSDL2.dylib`; `ffi.load` the first match or the fallback (`libSDL2-2.0.0.dylib`).

Frenzie:
> wouldn't the first loadlib incantation failing just throw

I had assumed it was tested and that it returned false if it couldn't find anything (as distinct from trying something with the wrong headers or whatnot).

PR #1941, opened 2024-09-23
ffi/jpeg: fix turbojpeg library load fallback

benoit-pierre:
Fix the fix…

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1941)


Frenzie:
https://github.com/koreader/koreader-base/pull/1940#issuecomment-2369037035

PR #1946, opened 2024-09-30
build sdl2 for appimage

benoit-pierre:
Getting an updated version fix an issue with some erroneous   key events being processed when using a X11 window manager:
 - start KOReader on a virtual desktop, say the first one
 - switch to another desktop, and back using the WM shortcut  (`Super+1`)
 - on getting back focus, KOReader somehow end up processing a `1` key press event, jumping to the beginning of the current book

Note: will need an updated `koappimage` docker image.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1946)


Frenzie (approved):

Frenzie (approved):
@NiLuJe any comments?

PR #1948, opened 2024-10-03
macos: fix wrap-mupdf library extension

benoit-pierre:
Cf. https://github.com/koreader/koreader/issues/12546#issuecomment-2391710914.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1948)


Frenzie:
Thanks!

Frenzie:
Did you also see the others in https://github.com/koreader/koreader/issues/12546#issuecomment-2391718387 ?

benoit-pierre:
Yes, but those are Lua modules, so they get a `.so` extension.

Frenzie:
They do? Hm, ok.

benoit-pierre:
Yeah, apart from Windows, LuaJIT default package path uses `?.so`.

PR #1954, opened 2024-10-04
libk2pdfopt: fix macOS build

benoit-pierre:
Make sure we use our libpng headers, and not an older framework version.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1954)


dmalinovsky:
@benoit-pierre, thanks for fixing macOS builds! Now it's the first time I can actually build and run KOReader on my macOS, makes it much easier testing the changes.

PR #1956, opened 2024-10-04
self sufficient testsuite + macos ci run tests

benoit-pierre:
I wanted to run the tests on the macOS CI, but then I remembered lua5.1 is not supported by brew anymore, and figured I might as well make our testsuite self-sufficient (like with the meson branch): install luarocks to staging and the necessary Lua packages to a dedicated tree under the output `spec` directory.

This led to discovering 2 macOS specific issues: see #1953 & #1954.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1956)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -193,4 +193,7 @@ jobs:
 
       # }}}
 
+      - name: Test
+        run: ./utils/fake_tty.py make --assume-old=all test
+
 # vim: foldmethod=marker foldlevel=0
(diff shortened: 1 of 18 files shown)


Frenzie:
We used to do something like that with https://github.com/torch/luajit-rocks btw. (I'm just mentioning it for potential onlookers who might be interested in random historical remarks.)

NiLuJe:
Random remark about macOS: the last few LuaJIT commits upstream deal with XCode 16 & macOS 15 stupidity, so we might wan tto bump LuaJIt again, too ;).

(Will do so tomorrow if nobody beats me to it, as I'm barely awake rn)

benoit-pierre:
The new macOS job using XCode 16.0 would hang during the build, but pass with the LuaJIT update from #1957, so it clearly has an impact.

PR #1964, opened 2024-10-14
dropbear: update to 2024.85

benoit-pierre:
I simplified the password-less patch, but do we want to always keep allowing connection from 127.0.0.1? That's not advertised on the tin.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1964)


thirdparty/dropbear/CMakeLists.txt
```diff
@@ -1,22 +1,45 @@
...
+set(LOG_COMMANDS 0)
+# Paths.
+set(DEFAULT_PATH "/sbin:/usr/sbin:/bin:/usr/bin")
+set(DROPBEAR_PIDFILE "settings/SSH/dropbear.pid")
+if(PLATFORM_KOREADER_DIR)
+    set(DBSCP_PATH ${PLATFORM_KOREADER_DIR})
+    set(SFTPSERVER_PATH "${PLATFORM_KOREADER_DIR}/sftp-server")
+endif()
+# Keys.
+set(DSS_PRIV_FILENAME "settings/SSH/dropbear_dss_host_key")
```
pazos, line 29: Just commented https://github.com/koreader/koreader/issues/12637#issuecomment-2411651938

Maybe `/mnt/public/.ssh` for cervantes?

benoit-pierre: I used the same value as before, if we change it, it'll break some existing configs.

pazos: Ok, nvm as long as it can be made passwordless.

Frenzie (approved):
lgtm!

NiLuJe (approved):
> Reviewed 14 of 15 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1964)* status: all files reviewed, 1 unresolved discussion (waiting on @benoit-pierre)

benoit-pierre:
> > do we want to always keep allowing connection from 127.0.0.1? That's not advertised on the tin
> 
> I don't recall why I added this, but it's in the original Kindle USBNet patch, too, FWIW. And, no, I can't think of an actual use-case for it right now ;D.

OK, so all in favor of simplifying the code to this:
```c
/*
 * Hack. Always auth successfully if we were launched with the -n flag.
 */
if (svr_opts.nopasschk) {
	dropbear_log(LOG_WARNING, "Password-less mode enabled, you're in!");
	send_msg_userauth_success();
	return;
}
```
?

Frenzie:
Perhaps it's useful for certain types of automation? It wasn't there until @NiLuJe added it in https://github.com/koreader/koreader-base/pull/706 so, er, fine by me.

Frenzie:
Where did those originate?

NiLuJe:
The USBNet hack on Kindle, way back when:

```
25861   │ ------------------------------------------------------------------------
25862   │ r6866 | NiLuJe | 2010-09-26 00:25:07 +0200 (Sun, 26 Sep 2010) | 3 lines
25863   │ 
25864   │ Kindle Hacks:
25865   │     * Add a dropbear patch to skip passwd check with a cmdline switch
25866   │ 
25867   │ ------------------------------------------------------------------------
```

Which got me curious, as that one doesn't have the weird localhost case...

NiLuJe:
> Which got me curious, as that initial one doesn't have the weird localhost case...

```
25635   │ ------------------------------------------------------------------------
25636   │ r7109 | NiLuJe | 2010-11-07 13:56:13 +0100 (Sun, 07 Nov 2010) | 5 lines
25637   │ 
25638   │ Kindle Hacks:
25639   │     * X-TC:
25640   │         * Passwordless login from localhost for dropbear, to make
25641   │           nice with AjaxTerm.
25642   │ 
25643   │ ------------------------------------------------------------------------
```

I don't even remember what AjaxTerm was, but it was definitely a Kindle-specific thingy, so, whee, off it goes! ;o).

Frenzie:
> I don't even remember what AjaxTerm was, exactly, but it was definitely a weird Kindle-specific thingy, so, whee, off it goes! ;o).

Nah, it's in the Debubuntu repos — or once was. So it's basically as I said. :-) It'd also allow you to use the Nickel browser to access the terminal with AjaxTerm or something like it for example. https://github.com/butlerx/wetty claims to be better.

PR #1965, opened 2024-10-16
leptonica: update to 1.85.0

benoit-pierre:
- https://github.com/DanBloomberg/leptonica/compare/1.84.1...1.85.0
- http://www.leptonica.org/source/version-notes.html

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1965)


benoit-pierre:
Let's wait a little before merging, as the release tag has already been forcibly updated 3 times…

PR #1970, opened 2024-11-04
Addressed a PocketBook SDK v6 breaking change related to HID events

tommiv:
Discussion is in https://github.com/koreader/koreader/issues/12703

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1970)

### DIFF
--- a/ffi/inkview_h.lua
+++ b/ffi/inkview_h.lua
@@ -749,9 +749,15 @@ static const int EVT_FOCUS = 36;
 static const int EVT_UNFOCUS = 37;
 static const int EVT_ACTIVATE = 38;
 static const int EVT_MTSYNC = 39;
-static const int EVT_TOUCHUP = 40;
-static const int EVT_TOUCHDOWN = 41;
-static const int EVT_TOUCHMOVE = 42;
+// NB: these constants reflect the current (v6) state of SDK.
+// A breaking change was introduced there.
+static const int EVT_KEYPRESS_EXT = 40;
+static const int EVT_KEYRELEASE_EXT = 41;
+static const int EVT_KEYREPEAT_EXT = 42;
+static const int EVT_TOUCHUP = 47;
+static const int EVT_TOUCHDOWN = 48;
+static const int EVT_TOUCHMOVE = 49;
+// end;
 static const int EVT_QN_MOVE = 51;
 static const int EVT_QN_RELEASE = 52;
 static const int EVT_QN_BORDER = 53;
--- a/ffi/input_pocketbook.lua
+++ b/ffi/input_pocketbook.lua
@@ -161,16 +161,16 @@ end
 local function getMajorSoftwareVersion()
     -- format is $model.$major.$minor.$build, like "U743g.6.8.4143"
     local fwVersion = inkview.GetSoftwareVersion()
-    
+
     local counter = 0
-    
+
     for segment in string.gmatch(fwVersion, "[^.]+") do
-      if counter == 1 then
-        local major = tonumber(segment) or 0
-        return major
-      end
-      
-      counter = counter + 1
+        if counter == 1 then
+            local major = tonumber(segment) or 0
+            return major
+        end
+
+        counter = counter + 1
     end
 
     return 0
--- a/ffi/input_pocketbook.lua
+++ b/ffi/input_pocketbook.lua
@@ -64,6 +64,9 @@ local pb_event_map = {
     [C.EVT_KEYUP] = "EVT_KEYUP",
     [C.EVT_KEYRELEASE] = "EVT_KEYRELEASE",
     [C.EVT_KEYREPEAT] = "EVT_KEYREPEAT",
+    [C.EVT_KEYPRESS_EXT] = "EVT_KEYPRESS_EXT",
+    [C.EVT_KEYRELEASE_EXT] = "EVT_KEYRELEASE_EXT",
+    [C.EVT_KEYREPEAT_EXT] = "EVT_KEYREPEAT_EXT",
     [C.EVT_POINTERUP] = "EVT_POINTERUP",
     [C.EVT_POINTERDOWN] = "EVT_POINTERDOWN",
     [C.EVT_POINTERMOVE] = "EVT_POINTERMOVE",
@@ -240,15 +243,15 @@ local function translateEvent(t, par1, par2)
             genEmuEvent(C.EV_SYN, C.SYN_REPORT, 0)
             setContactDown(0, false)
         end
-    elseif t == C.EVT_KEYDOWN then
+    elseif t == C.EVT_KEYDOWN or t == C.EVT_KEYPRESS_EXT then
         updateTimestamp()
 
         genEmuEvent(C.EV_KEY, par1, 1)
-    elseif t == C.EVT_KEYREPEAT then
+    elseif t == C.EVT_KEYREPEAT or t == C.EVT_KEYREPEAT_EXT then
         updateTimestamp()
 
         genEmuEvent(C.EV_KEY, par1, 2)
-    elseif t == C.EVT_KEYUP then
+    elseif t == C.EVT_KEYUP or t == C.EVT_KEYRELEASE_EXT then
         updateTimestamp()
 
         genEmuEvent(C.EV_KEY, par1, 0)
--- a/ffi/input_pocketbook.lua
+++ b/ffi/input_pocketbook.lua
@@ -158,6 +158,24 @@ local function setContactDown(slot, down)
     return false
 end
 
+local function getMajorSoftwareVersion()
+    -- format is $model.$major.$minor.$build, like "U743g.6.8.4143"
+    local fwVersion = inkview.GetSoftwareVersion()
+    
+    local counter = 0
+    
+    for segment in string.gmatch(, "[^.]+") do
+      if counter == 1 then
+        local major = tonumber(segment) or 0
+        return major
+      end
+      
+      counter = counter + 1
+    end
+
+    return 0
+  end
+
 -- Translate event from inkview EVT_* into emulated linux evdev events
 local function translateEvent(t, par1, par2)
     -- Much like Input does, detail what we catch when debug logging is enabled
@@ -169,6 +187,8 @@ local function translateEvent(t, par1, par2)
         -- Caught something post closeAll? (We're an InkView callback, remember).
         return 0
     end
+
+    local isV6Plus = getMajorSoftwareVersion() >= 6
 
     if t == C.EVT_INIT then
         inkview.SetPanelType(C.PANEL_DISABLED)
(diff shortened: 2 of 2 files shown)


tommiv:
@Frenzie yes. I've just dropped a more detailed comment here https://github.com/koreader/koreader/issues/12703#issuecomment-2454538535

benoit-pierre:
It would be better to update `ffi-cdecl/inkview_decl.c` and re-generate `ffi-cdecl/inkview_decl.c` with ffi-cdecl.

Do you have a build environment setup?

benoit-pierre:
Can some pocketbooks be running on old firmware with the old definitions?

tommiv:
> Do you have a build environment setup?
Unfortunately I don’t.

> Can some pocketbooks be running on old firmware with the old definitions?
Absolutely. Though it looks like SDK v6 is out and about for 2 years now

tommiv:
It probably may be done in a more safe way if I react on _EXT events only when the device runs v6.x firmware, but unfortunately I have no idea how to check the version in runtime

benoit-pierre:
There's `getSoftwareVersion()`, there should be a trace in the log at startup: `SoftwareVersion: …`, can you check yours?

tommiv:
Indeed, the log entry looks like this:
```
11/04/24-13:45:08 INFO  SoftwareVersion:  U743g.6.8.4143
```

benoit-pierre:
Have you actually checked that `par1 / par2` values don't overlap, between ext and non-ext events?

NiLuJe:
> Have you actually checked that `par1 / par2` values don't overlap, between ext and non-ext events?

I suspect that they theoretically can, but IV has historically been using custom keycodes for its basic needs, AFAICT.

(c.f., https://github.com/pocketbook/SDK_6.3.0/blob/dd8de463733a2bc0d53d1214b64734586fac427e/SDK-B288/usr/arm-obreey-linux-gnueabi/sysroot/usr/local/include/inkview.h#L475-L529)

I'm going to assume that if it currently works, eeeh, should be fine.

Since we don't actually use the TOUCHUP/TOUCHDOWN/TOUCHMOVE stuff, the breaking API change shouldn't really concern us.

NiLuJe:
> And since we don't actually use the TOUCHUP/TOUCHDOWN/TOUCHMOVE stuff, the breaking API change shouldn't really concern us either.

Unless these events are actually common on older FW, in which case, yeah, that's going to be a problem on FW < 6 devices ;D.

tommiv:
I actually do hope they have the same values (ie linux virtual keycodes) can’t prove it though, there is no hardware keyboard on-device, only 4 functional buttons. However, for on-board buttons the event is still without _EXT postfix, while HID keyboards always use _EXT events. I wonder if this ext actually stands for “external” and not “extended” as I thought initially

NiLuJe:
> I suspect that they theoretically can, but IV has historically been using custom keycodes for its basic needs, AFAICT.

I'm going by the assumption that the _EXT events are reserved for external input devices (e.g., OTG, BT), as opposed to the device's actual buttons for the non-_EXT variants.

tommiv:
> Unless these events are actually common on older FW, in which case, yeah, that's going to be a problem on FW < 6 devices ;D.

this can be mitigated nevertheless, if new event codes are only consumed if fw major version is >= 6. Allow me some time, I’ll add the check

benoit-pierre:
> > Have you actually checked that `par1 / par2` values don't overlap, between ext and non-ext events?
> 
> I suspect that they theoretically can, but IV has historically been using custom keycodes for its basic needs, AFAICT.
> 
> (c.f., https://github.com/pocketbook/SDK_6.3.0/blob/dd8de463733a2bc0d53d1214b64734586fac427e/SDK-B288/usr/arm-obreey-linux-gnueabi/sysroot/usr/local/include/inkview.h#L475-L529)
> 
> I'm going to assume that if it currently works, eeeh, should be fine.

Because in traces of https://github.com/koreader/koreader/issues/12703#issue-2631378050, there's:
```
11/03/24-18:41:09 DEBUG PB event => 26 (EVT_KEYRELEASE), par1: 24, par2: 0 
```
and
```
11/03/24-18:41:49 DEBUG PB event => 40 (EVT_TOUCHUP), par1: 23, par2: 0 
```
That's dangerously close…

NiLuJe:
Oh, right, we got away with it because the old SDK doesn't use custom constants, but just undefs & redefs the Linux KEY_ ones...

But you're right, we need the new mapping with custom names, make non-_EXT only handle the IV_ constants, and EXT handle the standard Linux ones.

NiLuJe:
i.e., compare https://github.com/blchinezu/pocketbook-sdk/blob/64e9fa210dd90569f8eabea92da4d40f3d072fc1/FRSCSDK/arm-none-linux-gnueabi/sysroot/usr/include/inkview.h#L283-L347 to https://github.com/pocketbook/SDK_6.3.0/blob/dd8de463733a2bc0d53d1214b64734586fac427e/SDK-B288/usr/arm-obreey-linux-gnueabi/sysroot/usr/local/include/inkview.h#L475-L529

(The good news is the actual values don't appear to have changed when they switched to a custom `IV_` namespace for the same constants in FW 5.x).

NiLuJe:
Following these old, old headers in base translates to: https://github.com/koreader/koreader-base/blob/ddcec0afa634e590c584e1fb76cc7babeb63a284/ffi/inkview_h.lua#L815-L848

(And I'm guessing we avoided a conflict with our https://github.com/koreader/koreader-base/blob/master/ffi/linux_input_h.lua by blind luck ;p)

which meant front could happily shove this mess under the carpet in front: https://github.com/koreader/koreader/blob/1e55dda4c765be3316d7ca87be3052943cb1be86/frontend/device/pocketbook/device.lua#L191-L200

NiLuJe:
> And then update the `event_map` in front to point to the `IV_` ones where relevant...

Or, perhaps saner for future readers, make the `non-_EXT` handler translate the `IV_`constants to their standard Linux counterparts, and update the event_map accordingly.

tommiv:
Well guys, I'm completely lost in your conversation, looks like I bit more than I can swallow. I've added v6+ check nevertheless, but it seems you have better ideas of how to incorporate v6 changes into the codebase. Anyway, let me know if I can help.

benoit-pierre:
@NiLuJe: that part seems really ugly:
```lua
        if band(poll_fds[0].revents, C.POLLIN) ~= 0 then
            updateTimestamp() -- single 'ts' copy for genEmuEvent inside the loop
            while rt.mq_receive(poll_fds[0].fd, ffi.cast("char*", hwmsg), hwmsg_len, nil) > 0 do
                local m = hwmsg[0]
                -- If there's no raw keymapping, emit this one instead
                if (m.type == C.EVT_KEYDOWN or m.type == C.EVT_KEYUP or m.type == C.EVT_KEYREPEAT) and not raw_keymap then
                    genEmuEvent(m.type, m.common.par1, m.common.par2)
                end
            end
        end
```

Since `EV_MAX` in `base/ffi/linux_input_h.lua` is 31, and those `EVT_KEYXXX` values start at 25.

Anyway, @tommiv, does this work:  [inkview_sdk_v6_update.zip](https://github.com/user-attachments/files/17620625/inkview_sdk_v6_update.zip)?

tommiv:
@benoit-pierre let me revert my changes for a clear experiment, I'll be back soon

tommiv:
Ah, there's no need to revert anything, files in the archive replacing ones I edited. Unfortunately, it crashes after the replacement:
```
 [*] Current time: 11/04/24-18:05:18
 [*] Version: v2024.07

ffi.load: libs/libutf8proc.so.3
ffi.load: blitbuffer
11/04/24-18:05:18 DEBUG fts: Preferred MONOTONIC clock source is CLOCK_MONOTONIC 
11/04/24-18:05:18 DEBUG fts: Preferred REALTIME clock source is CLOCK_REALTIME 
11/04/24-18:05:18 DEBUG fts: BOOTTIME clock source is supported 
ffi.load: inkview
./luajit: ./ffi/inkview_h.lua:141: size of C type is unknown or too large at line 69
stack traceback:
	[C]: in function 'cdef'
	./ffi/inkview_h.lua:141: in main chunk
	[C]: in function 'require'
	frontend/device/pocketbook/device.lua:14: in main chunk
	[C]: in function 'probeDevice'
	frontend/device.lua:61: in main chunk
	[C]: in function 'require'
	./reader.lua:151: in main chunk
	[C]: at 0x00013f61
!!!!
Uh oh, something went awry... (Crash n°5: 11/04/24 @ 18:05:18)
Too many consecutive crashes, aborting . . .
!!!! ! !!!!
```

benoit-pierre:
OK, I screwed the FFI declarations, new version: 
[inkview_sdk_v6_update.zip](https://github.com/user-attachments/files/17620918/inkview_sdk_v6_update.zip)

tommiv:
Still crashes, but this time the log is way longer than previous one, many things happen before the error. The error itself is also different:
```
11/04/24-18:19:11 DEBUG triggering refresh {
  mode = "full",
  region = "1404x1872+0+0"
} --[[table: 0xa1ed6e00]] 
11/04/24-18:19:11 DEBUG refresh: inkview full 0 0 1404 1872 nil 
./luajit: ./ffi/input_pocketbook.lua:429: missing declaration for symbol 'ProcessEventLoop'
stack traceback:
	[C]: in function '__index'
	./ffi/input_pocketbook.lua:429: in function 'waitForEvent'
	frontend/device/input.lua:1399: in function 'waitEvent'
	frontend/ui/uimanager.lua:1535: in function 'handleInput'
	frontend/ui/uimanager.lua:1590: in function 'run'
	./reader.lua:265: in main chunk
	[C]: at 0x00013f61
```

benoit-pierre:
And I was missing the compat declarations: 
[inkview_sdk_v6_update.zip](https://github.com/user-attachments/files/17621045/inkview_sdk_v6_update.zip)

NiLuJe:
> @NiLuJe: that part seems really ugly:

`waitForEventRaw` is barely used anywhere, as it both requires root, *and* a devcap flag, so, IIRC there's basically one device flagged as such: the one the author tested this on.

Add to that that this bypasses InkView entirely, and is actually polling an evdev device, so the InkView mapping is no longer relevant anyway ;).

NiLuJe:
Eh, ignore the above, not on my machine, clearly I can't read ;p.

benoit-pierre:
OK, did another pass on constants and added the missing declaration for `pocketbooksync.koplugin` [inkview_sdk_v6_update.zip](https://github.com/user-attachments/files/17621226/inkview_sdk_v6_update.zip).

NiLuJe:
> Eh, ignore the above, not on my machine, clearly I can't read ;p.

Okay, maybe I wasn't *entirely* off. Sleep deprivation FTW, yay!

Anyway, yeah, not quite sure what this does, but it is barely used anywhere because of the root requirement. I assume this specific monitor fd only outputs a similar InkView-like mess like the standard codepath, but I really have no idea, since Ive never seen that codepath exercised anywhere.

tommiv:
@benoit-pierre no errors anymore (including the one from the plugin). On-device buttons work, gamepad works too. Here are some logs just in case.

On-device "back" button (negative code is something new, but I assume it's your solution to separate non-EXT and EXT codes):
```
11/04/24-18:44:01 DEBUG PB event => 25 (EVT_KEYPRESS), par1: 24, par2: 0 
11/04/24-18:44:01 DEBUG key event => code: -24 (LPgBack), value: 1, time: 1730738668.124038 
...
11/04/24-18:44:01 DEBUG PB event => 26 (EVT_KEYRELEASE), par1: 24, par2: 0 
11/04/24-18:44:01 DEBUG key event => code: -24 (LPgBack), value: 0, time: 1730738668.284355 
```
Same behavior for on-device "forward" button, just a different code (-25)

BT gamepad "dpad-up" button (mapped in input_map.lua):
```
11/04/24-18:44:06 DEBUG PB event => 40 (EVT_KEYPRESS_EXT), par1: 46, par2: 0 
11/04/24-18:44:06 DEBUG key event => code: 46 (LPgFwd), value: 1, time: 1730738673.856902 
...
11/04/24-18:44:06 DEBUG PB event => 41 (EVT_KEYRELEASE_EXT), par1: 46, par2: 0 
11/04/24-18:44:06 DEBUG key event => code: 46 (LPgFwd), value: 0, time: 1730738673.958813 
```
Same for "dpad-down", just a different code (32) and mapped command (LPgBack).

tl;dr: works like a charm!

tommiv:
@benoit-pierre I guess I'm closing this PR in favor of your solution, since it's much more architecturally correct, and continue to run the locally patched version on my Pocketbook until the new koreader release is out, right?

Frenzie:
> until the new koreader release is out, right?

The one after, I'd say. ;-)

tommiv:
Closed in favor of https://github.com/koreader/koreader-base/pull/1971

PR #1977, opened 2024-11-20
ci/circle: bump docker images

benoit-pierre:
- koxtoolchain update to 2024.10 (https://github.com/koreader/koxtoolchain/releases/tag/2024.10)
- android image bump (with SDK build tools update to 34.0.0)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1977)

PR #1980, opened 2024-11-20
tesseract: update to 5.5.0

benoit-pierre:
https://github.com/tesseract-ocr/tesseract/releases/tag/5.5.0

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1980)


benoit-pierre:
Build failures:
```
build/thirdparty/tesseract/source/src/ccutil/ccutil.cpp:18:10: fatal error: filesystem: No such file or directory
 #include <filesystem> // for std::filesystem
          ^~~~~~~~~~~~
compilation terminated.
```

The `<filesystem>` header is part of C++17, so no increase to the standard.

And no issue building locally with latest koxtoolchain, so should be fixed with #1977.

PR #1969, opened 2024-10-31
crengine: add markdown support using md4c

benoit-pierre:
Cf. https://github.com/koreader/koreader/issues/12683.

2024-10-31-233329_540x720_scrot

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1969)

PR #1985, opened 2024-11-22
bump crengine

benoit-pierre:
- https://github.com/koreader/crengine/pull/603

Preliminary bump before #1969.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1985)


poire-z:
(I don't mind not having individual frontend bumps for this "normal line-height" stuff and you markdown support - the "normal line-height" frontend PR (requires it, but frontend won't break if it is not merged) can be bumped individually after a complex base is merged to frontend.)

PR #1986, opened 2024-11-22
cmake: fix external projects cmake bug

benoit-pierre:
Work around cmake bug https://gitlab.kitware.com/cmake/cmake/-/issues/25428 (introduced in 3.23.0, fixed in 3.27.9): do not create `build.d` dependency files with phony targets (which would keep triggering unnecessary rebuilds of some external projects).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1986)


Frenzie (approved):
Well found!

PR #1987, opened 2024-11-22
dropbear: add missing server key location override

benoit-pierre:
The ED25519 variant was missing, resulting in a failure to automatically generate the host key:
```
[15145] Nov 22 17:13:15 Couldn't create new file /etc/dropbear/dropbear_ed25519_host_key.tmp15145: No such file or directory
[15145] Nov 22 17:13:15 Exit before auth from <192.168.0.1:60314>: Couldn't read or generate hostkey /etc/dropbear/dropbear_ed25519_host_key
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1987)


Frenzie:
Cross-ref to koreader/koreader#12778

PR #1989, opened 2024-11-24
tests: add new runner

benoit-pierre:
Cf. https://github.com/koreader/koreader/pull/12789.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1989)

### DIFF
--- a/test-runner/busted_config.lua
+++ b/test-runner/busted_config.lua
@@ -1,0 +1,23 @@
+local lfs = require("lfs")
+
+local testsuites = {}
+local roots = {}
+local lpaths = {}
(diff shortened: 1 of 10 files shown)


test-runner/busted_config.lua
```diff
@@ -0,0 +1,25 @@
+-- vim: ft=lua
```
Frenzie, line 1: But it has a .lua extension?

Frenzie (approved):

NiLuJe (approved):
Sounds cool and looks sound at a very quick glance, but it's been a while since I've played with any of that ;).

PR #1997, opened 2024-11-27
ci/circle: fix ccache cache, take 2

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1997)

### DIFF
--- a/.ci/build_script.sh
+++ b/.ci/build_script.sh
@@ -1,16 +1,8 @@
 #!/usr/bin/env bash
 
 CI_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-DOCKER_HOME=/home/ko
 # shellcheck source=/dev/null
 source "${CI_DIR}/common.sh"
-
-if [[ -z "${CCACHE_DIR}" ]]; then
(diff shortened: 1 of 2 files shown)


Frenzie:
I imagine this should be `${DOCKER_HOME}/.cache/ccache`?
https://github.com/koreader/koreader-base/blob/538f02435ffc80a6dbb8f84c10c20a8bd2443a78/.ci/build_script.sh#L24

benoit-pierre:
> I imagine this should be `${DOCKER_HOME}/.cache/ccache`?
> 
> https://github.com/koreader/koreader-base/blob/538f02435ffc80a6dbb8f84c10c20a8bd2443a78/.ci/build_script.sh#L24

Not necessarily, since ccache will honor `~/.ccache`.

benoit-pierre:
Wait, no, since it won't already exists.

Frenzie:
> Wait, no, since it won't already exists.

ccache doesn't do that? That seems a little odd.

Frenzie:
> Not that it mattered in that instance, since the directory is guaranteed to exists (as a mountpoint for docker), so I could have left it at that.

I meant for consistency and to prevent future confusion, unless there was a reason to have it different.

PR #1999, opened 2024-12-08
sqlite: update to 3.47.2

benoit-pierre:
https://sqlite.org/releaselog/3_47_2.html

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1999)


Frenzie:
On a somewhat related note, I just saw this: https://joht.github.io/johtizen/automation/2022/08/03/keep-your-cpp-dependencies-up-to-date.html

PR #2001, opened 2024-12-08
mupdf: fix finalizers

benoit-pierre:
Some of the MuPDF tests segfault on quit pretty regularly (e.g.: `make --assume-old=all test T='mupdf -f jbig'`), this is because there's no guarantee that the context used to open a document or page won't be dropped before that document / page is.

Use the context private user pointer to keep a reference count and ensure it's only dropped last.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2001)

### DIFF
--- a/cache-key.base
+++ b/cache-key.base
@@ -19,6 +19,6 @@
 
 # Exceptions.
 /ffi-cdecl/koptcontext_cdecl.c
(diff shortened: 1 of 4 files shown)


NiLuJe (approved):
Neat and elegant ;).

:lgtm:

PR #2004, opened 2024-12-12
curl: build a static library

benoit-pierre:
Since there's only one user: zsync2.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2004)


Frenzie:
Minor note that some desire to do more things with cURL has been voiced, for example in https://github.com/koreader/koreader/issues/3193. On the flip side, it's been four years and no one has.

PR #2005, opened 2024-12-12
input: drop unused pocketbook leftovers

benoit-pierre:
Pocketbook devices don't even load `koreader-input`, cf. `frontend/device/input.lua`:
```lua
    elseif self.device:isAndroid() then
        self.input = require("ffi/input_android")
    elseif self.device:isPocketBook() then
        self.input = require("ffi/input_pocketbook")
    else
        self.input = require("libs/libkoreader-input")
    end
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2005)


Frenzie (approved):
There was the comment saying it's very dead code of course. Fine by me in any case.

PR #2007, opened 2024-12-14
fix linux target architecture flags

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2007)

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -601,7 +601,7 @@ ifdef APPIMAGE
 	HOST_ARCH:=-mtune=generic -march=x86-64
 endif
 
-ifdef LINUX
+ifeq ($(TARGET), linux)
 	HOST_ARCH:=-mtune=generic
 endif
 


Makefile.defs
```diff
@@ -601,7 +601,7 @@ ifdef APPIMAGE
 	HOST_ARCH:=-mtune=generic -march=x86-64
 endif
 
-ifdef LINUX
+ifeq ($(TARGET), linux)
```
Frenzie, line 604: I assume this will also affect the emulator, which isn't necessarily the biggest deal but if it's not too complicated I'd prefer to avoid that.

benoit-pierre: Why do you mean?

Frenzie: Never mind, I assume TARGET is empty for the emulator. I hadn't thought it through. What I mean is the emulator would like to have -march=native (not as a requirement, merely preferentially).

benoit-pierre: > Never mind, I assume TARGET is empty for the emulator.

Yes, we're still compiling the emulator with `-march=native`.

Frenzie (approved):

Makefile.defs
benoit-pierre: And yes, `linux` is not a great target name, considering most of our targets are Linux targets. :P

pazos (approved):

PR #2008, opened 2024-12-15
speedup testsuite

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2008)


spec/unit/mupdf_spec.lua
```diff
@@ -230,24 +223,10 @@ describe("mupdf module", function()
         end)
     end)
     describe("image API", function()
-        it("should render an image", function()
+        it("image API should render an image", function()
```
Frenzie, line 226: describe already covers that

PR #2009, opened 2024-12-15
openssl: update to 3.4.0

benoit-pierre:
https://github.com/openssl/openssl/blob/openssl-3.4.0/NEWS.md

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2009)

PR #2010, opened 2024-12-16
switch to libressl

benoit-pierre:
Alternative to #2009.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2010)


benoit-pierre:
Doh! So close…
```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.16.4 or higher is required.  You are running version 3.16.3
```

benoit-pierre:
I don't see anything in the [CMake 3.16 release notes](https://cmake.org/cmake/help/latest/release/3.16.html#updates) that would prevent downgrading the requirement to 3.16.3.

Frenzie (approved):

PR #2011, opened 2024-12-17
openssh: drop openssl and zlib dependencies

benoit-pierre:
We actually don't need them for building the only 2 executables we care about (scp & sftp-server).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2011)

PR #2012, opened 2024-12-19
dropbear & openssh: fix scp binary

benoit-pierre:
We can't use the openssh one without shipping the `ssh` binary too, so instead, ship the one provided by dropbear (and the `dbclient` binary, which is dropbear equivalent for `ssh`).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2012)


benoit-pierre:
Cf. #2011 

For the kindlepw2 release build:
| | Code size |
|:-|--------:|
| dbclient | 150KB |
| dropbear | 158KB |
| scp      | 17KB |
| sftp-server | 84KB |
| TOTAL | 409KB |

PR #2013, opened 2024-12-27
xtext: getSelectedWords returns with indices

TnS-hun:
Rename getSelectedWords to getSelectedWordIndices, and return the word boundary indices instead of the words. This is needed for highlighting. The words themselves can be obtained by passing the indices to the getText function.

Warning! Breaking change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2013)

### DIFF
--- a/xtext.cpp
+++ b/xtext.cpp
@@ -2063,9 +2063,8 @@ public:
         free(s_utf8);
     }
 
-    // Get (as a single UTF-8 string) the segment of m_text, extended to include
-    // the full words that may be cut at boundaries (start, end).
-    void getSelectedWords(int start, int end, int context) {
+    // Get start and end indices of an extended segment of m_text that includes the full words that may be cut at boundaries (start, end).
+    void getSelectedWordIndices(int start, int end, int context) {
         // Not much documentation about libunibreak word breaking,
         // some insight in:
         //  http://www.unicode.org/reports/tr29/tr29-25.html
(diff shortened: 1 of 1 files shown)


xtext.cpp
```diff
@@ -2063,9 +2063,8 @@ class XText {
         free(s_utf8);
     }
 
-    // Get (as a single UTF-8 string) the segment of m_text, extended to include
-    // the full words that may be cut at boundaries (start, end).
-    void getSelectedWords(int start, int end, int context) {
+    // Get extended start and end indices that include the full words that may be cut at boundaries (start, end).
```
poire-z, line 2066:
```suggestion
    // Get start and end indices of an extended segment of m_text that includes the full words that may be cut at boundaries (start, end).
```
(Just for clarity, "extended indices" reads odd to me.)

poire-z (approved):
Fine with me. I'll review the frontend PR later.

xtext.cpp
TnS-hun: I updated the comments.

TnS-hun:
I introduced and off by one error, so I added a small fix commit. (I did not want to use force push because it is two character change.)

Frenzie:
We'll just squash it. :-)

poire-z:
> Warning! Breaking change.

You may have been too quick at merging this :/
If you need to bump base into frontend, we'll need to either finalize #12948, or an interim commit to just translate bits to the xtext api change.

Frenzie:
Hm, indeed. If necessary I'll force push or revert.

PR #2014, opened 2025-01-03
KindleHF: Downgrade mtune to A9

NiLuJe:
Turns out we can't have nice things because PW4 ;'(.

Might require a similar fix in the actual kindlehf TC.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2014)

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -533,8 +533,9 @@ else ifeq ($(TARGET), kindlepw2)
 		COMPAT_CXXFLAGS:=$(UBUNTU_COMPAT_CFLAGS)
 	endif
 else ifeq ($(TARGET), kindlehf)
-	# We bias towards in-order CPUs (e.g., A53)
-	ARM_ARCH:=$(ARMV7_A7_ARCH)
+	# While we *would* like to bias towards in-order CPUs (e.g., A7/A53), we can't because we have to support the PW4,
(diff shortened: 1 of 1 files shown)


NiLuJe:
Re: https://github.com/koreader/koreader/issues/12988

Rexathion1:
> Turns out we can't have nice things because PW4 ;'(.
> 
> Might require a similar fix in the actual kindlehf TC.
> 
> This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2014)

Thanks for fixing this despite being busy rn 
Is the issue specific to PW4? My KT4 has the same issue & error from  [koreader/koreader#12988](https://github.com/koreader/koreader/issues/12988)

Commodore64user:
> we can't have nice things because PW4 ;'(.

As far as I am concerned, PW4 is the worst kindle ever made. yeah, worse than even the much older ones.

NiLuJe:
> > Turns out we can't have nice things because PW4 ;'(.
> > Might require a similar fix in the actual kindlehf TC.
> > This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2014)
> 
> Thanks for fixing this despite being busy rn Is the issue specific to PW4? My KT4 has the same issue & error from [koreader/koreader#12988](https://github.com/koreader/koreader/issues/12988)

Oh, yeah, makes sense, forgot that the KT4 also had support for 5.17.

Yep, likely running on a similar board/SoC.

PR #2015, opened 2025-01-07
thirdparty/lua-Spore 0.4.0

Frenzie:
https://framagit.org/fperrad/lua-Spore/-/tags/0.4.0
https://framagit.org/fperrad/lua-Spore/-/compare/0.3.4...0.4.0

Edit: of note, "switch to dkjson"
https://framagit.org/fperrad/lua-Spore/-/commit/8e7ad16a772ce7e13a4fbc33de33b045426dcb51

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2015)

PR #2016, opened 2025-01-08
tests: improve test runner and testsuite isolation

benoit-pierre:
- initial LUA paths are now setup by the runner itself, so it can be executed directly without needing to manually set them before
- patch the LUA paths before actually executing the tests to remove the testsuite specific entries (e.g. `spec/rocks/…`), ensuring tests only have access to shipped packages

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2016)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,20 +134,18 @@ download-all: test-data
 $(OUTPUT_DIR)/spec/base: | $(OUTPUT_DIR)/spec/
 	$(SYMLINK) $(KOR_BASE)/spec $@
 
-$(OUTPUT_DIR)/spec/config.lua: | $(OUTPUT_DIR)/spec/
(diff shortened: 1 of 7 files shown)


benoit-pierre:
With this PR the testsuite would fail with the initial version of #2015 (prevent importing a test-only package: `dkjson`), but it would not help when using `lfs` instead of `koreader-lfs` like in https://github.com/koreader/koreader/pull/13004, even after updating busted to master (https://github.com/lunarmodules/busted/commit/a144124839f027a2d0a95791936c478d047126fc, to get https://github.com/lunarmodules/busted/pull/654, which does not seem to work, as even some of the added busted tests don't pass).

PR #1984, opened 2024-11-21
tests: install luacov & cluacov to spec tree

benoit-pierre:
This fix coverage on master (https://app.circleci.com/pipelines/github/koreader/koreader/11659/workflows/527da8a7-fe87-4125-8656-dd1a707a2a9e/jobs/17231). The addition of cluacov (which must be compiled for the luajit version used during coverage, or things go boom) also greatly speeds up things (see https://app.circleci.com/pipelines/github/benoit-pierre/koreader/528/workflows/59965b76-8cfa-4e07-a75e-454aefc6c088/jobs/547).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/1984)

PR #2019, opened 2025-01-12
cmake: fix cleaning an external project source tree

benoit-pierre:
Cleaning directories need to be done in reverse order.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2019)


benoit-pierre:
Cf. https://github.com/koreader/koreader-base/pull/1984#issuecomment-2585850937.

poire-z (approved):
It solved my issue, bravo!

PR #2022, opened 2025-01-26
always compile with debug information

benoit-pierre:
Each binary's debug information are extracted and linked to a separate `.dbg` file. Those `.dbg` files are ignored when bundling a release, but otherwise transparently used by gdb, addr2line, and other tools when present.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2022)


benoit-pierre:
Note: for macOS, `dsymutil` is used to split the debug information into a separate `.dSYM` bundle that should work the same (untested, as I don't have access to a machine for checking).

Frenzie:
I suppose it might be a tad slower to build, but the runtime performance is identical?

benoit-pierre:
Yep.

PR #2026, opened 2025-01-27
lodepng: update to 0b1d9cc (2024-12-28)

benoit-pierre:
https://github.com/lvandeve/lodepng/compare/d398e0f10d152a5d17fa30463474dc9f56523f9c...0b1d9ccfc2093e5d6620cd9a11d03ee6ff6705f5

Additionally, compile without support for ancillary / unknown chunks.

Code size difference in release mode:
- Android ARM64: -13.6 KB
- Emulator x86_64: -14.2 KB
- Kindle PW2: -8.1 KB

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2026)


Frenzie:
For reference:
https://www.w3.org/TR/png/#11Ancillary-chunks

It would seem that in this case it only directly refers to suggested background color and the text values.
https://github.com/fhanau/Efficient-Compression-Tool/blob/9aabc23d73899ae55c1de292592fed6eb6217f66/src/lodepng/lodepng.h#L268-L348

NiLuJe (approved):

PR #2038, opened 2025-02-20
zstd: update to 1.5.7

benoit-pierre:
https://github.com/facebook/zstd/releases/tag/v1.5.7

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2038)


benoit-pierre:
macOS jobs were not run:
> The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings.

benoit-pierre:
Are we paying for anything related to Github Actions?

Frenzie:
No, GH offered all that for free. Maybe it's because the three different Mac versions add quite a lot more minutes. (Also all Mac minutes are multiplied by 10.)

benoit-pierre:
Some weird transient issue, then. ¯\\_(ツ)_/¯

PR #2045, opened 2025-03-09
Use /sys/power/autosleep and wakelocks

ghost:
This is the backend, so that the frontend can use the kernels 'autosleep' functionallity.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2045)

### DIFF
--- a/input/input.c
+++ b/input/input.c
@@ -530,7 +530,8 @@ static int waitForInputWithEpoll(lua_State* L)
     struct epoll_event event;
     struct epoll_event events[MAX_EPOLL_EVENTS];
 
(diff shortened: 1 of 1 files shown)


NiLuJe:
Looks good at a glance, thanks!

(Won't have time to test for a while, though).

NiLuJe:
Did not forget about it, hopefully I'll have more time in a few weeks, thanks ;).

NiLuJe:
Haven't forgotten about this, but I'm in the process of moving right now, so free time is in short supply ;).

PR #2047, opened 2025-03-29
BlitBuffer: add partial support for paintRoundedRect using RGB colours

ReneeIbeji:
Add RGB32 variants for paintBorder and paintRoundedRect functions, currently the paintborder variant doesn't have support to rounded corners.

Allows for framecontainers with coloured backgrounds to be used io statistics plugin calendar view.
See: https://github.com/koreader/koreader/pull/13453#issue-2944166320

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2047)

### DIFF
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -2078,7 +2078,7 @@ function BB_mt.__index:paintBorder(x, y, w, h, bw, c, r, anti_alias)
 end
 
 --[[
-paintBorder varient that uses ColorRGB32 instead of a luminance value,
+paintBorder variant that uses ColorRGB32 instead of a luminance value,
 no support for rounded corners therefore no need for anti aliasing flag
 --]]
 function BB_mt.__index:paintBorderRGB32(x, y, w, h, bw, c)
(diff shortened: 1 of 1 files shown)


ffi/blitbuffer.lua
```diff
@@ -2077,6 +2077,19 @@ function BB_mt.__index:paintBorder(x, y, w, h, bw, c, r, anti_alias)
     end
 end
 
+--[[
+paintBorder varient that uses ColorRGB32 instead of a luminance value,
```
NiLuJe, line 2081:
```suggestion
paintBorder variant that uses ColorRGB32 instead of a luminance value,
```

ffi/blitbuffer.lua
```diff
@@ -2122,6 +2135,21 @@ function BB_mt.__index:paintRoundedRect(x, y, w, h, c, r)
     end
 end
 
+--[[
+paintRoundedRect varient that uses ColorRGB32 instead of a luminance value
```
NiLuJe, line 2139:
```suggestion
paintRoundedRect variant that uses ColorRGB32 instead of a luminance value
```

ffi/blitbuffer.lua
```diff
@@ -2122,6 +2135,21 @@ function BB_mt.__index:paintRoundedRect(x, y, w, h, c, r)
     end
 end
 
+--[[
+paintRoundedRect variant that uses ColorRGB32 instead of a luminance value
```
NiLuJe, line 2139:
```suggestion
paintRoundedRect variant that uses ColorRGB32 instead of a luminance value
NOTE: Be aware that `paintBorderRGB32` currently does *NOT* support rounded corners, so this implementation is *also* incomplete!
```

NiLuJe (approved):
Sorry for the delay, thanks!

PR #2064, opened 2025-05-01
libk2pdfopt: update to v2.55-168-gb98b2ff

benoit-pierre:
https://github.com/koreader/libk2pdfopt/compare/0be83e5...b98b2ff


benoit-pierre:
Damn, the gcc-lua plugin is also broken by the update to GCC 15…

NiLuJe:
> Damn, the gcc-lua plugin is also broken by the update to GCC 15…

Ouch :/

benoit-pierre:
Yeah… I sent a patch to the maintainer. Hopefully I got it right and we'll bump gcc-lua.

PR #2066, opened 2025-05-03
cmake: fix checks for iconv / gettext support

benoit-pierre:
Can't use the same variable for 2 different checks: it's going to be cached after the first check, so the second one will never run…

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2066)


Commodore64user:
👍

PR #2070, opened 2025-05-04
ffi/crypto: AES ECB decryption, PKCS#7 unpadding

TnS-hun:
This commit makes it possible to decrypt buffers encrypted with AES using the ECB mode, and unpad buffers using the PKCS#7 algorithm.

It uses OpenSSL's EVP cipher routines, which are a high-level interface to certain symmetric ciphers. So it should be easy to add support for other ciphers beside AES if needed.

### DIFF
--- a/ffi-cdecl/crypto_decl.c
+++ b/ffi-cdecl/crypto_decl.c
@@ -3,3 +3,13 @@
 #include "ffi-cdecl.h"
 
 cdecl_func(PKCS5_PBKDF2_HMAC_SHA1)
+cdecl_func(EVP_CIPHER_CTX_new);
+cdecl_func(EVP_CIPHER_CTX_free);
+cdecl_func(EVP_CIPHER_CTX_set_padding);
+cdecl_func(EVP_CIPHER_block_size);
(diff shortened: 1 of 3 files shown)


TnS-hun:
This will be needed for the Kobo books plugin. I'll add more details there.

Frenzie:
(I'm just clicking on the Copilot thing to see if it can make any sense of this or if it's too complicated for it.)

benoit-pierre:
Did you manually update `ffi/crypto_h.lua`?

benoit-pierre:
You should use the `libressl-ffi` target to update `ffi/crypto_h.lua`.

TnS-hun:
> Did you manually update `ffi/crypto_h.lua`?

Yes, I added those manually. I could not get ffi-cdecl to work. :(

TnS-hun:
> What was the issue?
> 
> Which target are you using for testing this PR?

I did not realize that I had to just use `make libressl-ffi`. :) (The problem was that I tried to install ffi-cdecl manually with that old gcc-4.7-plugin-dev package...)

PR #2071, opened 2025-05-04
New zipreader module

TnS-hun:
Just like zipwriter but for reading zip files.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2071)


TnS-hun:
This will be needed for the Kobo books plugin. I'll add more details there.

Frenzie:
Fine by me, but should it be named something like `zlib_wrapper` to prevent potential confusion?

benoit-pierre:
Yes, and loose the header `zlib.h`, no need for it.

But really, I don't know if it might not be better at this point to integrate libarchive and use that for all our archive related needs.

Frenzie:
There are other advantages to that too. Building MuPDF with libarchive should open up CBR support iirc.

benoit-pierre:
Yeah, and we could drop tar (and use the libarchive provided bsdtar).

TnS-hun:
I renamed zlib.c to zlib_wrapper.c to make it clearer.

> Yes, and loose the header `zlib.h`, no need for it.

If I remove it then ffi-cdecl complains:
> ffi-cdecl/zlib_decl.c:9:12: error: ‘uncompress_raw’ undeclared here

benoit-pierre:
Oh, right, we need a header for the cdecls… 

I'm not sure `uncompress_raw` belongs in `zlib_decl.c`.

TnS-hun:
With libarchive added this no longer makes sense.

PR #2074, opened 2025-05-05
macos: automatically set `MACOSX_DEPLOYMENT_TARGET`

benoit-pierre:
NOTE: contrary to what CMake's documentation says, we can't rely on `CMAKE_OSX_DEPLOYMENT_TARGET` being "computed based on the host platform" if the `MACOSX_DEPLOYMENT_TARGET` environment variable is not set, so we have to set the later to a default value ourselves.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2074)


Frenzie:
Thanks. 👍

PR #2084, opened 2025-05-10
ci: bump docker images

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2084)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ executors:
 
   base: &BASE_EXE
     docker:
-      - image: koreader/kobase:0.3.5-20.04
+      - image: koreader/kobase:0.4.1-22.04
 
   # Emulators. {{{
 
@@ -35,7 +35,7 @@ executors:
 
   emu_clang_ninja:
     docker:
-      - image: koreader/kobase-clang:0.3.6-20.04
+      - image: koreader/kobase-clang:0.4.1-22.04
     environment:
       <<: *EMU_ENV_LST
       USE_CLANG: "1"
@@ -46,7 +46,7 @@ executors:
 
   android-arm: &ANDROID_EXE
     docker:
-      - image: koreader/koandroid:0.8.4-20.04
+      - image: koreader/koandroid:0.9.1-22.04
     environment:
       TARGET: "android"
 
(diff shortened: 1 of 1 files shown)


benoit-pierre:
Patches for nightly-builds: [patches.zip](https://github.com/user-attachments/files/20144082/patches.zip).

benoit-pierre:
And for the nightswatcher script:
```diff
 nightswatcher/nightswatcher.py | 1 +
 1 file changed, 1 insertion(+)

diff --git i/nightswatcher/nightswatcher.py w/nightswatcher/nightswatcher.py
index dd3f8c2..767f681 100644
--- i/nightswatcher/nightswatcher.py
+++ w/nightswatcher/nightswatcher.py
@@ -138,6 +138,7 @@ ota_zsync_models = frozenset([
                         'build_kindlehf'
                         'build_kobo', 'build_pocketbook',
                         'build_remarkable',
+                        'build_remarkable_aarch64',
                         'build_sony_prstux'])
 
 
```

benoit-pierre:
Updated nightly-builds patches: 
[patches.zip](https://github.com/user-attachments/files/20150636/patches.zip).

Frenzie:
Something seems to have failed with the ccache update that I don't immediately understand. (Never push an update on a Friday or just before going to bed. :-P)

```
    ccache: error: failed to create temporary file for /builds/koreader/nightly-builds/ccache/tmp/cpp_stdout.tmp.kkKcJR.i: Permission denied
```

https://gitlab.com/koreader/nightly-builds/-/jobs/10000032401

Maybe the GitLab cache merely needs to be invalidated? I'm just not quite sure why that would be the case.

benoit-pierre:
But the new remarkable-aarch64 build seems fine?!

benoit-pierre:
Can you delete those caches?

Frenzie:
I don't think so, but of course I can add a `-1` at the end of the cache key or some such.

benoit-pierre:
I wonder if we're not seeing the issue on Circle CI because of the build directory cache and/or the way caching is scoped. I guess we'll see soon enough once #2085 is merged.

PR #2085, opened 2025-05-11
cmake: bump minimum required version to 3.17.5

benoit-pierre:
We need at least 3.17 for libarchive, and Ubuntu 22.04 provides 3.22.1. Additionally, CMake support for versions <3.17 is deprecated since Meson 0.62.0.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2085)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.16.3)
+cmake_minimum_required(VERSION 3.17.5)
 project(koreader LANGUAGES C CXX)
 
 include(${CMAKE_KOVARS})
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -796,7 +796,7 @@ endif
 # Native & cross definitions.
 define cmake_toolchain
 
-cmake_minimum_required(VERSION 3.16.3)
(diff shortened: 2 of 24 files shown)


Frenzie (approved):

Commodore64user:
is this the reason for my getting this now
```
CMake Error at CMakeLists.txt:99 (message):
  $MACOSX_DEPLOYMENT_TARGET / CMAKE_OSX_DEPLOYMENT_TARGET discrepancy:
  "15.4.1" != "15.0"
  ```

Commodore64user:
oh, I didn't see #2074, good to know I don't have to set the target ever again, it was a bit annoying sometimes. Might be worth updating the build emulator wiki though. I still don't have a working trackpad/mouse though and Frenzie one take one for the team...

benoit-pierre:
You mean [doc/Building.md](https://github.com/koreader/koreader/blob/master/doc/Building.md)?

I have no idea if the information about a "minimum deployment version" is still applicable… Ditto regarding the Catalina mention.

Frenzie:
To my annoyance XCode is largely tied to macOS versions, so I'd say there's a decent enough chance it's perpetually true. But by now that's also an ancient unsupported version, so we can just drop the remark.

PR #2090, opened 2025-05-17
ArcReader: new module for reading archives

benoit-pierre:


benoit-pierre:
Example use in `ArchiveViewer`:
```lua
function ArchiveViewer:openFile(file)
    local _, filename = util.splitFilePathName(file)

    self.arc = ArcReader:new()
    self.fm_updated = nil
    self.list_table = {}

    if self.arc:open(file) then
        self:getListTable()
    end
[…]
end

function ArchiveViewer:getListTable()
    local function parse_path(filepath, filesize)
[…]
    end

    for entry in self.arc:iterate() do
        if entry.mode == "file" then
            parse_path(entry.path, entry.size)
        end
    end
end

function ArchiveViewer:extractFile(filepath)
    local directory = util.splitFilePathName(self.arc.filepath)
    self.fm_updated = self.arc:extractToPath(filepath, directory .. filepath)
end

function ArchiveViewer:extractContent(filepath)
    return self.arc:extractToMemory(filepath)
end
```

benoit-pierre:
In `readerui`:
```lua
function ReaderUI:extendProvider(file, provider, is_provider_forced)
    -- If file extension is single "zip", check the archive content and choose the appropriate provider,
    -- except when the provider choice is forced in the "Open with" dialog.
    -- Also pass to crengine is_fb2 property, based on the archive content (single "zip" extension),
    -- or on the original file double extension ("fb2.zip" etc).
    local _, file_type = filemanagerutil.splitFileNameType(file) -- supports double-extension
    if not is_provider_forced and file_type == "zip" then
        local arc = ArcReader:new()
        if arc:open(file) then
            for entry in arc:iterate() do
                local ext = util.getFileNameSuffix(entry.path)
                if ext and entry.mode == "file" and entry.size > 0 then
                    file_type = ext:lower()
                    break
                end
            end
            arc:close()
        end
        local providers = DocumentRegistry:getProviders("dummy." .. file_type)
        if providers then
            for _, p in ipairs(providers) do
                if p.provider.provider == "crengine" or p.provider.provider == "mupdf" then -- only these can unzip
                    provider = p.provider
                    break
                end
            end
        end
    end
    provider.is_fb2 = file_type:sub(1, 2) == "fb"
    provider.is_txt = file_type == "txt"
    return provider
end
```

benoit-pierre:
And `Device:unpackArchive`:
```lua
function Device:unpackArchive(archive, extract_to, with_stripped_root)
    require("dbg").dassert(type(archive) == "string")
    local BD = require("ui/bidi")
    local arc = ArcReader:new()
    local ok = arc:open(archive)
    if ok then
        for entry in arc:iterate() do
            local dest_path = entry.path
            if with_stripped_root then
                local __, tail = dest_path:match("([^/]*)/*(.*)")
                if tail then
                    -- Non-root: strip one level.
                    dest_path = tail
                elseif entry.mode == 'directory' then
                    -- Root directory: ignore.
                    goto continue
                else -- luacheck: ignore 542
                    -- Root non-directory: don't strip.
                end
            end
            if not arc:extractToPath(entry.path, extract_to.."/"..dest_path) then
                break
            end
            ::continue::
        end
        ok = not arc.err
    end
    if not ok then
        return false, T(_("Couldn't extract archive:\n\n%1\n\n%s."), BD.filepath(archive), arc.err)
    end
    return true
end
```

`Device:untar` implementations can be removed. This would fix installing dictionaries on Android < 19 (i.e. JellyBean, cf. https://github.com/koreader/koreader-base/pull/1617#issuecomment-1601819110).

hius07:
Great, thanks!

ffi/arcreader.lua
```diff
@@ -0,0 +1,178 @@
...
+
+function ArcReader:next()
+    self.err = nil
+    if self.archive == nil then
+        return
+    end
+    -- We can't consume the data twice: tweak the index
+    -- to trigger a reset in case of subsequent attempt.
+    self.index = math.floor(self.index) + 0.1
+    local err = libarchive.archive_read_next_header(self.archive, self.archive_entry)
```
Frenzie, line 67: Do they have some sample C code somewhere or some such btw? Or is it entirely from scratch?

Frenzie (approved):

ffi/arcreader.lua
benoit-pierre: There are some examples on the [wiki](https://github.com/libarchive/libarchive/wiki/Examples), and in some of the [manpages](https://github.com/libarchive/libarchive/wiki/ManualPages).

Frenzie: Thanks.

ffi/arcreader.lua
```diff
@@ -0,0 +1,178 @@
...
+        return
+    end
+    local entry
+    self.index = math.floor(self.index) + 1
+    if self.index > self.size then
+        local archive_entry = self.archive_entry[0]
+        entry = {
+            path = ffi.string(libarchive.archive_entry_pathname(archive_entry)),
+            mode = ENTRY_MODE[libarchive.archive_entry_filetype(archive_entry)] or "other",
+            size = libarchive.archive_entry_size(archive_entry),
```
NiLuJe, line 82: I don't recall the libarchive types (on mobile right now) but If it's a 64bit wide type on 32bit arches, it'll need to be truncated (e.g., tonumber).

Or, maybe better given that it is sort of critical not to screw this one up, keep all the maths on it in C ;). (e.g., keep self.size as a C type).

NiLuJe: Nevermind me, clearly haven't gotten enough sleep ;).

Confused the index size with this entry size field ;p.

hius07:
Sorry for disturbing you, can this be finished?
After recent updates to ZipWriter the system unzip on my Kindle cannot open archives created by ZipWriter.
`unzip: zip flags 1 and 8 are not supported`

benoit-pierre:
OK, so `ffi/arcreader.lua` → `ffi/archiver.lua`. Now support both reading and writing archives (replacing `ffi/zipwriter.lua`).

Frenzie:
Ready to merge?

benoit-pierre:
Yes.

PR #2093, opened 2025-05-24
fbink / fbdepth / libfbink_input: fix duplicate downloads

benoit-pierre:
The intent was always to use the same git clone.

Je devais pas avoir les yeux en face des trous quand j'ai codé cette portion… :P

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2093)


Frenzie:
Eyes in front of holes? Does that mean holes instead of eyes?

benoit-pierre:
"En face": aligned, so eyeballs not aligned with their sockets.

Frenzie:
Ah, I'm only familiar with "je suis en face de de la banque" or "je suis en face de toi à la table" or something like that. Thanks!

PR #2096, opened 2025-05-29
sqlite: update to 3.50.0 + disable unused extensions / features

benoit-pierre:
Release notes: https://sqlite.org/releaselog/3_50_0.html

Disable:
- (de-)serializing support
- FTS4 & FTS5 extensions (full text search)
- Geopoly & R-Tree extensions
- JSON support
- UTF-16 support

(hopefully no external plugin / user patch relies on those)

Code size reduction:
- android-arm: -372.3 KB
- android-arm64: -551.3 KB
- emulator: -459.5 KB
- kindlepw2: -261.6 KB

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2096)


Frenzie (approved):
Presumably JSON and full text search probably aren't currently used, but they do sound like something people might want to use. But they're easy to reenable if it comes up.

(Random search result for <https://blog.sqlitecloud.io/sqlite-extensions-intro-to-geopoly> since I didn't know what that was.)

PR #2111, opened 2025-06-20
curl: disable a bunch of unused features/protocols & update to 8.14.1

benoit-pierre:
https://curl.se/changes.html

Code size reduction for kindlepw2's zsync: -95.0 KB.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2111)


Frenzie:
To clarify for other readers, cURL is currently only used as a dependency for zsync2. Hence why even something as basic as  cookies and form API is okay to disable.

PR #2112, opened 2025-06-23
sqlite: enable FTS5 extension

szendo:
Enable FTS3/FTS4 extensions for SQLite.

I recently started working on a custom dictionary plugin that stores entries in an SQLite database and uses FTS4 for search.
I noticed that FTS4 recently got disabled in https://github.com/koreader/koreader-base/pull/2096 so I would like to have it re-enabled.

The commit references only `SQLITE_ENABLE_FTS3` as that is the official option to enable both versions (`SQLITE_ENABLE_FTS4` is only an alias as noted here: https://www.sqlite.org/src/info?name=d0f70d602be987501cf6eff86a06a0b5d0389929c6abe24972d59594e6f4cb03&ln=34-41)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2112)


benoit-pierre:
What about [FTS5](https://sqlite.org/fts5.html#_summary_of_technical_differences_)?

PR #2115, opened 2025-06-26
White threshold for scanned PDFs (MuPDF)

kerivin:
This PR is currently open as #2358

I left a comment in issue [#13294](https://github.com/koreader/koreader/issues/13294#issuecomment-2994186500) about brightness/contrast of scanned PDFs. But then I saw the amount of open issues in the main `koreader` repo and decided to experiment myself, and I think I might have found a decent solution for this.
The solution would contain 3 pull requests (for `koreader`, `koreader-base` (this one) and `libk2pdfopt`) and would add [option 'white threshold'](https://github.com/koreader/koreader/pull/13994). `libk2pdfopt` natively has it ([and I created a pull request for it as well](https://github.com/koreader/libk2pdfopt/pull/62)), but for MuPDF (here in `koreader-base`) I had to write my implementation, and it actually looks quite good:

I know literally nothing about Lua or contributing to open source, so chances I've made a mistake somewhere are not zero, please keep in mind.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2115)


Frenzie:
I'm not sure if a custom implementation necessarily makes sense if libk2pdfopt already has it? If a custom implementation does make sense it feels more like a generic thing than something specific for MuPDF.

kerivin:
I also experimented with libk2pdf because yeah, it does have native white threshold (see [this pull request](https://github.com/koreader/libk2pdfopt/pull/62)), but it doesn't work the same way as the original k2pdfopt does. Most likely I missed something, but I thought it's a good thing to have whether Dewatermark is on and off.

PR #2125, opened 2025-07-09
Add Lualanes thirdparty module

bateast:
Provides full pthread and inter-thread messaging

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -372,6 +372,9 @@ declare_project(thirdparty/minizip EXCLUDE_FROM_ALL)
 # mupdf
 declare_project(thirdparty/mupdf DEPENDS freetype2 harfbuzz libarchive libjpeg-turbo libwebp minizip zlib)
 
+# mupdf
+declare_project(thirdparty/Lualanes)
+
 # nanosvg
 declare_project(thirdparty/nanosvg)
 
--- a/thirdparty/Lualanes/CMakeLists.txt
+++ b/thirdparty/Lualanes/CMakeLists.txt
@@ -1,0 +1,30 @@
+set(MAKE_CMD
+    make -C ${SOURCE_DIR}
+)
+
+set(LUA_INCLUDE_DIR "${STAGING_DIR}/include/luajit-2.1/")
+set(MODIFIED_CFLAGS "${CFLAGS} -I${LUA_INCLUDE_DIR}")
+
+
+list(APPEND BUILD_CMD ${MAKE_CMD}
+    CROSS_COMPILE=${CROSS_COMPILE}
+    CC=${CC}
+    CXX=${CXX}
+    AR=${AR}
+    RANLIB=${RANLIB}
+    CFLAGS=${MODIFIED_CFLAGS}
+    LDFLAGS=${LDFLAGS}
+    LUA_PATH= LUA_CPATH= ${PROGRAM})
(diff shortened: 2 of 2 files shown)


benoit-pierre:
My experience with lua-lanes, when trying to build it for use by luacheck in our docker images, is a whole lot of crashing, including when trying to run their testsuite to determine if the build was functional. And newer / development versions necessitating a C/C++ standard version we don't support on all targets. But that was a while ago, before our latest koxtoolchain update. Have you tried building for, say, `kindle-legacy` for example?

benoit-pierre:
It still not quite what we want on Android: we need a static library so the module can be part of the monolibtic library.

See #2132 for that.

I can into other issues when trying you code on Android:
- I got an unknown lanes error, and then the code crashed when trying to dereference a nil `async_result`:
```diff
diff --git i/frontend/ui/translator.lua w/frontend/ui/translator.lua
index 732ede478..2dbb08201 100644
--- i/frontend/ui/translator.lua
+++ w/frontend/ui/translator.lua
@@ -473,8 +473,10 @@ function Translator:loadPage(text, target_lang, source_lang)
         end
 
         if trapper_status == "error" then
-            logger.warn("Translator async operation failed:", async_result.error_message or "Unknown error")
-            logger.dbg("Error details:", async_result.code, async_result.headers, async_result.http_status, async_result.content)
+            logger.warn("Translator async operation failed:", async_result and async_result.error_message or "Unknown error")
+            if async_result then
+                logger.dbg("Error details:", async_result.code, async_result.headers, async_result.http_status, async_result.content)
+            end
             return false, nil
         end
 
```
- monolibtic mode does not work (can be tested with the emulator by setting `MONOLIBTIC=1` in the environment and rebuilding), because `ffi/loadlib` is not loaded (which indicate a potential issue in non-monolibtic mode too, if `ffi.load` or `ffi.loadlib` is used):
```diff
diff --git i/frontend/ui/trapper.lua w/frontend/ui/trapper.lua
index 6224466b0..52b47cb2a 100644
--- i/frontend/ui/trapper.lua
+++ w/frontend/ui/trapper.lua
@@ -694,8 +694,12 @@ end
 --- It integrates with a UI manager for scheduling and displaying messages.
 ---
 
-local lanes       = require("lanes").configure()
-local linda       = lanes.linda()
+local lanes = require("lanes").configure{
+    on_state_create = function()
+        require "ffi/loadlib"
+    end
+}
+local linda = lanes.linda()
 
 -- ticket generator
 local _next_ticket = 0
```
- but then Android still does not work, because our `android` module is not available, and this module cannot be loaded again, or copied (multiple `cdata` fields), I did manage to get it running with a hack:
```diff
diff --git i/frontend/ui/trapper.lua w/frontend/ui/trapper.lua
index 6224466b0..eee27da03 100644
--- i/frontend/ui/trapper.lua
+++ w/frontend/ui/trapper.lua
@@ -694,8 +694,17 @@ end
 --- It integrates with a UI manager for scheduling and displaying messages.
 ---
 
-local lanes       = require("lanes").configure()
-local linda       = lanes.linda()
+local ffi = require "ffi"
+local android_nativeLibraryDir = ffi.os == "Linux" and os.getenv("IS_ANDROID") and require("android").nativeLibraryDir
+local lanes = require("lanes").configure{
+    on_state_create = function()
+        if android_nativeLibraryDir then
+            package.loaded.android = { nativeLibraryDir = android_nativeLibraryDir }
+        end
+        require "ffi/loadlib"
+    end
+}
+local linda = lanes.linda()
 
 -- ticket generator
 local _next_ticket = 0
```
But this obviously won't work if some code depends on the real `android` module.

Consequently, I don't think lanes is a viable option.

PR #2140, opened 2025-07-22
make: use clang on macOS

benoit-pierre:
When compiling the emulator or for the `macos` target: while gcc is already an alias for clang, this might prevent the build system from picking another GCC install (e.g. from homebrew).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2140)

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -201,6 +201,7 @@ else ifeq ($(TARGET), macos)
 	SDL=1
 	MACOS=1
 	EMULATE_READER=1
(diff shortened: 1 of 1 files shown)


Frenzie (approved):

benoit-pierre:
- it's not the system default compiler
- I don't know if `MACOSX_DEPLOYMENT_TARGET` is handled the same
- possibly different search paths (if it's homebrew)
- etc…

PR #2147, opened 2025-08-15
tests: fix test runner shellcheck 0.11.0 false positives

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2147)

### DIFF
--- a/test-runner/runtests
+++ b/test-runner/runtests
@@ -35,6 +35,7 @@ if [[ "${BASH_VERSINFO[0]}" -lt 4 ]]; then
     exit 1
 fi
 
+# shellcheck disable=SC2329 # shellcheck 0.11.0 false positive / regression
 run() {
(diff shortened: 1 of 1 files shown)


Frenzie:
Possibly similar to <https://github.com/koalaman/shellcheck/issues/3273>? Or in any case due to the same commit I imagine.

It does seem to work out if you make it much smaller though.

```sh
#!/usr/bin/env bash

runtests_meson() {
    echo hello
}

use_meson=1
which meson >/dev/null 2>&1 || use_meson=0

if [[ "${use_meson}" -ne 0 ]]; then
    runtests_meson "$@"
else
    runtests_busted "$@"
fi
```

benoit-pierre:
Yep, I subscribed to this issue earlier today.

PR #2157, opened 2025-09-04
sdl2: update to 2.23.10

benoit-pierre:
https://github.com/libsdl-org/SDL/releases/tag/release-2.32.10

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2157)


Frenzie:
What's it necessary for precisely?

Frenzie:
I'll just merge it as is for now. We can always remove the patch later.

PR #2169, opened 2025-09-16
ci/macos: tweaks

benoit-pierre:
Reduce differences with koreader.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2169)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -127,14 +127,16 @@ jobs:
 
       # Restore / setup build cache. {{{
 
-      - name: Restore build cache
+      - name: Lookup or restore build cache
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -103,7 +103,7 @@ jobs:
       # Checkout / fetch. {{{
 
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
```
Frenzie, line 106: Drat, I forgot v5 was already out. (See <https://github.com/koreader/koreader.github.io/pull/78>.)

PR #2183, opened 2025-10-04
bump crengine, cre.cpp: add a few more PageMap / synthetic pages methods

poire-z:
#### bump crengine: fix buildSyntheticPageMap()

Includes:
- lvtinydom: buildSyntheticPageMap(): clean page info https://github.com/koreader/crengine/pull/635

#### cre.cpp: add a few more PageMap / synthetic pages methods

Will be useful to https://github.com/koreader/koreader/pull/14405

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2183)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -1227,6 +1227,36 @@ static int isPageMapSynthetic(lua_State *L) {
     return 1;
 }
 
+static int buildSyntheticPageMap(lua_State *L) {
+    CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+    int chars_per_synthetic_page = luaL_checkint(L, 2);
+    if (doc->dom_doc) {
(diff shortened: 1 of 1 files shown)


hius07:
We can keep CreDocument:buildSyntheticPageMapIfNoneDocumentProvided() and substitute the call in it.

hius07:
`CreDocument.buildSyntheticPageMapIfNoneDocumentProvided = CreDocument.buildSyntheticPageMap`

poire-z:
It would replace real pagemap with synthetic one for the people with the user-patch, wouldn't it ?
You'd need to quickly write a one that would work this evening so we can merge it in frontend (for you to advance on your work) without being able to test it.
Let's keep it and clean later ?

hius07:
> Let's keep it and clean later ?

Yes, agree.

PR #2201, opened 2025-10-25
libarchive: ignore locale and force UTF-8

benoit-pierre:
We don't use `setlocale` anyway, so libarchive's encoding would default to ASCII.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2201)


NiLuJe:
Plus, we have a bunch of platforms (Kindle/Kobo) where `setlocale` would just either fail, or worse, silently do some even weirder broken crap ;o).

PR #2217, opened 2025-11-16
cmake: tweak download command line

benoit-pierre:
To work around MuPDF' slow server:
- increase connection timeout and maximum allowed total time
- drop curl builtin retries: we already manually retrying 2 times anyway (in case of possible incorrect checksum)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2217)

### DIFF
--- a/thirdparty/cmake_modules/koenv.sh
+++ b/thirdparty/cmake_modules/koenv.sh
@@ -83,7 +83,7 @@ download_archive() { (
     for timeout in 0 2 4; do
         sleep ${timeout}
         for url in "$@"; do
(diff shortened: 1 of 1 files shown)


benoit-pierre:
> Wasn't there some fallback mechanism? https://github.com/ArtifexSoftware/mupdf-downloads/releases

Nope! I missed that (I follow https://github.com/ArtifexSoftware/mupdf/releases.atom in my RSS reader, which only has the automatically generated source artifacts, sans sub-modules).

benoit-pierre:
I don't know, I think it still make sense if a mirror is particularly slow. Plus there's no need for retrying within individual curl calls if we do retry ourselves anyway (and potentially alternate with a fallback).

benoit-pierre:
The increased timeout might make things a little slower… ¯\\_(ツ)_/¯

PR #2219, opened 2025-11-17
utf8proc: update to 2.11.2

benoit-pierre:
- https://github.com/JuliaStrings/utf8proc/releases/tag/v2.11.2
- https://github.com/JuliaStrings/utf8proc/releases/tag/v2.11.1

Additionally, trim cdecls to keep only the strict minimum, for a minor reduction in monilibtic builds' code size:
- android-arm: -3.1 KB
- android-arm64: -4.2 KB

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2219)


Frenzie:
It kind of took me by surprise a few months ago because I'd forgotten Benoit added that (which means you're obliged to add things to `*_cdecl.c` or you won't be able to access them by tossing them in an `ffi.cdef` because they simply won't exist).

https://github.com/koreader/koreader-base/pull/1832

benoit-pierre:
[`gen_linker_exports.sh`](https://github.com/koreader/koreader-base/blob/master/utils/gen_linker_exports.sh) is used to parse a set of cdecls files and generate the appropriate linker arguments / script for only exporting what's needed (so you would just need to re-link). It's used for the `k2pdfopt`, `koreader-monolibtic`, and `wrap-mupdf` libraries. We can't use it for all libraries because of inter-dependencies between libraries (ours, potentially the system).

PR #2232, opened 2025-12-06
linux: build in monolibtic mode

benoit-pierre:
This will help improve isolation from system libraries (fixing flathub/rocks.koreader.KOReader#60 for example).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2232)

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -598,6 +598,7 @@ else ifeq ($(TARGET), arm-generic)
 	ARM_ARCH:=$(ARMV6_GENERIC_ARCH)
 	ARM_ARCH+=-mfloat-abi=softfp
 else ifeq ($(TARGET), linux)
+	MONOLIBTIC = 1
 	ifeq ($(LINUX_ARCH), arm)
 		ARM_ARCH:=$(ARMV7_GENERIC_ARCH)
 		ARM_ARCH+=-mfloat-abi=hard


Frenzie:
What concretely does this PR affect? Flatpak is a thing unto itself that should never be allowed to negativily affect non-Flatpak.

benoit-pierre:
> Currently it's very easy to try a (slightly) different library for example.

Meaning? From a deb or tar.xz install, removing the bundled library to switch to the system one?

poire-z:
No real opinions on .deb/flatpack (or AppImage).
Is the emulator affected ?

NiLuJe:
> It also allows any user to plug and play different library versions from nightlies and releases or provided by a developer to quickly do some trouble shooting of a sort that wouldn't be viable with a monolibtic build.

That's a fair point.

Who's currently the main audience for plain the .tar.xz builds?

(i.e., I could see keeping the isolation for the deb packages but not the plain ones, for instance)

benoit-pierre:
The AppImage can be extracted and provides everything (including a SDL2 library).

Frenzie:
It seems to keep failing at building in under an hour.
https://gitlab.com/koreader/nightly-builds/-/jobs/12345856477

Perhaps I should set `when: always` under `cache`?

https://gitlab.com/koreader/nightly-builds/-/blob/master/.gitlab-ci.yml?ref_type=heads#L11-15

Frenzie:
@NiLuJe 
> Who's currently the main audience for plain the .tar.xz builds?

I've always thought of it as slightly more advanced users and Lua-focused hackers. Admittedly, in practice they might prefer to work directly on their Kindle/Kobo, but in any case I've always seen it as equivalent to that and not to something like Flatpak.

Frenzie:
I'm definitely not a fan of this on account of the lack of alternative library use issues, but in practice the documentation's always gently guided people towards using the AppImage I suppose.

PR #2233, opened 2025-12-07
blitbuffer: RGB32 support for rounded corners

hius07:
To be used in colored buttons.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2233)


NiLuJe:
I'd obviously prefer that with the C implementation in, too ;).

(At a quick glance, it wouldn't look too gnarly to implement, just a bunch of code duplication).

PR #2248, opened 2026-01-06
miscellaneous FFI cdecl cleanups

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2248)


ffi-cdecl/freetype2_decl.c
```diff
@@ -12,6 +8,8 @@
 #include <freetype/tttables.h>
 #include <freetype/tttags.h>
 
+#include "ffi-cdecl.h"
```
Frenzie: What's the reason for moving it down btw?

PR #2257, opened 2026-01-21
feat(koptcontext): Add getAllPanelsFromPage() for panel detection

vmanojkumar:
Add a new method to detect all panels on a page using Leptonica's connected component analysis. This enables panel-by-panel navigation for comics and manga in KOReader.

Returns an array of panel bounding boxes {x, y, w, h} for panels that are at least 1/8 of page dimensions (filtering out small elements).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2257)

### DIFF
--- a/ffi/koptcontext.lua
+++ b/ffi/koptcontext.lua
@@ -82,6 +82,13 @@ local KOPTContext_mt = {__index={}}
 
 local __VERSION__ = "1.0.1"
 
+-- Panel detection constants
+local THRESHOLD_WHITE_BG = 225   -- Binary threshold for white/light backgrounds
+local THRESHOLD_DARK_BG = 30    -- Binary threshold for dark backgrounds (before invert)
+local MIN_PANEL_RATIO = 10       -- Minimum panel size: 1/10th of page dimension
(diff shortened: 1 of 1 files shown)


ffi/koptcontext.lua
```diff
@@ -460,6 +460,62 @@ function KOPTContext_mt.__index:getPanelFromPage(pos)
...
+            if w >= img_w / 8 and h >= img_h / 8 then
+                local box_x, box_y, box_w, box_h = boxGetGeometry(box)
+                table.insert(panels, {
+                    x = box_x,
+                    y = box_y,
+                    w = box_w,
+                    h = box_h,
+                })
+            end
+        end
```
Frenzie: Couldn't this be simplified to something like this (untested both functionally and performance-wise, regard as pseudocode)?
```suggestion
    for box_x, box_y, box_w, box_h in boxaIterBoxGeometries(bb) do
        if box_w >= img_w / 8 and box_h >=  img_h / 8 then
            panels[#panels + 1] = { x = box_x, y = box_y, w = box_w, h = box_h }
        end
    end
```

PR #2262, opened 2026-02-03
lua-Spore: Patch to accept more content-types as json

bartlibert:
Refer to discussion [14710](https://github.com/koreader/koreader/discussions/14710#discussion-9236796rl)
Fixes issue [14596](https://github.com/koreader/koreader/issues/14596#issue-3605343556)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2262)

### DIFF
--- a/thirdparty/lua-Spore/detect-more-json-content-types.patch
+++ b/thirdparty/lua-Spore/detect-more-json-content-types.patch
@@ -1,5 +1,5 @@
 diff --git i/src/Spore/Middleware/Format/JSON.lua w/src/Spore/Middleware/Format/JSON.lua
-index 225b25e..4b9305f 100644
+index 225b25e..d3419b5 100644
 --- i/src/Spore/Middleware/Format/JSON.lua
 +++ w/src/Spore/Middleware/Format/JSON.lua
 @@ -6,6 +6,7 @@ local decode = require'dkjson'.decode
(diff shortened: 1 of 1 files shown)


thirdparty/lua-Spore/detect-more-json-content-types.patch
```diff
@@ -0,0 +1,37 @@
...
+ local _ENV = nil
+ local m = {}
+
+ m['content-type'] = 'application/json'
++
++local has_json_mime_type = function(header)
++    if find(header, m['content-type'], 1, true) then
++        return true
++    end
++    if string.match(header, "application/[%w%p]-json") then
```
Frenzie: This strikes me as not quite right. It should at least be `"application/[%w%p]-%+json"`, though I'd possibly lean towards something even stricter like `"application/[%w.]-%+json"` (concretely, it shouldn't contain pretty much any of `()<>@,;:\\"/[]?={}` iirc).

As an aside but I think that's already an issue regardless, presumably it should split on `;` if present to ignore something like `; charset=UTF-8` if present *without* regarding `application/jsonbutnotreallyhaha` as JSON.

thirdparty/lua-Spore/detect-more-json-content-types.patch
```diff
@@ -0,0 +1,37 @@
...
+ 
+ local _ENV = nil
+@@ -13,6 +14,15 @@ local m = {}
+ 
+ m['content-type'] = 'application/json'
+ 
++local has_json_mime_type = function(header)
++    if find(header, m['content-type'], 1, true) then
++        return true
++    end
```
Frenzie: This old code shouldn't remain here since that features the `application/jsonbutnotreally` issue.

If such a shortcut is desirable maybe it could be made into something like this:
```lua
if header == m['content-type'] or find(header, m['content-type'] .. ';', 1, true) then
    return true
end
```

Though I'm inclined to think it shouldn't be more complicated than this, and leave further handling to the rest of the code.
```lua
if header == m['content-type'] then
    return true
end
```

PR #2263, opened 2026-02-03
lpeg: use a GitHub URL for downloading

benoit-pierre:
More dependable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2263)


benoit-pierre:
Irregular part?

PR #2271, opened 2026-02-14
ci/circle: bump docker images

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2271)


benoit-pierre:
Patch for gitlab: [0001-bump-docker-images.patch](https://github.com/user-attachments/files/25315084/0001-bump-docker-images.patch).

PR #2272, opened 2026-02-14
sdl: switch to SDL3 (3.4.0)

benoit-pierre:
When building the emulator, SDL3 is automatically built if the library is not already available.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2272)


Frenzie:
I just meant as a default, but simply updating the development instructions might suffice. Pinging @poire-z for an opinion.

poire-z:
I have no real opinion :/
I know not much about SDL, I'm just a end user.
I see it is available in debian13, so if I can just install these packages and it works (with an old Windows X server, https://vcxsrv.com/, which might not have all the recent capabilities/extensions), fine with me.
(And if I don't have them installed, it will just build against the libsdl2 I have, and no change, right?)
Dunno about the impact on other end users.

Frenzie:
This is a full upgrade/replacement, so without a prebuilt SDL3 from the system it'll have to be built as part of the program.

poire-z:
OK thanks.
So, still no real opinion :) (I'm allergic to change, but you both know better than me what's best on this topic.)

Frenzie:
I'm mainly hoping it'll make the "hidpi" situation less awful and not so infuriatingly inferior to the situation in X11 two decades ago. (That's more about Wayland adding some stuff in the meantime than about SDL, but only SDL 3 supports those features. The same also applies for Mac.)

Besides that there's also a new Dialog API which might be of some interest (see <https://github.com/koreader/koreader-base/pull/631> back when that was in the program for a while). Keyboard input should also be slightly better but it already seemed perfectly functional to me.

poire-z:
(I have a bit messy setup for compiling koreader - with some -dev packages manually extracted in some dedicated tree, and not apt-installed, including libsdl2-dev, and they are still from debian12, while I'm on debian13. Will just apt install libsdl3 (without the -dev) be enough to build and notice we don't need to build our own SDL3? Or do I need the libsdl3-dev ? (not avalable for debian12, so I'll have to update my mess.)

benoit-pierre:
The shared library is enough, no need for the dev package.

poire-z:
`-- Found libSDL3.so.0 full path: /usr/lib/gcc/x86_64-linux-gnu/14/../../../x86_64-linux-gnu/libSDL3.so.0`
Rebuilt went fine.
When launched, the luajit has opened `/usr/lib/x86_64-linux-gnu/libSDL3.so.0.2.10`.
No problem with my Windows X server, and looks like no problem either when using nxagent.
So, all good on my side.

PR #2279, opened 2026-02-19
Expose MuPDF's search function

Commodore64user:
### what's new

I was originally trying to do this in Lua but it turns out this function was there all along for anyone to C (sorry!)

massively simplifies https://github.com/koreader/koreader/pull/15001

* Added a C wrapper function `mupdf_fz_search_stext_page` that exposes the MuPDF `fz_search_stext_page` API for searching text on a page.
* Registered the new wrapper function with the FFI cdecl system to make it available to Lua.
* Declared the new function in the FFI Lua header for use in Lua code.
* Added a `search` method to the `page_mt` metatable in `mupdf.lua`, which uses the new C function to find all matches for a given string on a page and returns their bounding box coordinates.

<details>
  <summary>my effort ;( </summary>
  
  ```lua
function HtmlBoxWidget:_findMatchesOnPage(page_idx, search_text)
    local page = self.document:openPage(page_idx)
    if not page then return nil end

    local lines = page:getPageText()
    page:close()

    if not lines or type(lines) ~= "table" then return nil end

    -- Build searchable content with position mapping
    local content_parts = {}
    local word_map = {}
    local current_pos = 0

    for _, line in ipairs(lines) do
        if type(line) == "table" then
            for _, word_obj in ipairs(line) do
                if word_obj and word_obj.word then
                    -- Note: We lowercase words BEFORE building the position map because
                    --       case conversion can change UTF-8 byte lengths (e.g., Turkish İ→i: 2 bytes→1 byte).
                    --       If we lowercased after, our byte positions would be misaligned with the search string.
                    -- e.g.,
                    -- print("Original:", test, "Bytes:", #test)    -->    Original:     İİİİİ hello world       Bytes:     22
                    -- print("Lowered:", lower, "Bytes:", #lower)   -->    Lowered:      iiiii hello world       Bytes:     17
                    local word_text = Utf8Proc.lowercase(word_obj.word, false)
                    local start_pos = current_pos + 1

                    table.insert(content_parts, word_text)
                    table.insert(content_parts, " ")
                    current_pos = current_pos + #word_text + 1 -- +1 for space

                    table.insert(word_map, {
                        start_pos = start_pos,
                        end_pos = start_pos + #word_text - 1,
                        box = word_obj,
                        line = line
                    })
                end
            end
        end
    end
    local content = table.concat(content_parts)

    -- Search in the already-lowercased content
    local search_lower = Utf8Proc.lowercase(search_text, false)
    local match_rects = {}
    local search_start = 1
    local word_index = 1     -- Track position in word_map
    local last_word_idx = -1 -- Track last added word to prevent redundant rects

    -- Algorithm is O(n+m), n: number of words, m: matches. It makes a single forward pass
    -- through the word array, collecting rectangles for m matches along the way.
    while true do
        local match_start, match_end = content:find(search_lower, search_start, true)
        if not match_start then break end

        -- Skip words that end before this match starts, word_index only advances forward.
        while word_map[word_index] and word_map[word_index].end_pos < match_start do
            word_index = word_index + 1
        end

        -- Collect all words that overlap with this match
        local i = word_index
        while word_map[i] and word_map[i].start_pos <= match_end do
            local word_mapping = word_map[i]
            local overlap_start = math.max(word_mapping.start_pos, match_start)
            local overlap_end = math.min(word_mapping.end_pos, match_end)

            -- Ensure we only add the rect if there is an overlap and we haven't added this word yet
            if overlap_start <= overlap_end and i > last_word_idx then
                local rect = Geom:new{
                    x = word_mapping.box.x0,
                    y = word_mapping.line.y0,
                    w = word_mapping.box.x1 - word_mapping.box.x0,
                    h = word_mapping.line.y1 - word_mapping.line.y0,
                }
                table.insert(match_rects, rect)
                last_word_idx = i
            end
            i = i + 1
        end -- while word overlap
        -- Note: Moving to match_end + 1 avoids re-processing the same match, but
        --       it will skip overlapping results (e.g., searching "aba" in "ababa"
        --       will only return the first instance).
        search_start = match_end + 1
    end -- while true
    return #match_rects > 0 and match_rects or nil
end
  ```
  
</details> 

https://github.com/ArtifexSoftware/mupdf/blob/8c39e04a09c74fcc96ebb704bf6d1f59adb9b6fa/source/fitz/stext-search.c#L1254-L1285

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2279)

### DIFF
--- a/wrap-mupdf.c
+++ b/wrap-mupdf.c
@@ -154,7 +154,6 @@ int mupdf_get_cache_size() {
 int mupdf_error_code(fz_context *ctx) {
     return ctx->error.errcode;
 }
-
 char* mupdf_error_message(fz_context *ctx) {
     return ctx->error.message;
 }


ffi/mupdf_h.lua
```diff
@@ -314,6 +314,7 @@ void *mupdf_pdf_save_document(fz_context *, pdf_document *, const char *, pdf_wr
 fz_alloc_context *mupdf_get_my_alloc_context();
 int mupdf_get_cache_size();
 int mupdf_error_code(fz_context *);
+int mupdf_fz_search_stext_page(fz_context *ctx, fz_stext_page *text, const char *needle, int *hit_mark, fz_quad *hit_bbox, int hit_max);
```
Frenzie: The CI was complaining about this location not matching with the cdefs, but mind like I said I'd like to improve the location in the cdefs. ^_^

wrap-mupdf.c
```diff
@@ -154,6 +154,7 @@ int mupdf_get_cache_size() {
 int mupdf_error_code(fz_context *ctx) {
     return ctx->error.errcode;
 }
+
```
Commodore64user:
```suggestion
```

benoit-pierre (approved):

PR #2327, opened 2026-04-08
feat: Kindle: add uevent listener for input device hotplug

zampierilucas:
A user (@alllexx88, [kindle-hid-passthrough#40](https://github.com/zampierilucas/kindle-hid-passthrough/issues/40)) pointed out that keyboards connected via [kindle-hid-passthrough](https://github.com/zampierilucas/kindle-hid-passthrough) don't work in KOReader because KOReader never discovers the dynamically created `/dev/input/eventX` device.

On Kobo, input-kobo.h already uses libue to listen for netlink uevents and emit `CODE_FAKE_USB_DEVICE_PLUGGED_IN/OUT` when new input devices appear. Kindle's input-kindle.h had no equivalent, only the lipc-wait-event loop for power/charging events. This patch adds a uevent listener as a forked child alongside the existing lipc loop, using the same libue library already in the repo. When a new evdev node appears (e.g., a UHID keyboard), it emits the same fake event codes that the Kobo path uses, so the existing externalkeyboard plugin can handle it. Tested on a Kindle PW4 with kindle-hid-passthrough and a BLE keyboard, KOReader correctly shows the "Keyboard connected" banner and accepts input.

Ref koreader/koreader#14359

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2327)

### DIFF
--- a/input/input-kindle.h
+++ b/input/input-kindle.h
@@ -33,7 +33,7 @@ static void slider_handler(int sig)
         kill(pclose_arg.pid, SIGINT);
     }
     /* Also kill the uevent listener child */
-    if(ue_child_pid != 0) {
+    if (ue_child_pid != 0) {
         kill(ue_child_pid, SIGTERM);
     }
 }
@@ -67,7 +67,7 @@ static void sendEvent(int fd, struct input_event* ev)
 // c.f., input-kobo.h for the Kobo equivalent (which also handles USB OTG).
 static void ueventInputListener(int pipefd_w)
 {
-    // Die when our parent (the lipc process) exits
+    // Die when our parent (the lipc process) exits.
     prctl(PR_SET_PDEATHSIG, SIGTERM);
 
     struct uevent_listener listener = { 0 };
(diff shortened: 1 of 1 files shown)


input/input-kindle.h
```diff
@@ -53,6 +60,60 @@ static void sendEvent(int fd, struct input_event* ev)
...
+    while ((re = ue_wait_for_event(&listener, &uev)) == 0) {
+        // Match any input subsystem event with an evdev device node.
+        // On Kindle, dynamically created input devices (e.g., UHID keyboards
+        // from kindle-hid-passthrough) appear under /devices/virtual/input/.
+        // We intentionally don't filter on devpath (unlike Kobo which gates on
+        // /devices/platform/soc for USB OTG) because UHID devices live under
+        // /devices/virtual/, and built-in devices are static so won't generate
+        // add/remove uevents.
+        if (uev.subsystem && UE_STR_EQ(uev.subsystem, "input") &&
+            uev.devname && UE_STR_EQ(uev.devname, "input/event")) {
```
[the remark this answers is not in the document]
zampierilucas: Yeah, the name is a bit confusing but `UE_STR_EQ` is actually a prefix match, it'll match input/event0, input/event12, etc. Same usage as the Kobo code in input-kobo.h:97. See https://github.com/koreader/koreader-base/blob/master/input/libue.h#L91-L92

input/input-kindle.h
```diff
@@ -29,6 +32,10 @@ static void slider_handler(int sig)
         /* Be a little more gracious, lipc seems to handle SIGINT properly */
         kill(pclose_arg.pid, SIGINT);
     }
+    /* Also kill the uevent listener child */
+    if(ue_child_pid != 0) {
```
benoit-pierre:
```suggestion
    if (ue_child_pid != 0) {
```

input/input-kindle.h
```diff
@@ -53,6 +60,60 @@ static void sendEvent(int fd, struct input_event* ev)
...
 }
 
+// Watch for input device hotplug via netlink uevents.
+// Used to detect UHID devices created by kindle-hid-passthrough (or any other
+// userspace HID driver). Runs in a forked child, writes to the same pipe as
+// the lipc-wait-event loop.
+// c.f., input-kobo.h for the Kobo equivalent (which also handles USB OTG).
+static void ueventInputListener(int pipefd_w)
+{
+    // Die when our parent (the lipc process) exits
```
benoit-pierre:
```suggestion
    // Die when our parent (the lipc process) exits.
```

input/input-kindle.h
benoit-pierre: I would use continue to reduce the indentation level:
```suggestion
        if (!(uev.subsystem && UE_STR_EQ(uev.subsystem, "input") &&
            uev.devname && UE_STR_EQ(uev.devname, "input/event")))
            continue;
```

input/input-kindle.h
```diff
@@ -127,6 +201,12 @@ static void generateFakeEvent(int pipefd[2]) {
         }
     }
 
+    /* Clean up the uevent listener child */
```
benoit-pierre:
```suggestion
    /* Clean up the uevent listener child. */
```

benoit-pierre (approved):

zampierilucas:
@benoit-pierre Added your suggestions in the second commit!

input/input-kindle.h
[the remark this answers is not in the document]
Frenzie: I see. Yes, that name's deceptive. I'd call it UE_STARTS_WITH or UE_STR_PREFIX_EQ or something like that and I doubt any kind of comment would be needed in that case.

PR #2330, opened 2026-04-08
make: do not use `-march=native` for CI emulator builds

benoit-pierre:
Avoid potential build cache mismatches due to different runner hardware.

I don't think I've seen the issue on Circle CI, but on Github Actions, I've definitively ended up with some weird build with 0% ccache hit rate, or worse, all the tests crashing with an illegal instruction exception after restoring a build directory from cache.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2330)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ parameters:
     # Bump this to reset all caches.
     cache_epoch:
       type: integer
-      default: 4
+      default: 5
 
 # }}}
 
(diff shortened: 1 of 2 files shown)


Frenzie:
> I don't think I've seen the issue on Circle CI

You wouldn't, because the architecture is added to the cache key. If you put special casing for CI in the Makefile you should probably remove that to avoid unnecessary cache misses. (Or conversely, add it to GHA.)
https://github.com/koreader/koreader-base/blob/ffcf5e6b9fb5051e443b02f0936c89438b9d6e18/.circleci/config.yml#L161

Frenzie:
That is, `arch` comes out as something like `arch1-linux-amd64-1-1234` where everything after amd64 indicates the specific CPU.

benoit-pierre:
> > I don't think I've seen the issue on Circle CI
> 
> You wouldn't, because the architecture is added to the cache key. If you put special casing for CI in the Makefile you should probably remove that to avoid unnecessary cache misses. (Or conversely, add it to GHA.)
> 
> https://github.com/koreader/koreader-base/blob/ffcf5e6b9fb5051e443b02f0936c89438b9d6e18/.circleci/config.yml#L161

Right, forgot about that. There does not seem to be an equivalent variable on GA, but I think it does make more sense to indeed reduce cache misses.

PR #2326, opened 2026-04-07
ffi-cdecl: update to 9c9d46a

benoit-pierre:
NOTE: this include NiLuJe/FBInk#87 for now, as we are missing a number of other changes to master: https://github.com/NiLuJe/FBInk/compare/92e12700..master.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2326)

PR #2331, opened 2026-04-09
ffi-cdecl: rework to support all platforms

benoit-pierre:
When the target compiler is not GCC, switch to pre-processing the output with it before running the plugin with the build host GCC.

With a few tweaks to the pre-processed output (some Clang builtin float typedefs conflicting with GCC, and for some reason a duplicate ioctl prototype on Android), this make it possible to generate FFI cdecls for macOS and Android too.

Depends on #2326, and some Docker images will need to be updated.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2331)


Frenzie (approved):

benoit-pierre:
The `ffi-cdecl` build is broken again on my machine due to the GCC 16 update (the plugin builds but does not run…) *sigh*

I have a POC using tree-sitter instead, I think I'll polish that instead so we can drop the GCC plugin dependency altogether.

benoit-pierre:
Superseded by #2360.

PR #2332, opened 2026-04-13
ffi/util,SDL3: stop SDL3 gamepad thread around fork() to fix Linux UI freeze

WillScarlettOhara:
## Problem
On Linux (tested on CachyOS / Arch with GNOME on Wayland, koreader-appimage 2026.03-1), changing the font size or any other reflowable-rendering parameter causes KOReader to completely freeze ("luajit Is Not Responding" in GNOME).
Screenshot From 2026-04-13 08-27-43

### Root cause
When the font-size change triggers a background rerender, `ffi/util.lua:runInSubProcess` calls `C.fork()`. At that moment SDL3's gamepad/joystick subsystem has a background Linux thread that periodically calls `malloc()` for hotplug detection.
glibc's `pthread_atfork` *prepare* handler must lock every malloc arena before `fork()` can proceed. If the SDL thread holds an arena lock, the main UI thread blocks inside `fork()` until that lock is released.
While this is a glibc + Linux issue that can affect any distro, on schedulers like CachyOS's **BORE/LAVD**, background threads receive longer CPU time slices, so the SDL thread does not yield quickly. This makes the UI stall visibly and causes GNOME to flag the window as unresponsive.
## Fix
Add two new hook tables to `ffi/util.lua`:
- `_run_before_fork_funcs` — called in the **parent just before** `C.fork()`
- `_run_after_fork_parent_funcs` — called in the **parent just after** `C.fork()` returns (pid ≠ 0), unconditionally (even on failure)
`ffi/SDL3.lua` registers hooks that, **on Linux only**:
- **Before fork:** explicitly closes the open `SDL_Gamepad` handle (disarming its `ffi.gc` finalizer first to prevent a later double-free on the stale pointer), then calls `SDL_QuitSubSystem(SDL_INIT_GAMEPAD)` to tear down the background thread before `fork()` touches any arena locks.
- **After fork (parent):** calls `SDL_InitSubSystem(SDL_INIT_GAMEPAD)` and `openGameController()` to restore the subsystem.
`ffi/SDL3_h.lua` adds the two missing FFI prototypes: `SDL_InitSubSystem` and `SDL_QuitSubSystem`.
## Testing
Tested on:
- CachyOS (Arch Linux), kernel 6.19.11-1-cachyos, BORE/LAVD scheduler
- GNOME Shell 49.5 on Wayland
- `koreader-appimage 2026.03-1`
After the fix, changing font size triggers background rerendering in ~1 s with no UI freeze.
> Note: this patch was developed with AI assistance.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2332)


ffi/SDL3_h.lua
```diff
@@ -1029,7 +1029,9 @@ bool SDL_EnableScreenSaver(void);
 bool SDL_GetCurrentRenderOutputSize(SDL_Renderer *, int *, int *);
 bool SDL_GetWindowSize(SDL_Window *, int *, int *);
 bool SDL_Init(unsigned int);
+bool SDL_InitSubSystem(unsigned int);
 void SDL_Quit(void);
+void SDL_QuitSubSystem(unsigned int);
```
Frenzie, line 1034: Please also add this to cdecl.

PR #2345, opened 2026-04-22
bump crengine: MOBI and CSS fixes; cre.cpp: add getNearestWordFromPosition

poire-z:
#### bump crengine: MOBI and CSS fixes

Includes:
- https://github.com/koreader/crengine/pull/661
  - pdbfmt: don't overwrite extracted metadata
  - pdbfmt: extract more metadata from MOBI EXTH
  - pdbfmt: extract MOBI non-EXTH title if necessary
- https://github.com/koreader/crengine/pull/662
  - CSS first-line: fix possible crash
  - CSS width: fit-content: account for the element padding
  - getNodeByPoint(): improve fallback check

#### cre.cpp: add getNearestWordFromPosition()

Useful for improve text selection using DPAD/arrow keys.
Requested along https://github.com/koreader/crengine/issues/659.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2345)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -2698,6 +2698,52 @@ static int getWordBoxesFromPositions(lua_State *L) {
 	return 1;
 }
 
+static int getNearestWordFromPosition(lua_State *L) {
+    CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+    int x = luaL_checkint(L, 2);
+    int y = luaL_checkint(L, 3);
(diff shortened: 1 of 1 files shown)


poire-z:
@benoit-pierre : there are 20+ commits in koreader-base that have not yet reached frontend.
Are they safe to bump by you or me ?
I guess you might want to investigate that libressl issue I just saw before we do that.

PR #2343, opened 2026-04-21
libressl: update to 4.3.1

benoit-pierre:
- https://github.com/libressl/portable/releases/tag/v4.3.1
- https://github.com/libressl/portable/releases/tag/v4.3.0

Code size change:
- `android-arm`  : + 8.3 KB
- `android-arm64`: +21.1 KB
- `kindlepw2`    : + 4.5 KB
- `linux`        : + 6.8 KB

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2343)

PR #2346, opened 2026-04-22
make: explicitly disable executable stack

benoit-pierre:
To prevent a possible issue on Gentoo (cf. https://github.com/koreader/koreader-base/pull/2343#issuecomment-4299401787), and since no code we build should need it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2346)


benoit-pierre:
Damn outdated Ubuntu…

PR #2348, opened 2026-05-04
blitbuffer: fix GCC 16 warning (-Wunused-but-set-variable)

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2348)


Frenzie (approved):
I guess it's a leftover from the full antialiased form of the function?

PR #2351, opened 2026-05-06
Support SDL3 pen events

HackingGate:
## Summary

Adds SDL3 pen event translation to the SDL input shim:

- turns SDL3 pen down/move/axis/up events into Linux-style multi-touch slot events
- forwards pen hover/proximity as the stylus slot with `ABS_MT_TRACKING_ID = -1`, coordinates, and current tool type
- emits `BTN_TOOL_PEN` / `BTN_TOOL_RUBBER` and `ABS_MT_TOOL_TYPE` so KOReader can distinguish pen tips and eraser tips
- emits stylus side-button key events as `BTN_STYLUS` / `BTN_STYLUS2`
- suppresses pen-generated synthetic mouse events so pen input does not double-report as mouse/touch input
- clears active pen button/tool state when the pen leaves proximity

## Validation

- `luajit -b ffi/SDL3.lua /tmp/koreader-base-sdl3.lua.bin`
- tested on a Linux Wayland session with a real pen
- verified hover/proximity, contact, lift, proximity-out, and eraser behavior in the Pencil input debug log

## What I use it for

Add pen support on Linux tablet.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2351)

### DIFF
--- a/ffi/SDL3.lua
+++ b/ffi/SDL3.lua
@@ -216,6 +216,124 @@ end
 -- Mouse buttons are keyed by slot, while touch contacts are keyed by fingerID.
 local pointers = {}
 local finger_pointers = {}
+local pen_pointers = {}
+local pen_button_state = {}
+
+local SDL_TOUCH_MOUSEID = 0xFFFFFFFF -- ((SDL_MouseID)-1), generated as signed -1 in our cdefs.
+local SDL_PEN_MOUSEID = 0xFFFFFFFE -- ((SDL_MouseID)-2), not generated in our cdefs.
+local SDL_PEN_SLOT = 4
+local TOOL_TYPE_PEN = 1
+local TOOL_TYPE_ERASER = 2
+local SDL_PEN_INPUT_DOWN = bit.lshift(1, 0)
+local SDL_PEN_INPUT_BUTTON_1 = bit.lshift(1, 1)
+local SDL_PEN_INPUT_BUTTON_2 = bit.lshift(1, 2)
+local SDL_PEN_INPUT_ERASER_TIP = bit.lshift(1, 30)
+
+local function isMousePointer(which)
+    local pointer_id = tonumber(which)
+    return pointer_id ~= SDL_TOUCH_MOUSEID and pointer_id ~= SDL_PEN_MOUSEID
+end
+
+local function hasPenFlag(pen_state, flag)
+    return bit.band(tonumber(pen_state) or 0, flag) ~= 0
+end
+
+local function getPenTool(eraser, pen_state)
+    if eraser or hasPenFlag(pen_state, SDL_PEN_INPUT_ERASER_TIP) then
+        return TOOL_TYPE_ERASER
+    end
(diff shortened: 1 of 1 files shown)


ffi/SDL3.lua
```diff
@@ -344,7 +462,52 @@ function S.waitForEvent(sec, usec)
...
+    elseif event.type == SDL.SDL_EVENT_PEN_BUTTON_DOWN
+        or event.type == SDL.SDL_EVENT_PEN_BUTTON_UP then
+        local button_code
+        if event.pbutton.button == 1 then
+            button_code = C.BTN_STYLUS
+        elseif event.pbutton.button == 2 then
+            button_code = C.BTN_STYLUS2
+        end
+        setPenButtonState(button_code, event.type == SDL.SDL_EVENT_PEN_BUTTON_DOWN)
+    elseif event.type == SDL.SDL_EVENT_PEN_PROXIMITY_OUT then
```
Frenzie, line 637: Since you're already this far I think you might potentially also wish to send through in proximity. The frontend may then opt to show where the brush will start like in Krita. But there's no need to; it can always be added later if desired.

Frenzie (approved):
It's basically my own code from a few years ago with minor adaptations, so um, lgtm. ;-)

ffi/SDL3.lua
HackingGate: Thanks, I handled this in follow-up commits.

75c7b4c5 refines the SDL pen handling and suppresses synthetic mouse events, which could otherwise cause accidental touches.

e1e386ba then forwards hover/proximity to the frontend as slot=4 id=-1 with coordinates and tool type, so the frontend can see the pen position while it is in proximity but not touching.

Frenzie:
What's the purpose behind the `not not`s?

(GH wouldn't let me comment on a relevant line.)

HackingGate:
@Frenzie 
The type is already guaranteed to be a boolean at runtime, so `not not` is unnecessary and has been removed. The LLM agent was probably being too defensive about type coercion.

Frenzie:
Which LLM was involved? It didn't obviously strike me as LLM code, though then again like I said it was mostly just my own code from a few years ago with minor adjustments for pen instead of touch.

@benoit-pierre Did you have any thoughts on it, seeing how you touched this file not too long ago?

HackingGate:
To clarify, I used Codex GPT-5.5 xhigh as an assistant.

I tested the behavior before pushing, especially the pen-related behavior, and checked that it didn’t regress existing input handling like finger touch and mouse input. I gave Codex codebase context, logs, and feedback during iteration, but I didn’t know the full history of the surrounding code and didn’t do a full line-by-line manual review of the diff. That’s how small cleanup issues like the unnecessary `not not`s slipped through.

HackingGate:
I also noticed that SDL can leave an extra mouse cursor visible at the top-left when using pen input. That looks separate from the touch/mouse event translation itself, but it seems to be a side effect exposed by this PR.

PR #2357, opened 2026-05-07
make: change android targets to `android-xxx`

benoit-pierre:
Similar to kodev, making it easier to use shell loops like: `for t in android-arm kindlepw2; do make TARGET=$t; done`.

NOTE: Technically `make ANDROID_ARCH=xxx TARGET=android` is still supported (so things like `make TARGET=android-arm re` still work).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2357)


Frenzie (approved):
Alright.

PR #2360, opened 2026-05-17
ffi-cdecl: new implementation based on tree-sitter

benoit-pierre:
Alternative to #2331, as the update to GCC 16 locally (Arch Linux) broke the gcc-lua plugin again…

Implementation: the result of pre-processing with the real target compiler is parsed with tree-sitter to extract cdecls, and passed to the target compiler's assembler to extract constant values if needed.

Plus:
- work for all platforms
- no dependency on GCC plugin, so should not be impacted by updates
- better handling of constants (use unsigned for positive constants, instead of always int, so values > 2147483647 are correctly declared)

Minus:
- tree-sitter can fail to parse some C files (even if pre-processed), example: the czmq FFI cdecls can't be generated on macOS
- slightly slower than the GCC plugin based version

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2360)


ffi/crypto_h.lua
```diff
@@ -1,17 +1,18 @@
...
-void EVP_CIPHER_CTX_free(struct evp_cipher_ctx_st *);
-int EVP_CIPHER_CTX_set_padding(struct evp_cipher_ctx_st *, int);
-int EVP_CIPHER_block_size(const struct evp_cipher_st *);
-int EVP_DecryptInit_ex(struct evp_cipher_ctx_st *, const struct evp_cipher_st *, struct engine_st *, const unsigned char *, const unsigned char *);
-int EVP_DecryptUpdate(struct evp_cipher_ctx_st *, unsigned char *, int *, const unsigned char *, int);
-int EVP_DecryptFinal_ex(struct evp_cipher_ctx_st *, unsigned char *, int *);
-const struct evp_cipher_st *EVP_aes_128_ecb(void);
-const struct evp_cipher_st *EVP_aes_192_ecb(void);
-const struct evp_cipher_st *EVP_aes_256_ecb(void);
+require("ffi").cdef[[
```
Frenzie, line 3: Is that fully equivalent for this specific case? The LuaJIT documentation makes a big stink about using `local ffi = require("ffi")`, although presumably that's more regarding the fact that it's never global.

ffi/leptonica_h.lua
```diff
@@ -1,57 +1,56 @@
...
-l_int32 pixGetWidth(const PIX *);
-PIX *pixInvert(PIX *, PIX *);
-PIX *pixMultiplyByColor(PIX *, PIX *, BOX *, l_uint32);
-BOXA *pixSplitIntoBoxa(PIX *, l_int32, l_int32, l_int32, l_int32, l_int32, l_int32);
-PIX *pixThresholdToBinary(PIX *, l_int32);
-int pixWriteMemPng(l_uint8 **, size_t *, PIX *, l_float32);
-int pixWritePng(const char *, PIX *, l_float32);
+static const unsigned L_NOCOPY = 0;
+static const unsigned L_COPY = 1;
+static const unsigned L_CLONE = 2;
```
Frenzie, line 20: It changed from signed to unsigned? But isn't it defined as int in the leptonica headers?

ffi/libarchive_h.lua
```diff
@@ -1,34 +1,34 @@
...
 typedef ssize_t la_ssize_t;
-static const int AE_IFREG = 32768;
-static const int AE_IFLNK = 40960;
-static const int AE_IFSOCK = 49152;
-static const int AE_IFCHR = 8192;
-static const int AE_IFBLK = 24576;
-static const int AE_IFDIR = 16384;
-static const int AE_IFIFO = 4096;
-static const int ARCHIVE_EOF = 1;
+static const unsigned AE_IFREG = 32768;
```
Frenzie, line 8: Right, same everywhere regarding int vs unsigned int.

ffi/crypto_h.lua
Frenzie: > better handling of constants (use unsigned for positive constants, instead of always int, so values > 2147483647 are correctly declared)

I see, it relates to that I suppose. I do find the mismatch between what I see in the header and what's output a bit odd.

Frenzie:
> tree-sitter can fail to parse some C files (even if pre-processed), example: the czmq FFI cdecls can't be generated on macOS

I'm a bit confused since the main positive seems to be that it works for all platforms. ;-)

As long as it works I don't have any particular preference. If you think it's more dependable let's go for it. Pinging @NiLuJe

ffi/libarchive_h.lua
benoit-pierre: Yes, previously, int would always be used. For example:
```c
▸ printf '%s\n' '#include <limits.h>' '#include "ffi-cdecl.h"' 'cdecl_const(UINT_MAX);' | tee test.c
#include <limits.h>
#include "ffi-cdecl.h"
cdecl_const(UINT_MAX);

▸ ./build/x86_64-pc-linux-gnu-debug/staging/bin/ffi-cdecl test.c | tee test.lua
-- Automatically generated with ffi-cdecl.

local ffi = require("ffi")

ffi.cdef[[
static const int UINT_MAX = 4294967295;
]]

▸ luajit -l test -e 'print(require("ffi").C.UINT_MAX)'
-1
```
Which still works fine when passing the value to a C function, but can be problematic on the LUA side.

benoit-pierre:
> > tree-sitter can fail to parse some C files (even if pre-processed), example: the czmq FFI cdecls can't be generated on macOS
> 
> I'm a bit confused since the main positive seems to be that it works for all platforms. ;-)

Apart from those czmq cdecls, yes.

ffi/crypto_h.lua
benoit-pierre: Yeah, I don't think there's anything against doing it like this.

ffi/libarchive_h.lua
Frenzie: Whoops, I left https://github.com/koreader/koreader-base/pull/2360#discussion_r3255134124 in the wrong place.

So previously it had a similar issue of sorts and it merely coincidentally looked how I expected it to look. Since for positive numbers it doesn't matter, er, I suppose that's all good.

ffi/freetype_h.lua
```diff
@@ -1,109 +1,112 @@
...
-  FT_ENCODING_MS_BIG5 = 1651074869,
-  FT_ENCODING_MS_WANSUNG = 2002873971,
-  FT_ENCODING_MS_JOHAB = 1785686113,
-  FT_ENCODING_ADOBE_STANDARD = 1094995778,
-  FT_ENCODING_ADOBE_EXPERT = 1094992453,
-  FT_ENCODING_ADOBE_CUSTOM = 1094992451,
-  FT_ENCODING_ADOBE_LATIN_1 = 1818326065,
-  FT_ENCODING_OLD_LATIN_2 = 1818326066,
-  FT_ENCODING_APPLE_ROMAN = 1634889070,
+  FT_ENCODING_NONE = (((FT_UInt32)(unsigned char)(0) << 24) | ((FT_UInt32)(unsigned char)(0) << 16) | ((FT_UInt32)(unsigned char)(0) << 8) | (FT_UInt32)(unsigned char)(0)),
```
Frenzie, line 58: Incidentally, this seems like a very convoluted way of saying 0. But I see, that's what the header does under the hood:

```c
#define FT_ENC_TAG( value, a, b, c, d )                             \
          value = ( ( FT_STATIC_BYTE_CAST( FT_UInt32, a ) << 24 ) | \
                    ( FT_STATIC_BYTE_CAST( FT_UInt32, b ) << 16 ) | \
                    ( FT_STATIC_BYTE_CAST( FT_UInt32, c ) <<  8 ) | \
                      FT_STATIC_BYTE_CAST( FT_UInt32, d )         )

#endif /* FT_ENC_TAG */
```

```c
  typedef enum  FT_Encoding_
  {
    FT_ENC_TAG( FT_ENCODING_NONE, 0, 0, 0, 0 ),

    FT_ENC_TAG( FT_ENCODING_MS_SYMBOL, 's', 'y', 'm', 'b' ),
    FT_ENC_TAG( FT_ENCODING_UNICODE,   'u', 'n', 'i', 'c' ),

    FT_ENC_TAG( FT_ENCODING_SJIS,    's', 'j', 'i', 's' ),
    FT_ENC_TAG( FT_ENCODING_PRC,     'g', 'b', ' ', ' ' ),
    FT_ENC_TAG( FT_ENCODING_BIG5,    'b', 'i', 'g', '5' ),
    FT_ENC_TAG( FT_ENCODING_WANSUNG, 'w', 'a', 'n', 's' ),
    FT_ENC_TAG( FT_ENCODING_JOHAB,   'j', 'o', 'h', 'a' ),
```

benoit-pierre:
I've dropped the last 3 commits, since otherwise there's a bit of chicken and egg issue with the need for updated docker images.

ffi/loadlib.lua
```diff
@@ -18,7 +18,7 @@ ABI compatibility).
 
 local ffi = require("ffi")
 local android = ffi.os == "Linux" and os.getenv("IS_ANDROID") and require("android")
-local log = android and android.LOGI or print
+local log = ffi.log or android and android.LOGI or print
```
Frenzie, line 22: This reads a bit suspicious. Shouldn't it be something more like `android and android.LOGI or ffi.log or print`?

benoit-pierre: This is only used to suppress the traces (see `ffi-cdecl/ffi-cdecl.lua`).

Frenzie: So it's only suppressed in the very specific scenario of treesitter? Please add a comment to that effect or maybe name it something like ignoreLog instead which would be self-explanatory. Because on Android I'm extremely fond of seeing them, especially in user-provided logs. :-)

Frenzie (approved):

PR #2361, opened 2026-05-17
#7992 Isolate PDF SMask

kerivin:
The implementation of [#7992](https://github.com/koreader/koreader/issues/7992)
Suggested order of merges: this PR -> bump base in koreader -> [koreader PR](https://github.com/koreader/koreader/pull/15401)

Image

Image

Image

Image

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2361)


mupdf_utils.c
```diff
@@ -0,0 +1,26 @@
...
+    }
+}
+
+fz_device *fz_new_isolated_smask_device(fz_context* ctx, fz_device* dev)
+{
+    isolated_smask_device* smask_dev = fz_new_derived_device(ctx, isolated_smask_device);
+    smask_dev->default_device = dev;
+    smask_dev->super.fill_image = smask_fill_image;
+    return (fz_device*)smask_dev;
+}
```
Frenzie:
```suggestion
}

```

mupdf_utils.c
```diff
@@ -0,0 +1,26 @@
...
+    fz_device* default_device;
+} isolated_smask_device;
+
+static void smask_fill_image(fz_context* ctx, fz_device* dev, fz_image* img, fz_matrix ctm, float alpha, fz_color_params color_params)
+{
+    isolated_smask_device* smask_dev = (isolated_smask_device*)dev;
+    if (img->mask) {
+        float black[1] = { 0.f };
+        fz_fill_image_mask(ctx, smask_dev->default_device, img->mask, ctm,
+                           fz_device_gray(ctx), black, alpha, color_params);
```
Frenzie: We don't live in 1980, so there's no need to split things into two lines making life worse. ;-)

Frenzie (commented):
I'm not really a big fan of the whole separate utils file. That feels a bit like an LLM that read too much Java and therefore thinks every offhand thought should be split out into a separate file.

mupdf_utils.h
```diff
@@ -0,0 +1,15 @@
...
+typedef struct fz_page fz_page;
+
+/**
+	Create a device to draw an isolated soft mask on a pixmap.
+
+	dev: Default draw device
+*/
+fz_device* fz_new_isolated_smask_device(fz_context* ctx, fz_device* dev);
+
+#endif
```
Frenzie:
```suggestion
#endif

```

kerivin:
> I'm not really a big fan of the whole separate utils file. That feels a bit like an LLM that read too much Java and therefore thinks every offhand thought should be split out into a separate file.

There is a few C files with MuPDF stuff, but they all are about wrapping... I thought it would be not so obvious to put the non-wrapper to the wrappers

mupdf_utils.c
kerivin: Fixed it

Frenzie:
To my mind it's still wrapping some fz_ functions. Just because it's not the most minimal wrap possible doesn't invalidate that. See #2322 for my take on something similar. It's still "wrapping" fz_new_story and fz_story_document, just while also massaging the output into the desired format. I'd be more inclined to rename the file if the word wrap is an issue.

The difference with cre.c and dvju.c is that those use the Lua C API, so they're not doing any "wrapping". But API aside, they're otherwise very similar in concept.

kerivin:
Yeah ok, I'll move it to `wrap-mupdf.c`

ffi/mupdf.lua
```diff
@@ -670,16 +670,29 @@ function page_mt.__index:getPageLinks()
...
+    end
+    if not ok then
+        ok = W.mupdf_run_page(page.ctx, page.page, dev, ctm, nil)
+    end
+
     M.fz_close_device(page.ctx, dev)
     M.fz_drop_device(page.ctx, dev)
-    if ok == nil then merror(page.ctx, "could not run page") end
+
+    if not ok then merror(page.ctx, "could not run page") end
```
Frenzie: I'd rather do without this change.

wrap-mupdf.c
```diff
@@ -203,6 +203,30 @@ fz_rect *mupdf_fz_bound_page(fz_context *ctx, fz_page *page, fz_rect *r) {
...
+    fz_device super;
+    fz_device* default_device;
+} isolated_smask_device;
+
+static void smask_fill_image(fz_context* ctx, fz_device* dev, fz_image* img, fz_matrix ctm, float alpha, fz_color_params color_params)
+{
+    if (img->mask) {
+        isolated_smask_device* smask_dev = (isolated_smask_device*)dev;
+        float black[1] = { 0.f };
+        fz_fill_image_mask(ctx, smask_dev->default_device, img->mask, ctm, fz_device_gray(ctx), black, alpha, color_params);
```
Frenzie, line 217: Is it correct that fz_device_gray doesn't need any cleanup?

kerivin: Seems like it.
Here's an example from the official documentation:

```c
float cmyk[100][4] = { {1,0,0,0}, ...
float rgb[100][3];
fz_color_converter cc;
int i;

fz_find_color_converter(ctx,
	&cc,
	fz_device_cmyk(ctx),
	fz_device_rgb(ctx),
	NULL,
	fz_default_color_params(ctx)
);
for (i = 0; i < 100; ++i)
	cc.convert(ctx, &cc, cmyk[i], rgb[i]);
fz_drop_color_converter(ctx, &cc);
```

benoit-pierre:
```c
/**
	Retrieve global default colorspaces.

	These return borrowed references that should not be dropped,
	unless they are kept first.
*/
fz_colorspace *fz_device_gray(fz_context *ctx);
fz_colorspace *fz_device_rgb(fz_context *ctx);
fz_colorspace *fz_device_bgr(fz_context *ctx);
fz_colorspace *fz_device_cmyk(fz_context *ctx);
fz_colorspace *fz_device_lab(fz_context *ctx);
```

ffi/mupdf.lua
Frenzie: I'll just commit this so I can merge the PR.
```suggestion
    if ok == nil then merror(page.ctx, "could not run page") end
```

wrap-mupdf.c
```diff
@@ -203,6 +203,30 @@ fz_rect *mupdf_fz_bound_page(fz_context *ctx, fz_page *page, fz_rect *r) {
     return r;
 }
 
+
```
Frenzie:
```suggestion
```

ffi/mupdf.lua
kerivin: @Frenzie keep in mind that I initialize ok as `false`, not `nil`. Here it should work fine though

PR #2364, opened 2026-05-19
MuPDF: check if page has smask before applying it

kerivin:
@Frenzie This is the [diff](https://github.com/koreader/koreader/pull/15401#issuecomment-4485602076) you sent

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2364)


Frenzie (approved):
Pinging @benoit-pierre for a second pair of eyes.

ffi/mupdf_h.lua
```diff
@@ -249,6 +249,7 @@ fz_colorspace *fz_device_bgr(fz_context *);
 fz_device *mupdf_new_draw_device(fz_context *, const fz_matrix *, fz_pixmap *);
 fz_device *mupdf_new_bbox_device(fz_context *, fz_rect *);
 fz_device *mupdf_new_isolated_smask_device(fz_context *, fz_device *);
+int mupdf_page_has_smask(fz_context *, fz_page *);
```
benoit-pierre, line 252: Why not return a `bool`?

wrap-mupdf.c
```diff
@@ -203,6 +203,29 @@ fz_rect *mupdf_fz_bound_page(fz_context *ctx, fz_page *page, fz_rect *r) {
...
+        return 0;
+    }
+    const int n = pdf_dict_len(ctx, xobj);
+    for (int i = 0; i < n; i++) {
+        pdf_obj* val = pdf_dict_get_val(ctx, xobj, i);
+        if (pdf_dict_get(ctx, val, PDF_NAME(SMask))) {
+            return 1;
+        }
+    }
+    return 0;
```
benoit-pierre: Shouldn't this be book-ended between:
```c
fz_try(ctx) {
[…]
fz_catch(ctx) {
return 0; // or false
}

File: ffi/mupdf_h.lua (hunk 1 above)
Comment (kerivin) [reply ffi/mupdf_h.lua:252]: It's not a part of the regular C, and all the boolean functions in this file return `int`:
```
int fz_is_document_reflowable(fz_context *, fz_document *);
int fz_needs_password(fz_context *, fz_document *);
```
I guess it's better to follow the existing pattern

Comment (Frenzie) [reply ffi/mupdf_h.lua:252]: Correct, done that way because I'd find it more confusing to do things differently than MuPDF.

File: wrap-mupdf.c (hunk 2 above)
Comment (Frenzie) [reply wrap-mupdf.c]: Yes, quite true. The whole block should be.

Comment (kerivin) [reply wrap-mupdf.c]: `MUPDF_WRAP` does exactly that (adds try-catch), so maybe it does make sense if some of the lines can throw....

Comment (kerivin) [reply wrap-mupdf.c]: Yeah ok I changed it from `DLL_PUBLIC` (no try-catch) to `MUPDF_WRAP` (with try-catch)

File: wrap-mupdf.c [hunk 3]
```diff
@@ -203,6 +203,29 @@ fz_rect *mupdf_fz_bound_page(fz_context *ctx, fz_page *page, fz_rect *r) {
     return r;
 }
 
+int fz_page_has_smask(fz_context* ctx, fz_page* p)
```
Frenzie: mupfd_ is our stuff, fz_ is from MuPDF

kerivin: I mean the public function is `mupdf_page_has_smask` anyway because of the wrapper. I thought it's ok to pretend that these functions (including `fz_new_isolated_smask_device`) are from the library x)

Frenzie: Those are just passthrough. It's a type of pretend but it's transparent, not misleading.

kerivin: It's either `mupdf_fz` or no prefix. I would rather prefer no prefix but since some functions already use `mupdf_fz_` here I use this as well

Frenzie: We could also just write it like this:

```c
int mupdf_page_has_smask(fz_context* ctx, fz_page* p)
{
    int ret = 0;

    fz_try(ctx) {
        /* Other document types (EPUB, XPS, etc.) don't use smasks. */
        pdf_page* page = pdf_page_from_fz_page(ctx, p);
        if (!page) {
            break;
        }

        pdf_obj* resources = pdf_page_resources(ctx, page);
        pdf_obj* xobj = pdf_dict_get(ctx, resources, PDF_NAME(XObject));
        if (!xobj) {
            break;
        }

        const int n = pdf_dict_len(ctx, xobj);
        for (int i = 0; i < n; i++) {
            pdf_obj* val = pdf_dict_get_val(ctx, xobj, i);
            if (pdf_dict_get(ctx, val, PDF_NAME(SMask))) {
                ret = 1;
                break;
            }
        }
    }
    fz_catch(ctx) {
        ret = 0;
    }

    return ret;
}
```

Which do you prefer @benoit-pierre?

kerivin: > ```c
> mupdf_page_has_smask
> ```

Nah the wrapped functions use this prefix already. Their names are fine and I don't really want to change them... Changing internal implementations' names seems more reasonable but it's not that big of a problem anyway, so I'll rename it to whatever you like

kerivin: And having the function wrapped the same as all other functions is probably better than moving try-catch into it just for the sake of a nice internal name...

Frenzie: But the wrapped functions are actual wrapped functions. We have a clear delineation between real fz_ which is internal to MuPDF and our stuff. There shouldn't be fz_ in the name here.

kerivin: Dayumn ok.......

Frenzie: The mupdf_fz_ prefix is all just a wrapper around fz_ stuff but it's not custom, is what I'm trying to say.

And we depend on conventions like these to look at the headers.

kerivin: But no prefix suggests that it's a custom function

kerivin: Do I need to add try-catch to new_isolated_smask_device as well? Or is it fine wrapped?

Frenzie: Take a step back please. Why do you think it needs to say fz_ in it?

wrap-mupdf.h
```diff
@@ -115,8 +115,11 @@ MUPDF_WRAP(mupdf_new_draw_device, fz_device*, NULL,
     ret = fz_new_draw_device(ctx, transform ? *transform : fz_identity, dest),
     const fz_matrix *transform, fz_pixmap *dest)
 MUPDF_WRAP(mupdf_new_isolated_smask_device, fz_device*, NULL,
-    ret = fz_new_isolated_smask_device(ctx, dev),
+    ret = mupdf_fz_new_isolated_smask_device(ctx, dev),
```
Frenzie: Okay, I think I understand what you were trying to say. Just do it like this (if I understood correctly).


```suggestion
    ret = new_isolated_smask_device(ctx, dev),
```

wrap-mupdf.h
```diff
@@ -115,8 +115,11 @@ MUPDF_WRAP(mupdf_new_draw_device, fz_device*, NULL,
...
     fz_device *dev)
+MUPDF_WRAP(mupdf_page_has_smask, int, 0,
+    ret = mupdf_fz_page_has_smask(ctx, page),
```
Frenzie:
```suggestion
    ret = page_has_smask(ctx, page),
```

wrap-mupdf.c
kerivin: Those are just implementation details. Does it really matter whether a function comes from the original MuPDF library or added manually? It's an extension of the library, and I personally wouldn't want to remember that there are 'true' MuPDF functions and 'custom' ones, instead I would rather type `fz_` and see the entire autocompletion list...
But I see your point why you would want to separate them

Frenzie: It matters very much. I do not wish to be misled into thinking a function exists that doesn't. That's what they call technical debt, although I prefer to just call it bad for maintenance.

wrap-mupdf.h
kerivin: Yeah that's what I meant x)

kerivin: Done

wrap-mupdf.h
kerivin: Done

wrap-mupdf.c
Frenzie: This is the kind of thing that a decade ago I'd have ignored, but I've since learned that to do so is a guarantee that it'll come back to hound you in the future. No matter how remote or niche the functionality, it will be copied. And that's just talking about real human people; I don't even want to know what the current ilk of chatbots will dreck up and smear together. :-)

PR #2371, opened 2026-05-21
luacheck: align with koreader's configuration

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2371)


ffi-cdecl/ffi-cdecl.lua
Frenzie, line 52: Wouldn't  it make  more sense to also add the error?

benoit-pierre: There's no error, just the return code with Lua 5.1.

PR #2383, opened 2026-05-27
feat: add AVIF support

winstonma:
Implement AVIF image decoding support using libavif via FFI.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2383)


benoit-pierre:
Here's an idea: how about you at least check your AI slop builds locally before wasting our time?

PR #2393, opened 2026-06-08
android: fix getRotationMode() to return canonical DEVICE_ROTATED_* values

chofuhoyu:
Add a mapping from Android's SCREEN_ORIENTATION_* constants to KOReader's canonical DEVICE_ROTATED_* values. Required by [koreader/koreader#15507](https://github.com/koreader/koreader/pull/15507).

Changes:
- Add ANDROID_ORIENTATION_TO_ROTATION mapping table
- Fix getRotationMode() to return canonical DEVICE_ROTATED_* values
- Fix setRotationMode() to sync cur_rotation_mode

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2393)

### DIFF
--- a/ffi/framebuffer_android.lua
+++ b/ffi/framebuffer_android.lua
@@ -2,6 +2,7 @@ local ffi = require("ffi")
 local android = require("android")
 local BB = require("ffi/blitbuffer")
 local C = ffi.C
+local logger = require("logger")
 
 --[[ configuration for devices with an electric paper display controller ]]--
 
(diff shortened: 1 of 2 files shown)


ffi/framebuffer_android.lua
```diff
@@ -69,7 +81,15 @@ end
...
-        return android.orientation.get()
+        local ao = android.orientation.get()
+        local mapped = ANDROID_ORIENTATION_TO_ROTATION[ao]
+        if mapped ~= nil then
+            return mapped
+        else
+            -- For UNSPECIFIED(-1), USER(2), BEHIND(3), SENSOR(4), NOSENSOR(5)
+            -- fall back to cached value
+            return self.cur_rotation_mode
+        end
```
Frenzie: We generally just write something like:

```suggestion
        -- For UNSPECIFIED(-1), USER(2), BEHIND(3), SENSOR(4), NOSENSOR(5)
        -- fall back to cached value
        return mapped or self.cur_rotation_mode
```

Unless it'd turn into a forest of `or … or … or`.

chofuhoyu:
However, I seem to have come across a bug related to this `ANDROID_ORIENTATION_TO_ROTATION` mapping. I found a critical bug that will cause mismapping of the touching layer and the screen content layer in certain circumstances. In a word, my clicking will be shifted somehow, and I can't interact with what I'm clicking.

I'm trying to fix that.

PR #2397, opened 2026-06-13
utils: improve `inkview_h` generator script

benoit-pierre:
- improve output (thanks to `ffi-cdecl -M …`): more logical grouping, and ensure declarations are in the correct order
- make it more general so it be be reused

Depend on new ffi-cdecl from #2394.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2397)


Frenzie (approved):

poire-z:
(Can I bump crengine into frontend later today, and bring all the base stuff commited these last 2 days without issues ?)

benoit-pierre:
Yes.

PR #2398, opened 2026-06-13
ffi/posix_h: updated and corrected version

benoit-pierre:
Use the new 2.0.0 version of ffi-cdecl to generate POSIX cdecls for each platform variant, and our updated helper to merge those individual files.

This fix a number of issues, particularly with macOS (including a number of important constants: `EAGAIN`, `O_CLOEXEC`, `MAP_ANONYMOUS`, etc…).

Some unused declarations have been removed, some new declarations have been added:
- those used by the terminal plugin
- those necessary for turbo, so its code can be patched to rely on `ffi/posix_h`

NOTE:
- `ffi/pthread_h` has been removed: pthread declarations are now included in `ffi/posix_h`
- the frontend generic device code for retrieving network information will need updating: Linux only IOCTL values are not available to macOS anymore
- Linux x86 support has been dropped (Android x86 support is still provided)
- *BSD support has been dropped too

The new ffi targets depend on the ffi-cdecl from #2394 being available and the updated merge helper from #2397.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2398)

PR #2399, opened 2026-06-13
turbo: fix and trim the fat

benoit-pierre:
- add a very basic test to ensure the main module can at least be imported
- use our `ffi/posix_h` (instead of its possibly conflicting cdef)
- ensure its loading **our** SSL library
- drop unused APIs / code

Depend on #2398.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2399)


Frenzie (approved):

benoit-pierre:
Turn out the wrong SSL library was loaded by `turbo/hash.lua` too.

PR #2401, opened 2026-06-13
feat(ffi): invert screenshots in night mode

Euphoriyy:
Inverts screenshots that are taken in night mode to match the expected behavior.

 _Originally discussed by @Commodore64user in [#13041](https://github.com/koreader/koreader/issues/13041#issuecomment-2585491453)_

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2401)

### DIFF
--- a/ffi/framebuffer.lua
+++ b/ffi/framebuffer.lua
@@ -569,7 +569,11 @@ function fb:shot(filename)
     if self.device:hasBGRFrameBuffer() then
         bgr = true
     end
-    self.bb:writePNG(filename, bgr)
+    if self.night_mode then
+        self.bb:copy():invert():writePNG(filename, bgr)
+    else
+        self.bb:writePNG(filename, bgr)
+    end
 end
 
 -- Clear the screen to white


Frenzie:
Then wouldn't they be inverted again if you use them as a screensaver? That seems like rather unexpected behavior.

Euphoriyy:
When set as a screensaver, the screenshot remains dark (white on black) in both day and night mode because "original in night mode" is set. It gets inverted again (black on white) when opened with MuPDF in night mode but not the image viewer because of the same reason.

I guess it depends on the use case of the screenshot, where it might make more sense to keep the screenshot uninverted for reading later, even in night mode. But when screenshots are exported, it makes more sense to have them inverted. It does seem unintuitive though that screenshots can be viewed uninverted in night mode.

Frenzie (approved):
Anyway, fine by me. 👍

Euphoriyy:
~I forgot to handle BGR framebuffers...~ I'll just ignore that for now since I can't test it reliably.

hius07:
Can it be just in framebuffer.lua?
```lua
if self.night_mode then
    self.bb:copy():invert():writePNG(filename, bgr)
else
    self.bb:writePNG(filename, bgr)
end
```
-----

Before


-----

After

Euphoriyy:
> Can it be just in framebuffer.lua?

Yeah that would work. I also originally thought of doing that, but I thought it might be better to prevent redundant copies/blits if possible. Though that might be better in hindsight due to reducing complexity, especially for menial tasks like screenshots.

Euphoriyy:
I updated the PR to use your version.

hius07 (approved):
Thanks.

PR #2413, opened 2026-06-16
Background cleanup over dewatermark

kerivin:
I was testing scanned PDFs and discovered that a decision I've made earlier haunts me now x)
It was impossible to enable both Dewatermark and Background Cleanup at the same time since the kopt rendering used the default 'false' value for isolate_smask flag. Dewatermark effectively turned off Background Cleanup.

This PR also makes possible to have both Dewatermark and Background Cleanup enabled by default, which means every scanned PDF opened would be rendered nicely without extra settings adjustments. Especially considering extremely slow loading (or even freezing) of some PDFs which makes me want to have Background Cleanup always enabled no matter what.

Order of merges: this PR -> bump base in koreader -> [koreader PR](https://github.com/koreader/koreader/pull/15555)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2413)

### DIFF
--- a/djvu.c
+++ b/djvu.c
@@ -532,8 +532,7 @@ static int closePage(lua_State *L) {
 static int getPagePix(lua_State *L) {
 	DjvuPage *page = (DjvuPage*) luaL_checkudata(L, 1, "djvupage");
 	KOPTContext *kctx = (KOPTContext*) lua_topointer(L, 2);
-	void* options = (void*) lua_touserdata(L, 3);
-	ddjvu_render_mode_t mode = (int) luaL_checkint(L, 4);
+	ddjvu_render_mode_t mode = (int) luaL_checkint(L, 3);
 	ddjvu_rect_t prect;
 	ddjvu_rect_t rrect;
 	int px, py, pw, ph, rx, ry, rw, rh;
--- a/djvu.c
+++ b/djvu.c
@@ -532,7 +532,8 @@ static int closePage(lua_State *L) {
 static int getPagePix(lua_State *L) {
 	DjvuPage *page = (DjvuPage*) luaL_checkudata(L, 1, "djvupage");
 	KOPTContext *kctx = (KOPTContext*) lua_topointer(L, 2);
-	ddjvu_render_mode_t mode = (int) luaL_checkint(L, 3);
+	void* options = (void*) lua_touserdata(L, 3);
+	ddjvu_render_mode_t mode = (int) luaL_checkint(L, 4);
 	ddjvu_rect_t prect;
 	ddjvu_rect_t rrect;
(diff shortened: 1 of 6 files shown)


ffi/koptcontext_h.lua
benoit-pierre, line 1: This file is automatically generated, you can't modify the `KOPTContext` structure like this, it has to match the actually definition used when compiling our k2pfopt library.

kerivin: Yeah when I realized this I also thought I would need to change the k2pdfopt as well, and I absolutely don't want to do this. I'll think more on how to pass the parameters to 'getPagePix' without contamination of KOPTContext.
The PR is a draft for now

benoit-pierre: But did you actually test your code? Because  the k2pdfopt library still think it's receiving a struct with the old layout, so everything after your new member is shifted…

kerivin: I did, and it worked, at least looked like it x) But that might be just a lucky coincidence

djvu.c
```diff
@@ -532,7 +532,8 @@ static int closePage(lua_State *L) {
 static int getPagePix(lua_State *L) {
 	DjvuPage *page = (DjvuPage*) luaL_checkudata(L, 1, "djvupage");
 	KOPTContext *kctx = (KOPTContext*) lua_topointer(L, 2);
-	ddjvu_render_mode_t mode = (int) luaL_checkint(L, 3);
+	void* options = (void*) lua_touserdata(L, 3);
```
kerivin: Still better than modifying k2pdfopt x)
DjVu doesn't use the options MuPDF needs, so I simply ignore them here. I've changed the order so that  MuPDF's `getPagePix` could still skip `render_mode` as it does now

kerivin:
So I've tested this with both djvu (to see that nothing is broken and the render_mode is correct) and pdf. Seems to work fine.







(Before when both Dewatermark and Background Cleanup were enabled it was just Dewatermark without Background Cleanup)

kerivin:
@benoit-pierre @Frenzie Sorry for an unnecessary ping, but is there any chance this PR with the Background Cleanup fix could make it into the release? While I can adjust it quickly (and rebase against master in koreader PR) so that this new feature would actually work properly in the release

Frenzie:
If @benoit-pierre is okay with it.

benoit-pierre:
> no bool this time, keep int 

Or:
```c
#include <stdbool.h>
```

kerivin:
Yep that worked, thanks

kerivin:
I've just tested it, the result is [the same](https://github.com/koreader/koreader-base/pull/2413#issuecomment-4732532757)

PR #2414, opened 2026-06-16
ffi/netinfo: new module to retrieve network information

benoit-pierre:
Avoid using unsupported ioctls on macOS, plus some additional tweak to support more configurations:

- uses `getifaddrs()` results to retrieve the MAC address (from `AF_LINK` on macOS, `AF_PACKET` on Linux), instead of the Linux specific `SIOCGIFHWADDR` ioctl
- for checking the interface is wireless on Linux: do a filesystem check for `/sys/class/net/$IFACE/wireless/`, instead of using the `SIOCGIWNAME` ioctl, which is not always supported (it depends on the driver: no dice on my Kindle PW2)
- on macOS: checking if the interface is wireless is still unsupported, as is retrieving the SSID (which is not possible anyway on more recent macOS versions without location permission)

Depend on #2398.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2414)

### DIFF
--- a/ffi-cdecl/posix_cdecl.c
+++ b/ffi-cdecl/posix_cdecl.c
@@ -43,11 +43,15 @@
 #include <net/if.h>
 #if defined(__APPLE__)
 # include <net/if_dl.h>
+# include <net/if_types.h>
 #endif
 #include <net/route.h>
 #include <netdb.h>
 #include <netinet/ip.h>
 #include <netinet/ip_icmp.h>
+#if defined(__linux__)
+# include <netpacket/packet.h>
+#endif
 #include <poll.h>
 #include <pthread.h>
 #if defined(__linux__)
@@ -285,10 +289,17 @@ cdecl_type(__fsfilcnt_t);
 
 cdecl_struct(statvfs);
 
+#if defined(__APPLE__)
+cdecl_const(IFT_ETHER);
+#endif
+
 cdecl_const(AF_INET);
 cdecl_const(AF_INET6);
 #if defined(__APPLE__)
 cdecl_const(AF_LINK);
+#endif
+#if defined(__linux__)
+cdecl_const(AF_PACKET);
 #endif
 cdecl_const(AF_UNIX);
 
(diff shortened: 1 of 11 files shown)


Frenzie:
I still vehemently and fundamentally disagree with your description of the macOS situation. The generic device implementation is meant to be overridden by a more specific one where appropriate so there's simply no such thing as using Linux features on Mac but merely a lack of a Mac-specific implementation. The more obvious path is to disable it altogether on Mac rather than trying to make something of no import on the platform work.

Frenzie (approved):
Anyway, lgtm

backcountrymountains:
Tested on a **Nook Glowlight 4 Plus** (model `bnrv1300`, Android 8.1, Allwinner B300 SoC, RTL8723DS WiFi) using ADB probes. This testing was performed by Claude (Anthropic's AI assistant) using Claude Code, posting under my account for attribution transparency.

**What works:**
- `getifaddrs()` AF_INET — confirmed IP address retrieval (`192.168.1.224`) ✅
- `getifaddrs()` AF_PACKET — confirmed MAC address retrieval (`58:b0:d4:db:aa:fd`) ✅
- `/sys/class/net/wlan0/wireless/` — path exists, wireless interface detection works ✅
- `/proc/net/wireless` — does **not** exist, which confirms WEXT isn't available on this driver and your sysfs-based approach is the right one ✅

**SSID via nl80211:**
The RTL8723DS driver does support nl80211 — `wpa_supplicant` uses it to manage connections. However, Android 8.1's SELinux policy blocks `NETLINK_GENERIC` socket access from the app process (`u:r:untrusted_app`); even the ADB shell (`u:r:shell:s0`) gets `Permission denied` on some Netlink operations. The `pcall` around `_iface_ssid` will trigger silently on Android and `ssid` will be `nil` — which is the correct graceful degradation.

So on Android: IP and MAC with no new permissions, SSID always nil. Works exactly as designed.

NiLuJe:
> * /sys/class/net/$IFACE/wireless/

Do we actually have that everywhere?

This *seems* like something that should be there, but I've been bitten by weird sysfs crap in old kernels in the past... ;o).

benoit-pierre:
> > * /sys/class/net/$IFACE/wireless/
> 
> Do we actually have that everywhere?
> 
> This _seems_ like something that should be there, but I've been bitten by weird sysfs crap in old kernels in the past... ;o).

It was added to sysfs in 2.6.0, deprecated at some point (with an opt-in config option, replaced by an `phy82011` entry), completely removed and then re-added in 2012 because some Fedora scripts rely on it (so it should be unconditionally present on recent kernel, but it's empty).

PR #2417, opened 2026-06-28
feat(reMarkable): Oxide/libblight support

NoaHimesaka1873:
Another framebuffer/input backend for reMarkable, hooray! Its ecosystem is truly like https://xkcd.com/927/ lol.
It's used for Oxide, which gives it more bare-metal like experience than QTFB, no xochitl required! (Though it does depend on libqsgepaper, it's less clutter!)

Also it's my birthday today, more hooray!
DISCLAIMER: LLM-assisted, Gemini 3.5 Flash used.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2417)


ffi/framebuffer_blight.lua
```diff
@@ -0,0 +1,164 @@
...
+    if blight.is_rmpp then
+        width = 1620
+        height = 2160
+    elseif blight.is_rmppm then
+        width = 954
+        height = 1696
+    elseif blight.is_rmppure then
+        width = 1776
+        height = 2400
+    end
```
Frenzie, line 32: These can't be taken from it?

NoaHimesaka1873:
(also you can see some systemd/D-Bus bits as libblight uses D-Bus for communication, how cursed! and yes, it means that rM devices have full systemd!)

Frenzie:
Happy b-day! :partying_face:

ffi/framebuffer_blight.lua
NoaHimesaka1873: sadly no, libblight D-Bus interface doesn't provide one.

Frenzie (approved):

PR #2428, opened 2026-07-03
LuaSocket: fix CLOEXEC flag setting in enforce-cloexec.patch

zikasak:
O_CLOEXEC can be set only during FD creation and completely ignored in fcntl. As a result, the fd is not closed on exec in child processes. 

It leads to issues with, for example, http inspector when it restarts (and other process inherited a copy of the FD already). 

Solved by using fd_cloexec flag instead which is doing the same thing but can be applied after creation

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2428)


Frenzie (approved):
Reference https://github.com/koreader/koreader-base/pull/1827

Pinging @NiLuJe

NiLuJe (approved):
Oops, nice catch ;).

PR #2433, opened 2026-07-07
fix `TARGET=linux` ARM build

benoit-pierre:
Cf. https://github.com/koreader/koreader/pull/15617#issuecomment-4901561463.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2433)


Frenzie:
Thanks! Just "starts with" looks a tad suspicious in some ways but it should be more than good enough for now.

Issue #22, opened 2013-06-16
Building djvulibre For Kobo


rheaplex:
I'm trying to build koreader-base cc7f266 on 64-bit Debian Sid using the CodeSourcery toolchain from the KoboReader GitHub repo.

djvulibre fails to build with the following error when caling:

TARGET_DEVICE=KOBO make all customupdate

This is the error:

make[1]: Entering directory `/home/rob/src/koreader/koreader-base/djvulibre/build/tools'
arm-none-linux-gnueabi-g++ -DHAVE_CONFIG_H -I.. -I../.. -I../../libdjvu -I../../tools -DNDEBUG -Wall -O3 -Wno-non-virtual-dtor   -pthread -DTHREADMODEL=POSIXTHREADS    -ffast-math -pipe -fomit-frame-pointer -march=armv6j -mtune=arm1136jf-s -mfpu=vfp -mfloat-abi=softfp -marm -fno-stack-protector -U_FORTIFY_SOURCE -D_GNU_SOURCE -fno-finite-math-only -fno-use-cxa-atexit  -c ../../tools/bzz.cpp
/bin/sh ../libtool --mode=link \
      arm-none-linux-gnueabi-g++ -o bzz -DHAVE_CONFIG_H -I.. -I../.. -I../../libdjvu -I../../tools -DNDEBUG -Wall -O3 -Wno-non-virtual-dtor   -pthread -DTHREADMODEL=POSIXTHREADS    -ffast-math -pipe -fomit-frame-pointer -march=armv6j -mtune=arm1136jf-s -mfpu=vfp -mfloat-abi=softfp -marm -fno-stack-protector -U_FORTIFY_SOURCE -D_GNU_SOURCE -fno-finite-math-only -fno-use-cxa-atexit  bzz.o ../libdjvu/libdjvulibre.la -Wl,-O1 -Wl,--as-needed -lm 
[...]

PR #2441, opened 2026-07-08
sdcv: cache the .syn offset table in a .syn.oft file

asxelot:
Backport of https://github.com/Dushistov/sdcv/pull/129 as a fifth sdcv patch, as discussed in koreader/koreader#15645 (cc @Frenzie @xxyzz @Commodore64user).

## Problem

sdcv's `SynFile::load` rebuilds its entry pointer list by walking every 0-terminated entry of the mmap'd `.syn` file **on each process start** - there is no `.oft`-style cache for the `.syn`, only for the `.idx`. With a large inflection index (46 MB `.syn`, 1,725,572 entries) every lookup tap faults in the whole 46 MB: 5-15 s on e-ink devices (koreader/koreader#5437).

## Fix

Reuses sdcv's existing `.oft` cache mechanism for the `.syn` file: `OffsetIndex`'s cache load/save extracted into shared helpers, `SynFile` stores `guint32` offsets into the mmap instead of `gchar *` pointers (halves per-entry memory) and caches them in `<name>.syn.oft` next to the dictionary or in the user cache dir. Caches with a wrong size are rejected and rebuilt (this also stops a truncated `.idx.oft` from being `memcpy`'d past its mapping).

## Verification

- Fresh v0.5.5 tarball + the existing 4 patches + this one applied with `patch -p1 --forward` (same as `apply_patches`): all apply cleanly, builds, and sdcv's test suite passes 14/14 (includes a new `t_synfile_cache` covering cold/warm/garbage/truncated/stale cache paths).
- Output byte-identical to unpatched sdcv on 40 sampled inflected forms from two real dictionaries (cold and warm cache).
- Measured with the 46 MB `.syn` (macOS desktop, warm page cache): 55.6 ms -> 14.9 ms per invocation, peak RSS 64.3 MB -> 19.7 MB. On device the win is much larger since the per-launch scan dominates there.

If/when the upstream PR is merged and the pinned sdcv is bumped, this patch can simply be dropped.

---

This pull request was written by a human with AI assistance (implementation developed and reviewed with Claude).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2441)


Commodore64user:
same disclaimer applies: do no despair ;)

I am quite excited about it btw, so thank you!

asxelot:
@Commodore64user  you can already use a plugin https://github.com/asxelot/fastdict.koplugin

Commodore64user:
yes I know, I was trying to indirectly signal to @benoit-pierre and @Frenzie that this is a highly desirable gem... even if we can't properly review it ;)

Frenzie:
I'll wait a few days nonetheless. ;-)

Commodore64user:
@Frenzie could you please recompile `sdcv` for kindle, I tried cross-compiling on macOS and nothing but problems, I don't have it in me to figure out what went wrong right now.  

for the lads and testing purposes of course... ;)

Edit: https://github.com/koreader/koreader/blob/035497b509c998c8d67bf27ba74e9aa9308993ef/doc/Building_targets.md?plain=1#L5

thirdparty/sdcv/sdcv-syn-oft-cache.patch
```diff
@@ -0,0 +1,248 @@
...
+ 
+ private:
+     MapFile synfile;
+-    std::vector<gchar *> synlist;
++    std::vector<guint32> wordoffset;
+ };
+ 
+ class Dict : public DictBase
+diff --git a/tests/t_synfile_cache b/tests/t_synfile_cache
+new file mode 100755
```
Frenzie: Can you remove from here onwards from the patch?

Frenzie:
I'll just merge it in a couple of days if it's not merged there first so it can make it into the July release.

Commodore64user:
it seems like the lord is testing my patience ;) you sure you don't want a beta tester before sending it live?, and what is @benoit-pierre up to with all those cross references?? 6 so far.

Frenzie:
It hasn't even been a full 24 hours yet, lol.

Commodore64user:
okay, this without a doubt speeds up (on average) searches , I haven't technically benchmarked them but, on a PW5 searches tend to briefly show the info message (which shows up after 0.5s) before the DictQuick widget pops up, so we can safely say it is taking roughly 0.5+ secs (0.5-0.6 ish) to respond, before this change it could have taken either a similar amount of time or 4-5 seconds (never found out why).

It is also mostly consistent, so the majority of searches fall within that short reaction time however, when waking up the device from deep sleep, the first search takes a significantly longer time (5-7 secs). There are also times in which it takes just a little longer, I'd say about 1 sec, which doesn't seem bad but I can't seem to figure out why.

I am also curious to hear what the experience is like on non-kindle, before and after.

@asxelot are you using the Lua engine or this? Do you see any difference between the two on real hardware?

asxelot:
@Commodore64user I'm trying to debug it, but it takes so much time. I see it really takes longer after deep sleep.

Commodore64user:
Have you had any luck figuring it out? I know it’s annoying, you have to wait an hour between attempts.

Frenzie:
it sounds like whatever it is is simply an internal Kindle matter, doesn't it? (File system still warming up, essentially.)

Commodore64user:
I don't really know how it performs elsewhere, but that delay is significantly longer than without the patch at all, furthermore I don't really know the black magic that goes on during deep sleep but maybe one would expect the system to be fully back online after a while, yet every single first query (no matter how long after waking) suffers from it, I have even seen it on a second one as well.

asxelot:
@Commodore64user I can't reproduce it properly — the delay on my Kindle is short — so I wrote a probe so we can see where those 5–7 s actually go instead of guessing. [2-dictprobe.lua.zip](https://github.com/user-attachments/files/30598815/2-dictprobe.lua.zip) → copy to /mnt/us/koreader/patches/2-dictprobe.lua, restart KOReader. It logs a breakdown per lookup to /mnt/us/koreader/dictprobe.log: how much time is sdcv vs. spawning it vs. KOReader noticing it finished, how much sdcv actually computed vs. waited, how many MB came off storage, whether the .syn.oft got rebuilt, and how much of each dictionary file was still in the page cache before the lookup.

 And you don't have to wait an hour between attempts any more: look up the word dropcache (Search → Dictionary lookup). It doesn't search — it evicts the dictionaries from the page cache, so the next lookup is cold. That gets us the cold-cache half of the question in seconds; the hibernation wait is then only needed to check whether deep sleep does something beyond dropping the cache.

 Ideally: a few lookups right after starting (warm baseline), a dropcache + lookup, then one long sleep past the hibernation mark, wake, and two lookups in a row. Then send the log.

Commodore64user:
Now we are cooking with gas, [dictprobe.log.txt](https://github.com/user-attachments/files/30631245/dictprobe.log.txt)

I have to say, I am a little surprised by these values because 
`[2026-08-01 19:12:58] LOOKUP #12  word="hoarsely"  total=3.128s   <<< FIRST LOOKUP AFTER RESUME` 
it definitely takes way longer than 3 seconds from 'finger released from screen' to 'widget appears on screen'...

asxelot:
That log settles most of it, thanks. You have ~20 dictionaries, ~448 MB of dictionary data (OED_2nd alone is a 205 MB `.dict.dz`), on a device with ~250 MB usable RAM — so the page cache can never hold them, and suspend drops what it does hold (`Cached` goes ~105 MB → ~30 MB across a wake). The slow lookups each read **17–21 MB off storage**; the warm ones read 1.5–4 MB, and the one repeat lookup that read **0 MB** was the fastest in the log. sdcv only spends 0.23–0.43 s of CPU in all of them, so it's waiting on eMMC, not computing. Your CPU is also downclocked right after a wake (600–833 MHz vs 1000 MHz).

Worth noting: all 28 lookups say `.oft cache files: unchanged (the cache was accepted)` — the `.syn.oft` cache is never rebuilt, so it's doing its job. Without it, every one of these lookups would also have to linearly scan ~30 MB of `.syn` across those dictionaries. And one of your 2.9 s lookups happened with no resume at all, just after `dropcache` — so it's plain cold cache rather than something deep sleep does specially.

You're also right that 3.1 s doesn't match what you feel, and that was my bug: the probe stopped when the lookup returned, but that only *builds* the window — the layout and the e-ink refreshes come after, and a lookup fires several (the "Searching…" message, dismissing it, then the dictionary window), each a few hundred ms on a PW5. v2 attached: it now logs the whole path from finger to last pixels on screen, with every repaint timed. It also stops printing the page-cache column on your kernel, where `mincore()` reports nothing useful (it was showing 0% for files that were demonstrably cached).

If you'd like the lookups themselves faster in the meantime: disabling the dictionaries you rarely use is by far the biggest lever — sdcv pays for every enabled one on every lookup.

[2-dictprobe.lua.zip](https://github.com/user-attachments/files/30631918/2-dictprobe.lua.zip)

Commodore64user:
15 dicts, but not all are active at all times, admittedly all the demanding ones are, 12/15, all the bilingual ones xx-en are only available through a `Translate` preset (but yeah, they are the smaller ones). Nevertheless this time I added some spiciness, starting at 
`[2026-08-02 20:41:07] LOOKUP #25  word="chairman"  total=4.065s   <<< FIRST LOOKUP AFTER RESUME`
the OED (200MB one) is no longer in use.

[dictprobe.log.txt](https://github.com/user-attachments/files/30638314/dictprobe.log.txt)

asxelot:
@Commodore64user Correction to my last comment: a good chunk of those numbers was my own probe. v2 walked the dictionary tree twice per lookup, which costs 1-2 s warm and 5-7 s on cold storage — median 2.26 s, worst 7.36 s of pure measurement overhead, sitting right between your tap and the window. So the `PERCEIVED` figures in that log aren't usable.

Two things it did measure cleanly:

- A full-screen (`flashui`) refresh on the PW5 is **0.41-0.43 s**; every other refresh is 1-17 ms. So the "Searching…" message costs ~8 ms — I was wrong to blame it. Only the final window refresh is expensive.
- Real perceived latency is therefore roughly lookup + 0.42 s: **~1.3 s warm, ~2.8-4.5 s cold**.

Also: disabling the 205 MB OED barely moved anything (warm ~0.93 → ~0.87 s, cold reads 16-28 MB → 15-16 MB). The cost is spread across the other dictionaries' `.idx`/`.syn`/`.oft` files, not concentrated in the biggest `.dict.dz`.

And your log confirms the page-cache column was meaningless on your kernel — the self-test read a file in full and `mincore()` still reported it 0% resident.

v3 attached: the whole report now runs 3 s after the lookup instead of inside it, so its overhead is ~12 ms instead of ~2.3 s, and it no longer loses the window's refresh among UIManager's idle ticks.

[2-dictprobe.lua.zip](https://github.com/user-attachments/files/30695160/2-dictprobe.lua.zip)

Commodore64user:
from `[2026-08-04 21:37:35] LOOKUP #17  word="asked"  total=2.267s` until lookup 25, three dictionaries are in use, NOAD, ODE (does not have a `syn` file) and OED (200MB)

from `[2026-08-05 00:13:04] LOOKUP #25  word="right"  total=5.024s   <<< FIRST LOOKUP AFTER RESUME` onwards, only two dictionaries are in use: NOAD and ODE (not the big 200MB one)

[dictprobe.log.txt](https://github.com/user-attachments/files/30724389/dictprobe.log.txt)

asxelot:
@Commodore64user This log is the useful one — the probe's own overhead is down to 11-33 ms, so the numbers are finally the device's and not mine. And it kills my main theory.

**The dictionary count is not the cause.** Lookups `#21` and `#22` ran with three dictionaries and took 5.6 s and 6.4 s. `#25` ran with *two* and took 5.0 s. Going 15 → 3 → 2 dictionaries only moved the warm baseline (~1.0-1.4 s → 0.64-0.79 s); the cold spike is untouched. So "disable dictionaries" was bad advice for this bug, and I'll stop repeating it.

**It is not the amount of data read either.** Compare three lookups:

| # | dicts | read | sdcv wall | effective |
|---|---|---|---|---|
| 17 | 3 | 7.1 MB | 2.00 s | 3.6 MB/s |
| 26 | 2 | 4.0 MB | 0.79 s | 5.1 MB/s |
| **25** | **2** | **6.9 MB** | **4.26 s** | **1.6 MB/s** |

`#25` read the same amount as `#17` and took twice as long; per major fault it was 56 ms versus 7-19 ms for the warm lookups. Whatever happens after a wake makes each read several times more expensive, rather than making sdcv read more. Same story for `#11` (17.8 MB, 35.7 ms/fault) and `#21`.

**And it does not clearly look like disk wait.** During `#25` the system reported only 5% iowait while 52% of the CPU time went somewhere. sdcv used 0.5 s of CPU out of 4.26 s, so it was not computing — but I cannot tell from this log whether it was blocked on storage or simply not getting a CPU. That is the whole remaining question, and my instrument could not answer it.

So v4 answers it directly. Two additions:

1. **It watches the sdcv process itself.** sdcv is now started so the shell can report its pid, and DictProbe samples `/proc/PID/` every ~120 ms while you wait: process state (`D` = blocked on storage, `R` = wants a CPU but isn't getting one, `S` = sleeping), the kernel function it is parked in (`wchan`), and `read_bytes`, which is the block-device traffic of that one process rather than the whole device. One line per lookup, e.g. `D=67% R=8% S=25%  parked in: blk_io_schedule(14)`. That distinguishes slow storage from CPU contention, which the aggregate numbers cannot.

2. **It measures how fast storage actually is at that moment** — a 1 MB sequential read plus 24 scattered 4 kB reads of a scratch file it has just evicted from the page cache. The scattered number is the one that matters, since a binary search through an `.idx`/`.syn` is exactly that pattern. It runs at startup, at +1/+3/+8/+20 s after each wake, and ~3 s after each lookup, never before one — so if the storage really is cold and slow right after a wake, you will see it as MB/s and ms directly, independent of sdcv.

Also fixed: lookup `#1` in your log reported every `.oft` file as both `CREATED` and `DELETED`. That was cosmetic — the probe compared `./data/dict/x.oft` against `data/dict/x.oft` and saw two different files. Nothing was created or deleted; the cache was accepted in all 29 lookups, as in every log so far.

Same routine as before if you have the patience: a few warm lookups, then a long sleep, then a word, then the same word again. `dropcache` still works for forcing a cold lookup without waiting.

[2-dictprobe.lua.zip](https://github.com/user-attachments/files/30739409/2-dictprobe.lua.zip)

Commodore64user:
> So "disable dictionaries" was bad advice for this bug, and I'll stop repeating it.

yeah, I have instinctively known this to be false, even prior to this patch, searches could take ~5sec even with only one dictionary active.

> Also fixed: lookup 1 in your log reported every .oft file as both CREATED and DELETED. That was cosmetic — the probe compared ./data/dict/x.oft against data/dict/x.oft and saw two different files. Nothing was created or deleted; the cache was accepted in all 29 lookups, as in every log so far.

I figured that was the case.

> Same routine as before if you have the patience

at this point we should do what Apple did with iOS, and start calling them after the month.. ;) here is version-08

[dictprobe.log.txt](https://github.com/user-attachments/files/30761454/dictprobe.log.txt)

asxelot:
@Commodore64user version-08 it is. This log finally localises the problem, so it was worth it.

**The slow lookups are a stall in the middle, not slow reading.** All three of them (`#13` 6.16 s, `#16` 6.02 s, `#20` 4.38 s) have the same shape:

```
0.0s:R/0.1MB  0.5s:R/1.5MB  1.0s:S/1.5MB  1.5s:S/1.5MB  2.0s:R/1.5MB
2.5s:S/1.5MB  3.0s:S/1.5MB  3.4s:S/1.5MB  3.9s:D/2.1MB  4.4s:D/6.9MB
4.9s:R/8.8MB  5.4s:D/11.9MB
```

sdcv reads ~1.5 MB in the first half second, then its read counter sits **frozen for 2.5-4 s** while the process is asleep (`S` in 52-63% of samples) using about 0.3 s of CPU — and then it does the remaining ~10 MB in the last 1.5 s at a perfectly normal rate. It is not in `D`, so this is not ordinary disk wait; it is not in `R`, so it is not losing a fight for the CPU. Something puts it to sleep in the middle of the lookup, and everything else follows from that.

**Your storage does slow down, and now there is a number for it.** The calibration read has no sdcv in it at all, and it swings between 86-111 MB/s and 10-15 MB/s, hitting **3.3 MB/s one second after a resume**, back to ~11 MB/s by +3 s. So @Frenzie was right that the filesystem is still warming up — though it stays at 10-15 MB/s long after the wake, so that is not only a wake-up transient.

One more thing worth noting: on the worst lookup KOReader itself took **266 major faults** against a baseline of ~110. Some of what you wait for on a cold lookup is KOReader paging its own code back in, not sdcv at all.

Now the part I got wrong. Two of v4's instruments do not work:

- **`/proc/PID/wchan` came back empty in all 23 lookups** — either restricted or not built into this kernel. That was exactly the field that would have named the sleep.
- **The scattered-4 kB latency figure is meaningless.** It reads 0.0-0.1 ms whether the device is doing 10 MB/s or 107 MB/s, i.e. faster per byte than the sequential test, which cannot be true of real cold reads. The kernel was reading ahead into the cache faster than my strided reads could outrun it, so I was timing cache hits.
- And `/proc/PID/io` claims sdcv reads ~12 MB on *every* lookup, including warm repeats where `/proc/diskstats` says 0.0 MB. Those two disagree and I do not know which to believe, so please read the absolute byte counts in that log with suspicion. The freeze-then-jump *shape* within a single lookup does not depend on the scale, and that is what the finding rests on.

v5 fixes all three:

1. **`/proc/PID/syscall` on every sample**, which gives the syscall and its arguments. The log now prints e.g. `while parked, it was in: write(fd 1) x12` — fd 1 is the pipe back to KOReader, so if sdcv is stalling because KOReader is not draining its output, that is now stated outright. A task stopped in a page fault rather than a syscall shows up as `page fault / no syscall`, which is a different answer and just as useful.
2. **`/proc/PID/stack`**, the kernel backtrace of the wait, which names it precisely. Taken only while the process is actually parked, and capped per lookup because a stack walk is not free. The header line now reports whether these two files are readable at all, so this cannot silently produce nothing the way `wchan` did.
3. **The storage calibration now uses `O_DIRECT`**, which bypasses the page cache entirely — no readahead can prefetch for it and nothing can be served from RAM, so the scattered figure becomes an honest per-request latency. If your filesystem refuses the flag it falls back to the old method and says so, with the untrustworthy number suppressed rather than printed.
4. `read_bytes` is now reported as a delta rather than a running total, and the log flags it explicitly when a process appears to have read more than the whole device did.

Same routine, and `dropcache` is still the fast way to get a cold lookup. If the syscall line comes back pointing at `write(fd 1)`, the fix is in KOReader rather than in sdcv or this patch, and I will know where to look.

[2-dictprobe.lua.zip](https://github.com/user-attachments/files/30779147/2-dictprobe.lua.zip)

Commodore64user:
This time I also threw in some failures, `[2026-08-06 15:42:10] LOOKUP #19  word="Ivanovich"  total=1.011s` is with fuzzy search off (technically every single one is with fuzzy off) and `[2026-08-06 15:42:18] LOOKUP #20  word="Ivanovich"  total=2.718s` run with fuzzy on offered thorough https://github.com/koreader/koreader/blob/5584b5e57136977a35a8deca393e54fe5fb163cd/frontend/apps/reader/modules/readerdictionary.lua#L1501-L1505

Anyway here is the new one

[dictprobe.log.txt](https://github.com/user-attachments/files/30794436/dictprobe.log.txt)

asxelot:
@Commodore64user That fuzzy-search comparison of yours turned into the most useful thing in the log, and it is a real KOReader bug — nothing to do with this PR.

## The fuzzy search freezes the whole UI

`readerdictionary.lua:1592`, the "Fuzzy search" button in `showNoResultsDialog`, calls `self:stardictLookup(...)` **directly instead of inside `Trapper:wrap`**. That callback runs from UIManager's event loop, long after the coroutine of the lookup that showed the dialog has finished — so `Trapper:dismissablePopen` finds no coroutine, takes its `unwrapped dismissablePopen(), falling back to blocking io.popen()` path, and blocks KOReader's main loop for the entire search. Your lookup `#20` was 2.63 s of completely frozen UI: no "Searching…" message, no repaint, and no way to dismiss it. The sibling "search with preset" button right next to it goes through `onLookupWord`, which is wrapped, so it behaves.

I only noticed because `#20` is the one lookup in your log with no `sdcv process` block at all — my sampler needs UIManager to be alive to run, and it wasn't. Introduced in 99f4eaeaa (koreader#14399).

```diff
--- a/frontend/apps/reader/modules/readerdictionary.lua
+++ b/frontend/apps/reader/modules/readerdictionary.lua
@@ -1589,7 +1589,14 @@ function ReaderDictionary:showNoResultsDialog(word, dict_names, fuzzy_search, bo
                 if new_word == "" then return end
                 UIManager:close(dialog)
                 -- Re-run the lookup with the (possibly edited) word and fuzzy enabled.
-                self:stardictLookup(new_word, dict_names, true, boxes, link, dict_close_callback)
+                -- Wrapped through Trapper, as we may be using Trapper:dismissablePopen() in it:
+                -- this callback runs from UIManager's event loop, long after the coroutine of
+                -- the lookup that showed this dialog has finished, so without a wrap of its own
+                -- dismissablePopen() falls back to a blocking io.popen() and freezes the UI for
+                -- the whole search (seconds, on a device with many dictionaries and cold cache).
+                Trapper:wrap(function()
+                    self:stardictLookup(new_word, dict_names, true, boxes, link, dict_close_callback)
+                end)
             end,
         }
     elseif has_presets then
```

That belongs in koreader/koreader rather than here; I will open it separately. It does not explain the slow lookups, it just means fuzzy searches are worse than they need to be.

## The stall is still there, and it has its own signature

Your `#11` (6.2 s) froze at 1.5 MB from 0.5 s to 4.2 s, then read the remaining 11 MB in 1.5 s — same shape as last time. The new thing is that the kernel stack during the stall is *specific*:

| stack | in the 6.2 s stall | in all 21 other lookups |
|---|---|---|
| `0xc0380b4c<-0xc0383e44<-0xc0386324<-0xc0386434` | **12x** | 2x |
| `0xc022741c<-0xc0227d84<-0xc02282e4<-0xc0254eb4` | 1x | 33x |
| `...<-0xc0384978<-0xc0291b1c` | 0x | 27x |

So it is one particular wait, not a general slowdown. Resolving that one address would name the cause outright.

## Which is a problem, because your kernel hides almost everything

That is now three fields in a row it does not support: `mincore()` lies, `/proc/PID/wchan` is empty, and `/proc/PID/syscall` returns **0 for every parked sample** rather than the real number or -1. `/proc/PID/stack` does work, but prints raw addresses, so `CONFIG_KALLSYMS` looks to be off too. And `O_DIRECT` was refused by the filesystem, exactly the risk I flagged, so there is still no honest per-request latency figure.

v6 stops asking the kernel to name things and works it out instead:

1. **A kernel-stack translation table, built on your device.** At startup it spawns three short-lived children that are each blocked in a way *it chose* — a timer sleep, a read on a pipe with no data, and a read of a large uncached file — and records their kernel stacks. Your stall's stack is then matched against them, and the log says `== reading an uncached file (storage wait)` or `== blocked reading a pipe (no data)` or `(matches none of the fingerprints)`. No symbols needed. If it comes back as the pipe, the stall is KOReader's fault; if it comes back as storage, it is the device; if neither, both of my theories were wrong and that is worth knowing too.
2. **Read syscalls vs bytes off the device.** `read_bytes` was frozen during the stall — now the log also counts read *syscalls*, so "grinding through cached data" and "genuinely parked" become distinguishable.
3. **Voluntary context switches**, which separate one long block from thousands of short ones. Nothing sampled at 120 ms can tell those apart.
4. **Which file it has open and at what offset**, from `fdinfo` — so if it is stuck on one particular dictionary, that is now visible by name.
5. And if `/proc/kallsyms` happens to be usable after all, the addresses get resolved to symbols automatically and points 1-4 become unnecessary.

Everything above is verified against your last log's actual stack addresses, and the counters are checked against a known workload. The one thing I could not test is the fingerprint capture itself, because Docker's kernel does not expose `/proc/PID/stack` at all — but yours does, which is why this approach is worth a run.

[2-dictprobe.lua.zip](https://github.com/user-attachments/files/30798582/2-dictprobe.lua.zip)

Commodore64user:
> Introduced in 99f4eaeaa (koreader#14399).

that was yours truly, I've been a bad boy then.

I am actually pleasantly surprised it took this long but it finally happened

```
./luajit: ./patches/2-dictprobe.lua:822: attempt to index local 's' (a nil value)
stack traceback:
	./patches/2-dictprobe.lua:822: in function 'tail'
	./patches/2-dictprobe.lua:826: in function 'matchFingerprint'
	./patches/2-dictprobe.lua:974: in function 'childSampleLines'
	./patches/2-dictprobe.lua:1714: in function 'action'
	frontend/ui/uimanager.lua:1019: in function '_checkTasks'
	frontend/ui/uimanager.lua:1465: in function 'handleInput'
	frontend/ui/uimanager.lua:1585: in function 'run'
	./reader.lua:301: in main chunk
	[C]: at 0x00013ed5
[ue_wait_for_event:269] poll: Interrupted system call
```

I only managed to this far, if it helps

<details>
  <summary>Spoiler warning</summary>
  

  
  ```
================================================================================
DictProbe v6 started  2026-08-07 01:08:06
  KOReader v2026.07   device=KindlePaperWhite5
  dict dir=data/dict
  tmp=/tmp   /proc=true   mincore=true (self-test: a file we just read in full reads
                                  back as 0% resident -> residency column is MEANINGLESS on this kernel, ignoring it)
  storage calibration=ready   syscall=readable  kernel stack=readable  kallsyms=unusable (stack addresses stay raw; the fingerprints below are the key)
  cpu_freq=[600000,600000]kHz gov=ondemand  loadavg=0.21  MemFree=170MB MemAvailable=253MB Cached=71MB Dirty=0kB  cpu_spin=0.13ms (1.38x of best 0.09ms)
  storage now (fadvise+read): 1MB seq=98.1 MB/s (10ms)  [scattered reads unavailable]
================================================================================
[2026-08-07 01:08:16] inventory: 30 dictionary files, scanned in 5.333s (capped)
[2026-08-07 01:08:17]   kernel-stack fingerprints (this kernel prints raw addresses, so these are the
  translation table -- match a stall's stack against them):
    timer sleep (nanosleep)                    state=S syscall=syscall#0
      0xc01bb560<-0xc01bb6c4<-0xc0108a40<-0xffffffff
    blocked reading a pipe (no data)           state=S syscall=syscall#0
      0xc028f428<-0xc028faa4<-0xc0285ee8<-0xc0286c98
    reading an uncached file (storage wait)    (finished before it could be sampled)
  ```
  
</details>

asxelot:
Sorry about that — my patch crashed your reader, which is the one thing a diagnostic patch must not do. And no need to feel bad about the fuzzy-search line; it is a subtle trap, the wrapping is not visible from the call site, and it took a 6-version instrument to notice.

The crash was mine: a fingerprint whose child exits before it can be sampled has no stack recorded, and the matcher indexed it anyway. I introduced exactly that when I made stackless entries recordable so I could see the machinery working in Docker — and Docker could never catch the bug, because no *sampled* stack is readable there, so the matcher was never called with a real one. It also only fires on a stack that matches nothing, which is why it took until now.

Fixed, and then fixed properly: **UIManager does not `pcall` its scheduled tasks**, so any error in a patch callback takes KOReader down with it. Every callback DictProbe hands to UIManager is now wrapped, and the whole per-lookup report runs inside an `xpcall`. A failure writes `DictProbe failed to build its own report -- this is a bug in the patch, not in KOReader` to the log and carries on. I verified it by injecting a deliberate fault: contained, logged, lookups keep working.

## Your partial log already answered something

The two fingerprints that did capture:

```
timer sleep (nanosleep)              0xc01bb560<-0xc01bb6c4<-0xc0108a40<-0xffffffff
blocked reading a pipe (no data)     0xc028f428<-0xc028faa4<-0xc0285ee8<-0xc0286c98
```

Now compare the pipe one against the stacks in your previous log. Its two outer frames are `0xc0285ee8<-0xc0286c98` — and the stack that showed up **10 times** in that log was `0xc022741c<-0xc0227690<-0xc0285ee8<-0xc0286c98`. Same outer frames, so that one is a pipe read: sdcv waiting for KOReader, which is normal and cheap.

The stall's dominant stack was `0xc0380b4c<-0xc0383e44<-0xc0386324<-0xc0386434`, which matches **neither** fingerprint. So the 2.5-4 s freeze is *not* sdcv blocked on the pipe back to KOReader. That was one of my two hypotheses and it is now out, which leaves storage — and the third fingerprint, the one that would confirm it, is the one that failed to capture.

Also, your kernel's `/proc/PID/syscall` is now definitively broken rather than merely suspicious: a child parked in `sleep` reports `syscall#0`. v7 tests that at startup and stops reporting the field instead of printing a number that means nothing.

## Why the storage fingerprint missed, and what changed

Three separate mistakes, all mine:

1. It picked its target from the probe's own file inventory — which is capped and sorted index-files-first, so on a real installation **no `.dict` file is in it at all** and the biggest entry is a ~7 MB `.syn`. That was under my 8 MB threshold, so it silently fell back to a 4 MB scratch file. v7 does its own scan for a file over 32 MB.
2. The command was single-quoted, and your dictionary folders include names like `Merriam Webster's Collegiate`. An apostrophe ends a single-quoted shell word, so that would have failed on your device even once the size check passed. Now escaped properly, and tested against a 40 MB fixture in an apostrophe path.
3. It slept a fixed 0.4 s and then looked once. That is a race whose correct value depends on how slow the storage is — the very thing being measured. On your eMMC a cold 512-byte-block read of a 200 MB file takes many seconds; on my machine the same read takes 58 ms, which I measured. v7 polls every 10 ms for up to 2 s and takes the first sample where the process is actually parked, and it evicts the file's cache first so the wait is a real one.

So this run should fill in the third row of the table, and if it matches the stall, the answer is storage and I will go and look at why a page-cache read of an index file parks for 3 s on that device.

One honest caveat, same as last time: I cannot exercise the storage fingerprint locally, because nothing here is slow enough to park a process — Docker's kernel does not expose `/proc/PID/stack` at all, and its reads finish in 58 ms. The crash path, the guards, the matcher, the target selection and the quoting are all verified; the capture itself only happens on your hardware.

[2-dictprobe.lua.zip](https://github.com/user-attachments/files/30823443/2-dictprobe.lua.zip)

Commodore64user:
Here we go again, this whole investigation it's starting to feel like a [Dr] House episode, have we tested for Lupus yet ;)

just curiosity here: Are you using AI to help with all this? if so which one? and how much of it is you vs the AI?

[dictprobe.log.txt](https://github.com/user-attachments/files/30830086/dictprobe.log.txt)

asxelot:
@Commodore64user Lupus is off the table — I found it. And here is a drop-in KOReader patch you can test without rebuilding anything: `koreader/patches/2-dictprefetch.lua`, restart, done. Deleting the file removes it completely.

## What is actually wrong

It is not a stall. It is a fixed cost paid on **every** lookup, which only becomes visible once the page cache is cold.

Across all 20 lookups in your last log, sdcv asks for 0.5-1.3 MB through 110-210 read syscalls — and that is the same whether the lookup took 0.89 s or 6.02 s, and whichever word you looked up. The work is not proportional to the word. It is proportional to your *dictionary set*.

Your `.syn.oft` files total **8.2 MB across 12 dictionaries**, and sdcv reads all of them, in full, every time it starts. This is the last line of `load_offset_cache`:

```c
memcpy(&wordoffset[0], mf.begin() + strlen(CACHE_MAGIC) + sizeof(guint32),
       wordoffset.size() * sizeof(wordoffset[0]));
```

The whole table, copied out of the mapping, for every dictionary that has a `.syn`. sdcv is a fresh process per lookup, so nothing carries over except the OS page cache — and suspend drops that (your logs show `Cached` going 105 MB → 30 MB across a wake). Meanwhile the binary search that actually *uses* the table visits about 19 scattered entries per dictionary, roughly 0.9 MB for your whole set. So ~11x more is read than is used.

On your device that load is serialised at about one blocking wait per 4 kB page: across all 20 lookups, `voluntary_ctxt_switches ≈ read_bytes / 4096` (ratio 0.78-1.18). Three thousand sequential waits, against storage my probe measured at 3.3-11 MB/s just after a wake versus 86-111 MB/s warm. That is your 5-7 seconds.

And to say it plainly, since it was the original question: **the `.syn.oft` cache in this PR is a large win, not the cause.** I reproduced the pre-patch behaviour locally — a dictionary with a 46 MB `.syn` and no `.syn.oft` costs 44.36 MB, 44.28 MB and 44.49 MB for the words "hand", "water" and "книга": identical every time, because that is the linear `.syn` scan this PR removed. In over a hundred logged lookups on your device the cache was never once rejected.

## What the patch does

It cannot make your storage faster, so it does the next best thing: **it pays the cost before you tap the word.** A couple of seconds after each wake (and once at startup) it reads the offset tables back into the page cache in the background, in 128 kB slices, so the ~8-10 MB every lookup needs is already in memory by the time you look something up. If you do tap a word while it is still working, it stops immediately and finishes afterwards — a lookup you are waiting for always wins.

Worth flagging one thing I got wrong on the first attempt, because it is counter-intuitive. I first wrote it with `posix_fadvise(POSIX_FADV_WILLNEED)`, which is exactly what that call is for. Measured on a 44 MB file:

```
after DONTNEED                     :   0.0% resident
fadvise(WILLNEED, whole file)      : returned in 0.1 ms
  +0.2s                            :   2.8% resident
  +1.0s                            :   2.8% resident
  +3.0s                            :   2.8% resident
after simply reading it (33 ms)    : 100.0% resident
```

The kernel treats WILLNEED as a hint and clamps it to its readahead window, so it brought in 2.8% and stopped. Reading the file is what actually warms the cache, so that is what the patch does.

What I verified before sending it, since my last patch crashed your reader:

- every `.oft` goes from 0% to 100% resident; 8.4 MB warmed in 2.4 s
- **longest single scheduler tick: 1 ms** — it reads one slice per tick and yields between, so it cannot make the UI stutter
- it really does pause for a lookup in flight, and logs how long it waited
- loads and runs cleanly on a platform with no `/proc` and no `posix_fadvise`
- the underlying mechanism, measured end to end: warming the files sdcv is about to read drops its cold reads from **44.39 MB to 0.18 MB**

What I could not test is your exact case — a patched sdcv with twelve dictionaries on a PW5 — because I do not have that combination here. On paper it should turn the first lookup after a long sleep from 5-7 s into something close to your normal ~1 s, since the 8 MB it has to read will already be in memory. If it does not, the log line it writes (`dictprefetch: after resume -- warmed X MB across N files in Ys`) will say what it managed to do, and that plus a DictProbe log will tell us where the rest is going.

It is independent of the probe — you can run both, or just this one.

## The proper fix

The real fix belongs in sdcv: point at the offset table inside the mapping instead of copying it out, so a lookup faults in only the pages its binary search actually reaches. The reason it copies today is alignment — the magic is 30 bytes plus a 4-byte marker, so the offsets start at byte 34, two bytes off a 4-byte boundary — which means padding the header and bumping the cache version. I have that written and measured (page reclaims ~1640 → ~795, peak RSS 22.1 MB → 9.6 MB, output byte-identical across 11 words), but it needs an sdcv rebuild, so it is no use to you today. I will put it up separately.

[2-dictprefetch.lua.zip](https://github.com/user-attachments/files/30853360/2-dictprefetch.lua.zip)

P.S. It is all from Claude Code

Commodore64user:
hey @asxelot, sorry I had forgotten about this, as I had been reading on a NT kindle, are you waiting for me to do something here?

asxelot:
@Commodore64user try to install the latest patch, it should fix that issue

[2-dictprefetch.lua.zip](https://github.com/user-attachments/files/30853360/2-dictprefetch.lua.zip)

Commodore64user:
mmm... I have been working under the assumption that we want to solve the problem for everyone... not just me.

asxelot:
@Commodore64user sure, but I can't reproduce that issue on my Kindle, so you're helping me. If that patch works, I will create a PR.

Commodore64user:
I mean, it doesn't take 7 seconds anymore, but if you are in a controlled environment, I am not sure you need 15 dictionaries. Just set a hard memory limit on your container. 128MB RAM and your large dicts should do the trick, no?

PR #2444, opened 2026-07-11
mupdf: expose ink annotation support in the FFI layer

tomasAlabes:
### What

Wraps three MuPDF functions that `libwrap-mupdf` did not previously export, and adds a Lua helper on `page_mt` that mirrors the existing `addMarkupAnnotation`:

| wrapper | MuPDF function |
|---|---|
| `mupdf_pdf_set_annot_ink_list` | `pdf_set_annot_ink_list` |
| `mupdf_pdf_set_annot_border_width` | `pdf_set_annot_border_width` |
| `mupdf_pdf_update_annot` | `pdf_update_annot` |

Plus `page:addInkAnnotation(strokes, color, width, opacity)`, which creates a native ink annotation (`/Subtype /Ink`) with per-annotation colour, line width and opacity. Ink-list points use the same page-coordinate convention as the existing quad-point path.

`mupdf_pdf_update_annot` is exposed deliberately: it is what makes MuPDF synthesize the `/Rect` and `/AP` appearance stream, without which an annotation renders only in KOReader's own geometry renderer and is invisible in desktop PDF viewers (Preview, Acrobat, etc.).

### Why

There is currently no way to create ink annotations through the FFI layer — `PDF_ANNOT_INK` exists in the enum, but no setter for its geometry is exported, and `addMarkupAnnotation` only handles highlight/underline/strikeout. This adds the missing primitives so freehand annotations can be written into PDFs.

The immediate consumer is the [Pencil plugin's "save annotations to PDF" feature](https://github.com/mysticknits/pencil.koplugin) (which currently ships a locally-rebuilt `libwrap-mupdf` to get these symbols); landing this upstream means it — and any future in-tree use — works against a stock build.

### Testing

Cross-built for `TARGET=kobo` (MuPDF 1.27.2) with the `koreader/kokobo` toolchain image; all three symbols export cleanly and `libwrap-mupdf.so` links. The same change was verified end-to-end on a Kobo Libra Colour (KOReader v2026.03 / MuPDF 1.26.12): ink and markup annotations are written with a correct `/AP` and render in desktop PDF viewers.

### Notes

- Follows the existing `MUPDF_WRAP` macro style; `ffi/mupdf_h.lua` is updated to match `make ffi` output.
- Happy to trim `addInkAnnotation` and land just the wrapper primitives if you'd prefer to keep the Lua side minimal until there's an in-tree caller.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2444)


ffi/mupdf.lua
```diff
@@ -809,6 +809,55 @@ function page_mt.__index:addMarkupAnnotation(points, n, type, bb_color)
...
+    if ok == nil then merror(self.ctx, "could not set ink list") end
+
+    local color = ffi.new("float[3]")
+    color[0] = bb_color.r / 255
+    color[1] = bb_color.g / 255
+    color[2] = bb_color.b / 255
+    ok = W.mupdf_pdf_set_annot_color(self.ctx, annot, 3, color)
+    if ok == nil then merror(self.ctx, "could not set ink annotation color") end
+
+    ok = W.mupdf_pdf_set_annot_border_width(self.ctx, annot, width or 1.0)
```
Frenzie, line 857: Is the width necessarily constant? I'd somewhat expect writing with a pen to pick up pressure.

ffi/mupdf.lua
```diff
@@ -809,6 +809,55 @@ function page_mt.__index:addMarkupAnnotation(points, n, type, bb_color)
...
+    if ok == nil then merror(self.ctx, "could not set ink annotation border width") end
+
+    ok = W.mupdf_pdf_set_annot_opacity(self.ctx, annot, opacity or 1.0)
+    if ok == nil then merror(self.ctx, "could not set ink annotation opacity") end
+end
+
 function page_mt.__index:deleteMarkupAnnotation(annot)
```
Frenzie: Should there be a delete function to go along with it as well?

ffi/mupdf.lua
tomasAlabes: A PDF /Ink annotation only carries a single line width (the /BS border width); the /InkList itself is bare polylines with no per-vertex data, so genuine pressure/variable-width strokes can't be represented in a standard ink annotation without synthesising a custom /AP appearance stream (e.g. filled polygons per segment). This just exposes `pdf_set_annot_border_width`, whose signature is a single float, so I kept the API matching it.

On the caller side (the Pencil plugin) the stroke capture doesn't record pressure today and its own on-screen rendering uses a constant per-tool width, so a single width per stroke is faithful to what's drawn. If pressure capture lands later, the natural path is a custom AP stream rather than changing this setter — happy to revisit then.

ffi/mupdf.lua
tomasAlabes: Deletion is already covered by the existing `page_mt.__index:deleteMarkupAnnotation(annot)` just below — it calls `mupdf_pdf_delete_annot` / `pdf_delete_annot`, which is subtype-agnostic. An ink annotation is a markup annotation in the PDF spec, so it's deleted through the same path as highlights: enumerate with `first_annot`/`next_annot`, then pass the `pdf_annot*` in. So no ink-specific delete is needed. If the `Markup` in the name reads as too narrow now that ink is in the mix, I'm happy to rename it to `deleteAnnotation` (keeping an alias).

Frenzie: No need for an alias; just a name change is fine.

tomasAlabes: > No need for an alias; just a name change is fine.

Happy to rename it to `deleteAnnotation`. Just to confirm the sequencing: the only current caller is in the frontend repo (`koreader/koreader`, `frontend/document/pdfdocument.lua` — the highlight-clear path), so a plain rename here would break that call until a matching change lands there. Do you want me to open a companion PR against `koreader/koreader` to update the call site so they can be merged/bumped together? Or would you rather I keep the alias just for that transition and drop it once the frontend is updated?

Frenzie: > Do you want me to open a companion PR against koreader/koreader to update the call site so they can be merged/bumped together? 

Yes, that's how we work.

tomasAlabes: Done — renamed to `deleteAnnotation` here, and opened the companion frontend PR that updates the sole caller: koreader/koreader#15672. Ready to merge/bump together whenever.

benoit-pierre:
So where is `mupdf_pdf_update_annot` going to be used exactly?

wrap-mupdf.h
```diff
@@ -186,6 +186,15 @@ MUPDF_WRAP(mupdf_pdf_set_annot_color, void*, NULL,
...
     pdf_annot *annot, float opacity)
+MUPDF_WRAP(mupdf_pdf_set_annot_ink_list, void*, NULL,
+    { pdf_set_annot_ink_list(ctx, annot, n, count, v); ret = (void*) -1; },
+    pdf_annot *annot, int n, const int *count, const fz_point *v)
+MUPDF_WRAP(mupdf_pdf_set_annot_border_width, void*, NULL,
+    { pdf_set_annot_border_width(ctx, annot, width); ret = (void*) -1; },
+    pdf_annot *annot, float width)
+MUPDF_WRAP(mupdf_pdf_update_annot, void*, NULL,
+    { pdf_update_annot(ctx, annot); ret = (void*) -1; },
+    pdf_annot *annot)
```
benoit-pierre: Please rebase and switch to `MUPDF_WRAP_BOOL`:
```suggestion
MUPDF_WRAP_BOOL(mupdf_pdf_set_annot_ink_list,
    pdf_set_annot_ink_list(ctx, annot, n, count, v),
    pdf_annot *annot, int n, const int *count, const fz_point *v)
MUPDF_WRAP_BOOL(mupdf_pdf_set_annot_border_width,
    pdf_set_annot_border_width(ctx, annot, width),
    pdf_annot *annot, float width)
MUPDF_WRAP_BOOL(mupdf_pdf_update_annot,
    pdf_update_annot(ctx, annot),
    pdf_annot *annot)
```

ffi/mupdf.lua
```diff
@@ -809,9 +809,58 @@ function page_mt.__index:addMarkupAnnotation(points, n, type, bb_color)
...
+    if ok == nil then merror(self.ctx, "could not set ink annotation border width") end
+
+    ok = W.mupdf_pdf_set_annot_opacity(self.ctx, annot, opacity or 1.0)
+    if ok == nil then merror(self.ctx, "could not set ink annotation opacity") end
+end
+
+function page_mt.__index:deleteAnnotation(annot)
     local ok = W.mupdf_pdf_delete_annot(self.ctx, ffi.cast("pdf_page*", self.page), annot)
-    if ok == nil then merror(self.ctx, "could not delete markup annotation") end
+    if ok == nil then merror(self.ctx, "could not delete annotation") end
```
benoit-pierre: Once updated to use `MUPDF_WRAP_BOOL`:
```suggestion
    if not ok then merror(self.ctx, "could not delete annotation") end
```

benoit-pierre: Same for other similar calls.

wrap-mupdf.h
tomasAlabes: Done — rebased on master and switched all three wrappers to `MUPDF_WRAP_BOOL` (with the matching `bool` declarations in `mupdf_h.lua`).

ffi/mupdf.lua
tomasAlabes: Done — the ink-annotation calls (`set_annot_ink_list`, `set_annot_border_width`, `set_annot_opacity`, `update_annot`) and the renamed `deleteAnnotation` now all check `if not ok` against the `bool` return.

tomasAlabes:
`mupdf_pdf_update_annot` wraps `pdf_update_annot`, which is what makes MuPDF synthesize an annotation's `/Rect` and `/AP` appearance stream. Without it an annotation only has its raw geometry and renders in KOReader's own geometry renderer but is invisible in desktop viewers (Preview, Acrobat, ...).

I originally exposed it as a bare primitive, but per the review I now call it inside `page:addInkAnnotation` itself (right after setting the ink list / colour / width / opacity), so the in-tree API produces a complete, viewer-visible annotation and the wrapper has an in-tree caller. The downstream [Pencil plugin](https://github.com/mysticknits/pencil.koplugin) also uses it directly to retrofit an `/AP` onto annotations it writes.

benoit-pierre (approved):

PR #2449, opened 2026-07-14
mupdf : various improvements

benoit-pierre:
To be merged preferably before #2444 (so it can be updated to use `MUPDF_WRAP_BOOL(…)`).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2449)


benoit-pierre:
Damn, passthrough device support was added in 1.28.0 (which I'm already using on my branch, waiting for 1.28.1 before bumping on master). I've backported a minimal patch (just the `passthrough` member addition to `fz_device`).

PR #2452, opened 2026-07-17
MuPDF: enable Background Cleanup for PDFs with Mask (not SMask)

kerivin:
[Fixes Background Cleanup for archive.org PDFs that have Mask but no SMask](https://github.com/koreader/koreader/issues/15694)

image

image

Also, I'm not sure if I should rename `page_has_smask` to `page_has_mask` (and other smask-related names) now or leave it to the next PR?.. Ugh, another submodule updating x)

(The name `page_has_mask` is almost as confusing as `page_has_smask`, so it should be rather `page_has_masks` or `page_has_any_mask`?..)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2452)


Frenzie:
Just that suffices? Hm, hopefully there are no side effects, but since you have to explicitly enable it I suppose it doesn't matter too much.

`has_transparency_mask`?

kerivin:
I have tested some files with SMask, on which I had been testing SMask before, and Background CLeanup works the same. I guess it's the equivalent of `xobj.Mask | xobj.SMask` from the discussion, and it works mostly because `mask` field of an image passed to fz_device's `fill_image` function can be either SMask or Mask, so I don't need to change the processing, just the detection.
Although there might be other types of valid masks I don't know of.

Frenzie:
There's also some kind of color-key / chroma-key masking.
```
https://github.com/pymupdf/PyMuPDF/discussions/1259
```

kerivin:
I'll look into it and test some files, but if that's valid I'm not going to add them in this PR anyway, I would create a new one...........

Frenzie (approved):

PR #2453, opened 2026-07-19
jpeg: re-enable bmp writing support + miscellaenous improvements

benoit-pierre:
- re-enable BMP support in libjpeg-turbo
- change API to be closer to the PNG one (including better error reporting)
- support more input formats when encoding
- drop grayscale conversion code (do it in blitbuffer)
- add tests, including for `writeBMP`
- don't use fast upsampling / DCT: the drop in quality, particularly when encoding is not worth it

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2453)


Frenzie:
> On modern CPUs, these settings have little or no performance benefit and are retained solely for backward compatibility.

That's not necessarily a valid argument depending on what they mean by a "modern CPU." Iirc it might refer to AVX2 (post-2008ish), which doesn't play much of a role for us in ARM-device land. That being said, I'm usually in favor of trading a speed for quality in these matters, as long as the difference isn't too excessive.

PR #2454, opened 2026-07-20
[fix] blitfuffer.c: prevent SIGFPE in large anti-aliased rounded corners

Frenzie:
Use floating-point error terms to avoid integer overflow and division by zero.

Cf. <https://github.com/koreader/koreader/issues/15708>.

---

It seemed more elegant to me than the alternative quick workaround I sketched in https://github.com/koreader/koreader/issues/15708#issuecomment-5023828253 (which could theoretically still cause issues regardless), and floating point should be hardware-supported, but alternative solutions are welcome of course.

Refer to <http://members.chello.at/easyfilter/bresenham.html> for the algorithm.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2454)

PR #2455, opened 2026-07-21
fix(blitbuffer): fix inner rounded corner radius

tachibana-shin:
This can be verified by comparing it to `paintRoundedCorner` on line 2042—which uses the same midpoint circle algorithm but is written correctly:
| Function | Line |  Initialize delta2 | Correct? |
| --- | -- | -- | --- |
| paintCircle (before fix)  |1963 | 5/4 - r | Wrong — uses the outer radius for the inner circle |
| paintCircle (after fix) | 1963 | 5/4 - r2 | Correct | 
| paintRoundedCorner | 2042 | 5/4 - r2 | Correct (reference) |

The midpoint circle algorithm requires that delta be initialized using its own radius (5/4 - radius). The outer circle uses r, and the inner circle uses `r2 = r - w`. The original error caused the inner circle to be drawn in the wrong position when `w < r`

To be honest, this file is quite sensitive. Personally, I think this is a bug, but we still need to be cautious so we don’t end up with a situation like the recent Node.js release incident

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2455)

### DIFF
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -2039,7 +2039,7 @@ function BB_mt.__index:paintRoundedCorner(off_x, off_y, w, h, bw, r, c, anti_ali
         local r2 = r - bw
         local x2 = 0
         local y2 = r2
-        local delta2 = 5/4 - r
+        local delta2 = 5/4 - r2
 
         while x < y do
             -- decrease y if we are out of circle
--- a/ffi/blitbuffer.lua
+++ b/ffi/blitbuffer.lua
@@ -1960,7 +1960,7 @@ function BB_mt.__index:paintCircle(center_x, center_y, r, c, w)
     local r2 = r - w
     local x2 = 0
     local y2 = r2
-    local delta2 = 5/4 - r
+    local delta2 = 5/4 - r2
 
     -- draw two axles
     for tmp_y = r, r2+1, -1 do


benoit-pierre:
I'm confused:
- `paintCircle` before:
https://github.com/koreader/koreader-base/blob/2b9ab8527f67deb79783738b101909166dd00d92/ffi/blitbuffer.lua#L1954-L1963
- `paintRoundedCorner` before:
https://github.com/koreader/koreader-base/blob/2b9ab8527f67deb79783738b101909166dd00d92/ffi/blitbuffer.lua#L2033-L2042

So isn't your "reference" wrong too?

tachibana-shin:
In my opinion, we should fix this entire function (thanks for pointing it out). I’m confident in my suggestion to change it to r2 because I’m very familiar with this algorithm from my work with canvas

The more bug is in `paintCircle` at line 1963, where `delta2 = 5/4 - r` should be `delta2 = 5/4 - r2`.
The proof is in the C implementation (`blitbuffer.c:3128-3135`), which correctly initializes the inner circle parameters using r2:
```c
// r2 ... radius of inner circle; might be zero
const unsigned int r2 = r - bw;
int f2 = 1 - r2;          // inner circle uses r2
int ddF2_y = -2 * r2;     // inner circle uses r2
```

The midpoint circle algorithm [Bresenham's circle algorithm]( https://en.wikipedia.org/wiki/Midpoint_circle_algorithm) requires that delta be initialized with `5/4 - radius` where radius is the radius of that specific circle. The outer circle uses r, so` delta = 5/4 - r` is correct. The inner circle has radius `r2 = r - w`, so delta2 must be `5/4 - r2`.

Frenzie:
> To be honest, this file is quite sensitive. Personally, I think this is a bug, but we still need to be cautious so we don’t end up with a situation like the recent Node.js release incident

What are you referring to precisely?

tachibana-shin:
> > To be honest, this file is quite sensitive. Personally, I think this is a bug, but we still need to be cautious so we don’t end up with a situation like the recent Node.js release incident
> 
> What are you referring to precisely?

I'm sure this is a bug, but it has always worked in Koreader and its plugins. So, do you guys consider the incorrect output (due to the algorithm) to be a bug? It looks like this: https://www.linkedin.com/posts/matteocollina_hey-everyone-i-reported-a-nodejs-cve-activity-7478478614783528960-tkWI

Frenzie:
I'd suggest to simply include some before and after screenshots in relevant conditions.

tachibana-shin:
image
image

You can clearly see that on the left, `-r` is used, while on the right, `-r2` is used

Frenzie:
Then there's no need to be concerned about badly breaking anything. ;-)

benoit-pierre:
Are you not seeing the same problems as https://github.com/koreader/koreader/issues/15708#issuecomment-5035814926?

benoit-pierre:
Those rendering artifacts are due to #2454.

tachibana-shin:
> Are you not seeing the same problems as [koreader/koreader#15708 (comment)](https://github.com/koreader/koreader/issues/15708#issuecomment-5035814926)?

No, I think it's something else. I've never had a crash like that before

benoit-pierre:
Not the crash, the rendering artifacts. But you need to be on a version including #2454.

benoit-pierre:
Which is not yet available on a nightly.

tachibana-shin:
> Not the crash, the rendering artifacts. But you need to be on a version including #2454.



> Not the crash, the rendering artifacts. But you need to be on a version including #2454.

probably not yet. I installed the test plugin, but I didn't notice anything unusual. Could you share the parameters causing the display error?

benoit-pierre:
No need to change the default parameters:
2026-07-21-202311_354x289_scrot

But it is more visible with a bigger border size:
2026-07-21-202539_350x285_scrot


This patch seems to take care of the issue: https://github.com/koreader/koreader-base/pull/2454#issuecomment-5037281481.

tachibana-shin:
image
image

I don't see anything like that

tachibana-shin:
I noticed that the patch simply converts the number calculations to floats. It might fix integer overflow errors caused by multiplication, but I don't think that's relevant

poire-z:
Cross-ref https://github.com/koreader/koreader/issues/15753#issuecomment-5184059369 as I still have an issue with rounder corners on the emulator.

Frenzie:
Regression in 147d3c972882130627c347546c50332b63304187.

PR #2456, opened 2026-07-21
fix(harfbuzz): prevent leaks and division by zero

tachibana-shin:
- Use ffi.gc to ensure HarfBuzz sets are freed
- Initialize found index to prevent nil errors
- Add division-by-zero check for coverage

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2456)

### DIFF
--- a/ffi/harfbuzz.lua
+++ b/ffi/harfbuzz.lua
@@ -64,7 +64,7 @@ function hb_face_t:getCoverage()
     for script_id, tab in ipairs(coverage.scripts) do
         local hit, total = intersect(tab)
         -- for scripts, we do only rough majority hit
-        if 2*hit > total then
+        if total > 0 and 2*hit > total then
             scripts[script_id] = hit / total
         end
     end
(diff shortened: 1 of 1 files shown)


ffi/harfbuzz.lua
```diff
@@ -69,21 +70,19 @@ function hb_face_t:getCoverage()
...
+        local found = 1
         local hit, total = intersect(tab)
         -- for languages, consider predefined threshold by glyph count
         for i=1, #coverage_thresholds, 2 do
             if total > coverage_thresholds[i] then
                 found = i+1
             end
         end
-        if hit*100/total >= coverage_thresholds[found] then
+        if total > 0 and hit*100/total >= coverage_thresholds[found] then
```
Frenzie, line 81: Isn't this one the same?
https://github.com/koreader/koreader-base/blob/6232b2604873f1aa2b847f95d0364d6f90aa1d3a/ffi/harfbuzz.lua#L63-L69

tachibana-shin: yeah i fixed thanks

PR #2457, opened 2026-07-21
fix(mupdf): update annotations and fix BGR check

tachibana-shin:
- Synthesize appearance stream for desktop viewers
- Fix incorrect nil check on BGR pixmap conversion
- Drop original pixmap on conversion error

CHanges 1: `pdf_update_annot (line 808-810)` After setting the quadpoints, color, and opacity, you need to call `pdf_update_annot` so that MuPDF can composite the `/Rect` and `/AP` appearance streams. Without this line, desktop PDF viewers (Acrobat, Evince, etc.) will not be able to render the annotations

Change 2: bgr_pixmap == nil (line 993) The original code checks if pixmap == nil — wrong variable check bgr_pixmap because that is the result of mupdf_convert_pixmap. Add `fz_drop_pixmap(ctx, pixmap)` before merror to free the original pixmap when the conversion fails, to prevent a memory leak

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2457)


ffi/mupdf.lua
```diff
@@ -986,7 +990,8 @@ function mupdf.renderImage(data, size, width, height)
     local bb
     if mupdf.bgr and ncomp >= 3 then
         local bgr_pixmap = W.mupdf_convert_pixmap(ctx, pixmap, M.fz_device_bgr(ctx), nil, nil, M.fz_default_color_params, (ncomp == 4 and 1 or 0))
-        if pixmap == nil then
+        if bgr_pixmap == nil then
+            M.fz_drop_pixmap(ctx, pixmap)
             merror(ctx, "could not convert pixmap to BGR")
         end
         M.fz_drop_pixmap(ctx, pixmap)
```
benoit-pierre: Better changed to:
```lua
        M.fz_drop_pixmap(ctx, pixmap)
        if bgr_pixmap == nil then
            merror(ctx, "could not convert pixmap to BGR")
        end

        local p = M.fz_pixmap_samples(ctx, bgr_pixmap)
```
(like above)

ffi/mupdf.lua
```diff
@@ -805,6 +805,10 @@ function page_mt.__index:addMarkupAnnotation(points, n, type, bb_color)
     ok = W.mupdf_pdf_set_annot_opacity(self.ctx, annot, alpha)
     if not ok then merror(self.ctx, "could not set annotation opacity") end
 
+    -- Synthesize /Rect and /AP appearance stream, needed for visibility in desktop PDF viewers
+    ok = W.mupdf_pdf_update_annot(self.ctx, annot)
+    if not ok then merror(self.ctx, "could not update markup annotation") end
+
```
benoit-pierre, line 811: Doesn't this need to be called in `deleteAnnotation` annotation too?

benoit-pierre:
> * Synthesize appearance stream for desktop viewers
> 
> * Fix incorrect nil check on BGR pixmap conversion
> 
> * Drop original pixmap on conversion error

This would have been better as 2 separate commits.

Also, the PDF test files need to be updated: run the tests, and update `spec/unit/data/simple-out-xxx.pdf`  with the results from the run (in `build/x86_64-pc-linux-gnu-debug/spec/run/base_mupdf/`). After making sure visually the results are correct, of course. :P

benoit-pierre:
@tomasAlabes: ping!

ffi/mupdf.lua
benoit-pierre: Or better yet:
```lua
    -- Handle RGB->BGR conversion for Kobos when needed
    local bb
    if mupdf.bgr and ncomp >= 3 then
        local bgr_pixmap = W.mupdf_convert_pixmap(ctx, pixmap, M.fz_device_bgr(ctx), nil, nil, M.fz_default_color_params, (ncomp == 4 and 1 or 0))
        M.fz_drop_pixmap(ctx, pixmap)
        if bgr_pixmap == nil then
            merror(ctx, "could not convert pixmap to BGR")
        end
        pixmap = bgr_pixmap
    end
    local p = M.fz_pixmap_samples(ctx, pixmap)
    bb = BlitBuffer.new(p_width, p_height, bbtype, p):copy()
    M.fz_drop_pixmap(ctx, pixmap)
    return bb
```

benoit-pierre:
OK, I took the liberty to split the commit, apply my suggestions to the `renderImage` fix, and update the test PDFs.

I can confirm that with this change, annotations show correctly in the external viewers I tested (evince, okular, xpdf).

benoit-pierre (approved):

benoit-pierre:
To be marked as closed on bump: koreader/koreader#5010 and koreader/koreader#12186.

tachibana-shin:
thanks for update. I’d like to ask a somewhat personal question: Why do so many people like to split diffs that address the same issue into multiple commits? Personally, I’d prefer to combine them into one

benoit-pierre:
But they're not the same issue. It's 2 separate bugs (in the same subsystem).

benoit-pierre:
And smaller commits, as a rule of thumb, mean a better history: more convenient for blaming, running git bisect, or just reviewing the changes.

tachibana-shin:
Thanks for the answer. I'm feeling frustrated. I'm looking for a way to help my Kindle read EPUB and CBZ files faster, but I keep coming across unrelated bug Maybe I'm just having bad luck with EPUB.

PR #2458, opened 2026-07-21
posix: rework read / write API

tachibana-shin:
- Loop to handle partial reads and writes
- Prevent errors on incomplete I/O operations
- Return total bytes successfully processed

My plugin occasionally encounters memory-reading errors (I'm not sure if this is caused by this module, so I've adjusted it to comply with `IEEE Std 1003.1`):
- https://pubs.opengroup.org/onlinepubs/9699919799/functions/read.html
- https://pubs.opengroup.org/onlinepubs/9699919799/functions/write.html

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2458)

### DIFF
--- a/ffi/posix.lua
+++ b/ffi/posix.lua
@@ -31,35 +31,43 @@ function posix.open(path, flags, mode)
 end
 
 function posix.read(fd, ptr, len)
-    local total = 0
-    while total < len do
-        local ret = C.read(fd, ptr + total, len - total)
-        if ret < 0 then
-            error("read: "..strerror())
-        end
+    ptr = ffi.cast("uint8_t *", ptr)
+    local count = 0
+    while count < len do
+        local ret = C.read(fd, ptr + count, len - count)
         if ret == 0 then
-            break
+            error("read: "..string.format("short read, %u/%u", count, len))
         end
-        total = total + ret
+        if ret < 0 then
+            if ffi.errno() ~= C.EINTR then
+                error("read: "..strerror())
+            end
+        else
+            count = count + ret
+        end
     end
-    return total
+    return len
 end
 
 posix.strerror = strerror
 
 function posix.write(fd, ptr, len)
-    local total = 0
-    while total < len do
-        local ret = C.write(fd, ptr + total, len - total)
-        if ret < 0 then
-            error("write: "..strerror())
-        end
+    ptr = ffi.cast("uint8_t *", ptr)
+    local count = 0
+    while count < len do
+        local ret = C.write(fd, ptr + count, len - count)
         if ret == 0 then
-            break
+            error("write: "..string.format("short write, %u/%u", count, len))
         end
-        total = total + ret
+        if ret < 0 then
+            if ffi.errno() ~= C.EINTR then
+                error("write: "..strerror())
+            end
+        else
+            count = count + ret
+        end
     end
-    return total
+    return len
 end
 
 return posix
--- a/ffi/posix.lua
+++ b/ffi/posix.lua
@@ -31,18 +31,14 @@ function posix.open(path, flags, mode)
 end
 
 function posix.read(fd, ptr, len, deny_short)
-    if not deny_short then
-        local ret = C.read(fd, ptr, len)
-        if ret ~= len then
-            error("read: "..(ret < 0 and strerror() or string.format("short read, %u/%u", ret, len)))
-        end
-        return ret
-    end
     ptr = ffi.cast("uint8_t *", ptr)
     local count = 0
     while count < len do
         local ret = C.read(fd, ptr + count, len - count)
         if ret == 0 then
+            if deny_short then
+                error("read: "..string.format("short read, %u/%u", count, len))
+            end
             break
         end
         if ret < 0 then
@@ -59,18 +55,14 @@ end
 posix.strerror = strerror
 
 function posix.write(fd, ptr, len, deny_short)
-    if not deny_short then
-        local ret = C.write(fd, ptr, len)
-        if ret ~= len then
-            error("write: "..(ret < 0 and strerror() or string.format("short write, %u/%u", ret, len)))
-        end
-        return ret
-    end
     ptr = ffi.cast("uint8_t *", ptr)
     local count = 0
     while count < len do
         local ret = C.write(fd, ptr + count, len - count)
         if ret == 0 then
+            if deny_short then
+                error("write: "..string.format("short write, %u/%u", count, len))
+            end
             break
         end
         if ret < 0 then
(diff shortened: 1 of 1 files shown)


benoit-pierre:
> My plugin occasionally encounters memory-reading errors (I'm not sure if this is caused by this module […]

But are you using it? Because no KOReader code is currently using those read / write methods (the new upcoming OTA mechanism will).

You also changing the semantics of `posix.read` / `posix.write`: no error is raised on short read / write.

Additionally, incrementing `ptr` may not do the right thing depending on the type (e.g. `uint32_t *`).

Proposed alternative:
```lua
function posix.read(fd, ptr, len)
    ptr = ffi.cast("uint8_t *", ptr)
    local count = 0
    while count < len do
        local ret = C.read(fd, ptr + count, len - count)
        if ret == 0 then
            error("read: "..string.format("short read, %u/%u", count, len))
        end
        if ret < 0 then
            if ffi.errno() ~= C.EINTR then
                error("read: "..strerror())
            end
        else
            count = count + ret
        end
    end
    return len
end

function posix.write(fd, ptr, len)
    ptr = ffi.cast("uint8_t *", ptr)
    local count = 0
    while count < len do
        local ret = C.write(fd, ptr + count, len - count)
        if ret == 0 then
            error("write: "..string.format("short write, %u/%u", count, len))
        end
        if ret < 0 then
            if ffi.errno() ~= C.EINTR then
                error("write: "..strerror())
            end
        else
            count = count + ret
        end
    end
    return len
end
```

tachibana-shin:
Yes, a (private) plugin is using it. At first, I thought this module was standard POSIX so I used it, but it turned out not to be.

benoit-pierre:
So you're not using anymore? And you were seeing short reads / writes or `EINTR` errors?

tachibana-shin:
I'm also concerned about this because changes made in accordance with the rules aren't always correct (like the recent Node.js release issue). Since I've checked and found that this function is hardly used anywhere, I think it's safe to change it to comply with a standard

tachibana-shin:
> So you're not using anymore? And you were seeing short reads / writes or `EINTR` errors?

I'm thinking that if your suggestion is to keep the old code as is, I'll change my code

tachibana-shin:
> So you're not using anymore? And you were seeing short reads / writes or `EINTR` errors?

Yes, I run into this from time to time. In my opinion, it’s because the data hasn’t been fully recorded yet

benoit-pierre:
My suggestion is the code above, because your patch has 2 issues:
1. `ptr` may be incorrectly incremented (if pointing to a type whose size is not 1)
2. read / write can return `EINTR` if nothing was read / written before being interrupted

As for the difference in short read / write handling, are you relying on it? In what case?

benoit-pierre:
Are you using non-blocking FDs?

tachibana-shin:
> My suggestion is the code above, because your patch has 2 issues:
> 
> 1. `ptr` may be incorrectly incremented (if pointing to a type whose size is not 1)
> 2. read / write can return `EINTR` if nothing was read / written before being interrupted
> 
> As for the difference in short read / write handling, are you relying on it? In what case?



> Are you using non-blocking FDs?

yeah my plugin uses non-blocking FDs/sockets

benoit-pierre:
Well, then you're better of going through the FFI calls directly.

benoit-pierre:
Or we add an optional boolean flag to allow short reads / writes.

tachibana-shin:
I added the `deny_short` flag; by default, it will run the old code, otherwise it will run the new code. I also replaced the `error` with a `break` because the IEEE standard allows for missing bytes

benoit-pierre:
I'm confused, so you don't want `EINTR` to be handled automatically? What's the point of using this module? You're going to have to catch errors anyway to handle an `EINTR` when no bytes was read / written before interruption…

How about this instead:
```lua
function posix.read(fd, ptr, len, deny_short)
    ptr = ffi.cast("uint8_t *", ptr)
    local count = 0
    while count < len do
        local ret = C.read(fd, ptr + count, len - count)
        if ret == 0 then
            if deny_short then
                error("read: "..string.format("short read, %u/%u", count, len))
            end
            break
        end
        if ret < 0 then
            if ffi.errno() ~= C.EINTR then
                error("read: "..strerror())
            end
        else
            count = count + ret
        end
    end
    return count
end

function posix.write(fd, ptr, len)
    ptr = ffi.cast("uint8_t *", ptr)
    local count = 0
    while count < len do
        local ret = C.write(fd, ptr + count, len - count)
        if ret == 0 then
            if deny_short then
                error("write: "..string.format("short write, %u/%u", ret, len))
            end
            break
        end
        if ret < 0 then
            if ffi.errno() ~= C.EINTR then
                error("write: "..strerror())
            end
        else
            count = count + ret
        end
    end
    return count
end
```
?

So `EINTR` is automatically handled, and short reads / writes allowed by default (but the code still tries to read / write as much as possible, only stop on buffer / disk full). And I'll amend the OTA code to use `deny_short`.

tachibana-shin:
I realized that I would want to read/write continuously until I had enough data, rather than using a catch block, because even if an error occurred, I would still continue to write one more time

tachibana-shin:
in my opinion, if someone doesn't need this, the code will simply call `C.read` directly as before, without incurring any additional overhead (perhaps I'm a bit obsessed with changes that add overhead, even though no bugs have been reported yet)

tachibana-shin:
I think it would waste an additional 24 bytes if the deny_short flag isn't enabled

benoit-pierre:
> I think it would waste an additional 24 bytes if the deny_short flag isn't enabled

? I don't understand…

tachibana-shin:
> > I think it would waste an additional 24 bytes if the deny_short flag isn't enabled
> 
> ? I don't understand…

What I mean is that if the `deny_short` flag isn't set, `ffi.cast`, `count`, and `ret` become useless; they each take up 8 bytes, for a total of about 24 bytes I'm a little obsessed with them maybe it's OCD.

benoit-pierre:
Yes, you're being ridiculous…

benoit-pierre:
And they're not useless, I don't understand why you think that.

tachibana-shin:
> And they're not useless, I don't understand why you think that.

You're right maybe I just want to keep the original code as an option We should let this issue go

benoit-pierre:
No, this is a good change: handle `EINTR` automatically, and an API closer to native when not using `deny_short`, so it can be used with non-blocking fds for example.

benoit-pierre (approved):

Frenzie:
Can you provide an updated commit message to squash the commits with?

benoit-pierre:
@Frenzie:

posix: rework read / write API

- automatically handle `EINTR` errors
- allow short reads / writes by default, with an extra `deny_short` parameter to get the previous behavior (error on short operation)

PR #2459, opened 2026-07-21
djvu: guard against NULL pixelformat in setColorRendering

tachibana-shin:
…

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2459)

### DIFF
--- a/djvu.c
+++ b/djvu.c
@@ -214,6 +214,9 @@ static int setColorRendering(lua_State *L) {
 		doc->pixelsize = 1;
 		doc->pixelformat = ddjvu_format_create(DDJVU_FORMAT_GREY8, 0, NULL);
 	}
+	if (! doc->pixelformat) {
+		return luaL_error(L, "cannot create DjVu pixelformat");
+	}
 	ddjvu_format_set_row_order(doc->pixelformat, 1);
 	ddjvu_format_set_y_direction(doc->pixelformat, 1);
 	return 0;


tachibana-shin:
Also, I noticed something else: in `nnsvg.c`, on line 58, the `nsvgParse` function modifies the input by altering `\0`. Should we copy the input to prevent it from modifying the original data?

djvu.c
```diff
@@ -167,6 +167,10 @@ static int openDocument(lua_State *L) {
 		doc->pixelformat = ddjvu_format_create(DDJVU_FORMAT_GREY8, 0, NULL);
 	}
 	if (! doc->pixelformat) {
+		ddjvu_document_release(doc->doc_ref);
+		doc->doc_ref = NULL;
+		ddjvu_context_release(doc->context);
+		doc->context = NULL;
```
Frenzie: __gc will call closeDocument, so there's no leak here.

Frenzie: Mind, that's mainly a remark about the commit message. I'm not attached to waiting until gc gets around to it or anything.

djvu.c
```diff
@@ -214,6 +218,9 @@ static int setColorRendering(lua_State *L) {
 		doc->pixelsize = 1;
 		doc->pixelformat = ddjvu_format_create(DDJVU_FORMAT_GREY8, 0, NULL);
 	}
+	if (! doc->pixelformat) {
+		return luaL_error(L, "cannot create DjVu pixelformat");
+	}
```
Frenzie, line 219: That being said, I'd still prefer to just stick to this, or at least to spilt it up into two commits.

Frenzie:
> Also, I noticed something else: in `nnsvg.c`, on line 58, the `nsvgParse` function modifies the input by altering `\0`. Should we copy the input to prevent it from modifying the original data?

Pinging @poire-z

benoit-pierre:
FWIW, we always call `nnsvg_new` with a filename. But yes, `nsvgParse` does modify its input, so we should pass it a modifiable copy of the input string.

djvu.c
tachibana-shin: i deleted them. thanks

poire-z:
nnsvg.c is some old work from me, not familiar with it anymore, trusting @benoit-pierre .

tachibana-shin:
So that's how it is. I remember that in the past I had trouble displaying an inline SVG; I guess that might be the reason thanks @poire-z

PR #2464, opened 2026-07-22
Preserve screen buffer before adjusting its colors

### LINKED ISSUE #12527: Low quality image rendering artifacts
* KOReader version: 2024.07-100
* Device: Pocketbook Inkpad Color 3

#### Issue

For some low quality images, KOReader is rendering them with the rasterization artifacts. 

See comparison below:
| PB software | KOReader |
|--------|--------|
|  |  |

I tried to find any quality related settings in [`RenderImage:renderImageDataWithMupdf`](https://github.com/koreader/koreader/blob/03aa96dc7dc25c8d58977f0165630af7e4514891/frontend/ui/renderimage.lua#L88C10-L88C46) or [`mupdf.renderImage`](https://github.com/koreader/koreader-base/blob/9e16a13a6dfe66dcb6ac3ed0b4b57b4355c01f9e/ffi/mupdf.lua#L788), but alas.

Image Scaling is set to Best.

#### Steps to reproduce

Here's a sample ePub which can reproduce this issue: 
[Article.epub.zip](https://github.com/user-attachments/files/17032950/iOS.18s.new.home.screen.features.are.a.long.awaited.win.for.flexibility.epub.zip)

The image in question is located before the words "You can replace those controls with anything from this control picker".

dmalinovsky:
`inkview.adjustAreaDefault` updates buffer in-place, with additional refreshes mangling content. We need to restore the original buffer to make sure it won't be adjusted twice.

Also, this change prevents extra posterization artifacts we had with the low quality images before.

Fixes https://github.com/koreader/koreader/issues/12527, https://github.com/koreader/koreader/issues/11865, https://github.com/koreader/koreader/issues/11233

Tested on Pocketbook Inkpad Color 3, as well as b&w Pocketbook Era Lite to check for possible regressions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2464)

### DIFF
--- a/ffi/framebuffer_pocketbook.lua
+++ b/ffi/framebuffer_pocketbook.lua
@@ -28,46 +28,57 @@ local function _updatePartial(fb, x, y, w, h, dither, hq)
 
     fb.debug("refresh: inkview partial", x, y, w, h, dither)
 
-    if fb.device.hasColorScreen() and hq then
+    if fb.device.hasColorScreen() and dither then
         -- inkview.adjustAreaDefault updates buffer in-place, with additional refreshes mangling content.
         -- We need to restore the original buffer to make sure it won't be adjusted twice.
         fb:saveCurrentBB()
         _adjustAreaColours(fb)
-        
+    end
+
+    if fb.device.hasColorScreen() and hq then        
         inkview.PartialUpdateHQ(x, y, w, h)
-        fb:restoreFromSavedBB()
     else
         inkview.PartialUpdate(x, y, w, h)
+    end
+
+    if dither then
+        fb:restoreFromSavedBB()
     end
 end
 
 local function _updateFull(fb, x, y, w, h, dither)
     fb.debug("refresh: inkview full", x, y, w, h, dither)
 
-    if fb.device.hasColorScreen() then
+    if fb.device.hasColorScreen() and dither then
         fb:saveCurrentBB()
+    end
+    
+    if fb.device.hasColorScreen() then
         _adjustAreaColours(fb)
         
         inkview.FullUpdateHQ()
-        fb:restoreFromSavedBB()
     else
         inkview.FullUpdate()
+    end
+
+    if fb.device.hasColorScreen() and dither then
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
Ready to go?

dmalinovsky:
No more changes on my end :)

NiLuJe:
nit: Assuming there are no other callers, I'd just fold the save inside the `_adjustColor` funcs, and restore unconditionally at the tail end of the refresh wrappers, restoreBB already has a nil guard, so it's already a NOP if need be, no need to guard and branch twice.

(And I don't think there are *any* other callers of the save/restore thingy, in fact, I have no idea what it was originally written for, because it's ooooooold xD).

dmalinovsky:
> nit: Assuming there are no other callers, I'd just fold the save inside the `_adjustColor` funcs, and restore unconditionally at the tail end of the refresh wrappers, restoreBB already has a nil guard, so it's already a NOP if need be, no need to guard and branch twice.

Makes sense, I thought about that but then decided it's better to keep save/restore logic in the same func for future understanding. It's not as DRY, though.

> (And I don't think there are _any_ other callers of the save/restore thingy, in fact, I have no idea what it was originally written for, because it's ooooooold xD).

I stumbled upon these functions by accident while studying `framebuffer.lua` code to find a way to copy the screen buffer. It seems to be working well.

NiLuJe:
> Makes sense, I thought about that but then decided it's better to keep save/restore logic in the same func for future understanding.

I would generally agree, but given the lack of other callers, I'm torn between keeping the logic explicit and keeping it DRY ;p. Your call, though, this is a fairly personal stylistic nit.

> I stumbled upon these functions by accident while studying `framebuffer.lua` code to find a way to copy the screen buffer. It seems to be working well.

Yeah, the implementation is trivial. I was just curious because they didn't ring a bell, and was thinking you had implemented them in a previous PR, so I checked the blame, and, nooope, it dates back to pre-KOReader xD.

PR #2466, opened 2026-07-23
fix(input): reap fake ev process without blocking

tachibana-shin:
`closeAllInputDevices()` used `waitpid(-1, NULL, 0)` to reap the fake event
generator child process. This has two problems:

- `waitpid(-1, ...)` waits for *any* child
   process, not just the fake event generator. If other child processes are
   running (e.g., Wi-Fi scripts, USB mount helpers), they may be reaped
   instead — leaving the fake generator as a zombie, or causing the unrelated
   script's parent to lose track of it

= The fake generator may have already exited
   (via `PR_SET_PDEATHSIG` when the parent dies, or naturally). In that case,
   `waitpid(-1, NULL, 0)` blocks until a *different* child exits, causing
   KOReader to hang on shutdown

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2466)

### DIFF
--- a/input/input.c
+++ b/input/input.c
@@ -295,7 +295,7 @@ static int closeAllInputDevices(lua_State* L __attribute__((unused)))
     if (fake_ev_generator_pid != -1) {
         // Kill and wait to reap our child process.
         kill(fake_ev_generator_pid, SIGTERM);
-        waitpid(-1, NULL, 0);
+        waitpid(fake_ev_generator_pid, NULL, 0);
         fake_ev_generator_pid = -1;
     }
 


Frenzie (approved):

input/input.c
```diff
@@ -295,7 +295,7 @@ static int closeAllInputDevices(lua_State* L __attribute__((unused)))
     if (fake_ev_generator_pid != -1) {
         // Kill and wait to reap our child process.
         kill(fake_ev_generator_pid, SIGTERM);
-        waitpid(-1, NULL, 0);
+        waitpid(fake_ev_generator_pid, NULL, WNOHANG);
```
benoit-pierre: But now that we do wait on the right process only, we don't want to use `WNOHANG`, do we?

tachibana-shin: This is the final cleanup—`closeAllInputDevices` closes all file descriptors and clears all timers; it is the final cleanup function. The program will exit immediately afterward → `init` will reap all zombie processes, so I don’t see any reason to wait for it. Conversely, if you call `kill(SIGTERM)` and then `waitpid` with 0, it may still block if the process is slow to accept the signal or needs time to shut down

tachibana-shin: Or we could set it to 0 to be on the safe side and avoid any strange side effects 👀 I’d like to hear your thoughts

benoit-pierre: So the process is properly reaped?

benoit-pierre: It's just cleaner, IMHO.

tachibana-shin: No. With WNOHANG, it's not guaranteed.
- If the child has already exited by the time waitpid is called → reaped 
- If the child hasn't exited yet → returns 0, zombie is not reaped 
However, since this is final cleanup before program exit, init will reap it regardless. So in practice i think it's fine

tachibana-shin: Maybe I'll reset it to 0 to prevent something strange from happening

benoit-pierre: I think we should either do it properly (`waitpid(…, 0)`), or there's no point in calling `waitpid` and we let init do the cleanup.

First option is still cleaner, and does not assume users of the API only call `closeAllInputDevices` on exit.

tachibana-shin: reseted

benoit-pierre:
Can you fix the commit message? Something like: 
```
input: fix reaping of fake ev child process

Wait on the actual process pid, not on any child process.
```

tachibana-shin:
> Can you fix the commit message? Something like:
> 
> ```
> input: fix reaping of fake ev child process
> 
> Wait on the actual process pid, not on any child process.
> ```

yeah done

benoit-pierre (approved):

tachibana-shin:
@Frenzie sorry i have a side question: Is the Koreader Wi-Fi list pop-up on the Kindle broken? I’ve only seen it once—due to a crash—and I haven’t seen it appear since

Frenzie:
@mergen3107?

mergen3107:
Yeah, I never got to the bottom of this. It is kinda broken, but @notmarek says it works as it should. 

For me it sometimes fails to rejoin the wi-fi network after I enable wi-fi in koreader - and shows this popup. But it never worked for me to find a new network

tachibana-shin:
> Yeah, I never got to the bottom of this. It is kinda broken, but @notmarek says it works as it should.
> 
> For me it sometimes fails to rejoin the wi-fi network after I enable wi-fi in koreader - and shows this popup. But it never worked for me to find a new network

Who can I discuss this with, and where? When I call the `NetworkMgr:getNetworkList` function, it always returns an empty array. Is there a specific condition—such as the Wi-Fi not being connected—that would cause this function to return something?

Frenzie:
> For me it sometimes fails to rejoin the wi-fi network after I enable wi-fi in koreader - and shows this popup. But it never worked for me to find a new network

If you're simply enabling the network it's not supposed to show a popup most of the time (except with long-press; that's an open issue to be resolved into something more intuitive).

> Who can I discuss this with, and where? When I call the NetworkMgr:getNetworkList function, it always returns an empty array. Is there a specific condition—such as the Wi-Fi not being connected—that would cause this function to return something?

Perhaps I misunderstood your question as being Kindle-specific. WpaSupplicant only fires scans in order to find something to connect to.

tachibana-shin:
> Perhaps I misunderstood your question as being Kindle-specific. WpaSupplicant only fires scans in order to find something to connect to.

My situation is like this: I have a saved network but can't connect (router is dead). KoreaReader reports that it can't connect to the network. I press and hold on the Wi-Fi icon: it does nothing and sometimes even turns off the Wi-Fi.

There's another scenario where there's absolutely no network around. I'm using my phone to create a 2.4GHz Wi-Fi hotspot. I press and hold on the Wi-Fi icon, and a blank popup appears (it only has two lines, but I think it's a popup with no content).

In short, whether connected or not, I never see it listed in the Wi-Fi network list

I'm asking about Kindle because I've seen people with this popup on Kobo, I did the same as them and I still don't see it.

Frenzie:
What's the wifi network name? There could potentially be a character encoding issue of some sort. But I just remembered (after a quick grep) that on Kindle it's actually done through lipc, not through our own control of WpaSupplicant. So with any luck improvements may be possible.

tachibana-shin:
I have the following Wi-Fi networks: "Yoyazuki" "Kawasaki Coffee" "Tachibana Shin" "Yamaguchi Takezu" and "Host Club 98" They all look normal

All are dual-band and comply with the ASE-WPA2 standard

Frenzie:
That all sounds like basic ASCII. Then it's something funky about lipc but you'll probably be able to deal with that most quickly yourself. See in `frontend/device/kindle/device.lua`.

mergen3107:
> If you're simply enabling the network it's not supposed to show a popup most of the time

So I enable wi-fi, it fails to connect to previous network, and shows me this popup, but it doesn't work if I press connect from there. I usually exit KOReader and rejoin in Kindle UI and open it again

tachibana-shin:
> That all sounds like basic ASCII. Then it's something funky about lipc but you'll probably be able to deal with that most quickly yourself. See in `frontend/device/kindle/device.lua`.

Thank you. It seems my device has some kind of issue with liblipclua; when I completely switched the code to run on the fork, the problem reappeared

tachibana-shin:
> So I enable wi-fi, it fails to connect to previous network, and shows me this popup, but it doesn't work if I press connect from there. I usually exit KOReader and rejoin in Kindle UI and open it again

I've encountered this as well, but I think it's because the Kindle takes a long time to wake up the Wi-Fi module. I keep trying to connect to the network, and it usually works on the second or third try

PR #2468, opened 2026-07-23
fix(blitbuffer): memory safety, soft-float overhead, and variable shadowing

tachibana-shin:
### Change 1: fix stride check in BB8/BB32 fast-path blit

`BB_blit_to_BB8` and `BB_blit_to_BB32` both have a fast-path that copies
contiguous scanlines in a single `memcpy` instead of looping per scanline.
This optimization kicks in when source and destination have the same dimensions
and stride — but the old code only verified `src->stride == dst->stride`,
not that stride actually equals the width.

When a BlitBuffer is allocated with alignment padding (common on e-ink
devices where the framebuffer row stride is wider than the pixel width),
stride > width and the scanlines are not contiguous in memory. The single
`memcpy` would then read across scanline boundaries into padding bytes,
corrupting the output.

Additionally, the BBRGB32 fast-path computed the copy size as
`(w << 2U) * h` — derived from the width, not the stride. If stride != width*4,
this copies the wrong amount of data. Changed to `src->stride * h`, which is
consistent with the per-scanline fallback path already using `stride` for
addressing.

### Change 2: fix rounded corner AA (r<=0 guard, double→long long, variable shadowing)

`BB_paint_rounded_corner_AA` is called on every page render that uses rounded
corners. Three issues were found:

**Degenerate parameters.** When `r <= 0`, the function computes
`o_diam = 2*r` which is zero or negative. Subsequent arithmetic (multiplications
for `dx`, `dy`, `err`, and the modulo operations in the drawing loops) produces
undefined behavior. Added an early return for `r <= 0 || w == 0 || h == 0`.

**Soft-float overhead.** Most variables in the Bresenham-style ellipse walker
(`a2`, `dx`, `dy`, `err`, `dx2`, `dy2`, `e2`, `a1`) were declared `double` but
only used for integer multiplication and comparison — never for floating-point
division. On ARM devices without hardware FPU (Kindle legacy, most e-ink readers),
every `double` operation traps into a software float emulation library, adding
~100 cycles per operation. For a function that runs per-pixel on every page
with rounded corners, this was a significant and unnecessary overhead. Changed
these variables to `long long` with explicit `LL` constant suffixes. Only `ed`,
which participates in the `sqrt`-approximation division, remains `double`.

**Variable shadowing.** Both inner `for(;;)` loops declared
`double i = MIN(dx, dy)`, which shadowed the outer `int i` used as the loop
counter and break-value carrier. After the inner loop exits via `break`, the
outer `i` was expected to hold a value set by `i = x0` or `i = x0+1` — but
the compiler could reuse the same stack slot for the shadowed `double i`,
silently corrupting the outer variable. This could cause the fill-line loop
to iterate over wrong bounds, or the ellipse walker to use an incorrect
starting coordinate. Renamed the inner variable to `blend_i` and added explicit
`(uint8_t)` casts for the alpha value computations (where `double` → `uint8_t`
truncation needs to be well-defined).

### Change 3: set alpha=0xff after BB8A blend, guard stripe_width==0

**BB8A alpha consistency.** `BB_blend_rect`'s BB8A case blended the `a` channel
but left the `alpha` field untouched. Every other BB8A write path in the codebase
(`BB8A_BLEND_PIXEL_CLAMPED`, `BB8A_SET_PIXEL_CLAMPED`, the rounded-corner AA
macros) explicitly sets `alpha = 0xff` after modifying a pixel. The inconsistency
meant BB8A pixels could retain stale alpha values after a blend, which affects
downstream compositing that reads the alpha field.

**Hatch stripe_width==0.** `BB_hatch_rect` computes `sw2 = stripe_width * 2`
and then uses `(d_x + d_y) % sw2` to determine stripe positions. When
`stripe_width == 0`, `sw2` is zero and the modulo operation is undefined
behavior (integer divide by zero). Added `stripe_width == 0` to the existing
early-return guard.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2468)

### DIFF
--- a/blitbuffer.c
+++ b/blitbuffer.c
@@ -1255,7 +1255,7 @@ void BB_blit_to_BB8(const BlitBuffer * restrict src, BlitBuffer * restrict dst,
             // (i.e., setPixel, no rota, no invert).
             // The cbb codepath ensures setPixel & no invert, so we only check for rotation.
             if (sbb_rotation == 0 && dbb_rotation == 0) {
-                if (offs_x == 0 && dest_x == 0 && w == src->w && w == dst->w && src->stride == dst->stride) {
+                if (offs_x == 0 && dest_x == 0 && w == src->w && w == dst->w && src->stride == dst->stride && src->stride == w) {
                     // Single step for contiguous scanlines (on both sides)
                     //fprintf(stdout, "%s: full copy blit from BB8 to BB8\n", __FUNCTION__);
                     // BB8 is 1 byte per pixel
@@ -1708,13 +1708,13 @@ void BB_blit_to_BB32(const BlitBuffer * restrict src, BlitBuffer * restrict dst,
             // (i.e., setPixel, no rota, no invert).
             // The cbb codepath ensures setPixel & no invert, so we only check for rotation.
             if (sbb_rotation == 0 && dbb_rotation == 0) {
-                if (offs_x == 0 && dest_x == 0 && w == src->w && w == dst->w && src->stride == dst->stride) {
+                if (offs_x == 0 && dest_x == 0 && w == src->w && w == dst->w && src->stride == dst->stride && src->stride == (w << 2U)) {
                     // Single step for contiguous scanlines (on both sides)
                     //fprintf(stdout, "%s: full copy blit from BBRGB32 to BBRGB32\n", __FUNCTION__);
                     // BBRGB32 is 4 bytes per pixel
                     const uint8_t * restrict srcp = src->data + src->stride*offs_y;
                     uint8_t * restrict dstp = dst->data + dst->stride*dest_y;
-                    memcpy(dstp, srcp, (w << 2U)*h);
+                    memcpy(dstp, srcp, src->stride*h);
                 } else {
                     // Scanline per scanline copy
                     //fprintf(stdout, "%s: scanline copy blit from BBRGB32 to BBRGB32\n", __FUNCTION__);
--- a/blitbuffer.c
+++ b/blitbuffer.c
@@ -3079,9 +3079,9 @@ void BB_color_blit_from_RGB32(BlitBuffer * restrict dst, const BlitBuffer * rest
 }
 
 // Information about those three algorithms can be found on http://members.chello.at/~easyfilter/ (Zingl Alois)
-void BB_paint_rounded_corner_noAA(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, int bw, int r, uint8_t c);
-void BB_paint_rounded_corner_AA(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, int bw, int r, uint8_t c);
-void BB_paint_rounded_corner_AA_1px(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, int r, uint8_t c);
+void BB_paint_rounded_corner_noAA(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, unsigned int bw, unsigned int r, uint8_t c);
+void BB_paint_rounded_corner_AA(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, unsigned int bw, unsigned int r, uint8_t c);
+void BB_paint_rounded_corner_AA_1px(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, unsigned int r, uint8_t c);
 
 void BB_paint_rounded_corner(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, unsigned int bw, unsigned int r, uint8_t c, int anti_aliasing) {
     /*
@@ -3107,7 +3107,7 @@ void BB_paint_rounded_corner(BlitBuffer * restrict bb, unsigned int off_x, unsig
     }
 }
 
-void BB_paint_rounded_corner_noAA(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, int bw, int r, uint8_t c) {
+void BB_paint_rounded_corner_noAA(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, unsigned int bw, unsigned int r, uint8_t c) {
 
     const int bb_type = GET_BB_TYPE(bb);
     const int bb_rotation = GET_BB_ROTATION(bb);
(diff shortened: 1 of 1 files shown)


Frenzie:
> On soft-float ARM (Kobo/Kindle)

We don't support the very oldest Kobo devices without hard float, ftr.

Edit: by which I mean, please remove that from the commit message. Maybe just remove the parenthetical remark; we know it's mainly about older Kindle anyway.

tachibana-shin:
> > On soft-float ARM (Kobo/Kindle)
> 
> We don't support the very oldest Kobo devices without hard float, ftr.
> 
> Edit: by which I mean, please remove that from the commit message. Maybe just remove the parenthetical remark; we know it's mainly about older Kindle anyway.

I fixed it

blitbuffer.c
```diff
@@ -3351,6 +3351,9 @@ void BB_paint_rounded_corner_AA_1px(BlitBuffer * restrict bb, unsigned int off_x
 
 void BB_paint_rounded_corner_AA(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, int bw, int r, uint8_t c )
 {        // draw a black anti-aliased circle with width bw
+    if (r <= 0 || w == 0 || h == 0) {
+        return;
+    }
     const int bb_type = GET_BB_TYPE(bb);
```
Frenzie, line 3355: I think this feels like noise. The only realistic case is == 0 and that's already covered in the Lua leading up to this . But there seems to be a typo in these `_rounded_corner_*` functions: `r` should be unsigned like it is in its parent, shouldn't it?

tachibana-shin: Is this necessary? There might be quite a few changes, for example, here:
```c
int x = -r; // unsigned r → wraps to huge → UB
int err = 2-2*r; // unsigned r → wraps
r = 1-err; // set signed value to unsigned
```

Frenzie: Neither radius nor border width will ever be passed as negative,[1] nor would it make sense to do so since a radius has a starting and an end point. Which combines with if you're diving deep into these minor details shouldn't it simply match the parent, which already uses an unsigned int anyway?

NB Not a rhetorical question: there might certainly be valid approaches I'm overlooking, but a lack of consistency doesn't strike me as a good thing unless there's a pretty good reason for it.

> There might be quite a few changes

This is our library. While of course diffs shouldn't be unnecessarily large, what I said on k2pdfopt about keeping them as minimal as possible is only about patches. Everything in base and front is 100% ours unless clearly marked as being taken from somewhere.

[1] Though it looks kind of awesome if you try. Of course it'll already have turned positive along the way long before reaching this function. It's simply `r=412` here, which ultimately came out of an input of `-24` plus some DPI sizing. I don't think blocking that from happening would be particularly helpful since then you wouldn't see that you messed up.
Screenshot_20260724_003318

tachibana-shin: I've made the correction, and I've also added "unsigned" for bw since they're quite similar

ahmadadam96:
Hi, change 1 actually fixes an issue I was investigating and having with my own pull request (https://github.com/koreader/koreader/pull/15752/). The issue was that the screen wasn't drawing fully when Screen.bb:blitFrom() is used, leaving a small bottom part of the screen unchanged. I narrowed it to this same issue.

I applied a rough fix locally by applying a similar thing to ffi/blitbuffer.lua:

In blitTo8 (BB8): `width*height → self.stride*height`
In blitToRGB32 (BBRGB32): `lshift(width, 2)*height → self.stride*height
`
And then found this pull request, which was addressing it for the c file at least.

Frenzie:
I somewhat overlooked that, but we do like to keep the implementations in sync. :+1:

Frenzie:
It's fixed by reverting that unsigned stuff, without which the inner loop always fails. Overflow is probably of no concern since in C, `r` clamps itself to the buffer size and the Lua side doesn't leave it through if `r` is too large anyway. For future reference, breaking visuals counts a pretty good reason to leave it alone. ;-) I just want to know why you did or didn't do something; that doesn't necessarily mean any particular (coding) result.

```diff
diff --git a/blitbuffer.c b/blitbuffer.c
index 124da48a..3db61a95 100644
--- a/blitbuffer.c
+++ b/blitbuffer.c
@@ -3080,9 +3080,9 @@ void BB_color_blit_from_RGB32(BlitBuffer * restrict dst, const BlitBuffer * rest
 }
 
 // Information about those three algorithms can be found on http://members.chello.at/~easyfilter/ (Zingl Alois)
-void BB_paint_rounded_corner_noAA(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, unsigned int bw, unsigned int r, uint8_t c);
-void BB_paint_rounded_corner_AA(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, unsigned int bw, unsigned int r, uint8_t c);
-void BB_paint_rounded_corner_AA_1px(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, unsigned int r, uint8_t c);
+void BB_paint_rounded_corner_noAA(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, int bw, int r, uint8_t c);
+void BB_paint_rounded_corner_AA(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, int bw, int r, uint8_t c);
+void BB_paint_rounded_corner_AA_1px(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, int r, uint8_t c);
 
 void BB_paint_rounded_corner(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, unsigned int bw, unsigned int r, uint8_t c, int anti_aliasing) {
     /*
@@ -3108,7 +3108,7 @@ void BB_paint_rounded_corner(BlitBuffer * restrict bb, unsigned int off_x, unsig
     }
 }
 
-void BB_paint_rounded_corner_noAA(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, unsigned int bw, unsigned int r, uint8_t c) {
+void BB_paint_rounded_corner_noAA(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, int bw, int r, uint8_t c) {
 
     const int bb_type = GET_BB_TYPE(bb);
     const int bb_rotation = GET_BB_ROTATION(bb);
@@ -3270,7 +3270,7 @@ void BB_paint_rounded_corner_noAA(BlitBuffer * restrict bb, unsigned int off_x,
     setPixelAA_BBRGB32(off_x+r+x0,    off_y+h-r-1+y0); \
     setPixelAA_BBRGB32(off_x+w-r-1+x1,off_y+h-r-1+y0);
 
-void BB_paint_rounded_corner_AA_1px(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, unsigned int r, uint8_t c) { // draw a black anti-aliased circle with thickness 1
+void BB_paint_rounded_corner_AA_1px(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, int r, uint8_t c) { // draw a black anti-aliased circle with thickness 1
     const int bb_type = GET_BB_TYPE(bb);
     const int bb_rotation = GET_BB_ROTATION(bb);
     const unsigned int bb_width = BB_GET_WIDTH(bb);
@@ -3350,7 +3350,7 @@ void BB_paint_rounded_corner_AA_1px(BlitBuffer * restrict bb, unsigned int off_x
     } while (x < 0);
 }
 
-void BB_paint_rounded_corner_AA(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, unsigned int bw, unsigned int r, uint8_t c )
+void BB_paint_rounded_corner_AA(BlitBuffer * restrict bb, unsigned int off_x, unsigned int off_y, unsigned int w, unsigned int h, int bw, int r, uint8_t c )
 {        // draw a black anti-aliased circle with width bw
     const int bb_type = GET_BB_TYPE(bb);
     const int bb_rotation = GET_BB_ROTATION(bb);
```

tachibana-shin:
Lmao, I added unsigned as you suggested, and it seems that was a mistake

PR #2469, opened 2026-07-23
fix(djvu): validate page bounds and check errors

tachibana-shin:
- Add page bounds checks for DjVu functions
- Raise error if fetching page info fails

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2469)


tachibana-shin:
I'm confused. Why did Kobo fail?

Frenzie:
Looks more like a CircleCI-internal issue to me (which I don't think I've ever seen before). The build already succeeded; it was just something about retrieving or saving the build timings.

Frenzie (approved):

PR #2477, opened 2026-07-25
Add a LVContainerTransform for interaction between koreader and crengine

pawanjay176:
Related to https://github.com/koreader/koreader/issues/15649

Add a wrapper `LVContainerTransform` wrapper that allows koreader to provide replacement bytes for individual EPUB entries.

Exposes `loadEpubWithEntryTransform` function to be used in koreader. The transform object needs to provide the following interface:

```lua
{
  transformEntry = function(self, path) -- return replacement bytes, or nil to use the original entry
    end
}
```

Disclosure: Used AI assistance for the `openTransformedEntry` function. I don't have deep understanding of the lua C api.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2477)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -417,6 +417,116 @@ typedef struct CreDocument {
 	ldomDocument *dom_doc;
 } CreDocument;
 
+class LVContainerTransform : public LVContainer
+{
+    lua_State * _L;
+    int _transform_ref;
+    LVContainerRef _container;
+
+    LVStreamRef openTransformedEntry(lString32 path)
+    {
+        int top = lua_gettop(_L);
+        lua_rawgeti(_L, LUA_REGISTRYINDEX, _transform_ref);
+        lua_getfield(_L, -1, "transformEntry");
+        if (lua_isnil(_L, -1)) {
+            lua_settop(_L, top);
+            return LVStreamRef();
+        }
+
+        lua_insert(_L, -2);
+        lString8 path8 = UnicodeToUtf8(path);
+        lua_pushstring(_L, path8.c_str());
+        if (lua_pcall(_L, 2, 1, 0) != 0) {
+            CRLog::error("LVContainerTransform: transformEntry failed: %s", lua_tostring(_L, -1));
+            lua_settop(_L, top);
+            return LVStreamRef();
+        }
+        if (lua_isnil(_L, -1)) {
+            lua_settop(_L, top);
+            return LVStreamRef();
+        }
+
+        size_t data_size = 0;
+        const char * data = lua_tolstring(_L, -1, &data_size);
+        if (!data) {
+            CRLog::error("LVContainerTransform: transformEntry() must return nil or a byte string");
+            lua_settop(_L, top);
+            return LVStreamRef();
+        }
+
+        LVStreamRef stream = LVCreateMemoryStream((void*)data, (int)data_size, true);
+        if (!stream.isNull())
+            stream->SetName(path.c_str());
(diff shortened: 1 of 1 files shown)


poire-z:
Can you ask your AI to add more comments, especially for the Lua stack interactions ?
One good documentation of mine, and it is really needed to understand years after how it works, can be seen at:
https://github.com/koreader/koreader-base/blob/7aace426af8a156a2eb025056207c71fdab393a3/xtext.cpp#L536-L636

cre.cpp
```diff
@@ -417,6 +417,85 @@ typedef struct CreDocument {
...
+            return LVStreamRef();
+
+        if (mode == LVOM_READ) {
+            LVStreamRef transformed = openTransformedEntry(lString32(fname));
+            if (!transformed.isNull())
+                return transformed;
+        }
+
+        return _container->OpenStream(fname, mode);
+    }
```
poire-z, line 493: I'd like it (and similar) as condensed as:
```c
    virtual LVStreamRef OpenStream(const lChar32 * fname, lvopen_mode_t mode) {
        if (!fname)
            return LVStreamRef();
        if (mode == LVOM_READ) {
            LVStreamRef transformed = openTransformedEntry(lString32(fname));
            if (!transformed.isNull())
                return transformed;
        }
        return _container->OpenStream(fname, mode);
    }
```

cre.cpp
```diff
@@ -417,6 +417,85 @@ typedef struct CreDocument {
 	ldomDocument *dom_doc;
 } CreDocument;
 
+class LVContainerTransform : public LVContainer
```
poire-z, line 420: Add some introduction comment, what it is for, that it is not used by KOReader but available to plugins that would like to do xyz, what it allows, etc...

cre.cpp
```diff
@@ -704,6 +783,41 @@ static int loadDocument(lua_State *L) {
...
+static int loadEpubWithEntryTransform(lua_State *L) {
+	CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
+	const char *file_name = luaL_checkstring(L, 2);
+	luaL_checktype(L, 3, LUA_TTABLE);
+	bool only_metadata = false;
+	if (lua_isboolean(L, 4)) {
+		only_metadata = lua_toboolean(L, 4);
+	}
+
+	lString32 file_name32 = LocalToUnicode(lString8(file_name));
```
poire-z, line 795: Add some introduction comment, what it is for, that it is not used by KOReader but available to plugins that would like to do xyz, what it allows, etc...
Document (with a comment on each line checking one arg) what the provided arguments are expected to be (as it's not used by KOReader, hard to guess what this table at 3 should be).

cre.cpp
```diff
@@ -704,6 +783,41 @@ static int loadDocument(lua_State *L) {
...
+
+	LVContainerRef archive = LVOpenArchieve(stream);
+	if (archive.isNull()) {
+		lua_pushboolean(L, false);
+		return 1;
+	}
+
+	lua_pushvalue(L, 3);
+	int transform_ref = luaL_ref(L, LUA_REGISTRYINDEX);
+	LVContainerRef container(new LVContainerTransform(L, transform_ref, archive));
```
poire-z, line 810: Is that our arg 3 ? Explain what it is and how it will be used.
Really, without comments, I can't understand any of the magic :)

pawanjay176:
Apologies, got busy at my day job. Will update these as soon as I can!

PR #2480, opened 2026-07-29
netinfo: don't ask socket() for SOCK_CLOEXEC

iav:
`SOCK_CLOEXEC` needs Linux 2.6.27, so on older kernels `socket()` fails with `EINVAL`
and the SSID is never read: Network info reports `SSID: off/any` while connected.
Set the flag with `fcntl()` afterwards instead, as `Device:ping4` already does.

Kindle 3, kernel 2.6.26: `SSID: off/any` → `SSID: "ad33"`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2480)

### DIFF
--- a/ffi/netinfo.lua
+++ b/ffi/netinfo.lua
@@ -123,10 +123,12 @@ if ffi.os == "Linux" then -- {{{
 function NetInfo:_iface_ssid_ioctl(iface)
     local sd = self.sd
     if not sd then
-        sd = C.socket(C.AF_INET, bit.bor(C.SOCK_DGRAM, C.SOCK_CLOEXEC), C.IPPROTO_IP)
+        -- SOCK_CLOEXEC in socket() needs Linux 2.6.27; older kernels fail with EINVAL.
+        sd = C.socket(C.AF_INET, C.SOCK_DGRAM, C.IPPROTO_IP)
         if sd < 0 then
             error(string.format("socket(AF_INET) failed: %s", posix.strerror()))
         end
+        C.fcntl(sd, C.F_SETFD, ffi.cast("int", C.FD_CLOEXEC))
         self.sd = sd
     end
     local essid = ffi.new("char[?]", C.IW_ESSID_MAX_SIZE + 1)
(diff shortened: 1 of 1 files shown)


ffi/netinfo.lua
```diff
@@ -123,10 +123,12 @@ if ffi.os == "Linux" then -- {{{
 function NetInfo:_iface_ssid_ioctl(iface)
     local sd = self.sd
     if not sd then
-        sd = C.socket(C.AF_INET, bit.bor(C.SOCK_DGRAM, C.SOCK_CLOEXEC), C.IPPROTO_IP)
+        -- SOCK_CLOEXEC in socket() needs Linux 2.6.27; older kernels fail with EINVAL.
+        sd = C.socket(C.AF_INET, C.SOCK_DGRAM, C.IPPROTO_IP)
```
Frenzie, line 129: Potentially worth noting that it avoids race conditions.

https://man7.org/linux/man-pages/man2/open.2.html

ffi/netinfo.lua
```diff
@@ -123,10 +123,12 @@ if ffi.os == "Linux" then -- {{{
...
         if sd < 0 then
             error(string.format("socket(AF_INET) failed: %s", posix.strerror()))
         end
+        C.fcntl(sd, C.F_SETFD, ffi.cast("int", C.FD_CLOEXEC))
```
Frenzie: This should presumably check if it succeeded?

ffi/netinfo.lua
iav: Added: SOCK_CLOEXEC also sets the flag atomically, which fcntl cannot — another thread forking and exec'ing between socket() and fcntl() leaks the descriptor.

ffi/netinfo.lua
iav: It does now. Reported on stderr rather than through error(): a missing CLOEXEC only leaks the descriptor into a child process, the SSID read still works — same treatment as getifaddrs/getnameinfo above.

PR #2481, opened 2026-07-31
einkfb: refresh the screen even when its content has not changed

### LINKED ISSUE #12774: [Kindle NT] Full screen refresh doesn't work
* KOReader version: 2024.07-285-gfd3727274 (legacy)
* Device: Kindle 4 Non-Touch (FW 4.1.4)

#### Issue
Full screen refresh doesn't work, as reported here: https://github.com/koreader/koreader/pull/12484#issuecomment-2422787161

#### Steps to reproduce
Assign a hotkey to refresh the screen.
Use it to refresh - nothing happens.

Well, I found that it **_sometimes_** works, but cannot trace the lead.
See `crash.log` below.

##### `crash.log` (if applicable)
[crash_partial.log.txt](https://github.com/user-attachments/files/17851923/crash_partial.log.txt)
In this one, I did:
1) Use hotkey to refresh (1) - nothing
2) Jump back to previous chapter using Dpad left.
3) Use hotkey to refresh (2) - nothing
4) Use hotkey to refresh **again** (3) - it refreshed!
5) (outside `crash.log` Use hotkey to refresh n more time - nothing.

Although, logs looks exactly the same for all three refreshes above.
Need some more debugging...

iav:
The einkfb driver skips an update when every pixel matches what is already on screen, so
"Full screen refresh" on an unchanged screen did nothing. fx_update_slow is exempt from that
comparison and drives the panel at full fidelity.

Verified on a live Kindle 3: the refresh fires every time and clears ghosting; page turns and
UI flashes stay correct.

Fixes koreader/koreader#12774.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2481)

### DIFF
--- a/ffi/framebuffer_einkfb.lua
+++ b/ffi/framebuffer_einkfb.lua
@@ -26,7 +26,9 @@ function framebuffer:refreshPartialImp(x, y, w, h)
 end
 
 function framebuffer:refreshFullImp(x, y, w, h)
-    einkfb_update(self, C.fx_update_full, x, y, w, h)
+    -- The driver skips an update whose pixels all match what is already on screen, dropping
+    -- explicit full refreshes on an unchanged screen; fx_update_slow is exempt from that check.
+    einkfb_update(self, C.fx_update_slow, x, y, w, h)
 end
 
 return require("ffi/framebuffer_linux"):extend(framebuffer)


Frenzie:
I don't believe this is true on any of my devices. Pinging @NiLuJe who has a more complete mental overview.

iav:
Right — your devices wouldn't show it: the drop is a quirk of the legacy einkfb driver only
(K2/K3/K4/DX), and the changed module is only ever loaded on those. mxcfb devices are untouched
by this PR.

On a Kindle 3 (FW 3.4.3) the drop shows up in dmesg as `einkfb_print_buffers_equal_message`
whenever a full refresh hits an unchanged screen, and it reproduces outside KOReader too:
`fbink -fs` on a static screen flashes about 3 times out of 30.

Frenzie:
Hm, indeed.

Frenzie (approved):
It's self-contained so fine in that sense. Presumably slow has more of an overhead (since I believe that's intended for regions?) but then again we're not calling it every second.

iav:
Numbers from the K3: 0.156s for slow vs 0.111s for a normal full update. The extra ~45ms
doesn't show on a manual action, and it's off the hot path everywhere else.

NiLuJe:
Yep, I do remember that einkfb quirk ;).

Out of curiosity, because I don't remember how that affects it, does `FBIO_EINK_UPDATE_DISPLAY` + `fx_update_full` behave?

(Not that it matters here, as we do want to be able to flash areas and not the full screen ;))

NiLuJe (approved):

iav:
Checked on the K3 (FW 3.4.3): `FBIO_EINK_UPDATE_DISPLAY` + `fx_update_full` gets dropped the same way — 10 calls on an unchanged screen, 10× `einkfb_print_buffers_equal_message:def:full_screen(flashing)` in dmesg, zero flashes. The buffers-equal check sits below both ioctls. `fx_update_slow` through the full-display path is exempt as well: 5 calls, 5 flashes, dmesg silent.

PR #2482, opened 2026-07-31
Android: let the system handle power key events

hbast:
On a Tolino Shine 3 (firmware 16.2.0), short power-button presses were ignored by KOReader while working normally in the Tolino app.

`AKEYCODE_POWER` was consumed without being mapped to an action. Returning `0` lets Android handle it.

Tested with clean installations of KOReader v2026.07: the original build ignores the button, the patched build locks and wakes correctly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2482)

### DIFF
--- a/ffi/input_android.lua
+++ b/ffi/input_android.lua
@@ -179,17 +179,17 @@ local function keyEventHandler(key_event)
             return 0 -- event not consumed
         end
     elseif code == C.AKEYCODE_MEDIA_PLAY_PAUSE
-    or code == C.AKEYCODE_MEDIA_PLAY
-    or code == C.AKEYCODE_MEDIA_PAUSE
-    or code == C.AKEYCODE_MEDIA_STOP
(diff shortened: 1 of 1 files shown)


Frenzie:
That's odd; I don't believe that key can be captured. Which Android version is that?

hbast:
Android 4.4.2 (KitKat), with Tolino firmware 16.2.0

pazos:
that sounds like a broken system/rom

Frenzie:
Yes, I figured adding it couldn't hurt as long as there was a comment stating it's not needed, unless you disagree?

I think on those devices you have to install apps with adb and maybe other hacks, so it probably behaves a certain way on purpose for the stock app.

pazos:
Yeah, the workaround seems very safe for regular devices. It is just weird than a power event can be forwarded to and consumed by regular apps

PR #2484, opened 2026-08-03
ffi/downloader: fix range support check

benoit-pierre:
Also check for "Content-Range" header.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2484)

### DIFF
--- a/ffi/downloader.lua
+++ b/ffi/downloader.lua
@@ -45,17 +45,17 @@ function Downloader:fetch(url, callback, ranges, etag)
                 self.err = status_code
                 return false
             end
-            if not range_support_checked then
-                if resp_headers["accept-ranges"] ~= "bytes" then
-                    self.err = "server does not support range requests!"
-                    return false
-                end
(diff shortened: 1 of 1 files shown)


Frenzie:
How about something like this for the callback issue?

[response-headers-callback.patch](https://github.com/user-attachments/files/30671474/response-headers-callback.patch)

benoit-pierre:
> How about something like this for the callback issue?
> 
> [response-headers-callback.patch](https://github.com/user-attachments/files/30671474/response-headers-callback.patch)

OK, but I would change it to:
```diff
-local sink, abort = nreqt.response_headers(headers, code, status)
+local abort, sink = nreqt.response_headers(headers, code, status)
```

As replacing the sink is not going to be common.

Frenzie:
I'll leave it to you if you want to do anything with it. (But if you do, I suppose I'll bother upstream about it.)

PR #2485, opened 2026-08-03
ffi/downloader: improve fetch implementation

benoit-pierre:
Use new HTTP response headers callback to abort early in case of error.

Additionally, simplify check for range request support: only check the
status code, instead of headers (since Cloudflare response doesn't
include the expected "Accept-range" header).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2485)

### DIFF
--- a/ffi/downloader.lua
+++ b/ffi/downloader.lua
@@ -24,58 +24,48 @@ end
 
 function Downloader:fetch(url, callback, ranges, etag)
     assert(not (ranges and etag))
-    self.status_code = nil
-    self.etag = nil
-    local ok
     local sink = function(s)
         return s and callback(ffi.cast("uint8_t *", s), #s)
     end
-    local body, status_code, resp_headers, status_line
+    local abort
+    local response_headers = function(status_code, resp_headers, status_line)
+        if ranges then
+            abort = status_code ~= 206
+        else
+            abort = status_code ~= 200 and status_code ~= 304
+        end
+        return abort
+    end
+    local ok, status_code, resp_headers, status_line
     if ranges then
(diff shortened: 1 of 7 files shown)


Frenzie:
> (since Cloudflare response doesn't
include the expected "Accept-range" header)

Nginx. ;-) CF just passes them on.

benoit-pierre:
> > (since Cloudflare response doesn't
> > include the expected "Accept-range" header)
> 
> Nginx. ;-) CF just passes them on.

OK, so is there a way to get those so people that already are on kotasync can update?

Frenzie:
If adding that header suffices then yes, I can add it indiscriminately.

benoit-pierre:
Yeah, I see things like: `add_header accept_ranges bytes;`, we could try.

I'll downgrade `ffi/downloader` on my Kindle for testing.

Frenzie:
I've added it to all of `location ~* \.(apk|AppImage|targz|tar.gz|.tar.xz|zip)$`. (In all locations rather than just ota just to keep things the same.)

benoit-pierre:
And I can confirm I was able to download an update downgrading to v2026.07.2 with the initial `ffi/downloader` implementation! :+1:

Frenzie:
Alright, then I think the only remaining problem is the Kobo build. I'm not quite sure what to do about that one.

benoit-pierre:
The wrong OTA protocol? You can ask people to update `frontend/device/kobo/device.lua` manually. Or make a small `update.tar.xz` fixup with just that updated file and tell people to plop it down in `koreader/ota` and restart KOReader.

PR #2496, opened 2026-08-05
ffi/posix: fix Android and macOS declarations

benoit-pierre:
Also add a very basic test to sanity checks those intermediate generated `ffi-cdecl/posix_h_XXX.lua` files.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2496)


spec/unit/posix_h_spec.lua
```diff
@@ -0,0 +1,23 @@
+local lfs = require "lfs"
+
+describe("generated", function()
+    local ffi_cdecl_dir = lfs.symlinkattributes("ffi").target .. "-cdecl"
```
benoit-pierre, line 4: Not particularly pretty, but I could not think of another way to get the path to `base/ffi-cdecl`.

benoit-pierre:
It looks like CircleCI is having a fit…

spec/unit/posix_h_spec.lua
Frenzie: It doesn't seem ugly to me? It's a folder right next to wherever the symlink points.

If you really dislike it I suppose you could do something like this: https://github.com/koreader/koreader/blob/2749cf89170b90b44fae2c6c624352dd178e63c9/Makefile#L5 (otoh I don't think base has something like that right now)

Frenzie (approved):

spec/unit/posix_h_spec.lua
benoit-pierre: Yeah, let's leave it at that.

PR #2498, opened 2026-08-05
add `pocketbookhf` target

benoit-pierre:
For models like the InkPad One.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2498)


Frenzie:
Should it run a build test?

benoit-pierre:
If you want.

Frenzie (approved):

PR #2501, opened 2026-08-06
Android input: handle drawing

kerivin:
In the release, drawing is very laggy: turns out many input touch points are lost because most of them end up in the 'historical' events while a few get into the 'current' event. This is crucial for any future Android stylus plugins, but also other frequent events as such may occur I guess.
I said 'drawing' but as of now it doesn't exist in KOReader per-se, so I've tested it with an external plugin.
I hope this PR is correct, since I checked and re-checked it 5 times but can't be sure since I'm no Android expert x) From what I found it seems like processing 'historical' events before the 'current' event is actually the conventional way of processing touch events

This PR has [`AMotionEvent_getToolType`](https://developer.android.com/reference/android/view/MotionEvent#getToolType(int)) that was introduced in koreader/android-luajit-launcher#606

See [the explanation on historical events](https://developer.android.com/reference/android/view/MotionEvent#batching)

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2501)

### DIFF
--- a/ffi/input_android.lua
+++ b/ffi/input_android.lua
@@ -64,11 +64,22 @@ local function setPointerDown(slot, down)
     end
 end
 
+local TOOL_TYPE_MAP = {
+    [android.lib.AMOTION_EVENT_TOOL_TYPE_STYLUS]  = 1, -- TOOL_TYPE_PEN
+    [android.lib.AMOTION_EVENT_TOOL_TYPE_ERASER] = 2, -- TOOL_TYPE_ERASER
+}
+
+local function getToolType(event, index)
+    local tool = android.lib.AMotionEvent_getToolType(event, index)
+    return TOOL_TYPE_MAP[tool] or 0 -- TOOL_TYPE_FINGER
+end
+
 local function genTouchDownEvent(event, slot, index)
     local x = android.lib.AMotionEvent_getX(event, index)
     local y = android.lib.AMotionEvent_getY(event, index)
     local timev = genInputTimeval(android.lib.AMotionEvent_getEventTime(event))
     genEmuEvent(C.EV_ABS, C.ABS_MT_SLOT, slot, timev)
+    genEmuEvent(C.EV_ABS, C.ABS_MT_TOOL_TYPE, getToolType(event, index), timev)
     genEmuEvent(C.EV_ABS, C.ABS_MT_TRACKING_ID, slot, timev)
     genEmuEvent(C.EV_ABS, C.ABS_MT_POSITION_X, x, timev)
     genEmuEvent(C.EV_ABS, C.ABS_MT_POSITION_Y, y, timev)
@@ -80,6 +91,7 @@ local function genTouchUpEvent(event, slot, index)
     local y = android.lib.AMotionEvent_getY(event, index)
     local timev = genInputTimeval(android.lib.AMotionEvent_getEventTime(event))
     genEmuEvent(C.EV_ABS, C.ABS_MT_SLOT, slot, timev)
+    genEmuEvent(C.EV_ABS, C.ABS_MT_TOOL_TYPE, getToolType(event, index), timev)
     genEmuEvent(C.EV_ABS, C.ABS_MT_TRACKING_ID, -1, timev)
     genEmuEvent(C.EV_ABS, C.ABS_MT_POSITION_X, x, timev)
     genEmuEvent(C.EV_ABS, C.ABS_MT_POSITION_Y, y, timev)
@@ -91,6 +103,7 @@ local function genTouchMoveEvent(event, timev, slot, index)
     local x = android.lib.AMotionEvent_getX(event, index)
     local y = android.lib.AMotionEvent_getY(event, index)
     genEmuEvent(C.EV_ABS, C.ABS_MT_SLOT, slot, timev)
+    genEmuEvent(C.EV_ABS, C.ABS_MT_TOOL_TYPE, getToolType(event, index), timev)
     genEmuEvent(C.EV_ABS, C.ABS_MT_POSITION_X, x, timev)
     genEmuEvent(C.EV_ABS, C.ABS_MT_POSITION_Y, y, timev)
 end
@@ -149,15 +162,34 @@ local function motionEventHandler(motion_event)
 
         -- This effectively gives us the size of the current MotionEvent array...
         local pointer_count = tonumber(android.lib.AMotionEvent_getPointerCount(motion_event))
+        -- Collect the pointers that are still in contact, as (slot, pointer_index) pairs.
+        local active = {}
         for i = 0, pointer_count - 1 do
             -- So, loop through the array, and if that pointer is still down, move it
             local slot = android.lib.AMotionEvent_getPointerId(motion_event, i)
             if pointers[slot] then
-                genTouchMoveEvent(motion_event, timev, slot, i)
+                active[#active + 1] = { slot = slot, index = i }
             end
         end
 
-        -- Bundle everything in a single input frame
+        -- Useful for continious events like drawing
+        local history = tonumber(android.lib.AMotionEvent_getHistorySize(motion_event))
+        for h = 0, history - 1 do
+            local htimev = genInputTimeval(android.lib.AMotionEvent_getHistoricalEventTime(motion_event, h))
+            for __, ptr in ipairs(active) do
+                genEmuEvent(C.EV_ABS, C.ABS_MT_SLOT, ptr.slot, htimev)
+                genEmuEvent(C.EV_ABS, C.ABS_MT_TOOL_TYPE, getToolType(motion_event, ptr.index), htimev)
+                genEmuEvent(C.EV_ABS, C.ABS_MT_POSITION_X,
+                    android.lib.AMotionEvent_getHistoricalX(motion_event, ptr.index, h), htimev)
+                genEmuEvent(C.EV_ABS, C.ABS_MT_POSITION_Y,
+                    android.lib.AMotionEvent_getHistoricalY(motion_event, ptr.index, h), htimev)
+            end
+            genEndTouchEvent(motion_event, htimev)
+        end
+
+        for __, ptr in ipairs(active) do
+            genTouchMoveEvent(motion_event, timev, ptr.slot, ptr.index)
+        end
         genEndTouchEvent(motion_event, timev)
     elseif flags == C.AMOTION_EVENT_ACTION_CANCEL then
         -- Invalidate the pointers, and push a custom event to notify front to do the same.
--- a/ffi/input_android.lua
+++ b/ffi/input_android.lua
@@ -99,7 +99,6 @@ local function genTouchUpEvent(event, slot, index)
 end
 
 local function genTouchMoveEvent(event, timev, slot, index, x, y)
-    -- NOTE: May return a float for events w/ subpixel precision.
     genEmuEvent(C.EV_ABS, C.ABS_MT_SLOT, slot, timev)
     genEmuEvent(C.EV_ABS, C.ABS_MT_TOOL_TYPE, getToolType(event, index), timev)
     genEmuEvent(C.EV_ABS, C.ABS_MT_POSITION_X, x, timev)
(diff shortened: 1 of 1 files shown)


Frenzie (approved):
Looks okay. Do you have any idea what the performance impact is?

kerivin:
Historical events seem to still exist and being stored, it's just that KOReader doesn't read them. So shouldn't be much since I'm only adding the reading part. Didn't notice any changes on the device (2 GB RAM) but I'm going to keep an eye on the issues after it's merged x)

pazos:
>  Didn't notice any changes on the device (2 GB RAM) but I'm going to keep an eye on the issues after it's merged

Not very fan of this as it is showcased. Im going to assume that memory is not an issue and the overhead, if any, happens on cpu time.

Please at least test gestures extensively and make sure no performance penalty is introduced in regular motion events.

kerivin:
Disclaimer: data analyzed and put into table by AI

---

Metrics (and comparability):
- row = one ~1 s input window (a fresh row every ~1 s, only when input arrives).
- events = input events generated;
- hist = historical samples unrolled;
- tool = getToolType calls.
  -> scale with input volume: informative, but only a clean perf comparison when input count is identical.
- disp/s = avg per-window dispatch ms (~per second)  -> the normalized, comparable CPU rate.
- med/p90/max = per-window disp ms  -> comparable (window-normalized).
- µs/evt = disp / event count  -> comparable (event-normalized).

## Manual — identical input (branch vs baseline)

### Stylus drawing  (freehand — different input volume; events/hist/tool = scale, not perf)
| metric     | baseline | branch | delta |
|------------|---------:|-------:|------:|
| events     | 927 | 19969 | (branch 21.5x) |
| hist       | 0   | 3790  | - |
| tool       | 0   | 3989  | - |
| disp/s     |  9.0 |  16.3 | +84% |
| med/p90/max(ms) | 6.8 / 18.4 / 31.0 | 11.0 / 40.0 / 55.2 | - |
| µs/evt     | 204.1 | 13.9 | -93% |
| disp/frame (ms)    | 0.150    | 0.272  | +81%        |
| % of 16.7 ms frame | 0.90%    | 1.63%  | +0.73 pts   |


### Exactly 20 swipes  (same physical input count)
| Metric | baseline | branch | delta |
|--------|---------:|-------:|------:|
| events | 751 | 3435 | +2684 (+357%) |
| hist | 0 | 532 | +532 |
| tool | 0 | 678 | +678 |
| disp/s | 5.0 | 9.2 | +84% |
| med/p90/max(ms) | 3.6 / 10.7 / 12.7 | 6.8 / 16.9 / 27.9 | - |
| µs/evt | 166.3 | 56.3 | -66% |
| disp/frame (ms)    | 0.084    | 0.154  | +84%        |
| % of 16.7 ms frame | 0.50%    | 0.92%  | +0.42 pts   |


## Cost decomposition — 20 manual swipes (branch variant with unroll disabled)
| Metric | baseline | branch WITH-unroll | branch NO-unroll |
|--------|---------:|-------------------:|-----------------:|
| events | 751 | 3435 | 779 |
| hist | 0 | 532 | 0 |
| tool | 0 | 678 | 146 |
| disp/s | 5.00 | 9.21 | 5.84 |
| med/p90/max(ms) | 3.6/10.7/12.7 | 6.8/16.9/27.9 | 4.7/14.9/17.4 |
| µs/evt | 166.3 | 56.3 | 150.0 |
| disp/frame (ms) | 0.084 | 0.154 | 0.098 |
| % of 16.7 ms frame | 0.50% | 0.92% | 0.59% |

Per-second attribution of the branch's +4.21 ms/s:
- branch scaffolding + getToolType (tool 0->146): +0.84 ms/s (+16%).
- history-unroll loop only: +3.37 ms/s (+67% vs baseline) = ~80% of the increase.
- whole branch: +4.21 ms/s (+84%).

## Automatic (adb input) — same injection each build, hist=0 (branch only; no baseline: adb injects no history)
| Class | WITH-unroll µs/evt | NO-unroll µs/evt | hist |
|-------|------------------:|-----------------:|-----:|
| slow swipes | 128.8 | 96.2 | 0 |
| fast swipes | 67.6 | 35.0 | 0 |
| lots taps | 35.0 | 39.2 | 0 |
| slow taps | 29.7 | 28.6 | 0 |
| fast tap burst | 20.4 | 22.4 | 0 |

Note: real manual finger recorded hist 166-532; adb input hist 0.

## Takeaways
1. Input dispatch is ~1% of wall time (branch ~0.9%, baseline ~0.5%); med window stays <=16.7ms. The 12.7->27.9ms max coincide with render/GC single-wake events, not steady throughput.
2. Per event, branch is cheaper than baseline (-37% to -93%). Extra dispatch/s is from emitting 3-20x more events (history unroll).
3. Real finger swipes DO use history unroll (hist 166-532); not stylus-only.
4. The unroll loop is ~80% of the branch's added dispatch; the tool-type path + scaffolding is a small +0.8ms/s (+16%). Tool-type caching is not worthwhile.
5. Frame impact is negligible: even with ~2x dispatch (unroll), input processing consumes 0.84 -> 0.154 ms per 16.7 ms frame = 0.50% -> 0.92% of a frame. Doubling dispatch leaves ~99% of the frame free for rendering/UI. 

---

I like how it says 'real manual finger' x))

kerivin:
Thoughts on unrolling history _only_ for a non-finger tool? I doubt anyone would need a finger for drawing

UPD: Another option is to use some hidden (from the UI) setting, which plugins would use to enable history when drawing begins and disable history when it ends

Frenzie:
> Please at least test gestures extensively and make sure no performance penalty is introduced in regular motion events.

Though I believe this means that Android was the only platform apparently dropping up to 90% its input (although it doesn't feel like it).

NiLuJe:
> Though I believe this means that Android was the only platform apparently dropping up to 90% its input (although it doesn't feel like it).

IIRC, we also ignore jitter and some in-between contact/lift events on other platforms (we read them, but we don't act on them).

NiLuJe (approved):
Looks sound, and this only touches Android, so I'm happy ;p.

Frenzie:
> IIRC, we also ignore jitter and some in-between contact/lift events on other platforms (we read them, but we don't act on them).

Fair enough.

@kerivin a very simple test would also be to simply draw with your finger and see what it looks like.

kerivin:
## Finger drawing
| Metric            | baseline | WITH-unroll | NO-unroll |
|-------------------|---------:|-----------:|----------:|
| events            | 60       | 2742       | 101       |
| hist              | 0        | 441        | 0         |
| tool              | 0        | 536        | 19        |
| disp/s            | 2.03     | 6.02       | 3.73      |
| med(ms)           | 2.39     | 2.67       | 3.56      |
| max(ms)           | 2.45     | 72.88      | 5.87      |
| µs/evt            | 135.5    | 68.0       | 110.8     |
| disp/frame(ms)    | 0.034    | 0.100      | 0.062     |
| % of 16.7ms frame | 0.20%    | 0.60%      | 0.37%     |

And the actual visual difference between no-unroll and unroll:

current_page

(It's the same with stylus drawing)

Frenzie:
When I draw with my finger on Kobo iirc it looks like the latter, so I'd say the jitter prevention @NiLuJe mentioned isn't nearly as drastic. I didn't realize it was like that on Android; it seems smooth enough when panning a bit.

That being said, one does suspect that paring it down to, say, a thousand shouldn't have a visibly different result.

kerivin:
What do you mean by 'paring it down to a thousand'? Checking the number of the historical events and then only process every so many of them so that total would be 1000 max? Like skip every other historical event when there's 2000 of them?

Frenzie:
That was just a parenthetical remark, not a request. I don't literally mean any particular number, just that I'm not sure if subpixel differences for example are of relevance to us. But maybe they are for that pencil business?

kerivin:
Well, I can just test and see if the result is the same with some max cap 😀

kerivin:
I did test strokes with a max cap, but it's a bit tricky - the thousands you see are the accumulative value of historical events of all motion events. It turns out the average number of historical events per motion event (when casually drawing) is 17 on this device, and it has never exceeded 30. I tried drawing with cap 5, but the result looks jagged, not as bad as with zero historical but still I could see it... Cap 10 looks pretty much fine

kerivin:
My point is, the cap is probably not worth it x)

Frenzie:
It's also very much not what I meant by paring down. :p

kerivin:
But what did you mean then?.....................

kerivin:
~~Ok I've testes samples-by-time cap this time and it seems to work correctly (I'm still not sure if this is what you meant but the idea seems nice and it does reduce historical events volume while preserving the nice smooth look of the strokes)~~
~~I'll commit that soon~~

Nevermind, the cap of 1000 doesn't affect it at all, the events don't seem to happen that often so 1000 didn't change anything x)

I tried to skip move events older than 1ms from previous move event, but that didn't happen and zero was skipped

Frenzie:
I meant what I quoted from @NiLuJe earlier:

> IIRC, we also ignore jitter and some in-between contact/lift events on other platforms (we read them, but we don't act on them).

I wouldn't know if that's referring to base or front without investigating. In the latter case it may not matter.

Primarily I was referring to ignoring values that are really close together in distance and time. So a jitter threshold of 1 pixel (because input is normally much more precise than the screen)[1] and possibly some kind of low pass filter too. I once saved [this link](https://web.archive.org/web/20120709081205/http://blog.thomnichols.org/2011/08/smoothing-sensor-data-with-a-low-pass-filter), but that's about the much more jittery accelerometer so *if* it makes sense you'd be looking at something like 0.5, not 0.15. It's probably better to leave something like that to code further on if it wants to, like to subtly smoothen drawings. On that note, for a truly proper drawing app you'd want to do something based on speed as well, but that'd definitely be something for the frontend drawing code. 

[1] Though if you were drawing vector shapes you might want them to be more precise than pixels.

kerivin:
Ahh I see. Yeah that makes sense I guess. For now I just draw points (although I compress each stroke by skipping some)

I've just tested the 1-pixel filter, and it seems to reduce point count by 40%-50% (~1000 vs ~580 events per second) without losing quality, but with vector lines I'm not sure if it's worth it. Seems like a problem for a plugin to solve, I mean choose raster/vector and do whatever it wants with the info provided x)

I probably won't commit to this branch anymore so the PR is mergeable (unless you believe that 1-pixel filter is necessary in this case, I'll add it then, no problem)

NiLuJe:
> I wouldn't know if that's referring to base or front without investigating.

Note that I may entirely have dreamed up the jitter thing, but, yeah, that would be later down the chain, in the input state machine in front ;).

(There is a coordinates-based jitter filtering in the quick'n dirty [FBInk test tool](https://github.com/NiLuJe/FBInk/blob/886f25f13368859ad8a899b88d04c26e19cda32e/utils/finger_trace.c#L205-L220) though, so that may be the source of my confusion).

----

That said, given that Android is kind of a special snowflake, as we regenerate a Linux input stream from the Android machinery, the history stuff *might* warrant some kind of early filtering.

Then again, if you're only seeing something like 15 frames in a history batch, that's perhaps not worth the hassle ;).

TL;DR: I'm fine with this as-is, it's easy enough to address later if it turns out there are degenerate cases on some devices that risk flooding the input loop.

ffi/input_android.lua
```diff
@@ -80,21 +91,34 @@ local function genTouchUpEvent(event, slot, index)
...
     genEmuEvent(C.EV_SYN, C.SYN_REPORT, 0, timev)
 end
 
-local function genTouchMoveEvent(event, timev, slot, index)
-    -- NOTE: May return a float for events w/ subpixel precision.
-    local x = android.lib.AMotionEvent_getX(event, index)
-    local y = android.lib.AMotionEvent_getY(event, index)
+local function genTouchMoveEvent(event, timev, slot, index, x, y)
     genEmuEvent(C.EV_ABS, C.ABS_MT_SLOT, slot, timev)
+    genEmuEvent(C.EV_ABS, C.ABS_MT_TOOL_TYPE, getToolType(event, index), timev)
```
Frenzie: You don't need to send this in move, do you? Just in down. (Not in up either.)

kerivin: Yeah you are right, without it strokes still look correct

Frenzie: And you should save an insignificant but presumably at least measurable bit of time not calling both genEmuEvent and getToolType unnecessarily. We're talking about hundreds of them after all.

PR #2502, opened 2026-08-07
cre.cpp: tweak getLinkFromPosition()

poire-z:
Make the behaviour hardcoded in c0ae49f7 (#2497) toggable:
Frontend unit tests failed after this change, and it may actually be only helpful for the calls in the coherency check.
Ref: https://github.com/koreader/koreader/pull/15828#issuecomment-5209476714.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2502)

### DIFF
--- a/cre.cpp
+++ b/cre.cpp
@@ -1951,12 +1951,16 @@ static int getLinkFromPosition(lua_State *L) {
 	CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
 	int x = luaL_checkint(L, 2);
 	int y = luaL_checkint(L, 3);
+	// Providing forTextSelection=true may give more chances of success to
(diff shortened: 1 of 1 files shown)


cre.cpp
```diff
@@ -1951,12 +1951,16 @@ static int getLinkFromPosition(lua_State *L) {
 	CreDocument *doc = (CreDocument*) luaL_checkudata(L, 1, "credocument");
 	int x = luaL_checkint(L, 2);
 	int y = luaL_checkint(L, 3);
-
-	lvPoint pt(x, y);
-	// Providing forTextSelection=true gives more chances of success to
+	// Providing forTextSelection=true may give more chances of success to
 	// the link coherenrency checks done by frontend (eg. with negative
```
Frenzie: Might as well sneak in a  typo fix :-)

PR #2510, opened 2026-08-31
sdcv: read the .oft offset tables out of their mapping

asxelot:
Follow-up to #2441, and the fix I promised in [its discussion][disc] for the
remaining per-lookup cost that @Commodore64user has been chasing in
koreader/koreader#5437.

[disc]: https://github.com/koreader/koreader-base/pull/2441#issuecomment-5225410186

## Problem

#2441 stopped sdcv rescanning the whole `.syn` on every launch by caching its
offset table in a `.syn.oft`. But `load_offset_cache` reads that cache like
this:

```c
memcpy(&wordoffset[0], mf.begin() + strlen(CACHE_MAGIC) + sizeof(guint32),
       wordoffset.size() * sizeof(wordoffset[0]));
```

The whole table, copied out of the mapping into a heap vector, for every
dictionary that has one, on every launch — and sdcv is a fresh process per
lookup. A 46 MB inflection index has 1.7 M entries, so its table is 6.9 MB.
The binary search that uses it visits about twenty entries.

[@Commodore64user's suggestion on #2441][sug] was to reproduce it in a
container with a hard memory limit rather than by collecting 15 dictionaries.
That works, and it makes the shape of the problem very clear.

[sug]: https://github.com/koreader/koreader-base/pull/2441#issuecomment-5452173889

## Reproduction

14 dictionaries in one data dir (12 copies of a 46 MB `.syn` inflection index
plus two small dictionaries), 83 MB of `.oft` in total. One lookup per
container, `--memory=N --memory-swap=N`:

| `--memory` | sdcv today | with this patch |
|------------|------------|-----------------|
| 128m | ok, 18 ms | ok, 5 ms |
| 96m  | ok, 21 ms | ok, 6 ms |
| 64m  | **OOM-killed** | ok, 4 ms |
| 32m  | **OOM-killed** | ok, 3 ms |
| 16m  | **OOM-killed** | ok, 4 ms |
| 8m   | **OOM-killed** | ok, 4 ms |

Peak resident memory for a single lookup, split by kind:

| | RssAnon | RssFile | VmHWM |
|---|---------|---------|-------|
| sdcv today | 81.2 MB | 12.2 MB | 93.4 MB |
| with this patch | 1.8 MB | 18.8 MB | 20.7 MB |

The 81 MB of anonymous memory is those offset tables. Anonymous pages cannot be
reclaimed, are private to each sdcv process, and are what the OOM killer counts.

## Fix

Keep the `.oft` mapping for the lifetime of the dictionary and index into it,
instead of copying it out. `OffsetIndex` and `SynFile` share a small
`OffsetTable` that is backed either by that mapping (the normal case) or by a
vector (when the table had to be rebuilt because there was no usable cache).

The offsets have to be 4-byte aligned to be read in place, and today they start
at byte 34 — 30 bytes of magic string plus a 4-byte magic word. So the header
gains two bytes of padding, ending the magic word at byte 36, and the cache
format is bumped to 0.3. **A 0.2 cache fails the magic check and is rebuilt
once**, which costs one index scan per dictionary the first time a user runs
the new build; after that it is cheaper than before, forever.

While the mappings are being made honest about their access pattern, the ones
that are only ever probed by binary search are marked `MADV_RANDOM` — the
`.oft`s always, and a `.syn` once its table came from a cache (not on the branch
that has to scan the file). Without it each of the ~20 scattered probes pulls a
128 kB readahead window it never looks at.

`MapFile` gains move semantics so a mapping can be validated in a local and then
handed to its owner without being unmapped and remapped.

## Effect on a cold cache

A suspend/resume drops the page cache, which is what makes this visible on
device. Page cache dropped before each lookup, median of 6 words:

| | wall | read from storage |
|---|------|-------------------|
| sdcv today | 0.086 s | 106.8 MB |
| with this patch | 0.049 s | 11.1 MB |

Fuzzy search is not regressed by `MADV_RANDOM` (median of 4 words):
151.6 MB → 48.8 MB read, 0.133 s → 0.119 s.

With reads throttled to the rates DictProbe measured on a Kindle PW5 just after
a wake, dictionaries on a loop device, page cache dropped, one cold lookup:

| read rate | sdcv today | with this patch | |
|-----------|------------|-----------------|---|
| 3.3 MB/s | 30.84 s | 1.89 s | 16.3x |
| 4.0 MB/s | 25.46 s | 1.65 s | 15.4x |
| 6.0 MB/s | 16.93 s | 1.05 s | 16.1x |
| 11.0 MB/s | 9.25 s | 0.56 s | 16.5x |

## Verification

Fresh v0.5.5 tarball, all six patches applied with `patch -p1 --forward`
(exactly what `apply_patches` runs), in the order `CMakeLists.txt` lists them:
all apply cleanly, no `.rej`, and the result builds.

- **sdcv's own test suite: 13/13**, same as the unpatched build.
- **Output byte-identical** across 44 comparisons against the current build:
  15 words exact, 5 fuzzy and 2 misses, each on both the cold-rebuild and the
  warm-cache path.
- **Cache handling, 12 checks:** cold build, warm reuse, a 0.2 cache rejected
  and upgraded, a garbage cache, a truncated cache, an oversized cache, a stale
  cache, and a read-only dictionary directory where the cache cannot be written
  at all. Every one is detected and the answer never changes.

The alignment guarantee is enforced at compile time (`static_assert` on the
header size) and checked in the cache test against the file that is actually
written.

Upstream, this goes into the existing
[Dushistov/sdcv#129](https://github.com/Dushistov/sdcv/pull/129) rather than a
second PR — that one is still unreviewed, and there is no sense in having it
land a cache format that this immediately replaces. If it lands and the pinned
sdcv is bumped, both patches here can be dropped together.

## Thanks

Thanks to @Commodore64user for the on-device testing that turned this up and
kept it honest across many rounds on a PW5. The storage rates the table above
is throttled to are theirs, measured by the diagnostic patches they ran; the
observation that the cost was fixed per launch rather than proportional to the
word came out of their logs; and the memory-limited container that reproduces
the whole thing without a device or fifteen dictionaries was their suggestion.

---

This pull request was written by a human with AI assistance (implementation
developed and reviewed with Claude).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2510)


Commodore64user:
Have you put this up in the sdcv repo? I know the og hasn't been reviewed, but it's important that Dushistov gets this same code, after all, he knows this code better than anyone else.

asxelot:
Yes — just pushed it. Rather than a second PR it went onto the existing [Dushistov/sdcv#129](https://github.com/Dushistov/sdcv/pull/129), which now carries both commits: the `.syn.oft` cache and this zero-copy follow-up. Title and body rewritten to cover both, with the 0.2 → 0.3 format bump called out. Keeping them together seemed better than having #129 land a cache format this immediately replaces.

The upstream version also extends `tests/t_synfile_cache` to assert the aligned 0.3 layout on the caches sdcv actually writes, and that a 0.2 cache is rejected and rewritten — the patches here are source-only. 14/14 upstream tests pass.

Commodore64user:
@Frenzie do you have a plan here? Dushistov still nowhere to be seen...

Frenzie:
I can't really judge it, or at least not without investing a lot of time. I assume the regular sdcv regression tests were run?

asxelot:
Yes. sdcv's own ctest suite passes 13/13 both with koreader-base's current patch set and with this patch added (sdcv 0.5.5, `-DBUILD_TESTS=True`). The upstream version in Dushistov/sdcv#129 passes 14/14; the extra test asserts the 0.3 cache layout and the 0.2 → 0.3 rebuild.

Beyond the suite, I diffed lookup output between the two builds: 44 lookups (exact, fuzzy and misses, on both cold-rebuild and warm-cache paths) came out byte-identical. I also ran 12 cache-handling cases: cold build, warm reuse, a 0.2 cache rejected and upgraded, and garbage, truncated, oversized and stale caches, plus a read-only dictionary directory. All pass.

Frenzie (approved):
Alright, I'll merge it in a few days then, leaving some time for comments. You can ping me if I forget.

PR #2512, opened 2026-09-06
cre: add getVisibleWords to expose laid-out page glyphs

gkgangavarapu:
### Summary
Adds `getVisibleWords()` to the CREngine Lua binding: returns the visible words
of the current page, each as `{ text, chars = { {char, x0,y0,x1,y1}, ... } }`
(page/window coordinates, reading order).

### Motivation
Lua currently gets a finished page image but no per-glyph layout information, so
plugins cannot style or hit-test individual characters of a reflowable page.
This exposes that geometry using existing layout data only
(`getPageDocumentRange` + `getRangeWords` + per-offset `getRectEx`); layout and
rasterization are unchanged.

### Notes
- Purely additive & read-only; reversible.
- Words split across a line break are returned as separate words.
- Follow-up: `busted` spec (to be added in the koreader PR / emulator), and
  hardening for surrogate pairs / combining marks.

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2512)


gkgangavarapu:
Here's the intent behind this change.

Problem: KOReader hands Lua a finished, rasterized page image — Lua has no way to know where any individual character of a reflowable book is. So plugins can't style, emphasize, or hit-test a single glyph on the normal reading page.

What getVisibleWords does: returns each visible word's text together with its laid-out per-character rectangles ({ char, x0, y0, x1, y1 }) in page coordinates — read-only, built from the engine's existing layout data, with no change to formatting/rasterization.

Usability it unlocks:

reading aids that emphasize one letter of every word on the actual page (e.g. an Optimal Recognition Point highlighter),
more precise text-selection and highlight rectangles,
per-character dictionary lookup (helpful for CJK),
layout/metrics tools that need real glyph positions.
Because it's additive, generic geometry, other plugins can consume it too, not just this one use case. It pairs with frontend PR koreader#16022, which exposes the same method on Credocument.

Happy to add any tests or adjust the contract (e.g. coordinate space or word-splitting behavior) based on review.

Frenzie:
If I understand correctly it's not so much the words as the characters, which presumably aren't in any of the existing functions?

poire-z:
> Usability it unlocks:

Was going to ask for the real use case that **_you_** need it for - and I see it's at: https://github.com/koreader/koreader/pull/16022#discussion_r3944842241
(Clever people doing it for some other quick reading feature "Bionic reading" with only the start of words enbolden - you embolden the middle of the word :) strange - did it by having the font itself do it https://www.google.com/search?hl=en&num=25&q=font+bionic+reading.)

There probably could be ways to get it from the existing stuff (ie. get page xpointer, move by char until next page xpointer, getSegmentRect() of the existing char :) but it would be quote slow.
Althought it may already be quite slow, calling getRectEx() on each char. Is it?

Anyway, If you really want that for fun, I guess there is no other way than us including it - even if it looks excessive, and we would probably never do what you do :)
But let me think for a few days if there is better alternative.

Btw, if we include it in cre.cpp and don't use it in KOReader, no need for https://github.com/koreader/koreader/pull/16022 , you would use it by calling it directly `self..document._document:getVisibleWords()`.

Frenzie:
> But let me think for a few days if there is better alternative.

There's already the highlight/underline code. It feels similar. For this highly specific purpose presumably it could be expanded to bolden as well. Whether that's in any sense preferable I'll leave to you. ;-)

PR #2515, opened 2026-09-13
ci: don't bother with debug symbols

benoit-pierre:
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2515)


Makefile.defs
```diff
@@ -395,7 +395,7 @@ endif
 # BASE_CFLAGS = $(VECTO_CFLAGS) $(LTO_FLAGS)
 # Use this for debugging:
 ifdef KODEBUG
-  BASE_CFLAGS = -Og -g -pipe -fno-omit-frame-pointer
+  BASE_CFLAGS = $(if $(CI),-O2,-Og -g) -pipe -fno-omit-frame-pointer
```
Frenzie: This one seems to be opposed to the point of testing a debug build, doesn't it?

benoit-pierre: It ensure that all the thirdparty project options conditioned on a debug build (like extra traces) work.

Frenzie: Yes, but Og and O2 are different. Besides which O2 is slower to compile.

PR #2516, opened 2026-09-13
make: fix `bincheck` rule

benoit-pierre:
Fix `kobov5` error about missing `libcrypt.so.2` library: add its directory to the library search path (`/usr/lib`).

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2516)


benoit-pierre:
Is there a point in keeping sony-prstux support? We only build it on Circle CI, but not on nightly, and it's not part of the releases. It's also the only user of our `koreader/kodeb` Docker image.

Frenzie:
Pinging @protobits

protobits:
Hi, I'm not maintaining prstux anymore. If it is giving you trouble maybe it is best to remove support for it.

Frenzie:
Alright, in that case let's drop it.

PR #2518, opened 2026-09-14
[qrencode]: use upstream's newer improved encode algorithm

Commodore64user:
### what's new

This is a like for like copy of https://github.com/speedata/luaqrcode/blob/master/qrencode.lua which is the same current algorithm, rewritten for performance, changes described here by Claudine i.e., Claude's free tier ;) 

* binary(): was octal-based (%o + 3-bit table lookup), now hex-based (%x + 4-bit table). Fewer gsub passes.
* bit_xor: was a manually-written bit table (cclxvi) plus tbl_to_number; now a 256×256 xor_lookup table built at load time via a portable slow_xor, then O(1) lookups on the hot path.
* encode_string_numeric/_ascii/_binary: were string.gsub with anonymous callbacks; now plain indexed for loops with sub/byte. Same output, less closure overhead.
* add_pad_data: was repeated data .. string.rep(...) (quadratic string growth); now accumulates into a table and concats once.
* arrange_codewords_and_calculate_ec: was a repeat...until #arranged_data == #data loop re-scanning with string.sub each pass; now precomputes block lengths and walks in fixed 8-bit strides. This was the main O(n²)-ish hot spot for large versions.
* calculate_error_correction: was reconverting the entire polynomial to alpha/int representation on every iteration of the outer while-loop; now does it incrementally per step. For version-40-sized codes (which is exactly what multi-page QR generation exercises), this is the biggest real-world speedup — you're calling qrcode() N times now instead of once, so this matters more than it did before.
* add_data_to_matrix: was building a table of "next free positions" per byte via get_next_free_positions plus a gsub walk over the whole bitstring; now a single direct while loop toggling x_dir/y_dir. Same zigzag fill result, no per-byte table allocation.
* add_alignment_pattern: was 24 explicitly hardcoded cell assignments per alignment centre; now a 5×5 loop keyed on Chebyshev distance parity. I checked this reduces to the same pattern (I verified the parity arithmetic), not just "probably equivalent."
* add_timing_pattern, mask functions (maskFunc table replacing the elseif chain): same reduction — I checked mask 4, 6 and 7's rearranged arithmetic ((y%4-1.5)*(x%6-2.5)>0 etc.) and they're algebraically identical to the originals, just avoiding math.floor/math.fmod calls.

There is however one potential danger, the `assert` in `get_version_eclevel` could crash the app on failure (although we currently truncate text if overly long), which would require to use a `pcall` here https://github.com/koreader/koreader/blob/945470bf7f5f1db822c6e83b4914b03a6ab9fbeb/frontend/ui/widget/qrwidget.lua#L23

- - -
This change is [Reviewable](https://reviewable.io/reviews/koreader/koreader-base/2518)


Frenzie:
> There is however one potential danger, the assert

Beats the former freeze. Do we want to resolve the luacheck issue upstream?

Commodore64user:
pardon my ignorance, what was the former freeze?

Frenzie:
I was referring to this commit message: "assert when no QR version fits instead of looping forever"

https://github.com/speedata/luaqrcode/commit/5a9c34bc04d5113048df8528af72ef31d25b7300

Maybe I looked at the wrong assert but the point remains either way. :-)

Commodore64user:
okay, so this previous version was of course a carbon copy of that other repo, I don't want to claim like I understand QR codes but with the help of Gemini, I have been attempting to optimise this encoding for use with LuaJIT, I have this benchmark

<details>
  <summary>2-qrencode-benchmark.lua</summary>
  

  
  ```lua
local logger = require("logger")
local old_qr = require("ffi/qrencode_copy")
local new_qr = require("ffi/qrencode")

local function benchmark(name, qr_module, payload, iterations)
    collectgarbage("collect")

    -- GC left running this time: Time per QR now includes collection
    -- pauses, which is what the device will actually experience.
    local start_mem = collectgarbage("count")
    local start_time = os.clock()

    for _ = 1, iterations do
        qr_module.qrcode(payload, 1)
    end

    local end_time = os.clock()
    local end_mem = collectgarbage("count")

    local time_taken = end_time - start_time
    local total_mem_kb = end_mem - start_mem
    local mem_per_qr = total_mem_kb / iterations

    logger.info(string.format("=== QR Benchmark: %s ===", name))
    logger.info(string.format("  Iterations : %d", iterations))
    logger.info(string.format("  Total Time : %.4f seconds", time_taken))
    logger.info(string.format("  Time per QR: %.4f seconds", time_taken / iterations))
    logger.info(string.format("  Total RAM  : %.2f KB", total_mem_kb))
    logger.info(string.format("  RAM per QR : %.2f KB", mem_per_qr))
    logger.info("====================================\n")
end

local function profile_hotpath(name, qr_module, payload, iterations)
    local ok_jitp, jitp = pcall(require, "jit.p")
    if not ok_jitp then
        logger.warn("jit.p not available on this build, skipping profile for " .. name)
        return
    end

    collectgarbage("collect")
    logger.info(string.format("=== JIT Profile: %s ===", name))

    -- "Fi1" = flat, function-level, 1ms sample interval. Swap to "Fli1"
    -- if you want line-level detail inside the hottest function once you
    -- know which one it is.
    jitp.start("Fi1")
    for _ = 1, iterations do
        qr_module.qrcode(payload, 1)
    end
    jitp.stop()
    logger.info("====================================\n")
end

local function run_tests()
    local payload = "When I found that I was a prisoner a sort of wild feeling came over me. I rushed up and down the stairs, trying every door and peering out of every window I could find; but after a little the conviction of my helplessness overpowered all other feelings. When I look back after a few hours I think I must have been mad for the time, for I behaved much as a rat does in a trap. When, however, the conviction had come to me that I was helpless I sat down quietly—as quietly as I have ever done anything in my life—and began to think over what was best to be done. I am thinking still, and as yet have come to no definite conclusion. Of one thing only am I certain; that it is no use making my ideas known to the Count. He knows well that I am imprisoned; and as he has done it himself, and has doubtless his own motives for it, he would only deceive me if I trusted him fully with the facts. So far as I can see, my only plan will be to keep my knowledge and my fears to myself, and my eyes open. I am, I know, either being deceived, like a baby, by my own fears, or else I am in desperate straits; and if the latter be so, I need, and shall need, all my brains to get through."
    payload = payload .. payload
    local iterations = 1000

    logger.info("Starting QR Code RAM and CPU benchmarks...")
    logger.info("JIT status: " .. tostring(jit.status()))

    benchmark("Legacy string-concat implementation (GC on)", old_qr, payload, iterations)
    benchmark("New bit-packed integer implementation (GC on)", new_qr, payload, iterations)

    -- Profile only the new implementation - that's the one we're tuning.
    profile_hotpath("New bit-packed integer implementation", new_qr, payload, iterations)
end

local UIManager = require("ui/uimanager")
UIManager:nextTick(run_tests)

return {}
  ```
  
</details>

which currently produces

```
09/19/26-04:03:58 INFO  Starting QR Code RAM and CPU benchmarks... 
09/19/26-04:03:58 INFO  JIT status: true 
09/19/26-04:04:29 INFO  === QR Benchmark: Legacy string-concat implementation (GC on) === 
09/19/26-04:04:29 INFO    Iterations : 1000 
09/19/26-04:04:29 INFO    Total Time : 30.4069 seconds 
09/19/26-04:04:29 INFO    Time per QR: 0.0304 seconds 
09/19/26-04:04:29 INFO    Total RAM  : 6636.03 KB 
09/19/26-04:04:29 INFO    RAM per QR : 6.64 KB 
09/19/26-04:04:29 INFO  ====================================
 
09/19/26-04:04:41 INFO  === QR Benchmark: New bit-packed integer implementation (GC on) === 
09/19/26-04:04:41 INFO    Iterations : 1000 
09/19/26-04:04:41 INFO    Total Time : 12.3894 seconds 
09/19/26-04:04:41 INFO    Time per QR: 0.0124 seconds 
09/19/26-04:04:41 INFO    Total RAM  : 945.13 KB 
09/19/26-04:04:41 INFO    RAM per QR : 0.95 KB 
09/19/26-04:04:41 INFO  ====================================
 
09/19/26-04:04:41 INFO  === JIT Profile: New bit-packed integer implementation === 
59%  qrencode.lua:calculate_penalty_bb
20%  qrencode.lua:qrcode
12%  qrencode.lua:func
 4%  qrencode.lua:calculate_error_correction
 3%  qrencode.lua:set_bb_bit
09/19/26-04:04:54 INFO  ====================================
```

here is the current code https://github.com/Commodore64user/koreader-base/tree/qrencode/ffi/qrencode.lua

Frenzie:
It's slow enough to bother on the old Kindle?

Commodore64user:
It can take a few seconds on longer snippets, and if you want to remove the truncation and generate as many codes as needed it becomes n-times slower.

Of course most people won't use more than 1-3 at the time but still, less ram spikes and less GC.

Frenzie:
This heavily-AI assisted patch goes on top of yours.

[0001-qrencode-SWAR-bitboard-penalty-per-version-static-ca.patch](https://github.com/user-attachments/files/32416792/0001-qrencode-SWAR-bitboard-penalty-per-version-static-ca.patch)

>  ● Replaced the per-cell penalty scan (70% of runtime) with word-parallel SWAR over uint32 bitboards:                                                                         
>                                                                                                                                                                               
>    - P1: run detection via diff-bits (c = b ⊕ (b<<1 | carry)); contribution = popcount(A) + 2·popcount(A & c) where A marks 5-equal-window starts. Cross-word windows         
>      handled with per-distance shifts.                                                                                                                                        
>    - P3: 11-bit shift-AND finder matching (TA/NA, TB/NB) plus 15-bit overlap term for the A-then-B dedup the original did with last_p3.                                       
>    - P2 + dark ratio: word-parallel with SWAR popcount; byte-extract popcount, not the multiply trick — LuaJIT's double promotion silently breaks x*0x01010101 on high        
>      popcounts.                                                                                                                                                               
>    - Vertical pass: real 32×32 in-register bit-transpose (D&C, constant masks), so both directions run the same word-parallel code.                                           
>                                                                                                                                                                               
>    Plus per-version static caching: zigzag order, mask boards + transposes, per-EC fixed-dark boards — built once, not per call.                                              
>                                                                                                                                                                               
>    Byte-identical output verified by differential testing (3k+ cases, v1–40, all modes/EC). 6–11× faster; v1 371→46 µs, v40 15.5→2.6 ms.

Commodore64user:
I can't apply it `error: corrupt patch at line 521` can you share the whole qrencode.lua instead

one thing to verify is that the correct mask is applied

Frenzie:
[qrencode.zip](https://github.com/user-attachments/files/32417270/qrencode.zip)

Frenzie:
> I can't apply it error: corrupt patch at line 521 can you share the whole qrencode.lua instead

Is that git apply / git am? It's not meant for patch (though I don't see why it shouldn't work).

Commodore64user:
what the actual fuck? pardon my French

```
09/19/26-16:10:43 INFO  Starting QR Code RAM and CPU benchmarks... 
09/19/26-16:10:43 INFO  JIT status: true 
09/19/26-16:11:15 INFO  === QR Benchmark: Legacy string-concat implementation (GC on) === 
09/19/26-16:11:15 INFO    Iterations : 1000 
09/19/26-16:11:15 INFO    Total Time : 32.0191 seconds 
09/19/26-16:11:15 INFO    Time per QR: 0.0320 seconds 
09/19/26-16:11:15 INFO    Total RAM  : 6012.18 KB 
09/19/26-16:11:15 INFO    RAM per QR : 6.01 KB 
09/19/26-16:11:15 INFO  ====================================
 
09/19/26-16:11:17 INFO  === QR Benchmark: New bit-packed integer implementation (GC on) === 
09/19/26-16:11:17 INFO    Iterations : 1000 
09/19/26-16:11:17 INFO    Total Time : 1.9359 seconds 
09/19/26-16:11:17 INFO    Time per QR: 0.0019 seconds 
09/19/26-16:11:17 INFO    Total RAM  : 1570.83 KB 
09/19/26-16:11:17 INFO    RAM per QR : 1.57 KB 
09/19/26-16:11:17 INFO  ====================================
 
09/19/26-16:11:17 INFO  === JIT Profile: New bit-packed integer implementation === 
28%  qrencode.lua:calculate_error_correction
22%  qrencode.lua:qrcode
14%  qrencode.lua:popcount32
13%  qrencode.lua:scan_p1p3
 9%  qrencode.lua:func
 6%  qrencode.lua:calculate_penalty_bb
09/19/26-16:11:19 INFO  ====================================
```

Frenzie:
There's some input by me for what it calls "the big one," but hardly enough to take credit.

> ● Profiling first showed where the time went: ~70% scoring penalties, ~30% building the candidate matrices, ~0% elsewhere. So that's where the wins are.
>   
>   1. SWAR (the big one). The old code visited every pixel in a Lua loop for each of 8 masks — 8 × N² iterations, each with branches, table lookups, and state variables.
>   The new code packs 32 pixels into one uint32 (one bit each). XOR, AND, OR, and popcount now process 32 pixels per instruction, so the inner loop runs over words: ~N²/32
>   iterations, each straight-line bit arithmetic with zero branches. That alone turns the penalty scan from ~10 ops/pixel into ~1 op per 30 pixels.                           
>                                                                                                                                                                              
>   2. Each penalty rule became a few word ops.                                                                                                                                
>                                                                                                                                                                              
>   - P1 (runs): one XOR-with-shifted-self gives a "differs from my left neighbor" board. A run of length n≥5 scores n−2, which decomposes into "count 5-equal windows" + "2×  
>     count of runs" — both are popcounts over two masked words. No per-pixel state machine.                                                                                   
>   - P3 (finder patterns): instead of an 11-bit rolling window compared per pixel, shift-AND matching: AND together 11 shifted copies of the word (dark cells) with the       
>     inverse of the light cells. One such expression finds all match positions in the word at once. The overlap-dedup case becomes a 15-bit variant.                          
>   - P2 (2×2 blocks): equality of the four neighbors is a handful of XORs of the word with its shifted self and the row below; popcount × 3.                                  
>   - P4 (dark ratio): popcount per word.                                                                                                                                      
>                                                                                                                                                                              
>   3. The vertical pass stopped being a copy loop. The original built a transposed board by writing bits one at a time, then rescanned it per pixel. Now there's a real       
>   32×32 in-register bit-transpose (divide-and-conquer, 5 mask-swap stages), after which the vertical scan reuses the same word-parallel code.                                
>                                                                                                                                                                              
>   4. Caching what never changes. The zigzag free-cell order, the 8 mask bitboards, their transposes, and the fixed patterns depend only on the QR version — not on the       
>   data. Built once per version, then reused, eliminating most of the remaining 30%.                                                                                          
>                                                                                                                                                                              
>   Applying a candidate mask is now a single board XOR instead of a Lua closure call per pixel.                                                                               
>                                                                                                                                                                              
>   Net: the dominant cost went from ~N²/8 interpreted-ish branchy loops to ~N²/256 straight-line ops plus a transpose — hence 6–11×, biggest on small codes where fixed       
>   overheads dominated.

Commodore64user:
can you make it check the per-mask penalty scores? for both the current code and this one, we want to make sure that it not only is speedy, but it also chooses the right mask, every time. you also slightly bumped up the RAM usage ;)

the caching seems quite relevant, what I have currently is, user selects some text, we check if it fits into one qr code, if it doesn't we split the load equally so we create `n` qr codes of similar size (version), so that warm cache would make any extra codes effectively free

> but hardly enough to take credit.

the joys of free copilot tokens?

Frenzie:
> can you make it check the per-mask penalty scores? for both the current code and this one, we want to make sure that it not only is speedy, but it also chooses the right mask, every time.

All tests were already identical with the code above, but I'll put it to work a bit more for you.

> the joys of free copilot tokens?

Not really; those run out almost instantaneously. GLM-5.3 Flash operates at a price point where you can tell it to go ahead and do whatever and it'll probably cost less than a dollar.

Frenzie:
Here's the reduced RAM usage.
[qrencode (1).zip](https://github.com/user-attachments/files/32418341/qrencode.1.zip)

Commodore64user:
not as good as the `0.95 KB` but still better than the current encoding

```
09/19/26-17:20:48 INFO  Starting QR Code RAM and CPU benchmarks... 
09/19/26-17:20:48 INFO  JIT status: true 
09/19/26-17:21:19 INFO  === QR Benchmark: Legacy string-concat implementation (GC on) === 
09/19/26-17:21:19 INFO    Iterations : 1000 
09/19/26-17:21:19 INFO    Total Time : 31.0364 seconds 
09/19/26-17:21:19 INFO    Time per QR: 0.0310 seconds 
09/19/26-17:21:19 INFO    Total RAM  : 7693.65 KB 
09/19/26-17:21:19 INFO    RAM per QR : 7.69 KB 
09/19/26-17:21:19 INFO  ====================================
 
09/19/26-17:21:21 INFO  === QR Benchmark: New bit-packed integer implementation (GC on) === 
09/19/26-17:21:21 INFO    Iterations : 1000 
09/19/26-17:21:21 INFO    Total Time : 1.8562 seconds 
09/19/26-17:21:21 INFO    Time per QR: 0.0019 seconds 
09/19/26-17:21:21 INFO    Total RAM  : 1229.84 KB 
09/19/26-17:21:21 INFO    RAM per QR : 1.23 KB 
09/19/26-17:21:21 INFO  ====================================
 
09/19/26-17:21:21 INFO  === JIT Profile: New bit-packed integer implementation === 
25%  qrencode.lua:calculate_error_correction
24%  qrencode.lua:qrcode
14%  qrencode.lua:popcount32
14%  qrencode.lua:scan_p1p3
 9%  qrencode.lua:func
 7%  qrencode.lua:calculate_penalty_bb
09/19/26-17:21:23 INFO  ====================================
 
09/19/26-17:21:23 WARN  New bit-packed integer implementation - burst: no _debug_reset_cache exposed - cache state is whatever this session has already built, not guaranteed cold 
09/19/26-17:21:23 INFO  === QR Burst Benchmark: New bit-packed integer implementation - burst === 
09/19/26-17:21:23 INFO    Chunk 1/5 (1413 bytes): 0.00882 seconds 
09/19/26-17:21:23 INFO    Chunk 2/5 (1413 bytes): 0.00229 seconds 
09/19/26-17:21:23 INFO    Chunk 3/5 (1413 bytes): 0.00169 seconds 
09/19/26-17:21:23 INFO    Chunk 4/5 (1413 bytes): 0.00144 seconds 
09/19/26-17:21:23 INFO    Chunk 5/5 (1416 bytes): 0.00146 seconds 
09/19/26-17:21:23 INFO  ====================================
```

chunk 1 is cold cache here, 2-5 become very cheap after.

Commodore64user:
I am leaving here the mask penalty score tests,

seems like it all checks out, actually thanks to these tests, we found there is currently a bug in the current penalty p2 , it is off-by-one, but accounting for that

https://github.com/koreader/koreader-base/blob/9a8729713ab73f539b607af23ede6aa89d04cfed/ffi/qrencode.lua#L1035-L1038

we get `09/20/26-15:23:04 INFO  Differential test PASSED across 12 lengths - all penalty components and mask selections match`, that is to say. We are doing the same thing much, much faster and using far fewer resources.
Phone scans work as well, so seems ok.

<details>
  <summary>new version</summary>
  

  
  ```lua
-- Debug-only: same as calculate_penalty_bb, but returns p1/p2/p3/p4
-- separately instead of their sum, so a differential test can localize
-- a mismatch to a specific penalty rule instead of just "the total disagrees".
local function debug_penalty_components(bb, tbb, size, stride)
	local sm = get_size_masks(size, stride)
	local p1, p3 = scan_p1p3(bb, size, stride, sm)
	local p1v, p3v = scan_p1p3(tbb, size, stride, sm)
	p1 = p1 + p1v
	p3 = p3 + p3v

	local p2, dark = 0, 0
	local w_end = stride - 1
	local p2m = sm.p2
	for y = 0, size - 1 do
		local offA = y * stride
		for w = 0, w_end do
			dark = dark + popcount32(bb[offA + w])
		end
		if y < size - 1 then
			local offB = offA + stride
			for w = 0, w_end do
				local a = bb[offA + w]
				local b2 = bb[offB + w]
				local d = bor(bor(bxor(a, rshift(a, 1)), bxor(a, b2)), bxor(b2, rshift(b2, 1)))
				p2 = p2 + popcount32(band(bnot(d), p2m[w])) * 3
				if w * 32 + 31 <= size - 2 then
					local a31 = rshift(a, 31)
					local b31 = rshift(b2, 31)
					if a31 == b31 and a31 == band(bb[offA + w + 1], 1) and b31 == band(bb[offB + w + 1], 1) then
						p2 = p2 + 3
					end
				end
			end
		end
	end

	local p4 = floor(abs(dark / (size * size) * 100 - 50)) * 2
	return p1, p2, p3, p4
end

-- Debug-only: same setup as qrcode() up through mask selection, but
-- records every mask's p1/p2/p3/p4 breakdown instead of only the winner.
local function debug_mask_penalties(str, ec_level, mode_enc)
	local mode_num = mode_enc or get_mode(str)
	local version, ec = get_version_eclevel(#str, mode_num, ec_level)

	bw:reset()
	bw:write(mode_num, 4)
	write_length(#str, version, mode_num)
	encode_data(str, mode_num)
	add_pad_data(version, ec)

	local total_arranged_bytes = arrange_codewords_and_calculate_ec(version, ec, bw.buf)
	local size = generate_base_matrix(version)

	add_typeinfo_to_matrix(base_matrix, size, ec, 0)

	local stride = floor((size + 31) / 32)
	local total_words = size * stride

	local entry = static_cache[version] or build_static(version, size, stride)
	local count = entry.count

	local fixed_dark = entry.fixed[ec]
	if not fixed_dark then
		fixed_dark = ffi.new("uint32_t[?]", total_words)
		for yi = 1, size do
			local rowbase = (yi - 1) * size
			for xi = 1, size do
				if base_matrix[rowbase + xi] == 2 then
					set_bb_bit(fixed_dark, stride, xi, yi, 1)
				end
			end
		end
		entry.fixed[ec] = fixed_dark
	end

	ffi.copy(bb_base, fixed_dark, total_words * 4)

	local total_bits = total_arranged_bytes * 8
	for k = 1, count do
		local data_bit = 0
		if k <= total_bits then
			local bi = k - 1
			data_bit = band(rshift(arranged_data[floor(bi / 8) + 1], 7 - bi % 8), 1)
		end
		if data_bit == 1 then
			local xx, yy = entry.x[k], entry.y[k]
			local wi = yy * stride + floor(xx / 32)
			bb_base[wi] = bor(bb_base[wi], lshift(1, xx % 32))
		end
	end

	transpose_bb(bb_base, T_base, size, stride)

	local components = {}
	local min_penalty, best_mask = nil, 0
	for mask = 0, 7 do
		local mb = entry.masks[mask]
		local tb = entry.t_masks[mask]
		for i = 0, total_words - 1 do
			bb_scratch[i] = bxor(bb_base[i], mb[i])
			T_scratch[i] = bxor(T_base[i], tb[i])
		end
		add_typeinfo_both(bb_scratch, T_scratch, size, stride, ec, mask)

		local p1, p2, p3, p4 = debug_penalty_components(bb_scratch, T_scratch, size, stride)
		components[mask] = { p1 = p1, p2 = p2, p3 = p3, p4 = p4 }
		local penalty = p1 + p2 + p3 + p4
		if not min_penalty or penalty < min_penalty then
			min_penalty = penalty
			best_mask = mask
		end
	end

	return { version = version, ec = ec, components = components, best_mask = best_mask }
end
  ```

</details>

<details>
  <summary>current version</summary>
  

  
  ```lua
-- Debug-only: same as calculate_penalty, but returns p1/p2/p3/p4 separately.
local function debug_penalty_components(tab)
	local penalty1, penalty2, penalty3 = 0, 0, 0
	local size = #tab
	local number_of_dark_cells = 0

	local last_bit_blank, is_blank, number_of_consecutive_bits
	for x = 1, size do
		number_of_consecutive_bits = 0
		last_bit_blank = nil
		for y = 1, size do
			is_blank = tab[x][y] < 0
			if not is_blank then number_of_dark_cells = number_of_dark_cells + 1 end
			if last_bit_blank == is_blank then
				number_of_consecutive_bits = number_of_consecutive_bits + 1
			else
				if number_of_consecutive_bits >= 5 then penalty1 = penalty1 + number_of_consecutive_bits - 2 end
				number_of_consecutive_bits = 1
			end
			last_bit_blank = is_blank
		end
		if number_of_consecutive_bits >= 5 then penalty1 = penalty1 + number_of_consecutive_bits - 2 end
	end
	for y = 1, size do
		number_of_consecutive_bits = 0
		last_bit_blank = nil
		for x = 1, size do
			is_blank = tab[x][y] < 0
			if last_bit_blank == is_blank then
				number_of_consecutive_bits = number_of_consecutive_bits + 1
			else
				if number_of_consecutive_bits >= 5 then penalty1 = penalty1 + number_of_consecutive_bits - 2 end
				number_of_consecutive_bits = 1
			end
			last_bit_blank = is_blank
		end
		if number_of_consecutive_bits >= 5 then penalty1 = penalty1 + number_of_consecutive_bits - 2 end
	end
	for x = 1, size do
		for y = 1, size do
			if (y < size) and (x < size) and (
				(tab[x][y] < 0 and tab[x+1][y] < 0 and tab[x][y+1] < 0 and tab[x+1][y+1] < 0) or
				(tab[x][y] > 0 and tab[x+1][y] > 0 and tab[x][y+1] > 0 and tab[x+1][y+1] > 0)
			) then penalty2 = penalty2 + 3 end

			if (y + 6 <= size and
				tab[x][y] > 0 and tab[x][y+1] < 0 and tab[x][y+2] > 0 and tab[x][y+3] > 0 and
				tab[x][y+4] > 0 and tab[x][y+5] < 0 and tab[x][y+6] > 0 and
				((y + 10 <= size and tab[x][y+7] < 0 and tab[x][y+8] < 0 and tab[x][y+9] < 0 and tab[x][y+10] < 0) or
				 (y - 4 >= 1 and tab[x][y-1] < 0 and tab[x][y-2] < 0 and tab[x][y-3] < 0 and tab[x][y-4] < 0)))
			then penalty3 = penalty3 + 40 end

			if (x + 6 <= size and
				tab[x][y] > 0 and tab[x+1][y] < 0 and tab[x+2][y] > 0 and tab[x+3][y] > 0 and
				tab[x+4][y] > 0 and tab[x+5][y] < 0 and tab[x+6][y] > 0 and
				((x + 10 <= size and tab[x+7][y] < 0 and tab[x+8][y] < 0 and tab[x+9][y] < 0 and tab[x+10][y] < 0) or
				 (x - 4 >= 1 and tab[x-1][y] < 0 and tab[x-2][y] < 0 and tab[x-3][y] < 0 and tab[x-4][y] < 0)))
			then penalty3 = penalty3 + 40 end
		end
	end

	local dark_ratio = number_of_dark_cells / (size * size)
	local penalty4 = floor(abs(dark_ratio * 100 - 50)) * 2
	return penalty1, penalty2, penalty3, penalty4
end

-- Debug-only: mirrors qrcode() up through mask selection, recording
-- every mask's p1/p2/p3/p4 breakdown.
local function debug_mask_penalties(str, ec_level, mode_enc)
	local version, ec, data_raw, mode, len_bitstring
	version, ec, data_raw, mode, len_bitstring = get_version_eclevel_mode_bistringlength(str, ec_level, mode_enc)
	data_raw = data_raw .. len_bitstring
	data_raw = data_raw .. encode_data(str, mode)
	data_raw = add_pad_data(version, ec, data_raw)
	local arranged_data = arrange_codewords_and_calculate_ec(version, ec, data_raw)
	if #arranged_data % 8 ~= 0 then
		return nil, format("Arranged data %% 8 != 0: data length = %d, mod 8 = %d", #arranged_data, #arranged_data % 8)
	end
	arranged_data = arranged_data .. rep("0", remainder[version])

	local components = {}
	local min_penalty, best_mask = nil, 0
	for mask = 0, 7 do
		local tab = prepare_matrix_with_mask(version, ec, mask)
		add_data_to_matrix(tab, arranged_data, mask)
		local p1, p2, p3, p4 = debug_penalty_components(tab)
		components[mask] = { p1 = p1, p2 = p2, p3 = p3, p4 = p4 }
		local penalty = p1 + p2 + p3 + p4
		if not min_penalty or penalty < min_penalty then
			min_penalty = penalty
			best_mask = mask
		end
	end

	return { version = version, ec = ec, components = components, best_mask = best_mask }
end
  ```
  
</details>

<details>
  <summary>2-mask-test.lua</summary>
  

  
  ```lua
local logger = require("logger")
local old_qr = require("ffi/qrencode_copy")
local new_qr = require("ffi/qrencode")

local base_text = "The quick brown fox jumps over the lazy dog. "
local function make_payload(len)
    local s = base_text:rep(math.ceil(len / #base_text))
    return s:sub(1, len)
end

local test_lengths = { 10, 50, 100, 200, 350, 500, 800, 1200, 1700, 2200, 2700, 2953 }

local function run_differential_test()
    local mismatches = 0
    for _, len in ipairs(test_lengths) do
        local payload = make_payload(len)
        local old_result = old_qr._debug_mask_penalties(payload, 1)
        local new_result = new_qr._debug_mask_penalties(payload, 1)

        if not old_result or not new_result then
            logger.warn(string.format("LEN %d: one side returned nil (encode failure?) - skipping", len))
            mismatches = mismatches + 1
            goto continue
        end

        if old_result.version ~= new_result.version or old_result.ec ~= new_result.ec then
            logger.warn(string.format(
                "LEN %d: version/ec mismatch old=(%d,%d) new=(%d,%d)",
                len, old_result.version, old_result.ec, new_result.version, new_result.ec))
            mismatches = mismatches + 1
        end

        for mask = 0, 7 do
            local oc, nc = old_result.components[mask], new_result.components[mask]
            for _, key in ipairs({ "p1", "p2", "p3", "p4" }) do
                if oc[key] ~= nc[key] then
                    logger.warn(string.format(
                        "LEN %d version %d mask %d %s: mismatch old=%d new=%d",
                        len, old_result.version, mask, key, oc[key], nc[key]))
                    mismatches = mismatches + 1
                end
            end
        end

        if old_result.best_mask ~= new_result.best_mask then
            logger.warn(string.format(
                "LEN %d version %d: best_mask mismatch old=%d new=%d",
                len, old_result.version, old_result.best_mask, new_result.best_mask))
            mismatches = mismatches + 1
        end

        ::continue::
    end

    if mismatches == 0 then
        logger.info(string.format("Differential test PASSED across %d lengths - all penalty components and mask selections match", #test_lengths))
    else
        logger.warn(string.format("Differential test found %d mismatches - see warnings above", mismatches))
    end
end

local UIManager = require("ui/uimanager")
UIManager:nextTick(run_differential_test)

return {}
  ```
  
</details>

Frenzie:
> not as good as the 0.95 KB but still better than the current encoding

I think a few kB is nothing anyway, if it leads to significantly faster speeds.

In any case, I'll leave all the details in your hands. ;-)

Commodore64user:
if you have some free time https://youtu.be/c33AZBnRHks

> I think a few kB is nothing anyway, if it leads to significantly faster speeds.

a few on a MacBook, I imagine it will be a bit higher on actual hardware, but it is still good though, still doing some more sanity checks, unfortunately the standard is locked behind a [paywall](https://www.iso.org/standard/83389.html?utm_source=chatgpt.com) but oh well

Frenzie:
> if you have some free time

That's certainly a very effective clickbait opening. My guess based just on the first few seconds would be a lot of precomputed tables or some such (i.e., just like how we'd have all kinds of books full of tables of calculations before pocket Japanese showed up).

Commodore64user:
> That's certainly a very effective clickbait opening.

;)

there is another error in the code, p4 is wrong

https://github.com/koreader/koreader-base/blob/0a16b6a580d08bd7a3a1c22321d5156bcb39071d/ffi/qrencode.lua#L1087-L1088

The QR penalty-4 rule is in 5 percentage-point increments, not one penalty point per percentage point.

Anyway, Gemini tells me we can't quite claim to have a fully ISO complaint encoder but so far as the [unit tests](https://github.com/Commodore64user/koreader-base/blob/qrencode/spec/unit/qrencode_spec.lua) are concerned we have a pretty good one.

```
09/21/26-18:47:17 INFO  Starting QR Code RAM and CPU benchmarks... 
09/21/26-18:47:17 INFO  JIT status: true 
09/21/26-18:47:19 INFO  === QR Benchmark: New bit-packed integer implementation (GC on) === 
09/21/26-18:47:19 INFO    Iterations : 1000 
09/21/26-18:47:19 INFO    Total Time : 1.8259 seconds 
09/21/26-18:47:19 INFO    Time per QR: 0.0018 seconds 
09/21/26-18:47:19 INFO    Total RAM  : 1237.62 KB 
09/21/26-18:47:19 INFO    RAM per QR : 1.24 KB 
09/21/26-18:47:19 INFO  ====================================
 
09/21/26-18:47:19 INFO  === JIT Profile: New bit-packed integer implementation === 
24%  qrencode.lua:calculate_error_correction
15%  qrencode.lua:popcount32
15%  qrencode.lua:scan_p1p3
12%  qrencode.lua:prepare_boards
 9%  qrencode.lua:func
 8%  qrencode.lua:qrcode
 6%  qrencode.lua:compute_penalty_components
09/21/26-18:47:21 INFO  ====================================
 
09/21/26-18:47:21 WARN  New bit-packed integer implementation - burst: no _debug_reset_cache exposed - cache state is whatever this session has already built, not guaranteed cold 
09/21/26-18:47:21 INFO  === QR Burst Benchmark: New bit-packed integer implementation - burst === 
09/21/26-18:47:21 INFO    Chunk 1/5 (1413 bytes): 0.00853 seconds 
09/21/26-18:47:21 INFO    Chunk 2/5 (1413 bytes): 0.00285 seconds 
09/21/26-18:47:21 INFO    Chunk 3/5 (1413 bytes): 0.00265 seconds 
09/21/26-18:47:21 INFO    Chunk 4/5 (1413 bytes): 0.00197 seconds 
09/21/26-18:47:21 INFO    Chunk 5/5 (1416 bytes): 0.00174 seconds 
09/21/26-18:47:21 INFO  ====================================
```

Commodore64user:
> It's slow enough to bother on the old Kindle?

here is a legit run on a kindle 4

```
09/23/26-00:47:07 INFO  Starting QR Code RAM and CPU benchmarks... 
09/23/26-00:47:07 INFO  JIT status: true 
09/23/26-00:47:29 INFO  === QR Benchmark: Legacy string-concat implementation (GC on) === 
09/23/26-00:47:29 INFO    Iterations : 50 
09/23/26-00:47:29 INFO    Total Time : 22.0800 seconds 
09/23/26-00:47:29 INFO    Time per QR: 0.4416 seconds 
09/23/26-00:47:29 INFO    Total RAM  : 11241.86 KB 
09/23/26-00:47:29 INFO    RAM per QR : 224.84 KB 
09/23/26-00:47:29 INFO  ====================================
 
09/23/26-00:47:31 INFO  === QR Benchmark: New bit-packed integer implementation (GC on) === 
09/23/26-00:47:31 INFO    Iterations : 50 
09/23/26-00:47:31 INFO    Total Time : 1.9300 seconds 
09/23/26-00:47:31 INFO    Time per QR: 0.0386 seconds 
09/23/26-00:47:31 INFO    Total RAM  : 748.89 KB 
09/23/26-00:47:31 INFO    RAM per QR : 14.98 KB 
09/23/26-00:47:31 INFO  ====================================
 
09/23/26-00:47:31 INFO  === JIT Profile: New bit-packed integer implementation === 
22%  qrencode.lua:prepare_boards
18%  qrencode.lua:calculate_error_correction
16%  qrencode.lua:scan_p1p3
13%  qrencode.lua:qrcode
 6%  qrencode.lua:popcount32
 5%  qrencode.lua:compute_penalty_components
 5%  qrencode.lua:func
 3%  qrencode.lua:set_bb_bit
 3%  qrencode.lua:add_typeinfo_both
09/23/26-00:47:33 INFO  ====================================
 
09/23/26-00:47:33 WARN  New bit-packed integer implementation - burst: no _debug_reset_cache exposed - cache state is whatever this session has already built, not guaranteed cold 
09/23/26-00:47:33 INFO  === QR Burst Benchmark: New bit-packed integer implementation - burst === 
09/23/26-00:47:33 INFO    Chunk 1/5 (1413 bytes): 0.14000 seconds 
09/23/26-00:47:33 INFO    Chunk 2/5 (1413 bytes): 0.03000 seconds 
09/23/26-00:47:33 INFO    Chunk 3/5 (1413 bytes): 0.03000 seconds 
09/23/26-00:47:33 INFO    Chunk 4/5 (1413 bytes): 0.02000 seconds 
09/23/26-00:47:33 INFO    Chunk 5/5 (1416 bytes): 0.03000 seconds 
09/23/26-00:47:33 INFO  ====================================
```

I think the data speaks for itself, creating one relatively large QR code takes almost half a second currently, a cold run with the new engine ~140ms and then pretty much nothing afterwards, ~30ms. Dropping the memory allocation per QR code is arguably a bigger win here.

